PR #51, opened 2015-03-18
Addedd -z option to specify file size for filepaths fio files

liuk001:
If you are using normal files instead of special device files, the file size cannot be specified and will be missing in the output file.
This small patch add the -z option to specify the file size.


axboe:
Merged, thanks.

PR #53, opened 2015-04-08
genzipf not as helpfull as it should be

fbacchella:
The code of genzipf is not easy to undersant, too much variables used, results don't add up and the sample is wrong.

I tried to do a little cleaning, hope it help.

### DIFF
--- a/t/genzipf.c
+++ b/t/genzipf.c
@@ -171,7 +171,7 @@ static int node_cmp(const void *p1, const void *p2)
 int main(int argc, char *argv[])
 {
 	unsigned long offset;
-	unsigned long i, j, k, nr_vals, cur_vals, interval, total_vals, nnodes;
+	unsigned long i, j, k, nr_vals, cur_vals, interval_step, next_interval, total_vals, nnodes;
 	unsigned long long nranges;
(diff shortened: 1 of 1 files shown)


axboe:
The change itself looks good, but you have some badly messed up style. Tabs are tabs, not spaces. And there's leftover whitespace damage on some lines. And this is C, so variable declarations go at the top. If you compile with your patch, you get warnings:

    CC t/genzipf.o
t/genzipf.c: In function ‘main’:
t/genzipf.c:232:9: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
         unsigned long sum = 0;
         ^
t/genzipf.c:248:3: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
   unsigned long blocks = percentage * nranges / 100;
   ^
t/genzipf.c:287:9: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
         double hit_percent_sum = 0;
         ^

That's not acceptable. Fix up the patches, and I'll be happy to apply them. Include a better commit message for the patches describing the changes in more detail as well.

PR #54, opened 2015-04-08
genzipf is not as helpfull as it should be (take 2)

fbacchella:
The code of genzipf is not easy to undersant, too much variables used, results don't add up and the sample is wrong.

I tried to do a little cleaning, hope it help.

Sorry for the formating, I forgot the coding convention in kernel-style C. Hope this is better for you and commit messages are better


axboe:
This is a lot better - but I don't want the broken commits. Rebase and collapse your changes, then I can pull it in.

PR #55, opened 2015-04-08
fio-genzipf results output are slightly wrong, it doesn't help to understand how to use it

fbacchella:
fio-genzipf results output are slightly wrong, it doesn't help to understand how to use it:

 * The last row was not displayed
 * some useless variables were used
 * the calculation for the output table missed a few events, so the sum of 'No Hits' column was wrong
 * Wrong sample in genzipf.c header, used old style arguments
 * the help message for -o options was misleading.
 * useless define of DEF_NR.
 * default of default row number to a more common value of 20,
   instead of 23.

### DIFF
--- a/t/genzipf.c
+++ b/t/genzipf.c
@@ -3,10 +3,10 @@
  * what an access pattern would look like.
  *
  * For instance, the following would generate a zipf distribution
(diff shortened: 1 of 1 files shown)


axboe:
Pulled, thanks!

PR #63, opened 2015-05-05
ARM: Use generic assembly nop and barrier code for armv8-a

mtk10669:
Signed-off-by: Milton Chiang <milton.chiang@mediatek.com>


axboe:
Pulled, thanks!

PR #64, opened 2015-05-06
Fix fio perfomance on windows to match IOMETER.

iuliur:
On google cloud compute machines with local flash, we see fio do around 60k iops while IoMeter does 100k.
This change brings fio on par with iometer.


axboe:
Thanks, that's been bothering me for a while. Pulled!

PR #65, opened 2015-05-06
allow fio --client to run multiple hosts on a shared filesystem

bengland2:
This change is intended to allow fio --client to invoke multiple fio instances on different hosts accessing a shared filesystem, such as cephfs, glusterfs, or nfs.  The problem is that fio file pathnames are not unique across hosts.  So this change captures the IP addr of the fio workload generator process.  It assumes there is only one such process per host and it does not create a subdirectory for the host.  Consequently there will be more files per directory, depending on how many fio instances are accessing the shared directory.  All comments appreciated.

### DIFF
--- a/options.c
+++ b/options.c
@@ -825,6 +825,7 @@ int set_name_idx(char *target, char *input, int index)
 	unsigned int cur_idx;
 	int len;
 	char *fname, *str, *p;
+	extern char my_addr[INET6_ADDRSTRLEN];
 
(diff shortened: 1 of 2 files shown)


bengland2:
Thx for being clear, hopefully the most recent commit fixes it for you,  I unit tested with/without --client.

axboe:
Almost - put the array declaration in options.h and it should be good. Then fold the (now) three patches into one commit.

PR #67, opened 2015-05-07
allow --client parameter to be pathname containing client host IPs/names

bengland2:
This is particularly useful when you have large numbers of fio servers (example: virtual machine guests or containers) that participate in a test.  To use this, simply use the --client option but instead of specifying a host name or IP, provide a pathname for a file containing a list of host names/addresses, one host per line.  Should this be accompanied by man page or other doc updates?  It has been unit tested with --client being IP or being pathname, and with > _POSIX_HOST_NAME_MAX chars per record.

### DIFF
--- a/README
+++ b/README
@@ -172,7 +172,7 @@ $ fio
 	--warnings-fatal	Fio parser warnings are fatal
 	--max-jobs		Maximum number of threads/processes to support
 	--server=args		Start backend server. See Client/Server section.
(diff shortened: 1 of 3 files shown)


bengland2:
I pushed a commit to update doc. as you suggested.  It includes documentation of previous pull request to support filesystem sharing by fio servers, ok?

axboe:
Looks good, thanks! Pulled.

PR #69, opened 2015-05-07
make handle_text output to FD_NET instead of f_out 

bengland2:
client.c handle_*() output should not go to same output file as JSON, because it will get interleaved with it and break JSON syntax.  This output seems more like debugging output than necessary result output.  So use dprint(FD_NET,...) instead in handle_text().  This is solving problem in http://www.spinics.net/lists/fio/msg03860.html .


bengland2:
This is stale and would have to be rebased, don't seem to need it anymore, killing it.

PR #70, opened 2015-05-14
Fix warning from gmake on BSD

kusumi:
This pull request fixes -Wreturn-type compiler warning that i see on BSD.

stat.c: In function 'block_state_category':
stat.c:514:1: warning: control reaches end of non-void function
[-Wreturn-type]
 }

### DIFF
--- a/stat.c
+++ b/stat.c
@@ -511,6 +511,7 @@ static int block_state_category(int block_state)
 		return 2;
 	default:
 		assert(0);
+		return -1;
 	}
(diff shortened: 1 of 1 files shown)


kusumi:
This pull request fixes -Wreturn-type compiler warning that i see on BSD.

axboe:
assert(0) does not return at all, but we can add this to silence BSD compiles. I'll add a comment as well.

PR #71, opened 2015-05-14
Change (blank)cpu affinity macros to inline functions

kusumi:
Not sure if you would want to take this one, but I think it's better with inline functions for these cases where macros are used in either conditional or non-conditional so as to avoid compiler warnings.

---
gcc warns -Wunused-value on some environments (e.g. BSD) when
the following cpu affinity macros are used in non conditional code.

Also removed #ifdef FIO_HAVE_CPU_AFFINITY in gettime-thread.c
since this variable needs to be visible when calling the function.

gettime-thread.c: In function 'gtod_thread_main':
os/os.h:82:36: warning: statement with no effect [-Wunused-value]
 #define fio_setaffinity(pid, mask) (0)
                                    ^
gettime-thread.c:48:2: note: in expansion of macro 'fio_setaffinity'
  fio_setaffinity(gettid(), fio_gtod_cpumask);

### DIFF
--- a/gettime-thread.c
+++ b/gettime-thread.c
@@ -9,9 +9,7 @@
 struct timeval *fio_tv = NULL;
 int fio_gtod_offload = 0;
 static pthread_t gtod_thread;
-#ifdef FIO_HAVE_CPU_AFFINITY
 static os_cpu_mask_t fio_gtod_cpumask;
-#endif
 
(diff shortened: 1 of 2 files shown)


axboe:
I think this looks fine, that's actually the way I generally prefer to handle situations like this.

PR #73, opened 2015-05-16
Cleanup DragonFlyBSD support

kusumi:
Remove these define/undef since fio compiles on DragonFlyBSD
without them.

\<sys/rb.h> seems to have never existed in DragonFlyBSD history,
and it compiles without these define/undef. It seems this hack
was necessary on some older versions of NetBSD.


axboe:
Pulled, thanks.

PR #75, opened 2015-05-25
embed server address, not other end's address

bengland2:
This corrects a mistake I made in a previous pull request, which embedded the peer's IP address in the filename.  We want the fio --server IP address (not initiator's IP) to make filenames unique across distributed namespace.    I have run tests on 6-server distributed filesystem (Gluster) using this fix.

### DIFF
--- a/server.c
+++ b/server.c
@@ -21,6 +21,7 @@
 #endif
 
 #include "fio.h"
+#include "options.h"
 #include "server.h"
 #include "crc/crc16.h"
 #include "lib/ieee754.h"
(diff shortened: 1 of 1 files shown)


axboe:
I don't think we want to make the name lookup failure be a fatal error condition. Apart from that, looks fine.

bengland2:
Hopefully inet_ntop doesn't do reverse DNS lookup.  So I had assumed that it should not fail if the socket was still connected, and if the socket wasn't still connected then we probably wanted to error out anyway, but you are right that address conversion to string is not an essential feature of fio in all cases so why abort if it doesn't work.   I tried to do what you suggested in the 2nd commit.

axboe:
Thanks, pulled.

PR #78, opened 2015-06-01
Minor cleanups on cleanup()

kusumi:
Sync with the way other engines implement cleanup() that is regarded
as a better practice. td->io_ops->data should have null check before
dereference and its members can be freed without null check.


axboe:
Pulled, thanks.

PR #81, opened 2015-06-04
Disable libaio for ESXi build - bug#80

kennethburgener:
Suggested fix for bug#80 (libaio library not found on ESXi)

### DIFF
--- a/configure
+++ b/configure
@@ -471,8 +471,10 @@ int main(void)
 }
 EOF
 if compile_prog "" "-laio" "libaio" ; then
-  libaio=yes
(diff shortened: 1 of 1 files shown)


axboe:
No point in checking and compiling the libaio test, just to disable it if esx is set. That check should be moved further up, below where 'libaio' is set to "no".

kennethburgener:
Ezra Pedersen has the change you suggested.  He should be posting it soon...

kennethburgener:
Ezra Pedersen has the change you suggested: https://github.com/axboe/fio/pull/138

PR #87, opened 2015-06-09
Minor fix for an unlikely problem

szaydel:
I spotted this, and it is a rare problem, but could happen, so I made a quick change. Figured, perhaps you might be interested.


axboe:
Looks fine, not a bad idea to clean up the type.

PR #88, opened 2015-06-09
Fix compiler warning

kusumi:
The local variable off is uninitialized on 'goto fill'.
Initializing with 0 seems to be appropriate to fix following gcc warning
although off is actually always initialized as the function returns with
1 on 'if (!i)' case.

options.c:920: warning: 'off' may be used uninitialized in this function

### DIFF
--- a/options.c
+++ b/options.c
@@ -917,7 +917,7 @@ static int str_opendir_cb(void *data, const char fio_unused *str)
 static int pattern_cb(char *pattern, unsigned int max_size,
 		      const char *input, unsigned int *pattern_bytes)
 {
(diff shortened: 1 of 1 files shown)


axboe:
It's a bogus warning, your compiler is too stupid to realize this. Mine doesn't complain. But I'll pull your change, it's probably not the only one out there that will throw a warning for this.

kusumi:
thanks. btw I saw this on RHEL6 compatible distro using gcc 4.4.7.
> gcc version 4.4.7 20120313 (Red Hat 4.4.7-11) (GCC)

PR #90, opened 2015-06-11
Add get_fs_size() support for BSDs

kusumi:
This pull request submits 3 related commits on get_fs_size().

I've used POSIX statvfs(2) instead of statfs(2), but if there is a reason to use statfs(2) I'll resubmit it using statfs(2) (just like Linux version does).


axboe:
Looks good, thanks! Pulled.

PR #92, opened 2015-06-12
Enable FIO_HAVE_CHARDEV_SIZE on DragonFlyBSD

kusumi:
DragonFlyBSD no longer has block device just like FreeBSD got rid of
it at some point. Enable FIO_HAVE_CHARDEV_SIZE and implement get-size
functions with DragonFlyBSD's ioctl so fio can retrieve correct size
when targets are not regular files.

The following result verifies that df(1) shows the same fs size as
chardev_size() with regards to the character device (block device if
it were on Linux).

--
 # uname
 DragonFly
 # cat ./test3.c
 #include \<stdio.h\>
 #include \<stdlib.h\>
 #include \<fcntl.h\>
 #include "os/os-dragonfly.h"

 int main(int argc, char *argv[]) {
         struct fio_file f;
         unsigned long long bytes;

         f.fd = open(argv[1], O_RDONLY);
         if (f.fd < 0) {
                 perror("open");
                 exit(1);
         }

         if (chardev_size(&f, &bytes)) {
                 perror("ioctl");
                 exit(1);
         }
         printf("%s %llu\n", argv[1], bytes);
         close(f.fd);
         return 0;
 }
 # gcc -Wall -g ./test3.c -o test3
 # file /dev/da8
 /dev/da8: character special (30/504430663)
 # ./test3 /dev/da8
 /dev/da8 31004295168
 # newfs /dev/da8 > /dev/null
 # mount -t ufs /dev/da8 /mnt
 # df -TH /mnt
 Filesystem  Type   Size   Used  Avail Capacity  Mounted on
 /dev/da8    ufs     31G   2.0k    28G     0%    /mnt

### DIFF
--- a/os/os-dragonfly.h
+++ b/os/os-dragonfly.h
@@ -8,14 +8,15 @@
 #include <sys/param.h>
 #include <sys/sysctl.h>
 #include <sys/statvfs.h>
+#include <sys/diskslice.h>
 
 #include "../file.h"
 
 #define FIO_HAVE_ODIRECT
-#define FIO_USE_GENERIC_BDEV_SIZE
 #define FIO_USE_GENERIC_RAND
 #define FIO_USE_GENERIC_INIT_RANDOM_STATE
 #define FIO_HAVE_FS_STAT
+#define FIO_HAVE_CHARDEV_SIZE
 #define FIO_HAVE_GETTID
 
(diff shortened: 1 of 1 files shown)


axboe:
Pulled, thanks!

PR #94, opened 2015-06-23
Modify rdma engine to use proper arguments.

lsgunth:
The old rdma options are unintuitive and inflexible. We are also going
to do some experiments with this engine which require adding further
options. Thus we've improved it to take more normal arguments.

The technique is copied from the netio engine. It now requires a
hostname, port and verb option. The fio scripts in the example directory
have been updated. Compatability is also maintained for fio scripts that
use the old options.


axboe:
Nice work! Pulled.

PR #96, opened 2015-07-08
fio: allow for build in separate build directory

jsgf:
Change configure so that it will use its current directory for building,
and the path to the script itself as the source directory.

Change the Makefile to use VPATH to find the source files. Only a few other
things needed to be touched:
 - use the full path to the source in wildcard, then strip it off again
 - search both build and source for header files
 - FIO-VERSION-GEN is in source
 - make directories in build as needed
 - use $< to refer to input files
 - install non-executables from source

Signed-off-by: Jeremy Fitzhardinge <jeremy@exablox.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,9 @@
+ifeq ($(SRCDIR),)
+SRCDIR := .
+endif
+
+VPATH := $(SRCDIR)
+
 ifneq ($(wildcard config-host.mak),)
 all:
 include config-host.mak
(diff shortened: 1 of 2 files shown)


axboe:
Tested it here, works for me. Pulled, thanks!

jsgf:
Hm, configure is probably a little too eager to overwrite a Makefile...

axboe:
Heh yes, just noticed that too. This doesn't end well:

$ axboe@lenny:/home/axboe/git/fio $ ../fio/configure

PR #97, opened 2015-07-08
configure: don't be so eager to overwrite an existing Makefile

jsgf:


axboe:
That's an improvement, at least it wont overwrite an existing one. Pulled.

PR #99, opened 2015-07-16
re-seed based on rand_repeatable

DaveGlen:
Re-seeding the random generator will repeat "random" IO sequences, which in some cases may produce unexpected IO patterns.

if rand_repeatable is true (default), then the random generator will be re-seeded every iteration per normal. Otherwise the random generator will not be-reseeded and will continue to produce random IO offsets.

### DIFF
--- a/libfio.c
+++ b/libfio.c
@@ -116,9 +116,11 @@ void clear_io_state(struct thread_data *td)
 	}
 
 	/*
-	 * Set the same seed to get repeatable runs
+	 * Re-Seed random number generator if rand_repeatable is true
 	 */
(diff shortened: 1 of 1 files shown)


axboe:
Thanks Dave. I manually merged this since you had some style issues in your patch, but the end result should be the same.

DaveGlen:
Jens, do you have a style guideline so I can match it in the future?

axboe:
The code should be pretty consistent, so I'd just say look at that. No braces for single line ifs, spaces after closing '(', tabs are tabs, no spaces. That sort of thing.

PR #104, opened 2015-08-12
Potential Improvement in Rate Implimentation

DaveGlen:
The current implementation of rate is direction control inside of rate_ddir, using td->rate_pending_usleep[ddir] as feedback.  This feedback occurs during io-completion, which introduces a delay while IO is pending.  When IO-depth > 1, the result is that rate control may saturate the iodepth with a given ddir until IO completions occur and provide feedback, regardless of if the given iodepth is necessary to maintain rate. A simplified flow chart of the feedback delay can be found here:
https://docs.google.com/drawings/d/1EG-9eGlNvw-m9m0wMSb_C_lyJcaPlhEFbkHRsVpt4wo/edit?usp=sharing

The proposed change alters rate control to contain feedback on io issues rather than completions, in order to remove the feedback delay. A flow diagram shoeing the idea can be found here:
https://docs.google.com/drawings/d/1NphdZGjYGuOLWJzvXHv44zuy0nnSunvFrROCVfA67Y8/edit?usp=sharing

Here are the list of changes:
1.Replace td->rate_pending_usleep[ddir] with td->rate_next_io_time[ddir] to track rate control
2.Add td->rate_io_issue_bytes[ddir] to track submissions and update it whenever td->io_issue_bytes[ddir] is updated.
3.Modify usec_for_io() to use td->rate_io_issue_bytes and move it into backend.c
4.Update td->rate_next_io_time[ddir] using usec_for_io() after io enqueues and before “reap:” in do_io()
5.Check td->rate_next_io_time[ddir] compared to now in rate_ddir(). If it is in the past, we are behind rate and need to submit IO. if it is in the future, we are ahead of rate and can switch ddir or sleep. 
6.Remove rate control from io_completed and cleanup unused variables
7.Change _setup_rate in init.c for new thread variables
8.reset td->rate_next_io_time[ddir] and td->rate_io_issue_bytes[ddir] in reset_io_counters

I have performed some testing where a single randrw job that has rate control on writes, but not on reads.  The result is a 4x implement in 99.99% clat  read percentiles, and a 1.6x improvement in the same for writes.

Although the results appear promising, At this point It might be best to keep it on a separate branch pursuing possible performance improvements.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -763,6 +763,25 @@ static int io_complete_bytes_exceeded(struct thread_data *td)
 }
 
 /*
+ * used to calculate the next io time for rate control
+ *
+ */
+static long long usec_for_io(struct thread_data *td, enum fio_ddir ddir)
+{
+	uint64_t secs, remainder, bps, bytes;
+
+	assert(!(td->flags & TD_F_CHILD));
+	bytes = td->rate_io_issue_bytes[ddir];
+	bps = td->rate_bps[ddir];
+	if (bps) {
+		secs = bytes / bps;
+		remainder = bytes % bps;
+		return remainder * 1000000 / bps + secs * 1000000;
+	} else
+		return 0;
+}
+
+/*
  * Main IO worker function. It retrieves io_u's to process and queues
  * and reaps them, checking for rate and errors along the way.
  *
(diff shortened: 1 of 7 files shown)


axboe:
I like this change, good work. And I'm impressed with the level of detail that went into your analysis and presentation, great work!

PR #106, opened 2015-08-17
Allow ratemin to be used without also setting rate

DaveGlen:
only check if rate < ratemin when rate is set.


axboe:
Can you collapse the patches? I'd really prefer not pulling anything that breaks the build, that makes it that much harder to debug issues with git bisect. No concerns on the change itself, it's good.

DaveGlen:
sure. I will close and re-submit another pull request

PR #108, opened 2015-08-17
allow for ratemin to be used without setting rate

DaveGlen:
Only check if ratemin < rate when rate is set.


DaveGlen:
sorry, a tool change is causing confusion and auto-synchronizing... closing this pull request, I will make a new one once it is cleaned up

PR #110, opened 2015-08-18
Correct handleing of rate_iops_min and ratemin

DaveGlen:
change to allow ratemin to be set without rate, disabled ratemin checking in __check_min_rate.

Separate from the change to ratemin parsing, rate_iops_min check did not result in "return 1;" so jobs would not terminate on a failure to maintain rate_iops_min.


axboe:
Thanks, pulled!

PR #114, opened 2015-09-15
Reduced minimum for option log_gz

sbates130272:
Reduced the permissible minimum for log_sz from 32MB to 1KB to avoid
a latency log issue associated with the reallocation of the buffer logs.

Hi 

We noticed an issue in the latency logs when running on a low latency DRAM based SSD device. We saw a noticeable increase in latency on the 1024, 2048, 4192 etc IO and the increase in latency increases linearily (see attached image for an example).

Turns out this is an issue in the way the buffer is reallocated when it reaches those values.

Our fix allows the user to switch to the compression mode by reducing the compression minimum value from 32MiB to 1KiB. We don't think there is any downside in reducing this lower limit.

capture


axboe:
Looks fine to me, thanks. Pulled.

Issue #56, opened 2015-04-13
fio2gnuplot won't generate graphs - 'IOError: [Errno 21] Is a directory: './''


sammcj:
- fio-2.1.11
- gnuplot 4.6 patchlevel 6
- Debian 8

Running `# fio2gnuplot -p '*.log' -g` gives:

```
12 files Selected with pattern '*.log'
 |-> s1-san5-128k-md200-randread-para.results_bw.4.log
 |-> s1-san5-128k-md200-randread-para.results_iops.4.log
 |-> s1-san5-128k-md200-randwrite-para.results_bw.3.log
 |-> s1-san5-128k-md200-randwrite-para.results_iops.3.log
 |-> s1-san5-1m-md200-randread-para.results_bw.6.log
 |-> s1-san5-1m-md200-randread-para.results_iops.6.log
 |-> s1-san5-1m-md200-randwrite-para.results_bw.5.log
 |-> s1-san5-1m-md200-randwrite-para.results_iops.5.log
 |-> s1-san5-4k-md200-randread-para.results_bw.2.log
 |-> s1-san5-4k-md200-randread-para.results_iops.2.log
 |-> s1-san5-4k-md200-randwrite-para.results_bw.1.log
 |-> s1-san5-4k-md200-randwrite-para.results_iops.1.log

Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/pudb/__init__.py", line 77, in runscript
    dbg._runscript(mainpyfile)
  File "/usr/lib/python2.7/dist-packages/pudb/debugger.py", line 371, in _runscript
    self.run(statement, globals=globals_, locals=locals_)
  File "/usr/lib/python2.7/bdb.py", line 400, in run
    exec cmd in globals, locals
  File "<string>", line 1, in <module>
  File "/usr/bin/fio2gnuplot", line 517, in <module>
    sys.exit(main(sys.argv))
  File "/usr/bin/fio2gnuplot", line 497, in main
    compute_aggregated_file(fio_data_file, gnuplot_output_filename, gnuplot_output_dir)
  File "/usr/bin/fio2gnuplot", line 137, in compute_aggregated_file
    f = open(gnuplot_output_dir+gnuplot_output_filename, "w")
IOError: [Errno 21] Is a directory: './'
```

Running a trace, it looks like the variable `gnuplot_output_filename` is empty:

```
file: 's1-san5-4k-md200-randwrite-para.results_iops.1.log'
fio_data_file: list
[...]

PR #115, opened 2015-09-16
Update fio2gnuplot

Tfindelkind:
Will fix the https://github.com/axboe/fio/issues/56 issue and use the word "default" as filename. Can be changed to something more useful.


axboe:
Pulled, thanks!

PR #117, opened 2015-09-17
Update fio2gnuplot

Tfindelkind:
Should fix the "No log file..." with default pattern -b and -i and
 user generated a log file per thread/numjob with per_numjobs_log=1


axboe:
Thanks, pulled! In the future, please use the following format for commit messages:

1) First line, summary of the change. For instance, this could have been

fio2gnuplot: fix "no log file" error with -b and -i

2) After the first line, more exhaustive explanation of the change.

Tfindelkind:
Alright.

PR #118, opened 2015-09-29
Sgupdates

shimrot:
Had a project where the aio engine would not work and needed to use the SGEngine instead.  However, we encountered a couple of bugs in the SG Engine and needed support for larger LBA ranges.  Thus, the changes I've listed in the commits, however I'll elaborate below:

1) The use of queue depths greater than 1 was not working.   In fio_sgio_getevents, a seg fault was caused by attempting to read response events for all of the outstanding requests at one time, and only the 1st event was actually returned.  So, switched to doing multiple reads for the completed events which worked.

2) When a scsi operation failed, there was no attempt to read the sense data, and the failure was actually being ignored.  So added support to capture the sense data and ripple up to print an error detail message with the sense contents.

3) Many operations could actually timeout because the default scsi operation timeout was '0', which apparently does not mean 'indefinite'.  So, I hardcoded the timeout value to 30 seconds.   It would be better to be configurable, but I didn't attempt to add that (wasn't sure how to do it in fio).

4) sg engine was implemented to only use the 10 byte read/write operations.  Added support for the 8byte LBAs and 16 byte commands when needed.

5) The maximum size was not being set, so the user needed to know the exact lba size of the device to address the whole range.  I needed the exact range for some tests, which was variable by device.  So, added the read Capacity operation to get the maxlba and calculate the size from there.  
Ran into two difficulties however - the structures used within fio to manage the io and completions are not initialized at the time the file size is queried from the engine.  So I just used some local structures during the readcapacity operation, and let it be called twice - once for the block size query, and another for the capacity.  Each call tried the 10 byte readcapacity (mandatory scsi command) and if the returned capacity is 0xFFFFFFFF (the 4byte maximum), the 16 byte ReadCapacity is performed (for which support may be optional).  Thus, each file open may involve 4 ReadCapacity operations.

5)  I added line buffering to stdout.  When sense errors or debug messages were being written when using multiple threads or processes, the line content was being mixed making debug fairly difficult.  Line buffering was sufficient to ensure the contents were not being mixed.

I have only had these changes tested on CentOS 7 and Ubuntu 14.04 and with a couple of HDD SAS devices.  

Thanks
Kris Davis

### DIFF
--- a/engines/sg.c
+++ b/engines/sg.c
@@ -15,8 +15,15 @@
 
 #ifdef FIO_HAVE_SGIO
 
+#define MAX_10B_LBA  0xFFFFFFFFULL
+#define SCSI_TIMEOUT_MS 30000   // 30 second timeout; currently no method to override
+#define MAX_SB 64               // sense block maximum return size
+#define FALSE 0
+#define TRUE 1
+
 struct sgio_cmd {
-	unsigned char cdb[10];
+	unsigned char cdb[16];  	// increase to support 16 byte commands
+	unsigned char sb[MAX_SB];   // add sense block to commands
 	int nr;
 };
 
@@ -27,6 +34,7 @@ struct sgio_data {
 	int *fd_flags;
 	void *sgbuf;
 	unsigned int bs;
+	long long max_lba;
 	int type_checked;
 };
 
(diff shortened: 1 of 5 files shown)


axboe:
In general it looks fine, but there are lots of style issues and things like the new errdetails() hook not being reentrant, which it must be. I pulled your changes into a separate branch and did some fixes on top, please see:

http://git.kernel.dk/cgit/fio/log/?h=sg-updates

and let me know if that works for you.

shimrot:
Jens,

Wow, you're quick.  I reviewed your changes and don't have any objections.
I pulled and built this version locally.  No problems from a cursory check,
including forced errors.  One of my guys will do an overnight stability
run, but looks good as far as I'm concerned.

Thanks again
Kris Davis

axboe:
Collapsed and folded into master. Any further updates should be sent against that. Thanks!

PR #121, opened 2015-10-13
engines/rbd: potential re-use of active fio_rbd_iou

dillaman:
Running multiple concurrent RBD jobs can result in a segfault
attempting to retrieve the RBD AIO completion status from a
prematurely released fio_rbd_iou object.

Signed-off-by: Jason Dillaman <dillaman@redhat.com>


axboe:
Thanks, this looks better. At some point it'd be a good idea to get rid of this seen/complete issue, it doesn't fill me with confidence that it's safe...

ktdreyer:
@axboe , mind releasing a new version of fio with this change? It would be great to get this into the distros.

axboe:
There will be a new release pretty soon.

PR #124, opened 2015-10-27
Add mmapshared option to use mmaped files with the MAP_SHARED flag.

lsgunth:
This is primarily useful for benchmarking DAX files. If they are mmaped
with the MAP_PRIVATE flag they will still have copy on write semantics
and the test will not actually run against the memory that backs the
DAX file. This memory option allows tests to be constructed that use the
MAP_SHARED flag which allows running directly with the memory backing
a DAX file.


axboe:
Thanks, looks good - pulled.

PR #135, opened 2015-12-19
Rewrote libhdfs engine

fbacchella:


axboe:
That looks a lot better! Only question is why the changes to options.h?

fbacchella:
Because there is too much options. And you store them in a bit mask or something like than. By activating only used options, it reduced the needed length.

axboe:
I just extended the bitfields to 64-bit instead, see current -git. I think that's cleaner than having to provide ifdefs, and have different numbering depending on what is available. This could break client/server, for instance.

fbacchella:
I agree totally. That's much better this way. I will fix my pull request and resubmit.

fbacchella:
It should be better now. There is a little reformatting because of mixed space and tabulation in options.h. Do you want a separate commit to fix that

axboe:
No that all looks fine, pulled! Thanks.

PR #146, opened 2016-02-03
Compile fixes

robbat2:
- Fixup a warning re _FORTIFY_SOURCE
- ensure that libmtd stuff compiles on newer kernel headers


axboe:
Thanks, pulled.

PR #153, opened 2016-03-05
Log to parent instead of twice to child td

digikata:
In io_u_log_error, the duplicated log call looked strange, this patch is a guess at original intent.


axboe:
Yep, good catch!

PR #160, opened 2016-04-04
Modify RDMA engine to print strerror messages.

lsgunth:
In order to ease debugging it would be nice to know why certain
rdma/ib functions fail. I've added %m messages to a large
number of log_err lines.


axboe:
Added, thanks.

PR #161, opened 2016-04-05
add pmemblk engine

bgbhpe:
Seeking comments on a new IO engine utilizing the NVML libpmemblk.  More information on libpmemblk and the NVML is available here http://pmem.io/nvml/.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,9 @@ ifdef CONFIG_MTD
   SOURCE += oslib/libmtd.c
   SOURCE += oslib/libmtd_legacy.c
 endif
+ifdef CONFIG_PMEMBLK
+  SOURCE += engines/pmemblk.c
+endif
 
 ifeq ($(CONFIG_TARGET_OS), Linux)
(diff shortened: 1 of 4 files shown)


axboe:
I'll take a look when I'm back from vacation next week. But some initial comments:

1) If this is going into fio git, then you don't need to keep the external engine hooks. Get rid dlym checks, and the defines.
2) Add an example/ job file that uses it.

axboe:
Additionally, enable it by default on the appropriate platforms, and I don't think we need a --disable thing for configure either.

axboe:
I have merged this, but you still had missing bits. You didn't wire up the engine through options.c, I'm assuming you originally used this as an external engine? I've added some documentation bits as well, and cleaned up the coding style.

bgbhpe:
Thanks, Jens.  Sorry for the delayed reply, I'd addressed some of your feedback, but then got caught up with some other stuff.  Yes, it was originally an external engine.

axboe:
No worries! I've done some cleanups around it, when you have time it would be great if you could test it and provide any comments on things that you feel need changing.

PR #168, opened 2016-05-01
add clustername for rbd engine

jazeltq:
As the Ceph Jewel is released, ceph cluster name
may not be the default "ceph". So we should add
the clustername for rbd engine to support that.

Signed-off-by: Tianqing <tianqing@unitedstack.com>


axboe:
I'll pull this in, but you have a lot of coding style conflicts. Please try and follow them better next time, yours is not even consistent (tabs vs spaces).

PR #176, opened 2016-05-17
Fiologparser stats

bengland2:
I added a new option to fiologparser.py, -A, to generate more useful stats from latency logs.  I think it's a great idea to divide up the latency log into time intervals, but I wasn't getting a good picture of the latency distribution.   This option generates a csv format that can be directly loaded into a spreadsheet and graphed.   For example, if I run it on a real latency log with thousands of records, it looks like this:

[ben@bene-laptop ceph]$ ./fiologparser.py -A example-latency.log | head > [example.csv](https://s3.amazonaws.com/ben.england/fiologparser/example-latency.csv)

and here's a sample output, a [LibreOffice spreadsheet]( https://s3.amazonaws.com/ben.england/fiologparser/example-latency-graph.ods) and a [graph of it generated with LibreOffice]( https://s3.amazonaws.com/ben.england/fiologparser/example-latency-graph.png):


axboe:
Ben, change looks fine, but your tree is 5 merge commits and this one. Can you clean it up? I'd rather not pull that into my tree.

bengland2:
Sorry, I'll create a fresh PR with just the one change.  It was an old git tree, should have started fresh.

PR #179, opened 2016-05-20
The fixed CPU architecture in the Makefile will make failure on ppc64le.

davidzengxhsh:
The Makefile for building with libhdfs is fixed to amd64, which makes failure on ppc64le arch.


axboe:
Thanks!

PR #182, opened 2016-05-25
Allow build software to inject further C build flags

martin21:
This is useful for dpkg and oder packaging software to inject hardening
options like -fPIE -pie for position independent builds which allow for
address space randomization.

### DIFF
--- a/FIO-VERSION-GEN
+++ b/FIO-VERSION-GEN
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 GVF=FIO-VERSION-FILE
-DEF_VER=fio-2.9
+DEF_VER=fio-2.10
 
 LF='
 '
--- a/FIO-VERSION-GEN
+++ b/FIO-VERSION-GEN
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 GVF=FIO-VERSION-FILE
-DEF_VER=fio-2.10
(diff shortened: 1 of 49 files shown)


axboe:
There's already EXTFLAGS for this, can't we reuse that?

martin21:
On Mittwoch, 25. Mai 2016 08:53:12 CEST Jens Axboe wrote:
> There's already EXTFLAGS for this, can't we reuse that?

Well, I just used what works with the standard Debian hardening build 
mechanism using the path of least resistance. I could try to deviate from that 
and set the hardening flags in EXTFLAGS.

But what does speak against just adding to CFLAGS like the makefile does it 
with LDFLAGS already?

Thanks,
-- 
Martin

axboe:
We can do that, but I think we need to do a few cleanups for that too. Currently the configure script outputs CFLAGS, but we override them in the Makefile. If we make the CFLAGS statement in the Makefile additive, then we have duplicate flags. Would be nice to clean up that eyesore too.

martin21:
What do you suggest? Remove duplicate CFLAGS from Makefile?

axboe:
Yes. If we just merge this change, then we'll have -D_GNU_SOURCE -include config-host.h twice, for instance, since it's set in both CFLAGS and BUILD_CFLAGS.

PR #201, opened 2016-07-08
verify: Reset verify_state before verification phase.

charles-jacobsen:
This job with verification will incorrectly fail:

[my_job]
bs=64k
verify=md5
rw=write
size=1M
randrepeat=0
verifysort=0

verifysort is on by default, so the rand_seed check wasn't being done.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -603,6 +603,15 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
 	if (td->error)
 		return;
 
+	/*
+	 * verify_state needs to be reset before verification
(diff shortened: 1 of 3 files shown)


charles-jacobsen:
I could check if rand_repeatable is set, and not do the reset in that case.

axboe:
I think it'd be cleaner with that check, since it'd include more information to the reader (that this depends on randrepeat).

charles-jacobsen:
Ok great, I updated it. If the comment is too verbose, let me know and I'll fix that.

axboe:
Looks fine, thanks!

PR #203, opened 2016-07-11
workqueue: rename private to priv for compiling as c++

cbodley:
Given the main.cc source file:
```
#include <fio.h>
int main() { return 0; }
```
And the gcc command line:
```
$ gcc main.cc -Ifio -DCONFIG_HAVE_BOOL
In file included from ./ioengine.h:10:0,
                 from ./iolog.h:7,
                 from ./stat.h:4,
                 from ./thread_options.h:7,
                 from ./fio.h:18,
                 from main.cc:1:
./workqueue.h:19:8: error: expected unqualified-id before ‘private’
  void *private;
        ^
./workqueue.h:19:7: error: expected ‘;’ at end of member declaration
  void *private;
       ^
./workqueue.h:19:15: error: expected ‘:’ before ‘;’ token
  void *private;
               ^
```
Building as c++ is useful for external engines, such as the ceph objectstore engine (ceph/ceph#5943).

### DIFF
--- a/rate-submit.c
+++ b/rate-submit.c
@@ -14,7 +14,7 @@ static int io_workqueue_fn(struct submit_worker *sw,
 {
 	struct io_u *io_u = container_of(work, struct io_u, work);
 	const enum fio_ddir ddir = io_u->ddir;
-	struct thread_data *td = sw->private;
+	struct thread_data *td = sw->priv;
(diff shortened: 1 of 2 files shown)


axboe:
Pulled, thanks.

PR #208, opened 2016-07-17
gnuplot 5 fixes

sitsofe:
`fio2gnuplot` doesn't work with gnuplot 5 because gnuplot 5 changed the syntax used for passing arguments to scripts in a backwards incompatible way (which results in `fio2gnuplot` creating files named "$3.png"). This patch moves to new-style parameters while trying to also keep things working on older gnuplot versions (tested with gnuplot 4.6.6 and gnuplot 5.0.3).


axboe:
Thanks!

PR #211, opened 2016-07-20
gaps in IOPS logs because of time drift

cronburg:
Every time we set iolog->avg_last, we get a slight time drift. Namely we are taking samples *at least* every log_avg_msec milliseconds. This PR adjusts for the drift by subtracting the amount we drift by when updating avg_last.


axboe:
Let's see if that does it... Thanks for looking into this! Merged.

PR #212, opened 2016-07-22
Fix "exitall_on_error" option

YukiKita:
I found that the rest of the jobs did not stop after a job terminated with an error while _exit_on_all_ was enabled.
So I fixed _options.c_ so that _exitall_on_error_ could work as specified.


axboe:
Good catch! Thanks, merged.

PR #213, opened 2016-07-27
Fix memory leak in _fio_setup_rbd_data()

JevonQ:
Fix memory leak in _fio_setup_rbd_data().
Signed-off-by: Jevon Qiao <scaleqiao@gmail.com>

### DIFF
--- a/engines/rbd.c
+++ b/engines/rbd.c
@@ -110,8 +110,13 @@ static int _fio_setup_rbd_data(struct thread_data *td,
 	return 0;
 
 failed:
-	if (rbd)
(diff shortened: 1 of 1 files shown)


axboe:
Applied, thanks.

JevonQ:
@axboe BTW, is there any code style rules for fio we need to follow? If so, is there any tool to check it?

Thanks,
Jevon

axboe:
There's no set of rules for code style, but the code base should be pretty consistent as it is. It's pretty close to the linux kernel code style in general, basically the same.

PR #214, opened 2016-07-28
Fix segmentation fault and memory leak issue in rbd engine

JevonQ:
For the segmentation fault issue:
Fio with rbd engine will panic if clustername option is specified but
clientname is not. In that case, o->client_name will be a NULL pointer,
strlen(o->client_name) in _fio_rbd_connect() will lead to a segmentation
fault.

For the memory leak issue:
When running fio with rbd engine and specifying clustername, a piece of
memory will be allocated in _fio_rbd_connect(), but it never gets freed.

### DIFF
--- a/engines/rbd.c
+++ b/engines/rbd.c
@@ -134,15 +134,22 @@ static int _fio_rbd_connect(struct thread_data *td)
 		 * as a full type.id namestr
 		 */
 		if (o->client_name) {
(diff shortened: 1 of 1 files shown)


axboe:
Why was this closed? The segfault fix looks fine, it doesn't matter if the other patch is already merged.

Issue #183, opened 2016-05-26
No output file generated for write_bw_log or wirte_iops_log!


cashew250:
No output files are generated to /var/log/SSD/ when running the below config. However, when I kill the fio process I will suddenly see some. Issue is, I ran this overnight, it completed and no files were generated. Not sure what is going on, when is the output being dumped by each process? Fio 2.1.10, Debian version is Wheezy, I tried a 2.11 build of Fio but no luck.

[global]
ioengine=libaio
direct=1
buffered=0
log_avg_msec=1000
group_reporting
filename=/dev/sdb

[sequential-write-precondition]
stonewall
description=Sequential write precondition
rw=write
iodepth=16
bs=1M
size=100%
# log_avg_msec=60000

write_bw_log=/var/log/SSD/ssd_seqwrt_precon_bandwidth
write_iops_log=/var/log/SSD/ssd_seqwrt_precon_iop
loops=2

[random-write-precondition]
stonewall
description=Random write precondition phase
rw=randwrite
bs=4K
iodepth=4
numjobs=4
size=100%
refill_buffers
# log_avg_msec=60000

write_bw_log=/var/log/SSD/ssd_rndwrt_precon_bandwidth
write_iops_log=/var/log/SSD/ssd_rndwrt_precon_iop
norandommap

[random-write-steady]
stonewall
description=Random write steady state phase
rw=randwrite
bs=4K
iodepth=4
numjobs=4
time_based
runtime=18000
size=100%
refill_buffers
norandommap
# log_avg_msec=60000

write_bw_log=/var/log/SSD/ssd_rndwrt_results_bandwidth
write_iops_log=/var/log/SSD/ssd_rndwrt_results_iop

xjtuwjp:
We have a similar bug with latest fio master branch (6a89b4012)
iops file was not generated.
bw file was generated but a lot zeros after first sample, eg
2498, 615636, 1, 0
5497, 0, 1, 0
8495, 0, 1, 0
11494, 0, 1, 0
14492, 0, 1, 0

axboe:
xjtuwjp, what job file did you run? I can take a look at this, most easily with a job file or similar that reproduces the issue.

xjtuwjp:
@axboe, sorry missed you comments, I just retried with fio-2.12, same result.
[...]

PR #216, opened 2016-07-29
Fix overflow caused by signed long division by unsigned long.

YukiKita:
I have been following issue ( #183 ), and found the root cause.
If the value of 'log_avg_msec' is relatively large (maybe greater than 1000), the computed interval time in _mtime_since_ overflows.Then the values of bw and IOPS are set to zero by the overflow. 
I confirmed this fix have resoloved the issue ( #183 ).


axboe:
I've merged this. Might have been cleaner to just make the variables unsigned as well, but that would require other changes.

PR #222, opened 2016-08-15
Make histogram samples non-cumulative

cronburg:
by tracking a linked-list of the most recent histogram and differencing it when we print to the log file(s). Linked list of pointers used to minimize runtime impact on recording side, instead choosing to do subtraction on the logging (when logs get printed to file) side.

This helps cleanup `fiologparser_hist.py`, makes the log files easier to understand at a glance (the frequencies correspond to just that interval), reduces file size(s), and averts the possibility of integer overflow in the logs (for a long enough fio job).

I couldn't get the existing linked list implementation to work for me (`flist.h`), but I can try again / rework this to use that if that's preferred. All I needed was a write-once, read-once singly-linked list.

### DIFF
--- a/iolog.c
+++ b/iolog.c
@@ -576,6 +576,9 @@ void setup_log(struct io_log **log, struct log_params *p,
 	       const char *filename)
 {
 	struct io_log *l;
+	int i;
+	struct io_u_plat_entry *entry;
+	struct flist_head *list;
 
 	l = scalloc(1, sizeof(*l));
 	INIT_FLIST_HEAD(&l->io_logs);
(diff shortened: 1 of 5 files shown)


axboe:
Yes, you need to use the flist helpers. It's much easier to read than hand coded lists... It's not hard to do, something ala:

    struct some_entry
    {
        struct flist_head list;
        other members;
    };

    struct some_list
    {
        struct flist_head list;
        other members;
    };

    struct some_entry entry;
    struct some_list list;

    INIT_FLIST_HEAD(&list.list);
    flist_add(&entry.list, &list.list);`

and so on. Then you can extract the first and last members through using flist_first_entry() and flist_last_entry() on &list.list. Note that for both of those, you need to ensure that flist_empty() isn't true, or it'll return garbage. Similarly, flist_last_entry() can return the first member of the list, if there is only one entry. Hope that makes sense.

Additionally, don't use ?: constructs. I hate them. They hurt readability for no good reason.

cronburg:
Updated. I don't have any calls to `flist_empty()` because the pattern of adds and deletes should always result in there being one or more entry in the list at a time (same as the `malloc` / `free` for the list entries).

Removed the ternary ifs.

axboe:
That looks better! I fixed up some of the style, and the now redundant io_u_plat_list, since it's just holding a list. We can embed that directly.

cronburg:
What tab width do you use, and how are you determining how far to tab over when continuing arguments to a function on a new line?

axboe:
A tab is a tab, 8 spaces in size. Unless you want to line up (like I do for function arguments), then just tab in so it's beyond the function name and parenthesis. But it's more things like having empty spaces or tabs on empty lines, end of lines, or spaces before tabs that makes my editor unhappy.

PR #224, opened 2016-08-18
compatible with c++

cooljiansir:
compatible with c++

### DIFF
--- a/fio.h
+++ b/fio.h
@@ -563,7 +563,7 @@ enum {
 
 static inline enum fio_ioengine_flags td_ioengine_flags(struct thread_data *td)
 {
-	return (td->flags >> TD_ENG_FLAG_SHIFT) & TD_ENG_FLAG_MASK;
+	return fio_ioengine_flags((td->flags >> TD_ENG_FLAG_SHIFT) & TD_ENG_FLAG_MASK);
 }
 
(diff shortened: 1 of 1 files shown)


axboe:
This doesn't look right at all. Did you mean to cast it? Please don't send patches that haven't been tested, let alone compiled.

axboe:
I suspect this should fix it for you:

http://git.kernel.dk/cgit/fio/commit/?id=8f6d1311afd37bf70b5b0bd7ebb0944ed9460d34

Let me know if that's the case or not.

cooljiansir:
yes, just cast it. 
well, this has been compiled and runs okay on my computer.
How should I write here if I want to make my C++ shared lib (io engine) with it,thanks: )

axboe:
I don't see how that could work at all... It'd be a function call at that point.

Can you expand on your last question?

cooljiansir:
I'm testing libfio_ceph_objectstore.so, it is a plugin written in C++. This works okay with g++ (no warning).
it is something like function call in c, which got me some warning when I make fio.

axboe:
OK, so I think we can close this issue. Current -git should work for you, I believe. If not, reopen and let me know.

cooljiansir:
okay,i'll try it latter.
thx

PR #227, opened 2016-08-18
Unix epoch in log files

cronburg:
This adds an option to output Unix timestamps instead of zero-based ones.

This makes it easier to analyze the data in the context of other benchmarking tools running at the same time as fio.

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1681,6 +1681,10 @@ log_store_compressed=bool	If set, fio will store the log files in a
 		the --inflate-log command line parameter. The files will be
 		stored with a .fz suffix.
 
+log_unix_epoch=bool	If set, fio will log Unix timestamps to the log
+		files produced by enabling write_type_log for each log type, instead
+		of the default zero-based timestamps.
+
 block_error_percentiles=bool	If set, record errors in trim block-sized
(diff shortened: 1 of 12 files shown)


axboe:
Idea looks fine, but I think you're missing a few places where we set/reset ->epoch. See libfio for instance, and rate-submit. Would probably make sense to do this in one place, and have the existing users call that function.

cronburg:
Updated as such. Added `set_epoch_time` to `time.c` which updates both `epoch` and `unix_epoch`.

axboe:
Thanks! One final request - could you update HOWTO as well? Just needs the same section as the fio.1 change.

cronburg:
I think that's in there?

HOWTO
```diff
@@ -1681,6 +1681,10 @@ log_store_compressed=bool	If set, fio will store the log files in a
 		the --inflate-log command line parameter. The files will be
 		stored with a .fz suffix.
 
+log_unix_epoch=bool	If set, fio will log Unix timestamps to the log
```
cronburg: Do you mean this?

axboe:
Yep, my bad, it is in there!

axboe:
Merged, thanks.

PR #229, opened 2016-08-18
fiologparser_hist.py man page and Makefile updates

cronburg:
This moves the documentation for `fiologparser_hist.py` into a man page, and installs both the man page and the script.


axboe:
Pulled, thanks!

PR #234, opened 2016-08-25
fiologparser_hist.py updates

cronburg:
Update to handle `log_unix_epoch` from histogram parser, and auto detect `log_hist_msec` if it's given in a job file but not on the command line.


axboe:
Merged, thanks!

PR #235, opened 2016-08-26
Client / server code for handling histograms

cronburg:
If I did this right, it should only affect fio when `--write_hist_log` is given.

Things I'm not sure I've done fully:
- Network byte order conversions
- Does the server version get updated, or is that just between releases?

For now without zlib present, this prints an error in client/server mode.

This code is working for me / my code question on the mailing list is solved (see commit message).

### DIFF
--- a/client.c
+++ b/client.c
@@ -1251,6 +1251,44 @@ static void handle_eta(struct fio_client *client, struct fio_net_cmd *cmd)
 	fio_client_dec_jobs_eta(eta, client->ops->eta);
 }
 
+static void client_flush_hist_samples(FILE *f, int hist_coarseness, void *samples,
+				      uint64_t sample_size)
+{
+	struct io_sample *s;
+	int log_offset;
+	uint64_t i, j, nr_samples;
+	struct io_u_plat_entry *entry;
+	unsigned int *io_u_plat;
+
+	int stride = 1 << hist_coarseness;
+
+	if (!sample_size)
+		return;
+
+	s = __get_sample(samples, 0, 0);
+	log_offset = (s->__ddir & LOG_OFFSET_SAMPLE_BIT) != 0;
+
+	nr_samples = sample_size / __log_entry_sz(log_offset);
+
+	for (i = 0; i < nr_samples; i++) {
+
+		s = (struct io_sample *)((char *)__get_sample(samples, log_offset, i) +
+			i * sizeof(struct io_u_plat_entry));
+
+		entry = s->plat_entry;
+		io_u_plat = entry->io_u_plat;
+
+		fprintf(f, "%lu, %u, %u, ", (unsigned long) s->time,
+						io_sample_ddir(s), s->bs);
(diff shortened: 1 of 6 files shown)


server.h
```diff
@@ -183,6 +183,7 @@ struct cmd_iolog_pdu {
 	uint32_t log_type;
 	uint32_t compressed;
 	uint32_t log_offset;
+	uint8_t log_hist_coarseness;
```
axboe: Make this a uint32_t, otherwise the struct will have holes in it.

server.c
```diff
@@ -1805,6 +1904,7 @@ int fio_send_iolog(struct thread_data *td, struct io_log *log, const char *name)
 	pdu.nr_samples = cpu_to_le64(iolog_nr_samples(log));
 	pdu.thread_number = cpu_to_le32(td->thread_number);
 	pdu.log_type = cpu_to_le32(log->log_type);
+	pdu.log_hist_coarseness = log->hist_coarseness;
```
axboe: pdu.log_hist_coarseness = cpu_to_le32(log->hist_coarseness);

axboe: And I can't see where you do the reverse conversion on receiving?

init.c
```diff
@@ -1426,6 +1426,12 @@ static int add_job(struct thread_data *td, const char *jobname, int job_add_num,
 	}
 
 	if (o->hist_log_file) {
+#ifndef CONFIG_ZLIB
+		if (td->client_type) {
+			log_err("fio: --write_hist_log requires zlib in client/server mode\n");
+			goto err;
+		}
+#endif
```
cronburg: Forgot to push this too - this is where it checks if attempting write_hist_log without zlib.

client.c
```diff
@@ -1478,6 +1525,7 @@ static struct cmd_iolog_pdu *convert_iolog(struct fio_net_cmd *cmd,
 	ret->log_type		= le32_to_cpu(ret->log_type);
 	ret->compressed		= le32_to_cpu(ret->compressed);
 	ret->log_offset		= le32_to_cpu(ret->log_offset);
+	ret->log_hist_coarseness = le32_to_cpu(ret->log_hist_coarseness);
```
cronburg: Added client side conversion here.

cronburg:
All set with your most recent comments.

bengland2:
Jens, thanks for advice on the histogram logging changes.  Are there any additional concerns that you have at this point?  Karl is transitioning back to PhD thesis so I'm going to try to fill in for him from this point onward if any tweaks are needed.   If there are no further concerns about this PR, at your convenience it would be great to have this in upstream master.   The RH perf team plans to use it for Ceph latency analysis under heavy load at scale when various operational events such as hardware addition/removal occur.   I've tried this branch with fio --client and it seems to get the histograms correctly.

axboe:
Lost track of it, sorry. I think it looks fine now, I'll pull it in.

PR #238, opened 2016-09-01
smalloc.c: simplify the process in sinit()

atheism:
Signed-off-by: Weibing Zhang <zhangweibing@unitedstack.com>


axboe:
That's basically changing code that was perfectly readable, to something that is not. Not going to pull this, sorry.

atheism:
Got it.
Thanks

PR #240, opened 2016-09-06
fiologparser_hist configparser bug fix

cronburg:
Wasn't handling older version(s) of configparser correctly.


axboe:
Pulled, thanks.

PR #241, opened 2016-09-06
generate unique pathname for each fio --client log file

bengland2:
fio --client implementation in client.c is not yet generating log filenames that are unique across hosts.  This fix simply appends .hostname to the pathname.  It also logs the filename if there is an error opening the logfile.  It does not affect filename at all unless --client is used.

### DIFF
--- a/client.c
+++ b/client.c
@@ -1294,6 +1294,7 @@ static int fio_client_handle_iolog(struct fio_client *client,
 {
 	struct cmd_iolog_pdu *pdu;
 	bool store_direct;
+	char * log_pathname;
 
 	pdu = convert_iolog(cmd, &store_direct);
 	if (!pdu) {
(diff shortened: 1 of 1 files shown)


client.c
```diff
@@ -1294,22 +1294,32 @@ static int fio_client_handle_iolog(struct fio_client *client,
...
+	char * log_pathname;
 
 	pdu = convert_iolog(cmd, &store_direct);
 	if (!pdu) {
 		log_err("fio: failed converting IO log\n");
 		return 1;
 	}
 
+	/* generate a unique pathname for the log file using hostname */
+	log_pathname = malloc(PATH_MAX+10);
```
axboe: Should this be PATH_MAX+MAXHOSTNAMELEN?

client.c
```diff
@@ -1294,22 +1294,32 @@ static int fio_client_handle_iolog(struct fio_client *client,
 {
 	struct cmd_iolog_pdu *pdu;
 	bool store_direct;
+	char * log_pathname;
```
axboe: No space between '*' and variable name.

bengland2:
on your first comment: It looks to me like this symbol is defined in /usr/include tree not fio tree.  cscope led me to /usr/include/linux/limits.h where it is defined as 

    #define PATH_MAX        4096    /* # chars in a path name including nul */

So it seems to include both the directory and the filename, right?

axboe:
Pulled, thanks.

PR #243, opened 2016-09-11
Add truncate engine

dmonakhov:
Hi,
Fio was designed to be swiss knife for io performance testing, but it is appeared to be good stress testing tool too. For that reason it is good to have such things as truncate.

BTW: If you do not mind I also  would like to add more engines for zero_range,insert_range, collapse_range ioctls.

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1675,6 +1675,11 @@ I/O engine
 			DDIR_TRIM
 				does fallocate(,mode = FALLOC_FL_KEEP_SIZE|FALLOC_FL_PUNCH_HOLE).
 
+		**ftruncate**
+			I/O engine that sends :manpage:`ftruncate(2)` operations in response
+			to write (DDIR_WRITE) events. Each ftruncate issued sets the file's
+			size to the current block offset. Block size is ignored.
+
 		**e4defrag**
 			I/O engine that does regular EXT4_IOC_MOVE_EXT ioctls to simulate
 			defragment activity in request to DDIR_WRITE event.
--- a/examples/ftruncate.fio
+++ b/examples/ftruncate.fio
@@ -1,0 +1,27 @@
+# Example ftruncate engine jobs
+
+[global]
+ioengine=ftruncate
+directory=/scratch
+size=102404k ; 100Mb+4k
+stonewall
+filename=truncate
+runtime=10s
+time_based
+direct=1
+#
+# bs option is stub here. Truncation is performed on the current block offset.
+# blocksize value is ignored
+bs=4k
+
+# truncate the file to 4Kbytes then repeatedly grow the file back to just over
+# its original size using subsequent truncates
+[grow-truncate]
+rw=write
+
+# Repeatedly change a file to a random size between 0Kbytes and 100Mb
+# using truncates
+[rand-truncate]
+rw=randwrite
+norandommap
+
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ SOURCE :=	$(patsubst $(SRCDIR)/%,%,$(wildcard $(SRCDIR)/crc/*.c)) \
 		eta.c verify.c memory.c io_u.c parse.c mutex.c options.c \
 		smalloc.c filehash.c profile.c debug.c engines/cpu.c \
 		engines/mmap.c engines/sync.c engines/null.c engines/net.c \
+		engines/ftruncate.c \
 		server.c client.c iolog.c backend.c libfio.c flow.c cconv.c \
 		gettime-thread.c helpers.c json.c idletime.c td_error.c \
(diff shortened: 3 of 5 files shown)


axboe:
Looks nice and clean. Could you add a sample job file as well? And the HOWTO and fio.1 man page should be updated as well.

sitsofe:
I've added a sample job and some docs along with some rejigged some text on https://github.com/sitsofe/fio/tree/defrag_ftruncate . Hopefully https://github.com/sitsofe/fio/commit/7606d7e350ee313ac8cf1f05816d0986780e2f44 captures the spirit of the original commit message - @dmonakhov do you know if it's correct?

sitsofe:
Just sending another ping to @dmonakhov after rebasing a branch - does https://github.com/sitsofe/fio/tree/defrag_ftruncate look correct? I've added the sample job and updated the HOWTO...

dmonakhov:
Sorry @sitsofe.  I've missed first one.
Thank you for this. I'll update branch shortly

sitsofe:
@dmonakhov no problem - just seemed a shame to let this go after the work you'd done...

sitsofe:
@dmonakhov Would you say your updated branch is now ready for pulling?

Some things I noticed about this ioengine while using it:
* Trying to verify "write" I/O done by this engine doesn't make sense but there's currently no way for the engine to tell fio this.
* The current code can never truncate a file to size 0 but it can grow a file just past its original size.

dmonakhov:
@sitsofe 
1)what do you mean by verify?  and data verify engine? Off course this does not make sense.  This engine was aimed enrich fio's stress functionality,  it is not about data integrity.
2)  truncate to i_size 0 may be important in some cases (it may helps to trigger some bugs), so yes it is reasonable change truncate to io_u->offset only. I'll update pool request

axboe:
The point that Sitsofe is trying to make is that if you set rw=write/randwrite and verify is turned on, fio will think it can verify what it wrote. But that makes no sense. So perhaps we need to be able to flag this somehow. For the null engine, we set FIO_FAKEIO and verify understands this. This engine should likely do the same.

dmonakhov:
Ohh... I've missed that.  fakeio is reasonable flag here, I'll be back with updated patch

examples/ftruncate.fio
```diff
@@ -0,0 +1,24 @@
...
+filename=truncate
+runtime=10s
+time_based
+direct=1
+
+# truncate the file to 4Kbytes then repeatedly grow the file back to just over
+# its original size using subsequent truncates
+[grow-truncate]
+rw=write
+bs=4k # truncation is performed on the current block offset + blocksize
```
sitsofe: The comments for this job don't match how the engine now works..

examples/ftruncate.fio
```diff
@@ -0,0 +1,24 @@
...
+
+# Repeatedly change a file to a random size between 4Kbytes and 106496Kbytes
+# using truncates
```
sitsofe: Same for this - the comment has become buggy

HOWTO
```diff
@@ -1675,6 +1675,11 @@ I/O engine
 			DDIR_TRIM
 				does fallocate(,mode = FALLOC_FL_KEEP_SIZE|FALLOC_FL_PUNCH_HOLE).
 
+		**ftruncate**
+			I/O engine that sends :manpage:`ftruncate(2)` operations in response
+			to write (DDIR_WRITE) events. Each ftruncate issued sets the file's
+			size to the current block offset + current block size.
```
sitsofe: The description sadly no longer correctly describes the engine's operation.

dmonakhov:
Fixed obsoleted comment about ftruncate.
Also added two independent patches which add default values for env-variables.  Should this patches goes via separate pull request?

sitsofe:
@dmonakhov I'd argue the env work is better off in a separate pull request.

Unfortunately I can't see the updated documentation/comments in https://github.com/dmonakhov/fio/find/pull-2 (e.g. https://github.com/dmonakhov/fio/blob/pull-2/HOWTO#L1694 )...

axboe:
@sitsofe This one:

https://github.com/axboe/fio/pull/243/commits/b414d1e543e7476a28b16fef4049d4291cd16911

Is everybody happy with this so we can pull it in?

dmonakhov:
@sitsofe  Oh.. you 100% right. I've fixed that in the patch, but push wrong branch.

I've updated the branch. Now it should be correct

sitsofe:
I'd say it's pretty much there - my only quibble is that I think

> truncate the file to 4Kbytes then repeatedly grow the file back to just over its original size using subsequent truncates

should be

> Truncate the file to 0 bytes then repeatedly grow the file back to its original size using subsequent truncates

and I don't think that should be enough to stop this going in.

axboe:
Merged, thanks all!

PR #245, opened 2016-09-15
add simple fio.service to start fio server with systemd

jan--f:
I put it under tools for lack of better idea.
Comments appreciated.

Signed-off-by: Jan Fajerski <jfajerski@suse.com>


axboe:
Pulled, thanks!

PR #258, opened 2016-10-18
use poll() and rbd_poll_io_events to speed up io retrieval.

liupan1111:
Hi,
In Aug. 2015, a new function rbd_poll_io_events is added in Ceph. I use it to speed up io retrieval. I tested by using P3700 SSD and ceph_jewel(bluestore), find the iops and latency has 10% improvement.

### DIFF
--- a/engines/rbd.c
+++ b/engines/rbd.c
@@ -13,9 +13,11 @@
 #include <zipkin_c.h>
 #endif
 
+#ifdef CONFIG_RBD_POLL
 /* add for poll */
 #include <poll.h>
 #include <sys/eventfd.h>
+#endif
 
 struct fio_rbd_iou {
 	struct io_u *io_u;
@@ -33,7 +35,9 @@ struct rbd_data {
 	rbd_image_t image;
 	struct io_u **aio_events;
 	struct io_u **sort_events;
-    int fd; /* add for poll */
+#ifdef CONFIG_RBD_POLL
+	int fd; /* add for poll */
+#endif
 };
 
 struct rbd_options {
@@ -109,9 +113,11 @@ static int _fio_setup_rbd_data(struct thread_data *td,
 	if (!rbd)
 		goto failed;
 
-	/* init fd: -1 */
+#ifdef CONFIG_RBD_POLL
+	/* add for poll, init fd: -1 */
 	rbd->fd = -1;
-    
+#endif
+
 	rbd->aio_events = calloc(td->o.iodepth, sizeof(struct io_u *));
 	if (!rbd->aio_events)
 		goto failed;
@@ -135,7 +141,6 @@ failed:
 
 }
 
-#define EVENT_SOCKET_TYPE_EVENTFD 2 /* add for rbd poll */
 static int _fio_rbd_connect(struct thread_data *td)
 {
 	struct rbd_data *rbd = td->io_ops_data;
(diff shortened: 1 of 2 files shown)


engines/rbd.c
```diff
@@ -29,6 +33,7 @@ struct rbd_data {
 	rbd_image_t image;
 	struct io_u **aio_events;
 	struct io_u **sort_events;
+    int fd; /* add for poll */
```
axboe: You used spaces, please use tabs.

engines/rbd.c
```diff
@@ -127,6 +135,7 @@ static int _fio_setup_rbd_data(struct thread_data *td,
 
 }
 
+#define EVENT_SOCKET_TYPE_EVENTFD 2 /* add for rbd poll */
```
axboe: Why do we have to define this? If the library is new enough to have support for image notification, doesn't it also include this define in the header?

engines/rbd.c
```diff
@@ -188,8 +197,26 @@ static int _fio_rbd_connect(struct thread_data *td)
 		log_err("rbd_open failed.\n");
 		goto failed_open;
 	}
+
+	/* add for rbd poll */
+    rbd->fd = eventfd(0, EFD_NONBLOCK);
+    if (rbd->fd < 0)
```
axboe: Spaces again, and you are using mixed styles of braces. Please the starting { on the same line as the if statement.

engines/rbd.c
```diff
@@ -188,8 +197,26 @@ static int _fio_rbd_connect(struct thread_data *td)
...
+	r = rbd_set_image_notification(rbd->image, rbd->fd, EVENT_SOCKET_TYPE_EVENTFD);
+	if (r < 0) {
+		log_err("rbd_set_image_notification failed.\n");
+		goto failed_notify;
+	}
+	
 	return 0;
 
+failed_notify:
+    close(rbd->fd);
```
axboe: Spaces...

engines/rbd.c
```diff
@@ -205,6 +232,13 @@ static void _fio_rbd_disconnect(struct rbd_data *rbd)
 	if (!rbd)
 		return;
 
+    /* close eventfd */
+    if (rbd->fd >= 0)
```
axboe: Spaces and braces.

engines/rbd.c
axboe: enum {                                                                          
  EVENT_TYPE_PIPE = 1,                                                          
  EVENT_TYPE_EVENTFD = 2                                                        
}; 

So it is there. I'd propose adding a configure check for whether set_image_type(..., EVENT_TYPE_EVENTFD) works or not, and use it in rbd.c if that is the case.

engines/rbd.c
liupan1111: Thanks. I will take care coding style next time.

One suggestion, how about adding a shell to check the coding style in fio package? If you agree, I could give help, draft a "check_format.sh", and let you review.

engines/rbd.c
liupan1111: I agree, please take a look at my new commit.

axboe:
This looks better. Now please collapse the three patches into one proper patch, I don't want to pull your intermediate breakage into my history.

liupan1111:
ok, I will close this pull request and create a new one with final changes.

axboe:
Great, thanks!

PR #259, opened 2016-10-19
use poll() and rbd_poll_io_events to speed up io retrieval.

liupan1111:
Hi,
In Aug. 2015, a new function rbd_poll_io_events is added in Ceph. I use it to speed up io retrieval. I tested by using P3700 SSD and ceph_jewel(bluestore), find the iops and latency has 10% improvement.

### DIFF
--- a/configure
+++ b/configure
@@ -1335,6 +1335,32 @@ fi
 echo "Rados Block Device engine     $rbd"
 
 ##########################################
+# check for rbd_poll
+rbd_poll="no"
+if test "$rbd" = "yes"; then
+cat > $TMPC << EOF
+#include <rbd/librbd.h>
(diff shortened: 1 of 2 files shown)


liupan1111:
Hi Axboe,

I also gave a comment in #258, I suggest we add a "check_format.sh" to check and adjust the coding style by using the tool "astyle". If you agree, I could give this help.

axboe:
Pulled this in, with a cleanup commit on top.

axboe:
As for something for checking the format, if you want to work on it, go ahead. I try not to be too picky about format, but I will complain about gross violations like tabs, braces, or style that isn't consistent.

liupan1111:
not picky, I appreciate. Thanks

PR #262, opened 2016-10-28
iolog: add support for 'replay_scale'

sitsofe:
Example patch for adding replay scaling to iolog replay and also fixes up a 64 bit number truncation issue in `iolog.c` too.

Not really a final version but more to get comment/ideas on how to add this in such a way that  `ipo_bytes_align()` can also be shared between `blktrace.c` and `iolog.c`.

### DIFF
--- a/blktrace.c
+++ b/blktrace.c
@@ -221,7 +221,7 @@ static void ipo_bytes_align(struct thread_options *o, struct io_piece *ipo)
 	if (!o->replay_align)
 		return;
 
-	ipo->offset &= ~(o->replay_align - 1);
+	ipo->offset &= ~(o->replay_align - (uint64_t)1);
 }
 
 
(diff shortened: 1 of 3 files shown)


sitsofe:
Updated branch to final version which should be suitable for pulling.

sitsofe:
Obvious bug spotted and corrected: previous version only aligns if also scaling an fio iolog replay whereas blktrace code allowed aligning without scaling. Latest revision makes fio iolog replay match blktrace replay align behaviour.

sitsofe:
Rebasing branch against current master to make any potential merge easier. Jens - is there anything in these patches that you object to?

axboe:
No, looks reasonable to me. Sometimes I just need a little reminder, if I'm been busy on other projects.

axboe:
Pulled, thanks.

sitsofe:
I thought you might be in "merge mode" because you recently pulled other branches (and I didn't want to miss the window) so I was just rechecking :-)

Thanks Jens!

axboe:
Sometimes that is exactly how it happens, and the pulls/merges tend to be bundled. In any case, always feel free to re-check or ping me, if it's been more than a few days. Chances are it's off the radar at that point, and I might only find it much later, or not at all...

PR #264, opened 2016-11-01
Fix minor spelling mistakes in HOWTO document

szaydel:
Over the years as the document is revised it incurred spelling errors as byproduct. I fixed some, probably not all. I also noticed that it might be good to clean-up some inconsistent casing, for example `IO` vs. `io` in places and `Fio` vs. `fio`. I noticed that while sentences always start with an Upcase `F`, when fio is the first word in the sentence, in some cases `Fio` is used in the middle of sentences, but in most cases I see `fio`. Getting consistent casing may not be a bad idea. I am happy to clean this up if we agree that it is desirable. Same is true for `IO` vs. `io`.


axboe:
Pulled, thanks.

szaydel:
Yeah, no problem. Glad to do a small part to help. If you'd like to see some consistency, I am more than happy to walk through and standardize things like `IO` instead of `io`, consistently using `Fio` or `fio` when not in the beginning of sentence, etc. If you don't mind seeing that happen, let me know. Thanks!

axboe:
That would be great, thanks.

szaydel:
Sounds good. Once I have changes will do pull req.

PR #267, opened 2016-11-12
Fix a crash: when the user create a rbd image with size zero, and

liupan1111:
specify the "size" larger than zero in fio option, rbd will crash.

For ceph image, if the size is zero, just report an error and quit.

Signed-off-by: Pan Liu <pan.liu@istuary.com>


liupan1111:
@axboe , please help take a look, thanks.

axboe:
Pulled - I re-applied, fixing up your formatting and the commit style. Thanks.

liupan1111:
thank you！

PR #268, opened 2016-11-14
Fix memory leak on tmp_buf

ColinIanKing:
tmp_buf is allocated but not free'd, causing a minor memory leak

Signed-off-by: Colin Ian King <colin.king@canonical.com>


axboe:
Pulled, thanks.

Issue #194, opened 2016-06-21
rate=[int] Windows


SimonHaywood:
Hi!

I'm testing performance of Win2008 Server against a NAS. I want to do some rate limited read tests. 

However, when I set the rate=[int] I see no data being read (and the estimated time to completion increases). If I do rate_limit=[int], I get surprising results (but that may be unrelated - I can't tell at this stage).

I've tested my syntax on an OSX machine reading from internal SSD - but I can't connect the OSX machine to the NAS.

So, is there something going wrong with rate=[int] on the Windows implementation of FIO, or is it me!?

Thanks!

Simon

sitsofe:
@SimonHaywood : is rate_limit even an fio parameter? I can't see it in https://github.com/axboe/fio/blob/master/HOWTO and Linux fio says "unrecognized option" if you attempt to use it. It might help if you can post the job file...

sitsofe:
@SimonHaywood : I've just tried the following on Windows 2012R2:

`& 'C:\Program Files\fio\fio.exe' --rw=read --filename \\.\PhysicalDrive1 --name=go --runtime=1h --time_based --rate=1M`

And it correctly limited the speed to around 1MByte/s (Windows' resource manager confirms the throughput was restricted to this to). In addition to my rate_limit question above, what fio job file were you using and which version of fio?

chrispeters1:
Hello,

I'm also having an issue with rate=[int] in Windows server 2012 R2. 

The rate option works correctly for about 1 min then it stops and disk performance goes to full speed. Below are the details of my fio jobfile any help is appreciated. 

[global]

time_based
ramp_time=10s
runtime=14400s
size=859G
bs=128k
ba=4k
rw=write
iodepth=4
rate=64M
ioengine=windowsaio
direct=1
numjobs=1
thread

[job1]

directory=/io.tst
filename=io.tst

Thanks,

Chris

sitsofe:
@chrispeters1 which version of fio?

chrispeters1:
I tried using a few different versions from http://bluestop.org/fio/

2.13
2.11
2.9

Then I downloaded a version 2.14 zip from here https://github.com/axboe/fio a couple days ago and compiled my own with Cygwin. With all versions I seem to have the same issue. I also tried to simplify the jobfile by commenting out a few options and also tried rate_iops instead but still get the same outcome.

sitsofe:
[...]

Issue #266, opened 2016-11-09
Windows FIO exit before time with high runt value


jasonjorge:
I've been running the latest windows version (2.15 x64) with the following command line, downloaded from Bluestop.  Every time I attempt to run the following workload FIO will always exit after ~25mins, and the IOPs reported are 0.

fio --filename="\\.\PhysicalDrive1" --time_based=1 --ramp_time=5s --runtime=7200s --direct=1 --bs=4KB --iodepth=1 --ioengine=windowsaio --thread=1 --output="4KB_RandRd_QD=1_1.out" --name="4KB RandRd QD=1" --rw=randread

![image](https://cloud.githubusercontent.com/assets/8410864/20146097/3700b44a-a668-11e6-9344-60aa085b39d8.png)

I looked in the output file, and the runt value is runt=18446744071562070msec

This same command line worked correctly with version 2.13. 

When we run with a runtime value of less than half hour the test completes successfully.  

for example --runtime=300s completes in runt=300068msec

4KB RandRd QD=1: (g=0): rw=randread, bs=4K-4K/4K-4K/4K-4K, ioengine=windowsaio, iodepth=1
fio-2.15
Starting 1 thread

output results.
4KB RandRd QD=1: (groupid=0, jobs=1): err= 0: pid=2268: Wed Nov 9 16:12:46 2016
  read : io=701144KB, bw=565B/s, iops=0, runt=18446744071562070msec
    slat (usec): min=7, max=6563, avg=16.80, stdev=22.67
    clat (usec): min=864, max=91597, avg=12229.06, stdev=4148.19
     lat (usec): min=881, max=91614, avg=12245.86, stdev=4148.23
    clat percentiles (usec):
     |  1.00th=[ 4128],  5.00th=[ 5856], 10.00th=[ 7072], 20.00th=[ 8768],
     | 30.00th=[10048], 40.00th=[11072], 50.00th=[12096], 60.00th=[13120],
     | 70.00th=[14144], 80.00th=[15552], 90.00th=[17280], 95.00th=[18560],
     | 99.00th=[27776], 99.50th=[29056], 99.90th=[30336], 99.95th=[30592],
     | 99.99th=[34048]
    lat (usec) : 1000=0.01%
    lat (msec) : 2=0.01%, 4=0.83%, 10=29.14%, 20=67.78%, 50=2.23%
    lat (msec) : 100=0.01%
  cpu          : usr=0.05%, sys=0.14%, ctx=0, majf=0, minf=0
[...]

PR #271, opened 2016-11-20
fio: Fix (unsigned) integer overflow issues

sitsofe:
This patch tries to address issues found when compiling fio with clang's
undefined behaviour sanitizer on 32 bit Linux. While this generated many
false positives it did identify a number of subtle issues when
rates/amounts/time calculations went above 32 bit values.

The following job was used:
```
[global]
time_based
runtime=50m
bs=256k
rw=read
rate=512M
ioengine=null
direct=1
iodepth=8
thread
numjobs=10

[null]
size=1G
```
This patch should also fix https://github.com/axboe/fio/issues/194 and
https://github.com/axboe/fio/issues/266 because 64 bit Windows is an
LLP64 platform so ints and longs still default to 32 bits.

This doesn't try to fix all potential truncation issues - turning on compiler warnings related to them generates large numbers of false positives so work would have to start to eliminate those...

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -1864,8 +1864,8 @@ static void dump_td_info(struct thread_data *td)
 /*
  * Run over the job map and reap the threads that have exited, if any.
  */
-static void reap_threads(unsigned int *nr_running, unsigned int *t_rate,
-			 unsigned int *m_rate)
+static void reap_threads(unsigned int *nr_running, uint64_t *t_rate,
+			 uint64_t *m_rate)
 {
 	struct thread_data *td;
 	unsigned int cputhreads, realthreads, pending;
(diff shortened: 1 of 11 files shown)


sitsofe:
Withdrawing due to 64 bit build warnings...

axboe:
You are so close, though. I think all you need is to change the calc_rate() last type to be a uint64_t pointer, and change the two cconv cases to use the __ version of cpu_to_le64(), __cpu_to_le64(). It's fine to use unsigned long long and uint64_t interchangably.

sitsofe:
Thanks for the hints Jens - I'll try and fix your suggestions and repost.

PR #279, opened 2016-12-12
iolog: Ignore re-add/re-open with replay_redirect

sitsofe:
The HOWTO entry for replay_redirect says:
"Replay_redirect causes all IOPS to be replayed onto
the single specified device regardless of the device it was
recorded from. i.e. replay_redirect=/dev/sdc would cause all
IO in the blktrace or iolog to be replayed onto /dev/sdc.
This means multiple devices will be replayed onto a single
device, if the trace contains multiple devices."

Unfortunately trying to replay the following iolog (saved as
redirtest.fiolog):
fio version 2 iolog
/dev/sdb add
/dev/sdc add
/dev/sdb open
/dev/sdc open
/dev/sdb read 0 512
/dev/sdc read 512 512
/dev/sdb close
/dev/sdc close

with an fio line like
fio --name=redirtest --replay_redirect=/dev/nullb0 --read_iolog=redirtest.fiolog
winds up hitting this assertion:
fio: ioengines.c:413: td_io_open_file: Assertion `!fio_file_open(f)' failed.
which occurs because after redirection is applied /dev/nullb0 is added and
opened twice.

Fix this by ignoring subsequent adds or opens for the same file during
iolog replay setup when replay_redirect is in effect.

### DIFF
--- a/iolog.c
+++ b/iolog.c
@@ -109,6 +109,11 @@ static int ipo_special(struct thread_data *td, struct io_piece *ipo)
 
 	switch (ipo->file_action) {
 	case FIO_LOG_OPEN_FILE:
+		if (td->o.replay_redirect && fio_file_open(f)) {
+			dprint(FD_FILE, "iolog: ignoring re-open of file %s\n",
+					f->file_name);
+			break;
+		}
 		ret = td_io_open_file(td, f);
 		if (!ret)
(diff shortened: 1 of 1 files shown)


axboe:
Thanks, applied!

PR #280, opened 2016-12-22
rbd: remove duplicate _fio_rbd_connect

liupan1111:
Signed-off-by: Pan Liu <pan.liu@istuary.com>

### DIFF
--- a/engines/rbd.c
+++ b/engines/rbd.c
@@ -36,6 +36,7 @@ struct rbd_data {
 	struct io_u **aio_events;
 	struct io_u **sort_events;
 	int fd; /* add for poll */
+	bool connected;
 };
 
 struct rbd_options {
@@ -110,6 +111,8 @@ static int _fio_setup_rbd_data(struct thread_data *td,
 	rbd = calloc(1, sizeof(struct rbd_data));
 	if (!rbd)
 		goto failed;
+
+	rbd->connected = false;
 
 	/* add for poll, init fd: -1 */
 	rbd->fd = -1;
@@ -529,6 +532,10 @@ failed:
 static int fio_rbd_init(struct thread_data *td)
 {
 	int r;
+	struct rbd_data *rbd = td->io_ops_data;
+	if (rbd->connected) {
+		return 0;
+	}
 
 	r = _fio_rbd_connect(td);
 	if (r) {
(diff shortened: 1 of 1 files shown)


axboe:
You'll have to explain this in more detail, what problem does the patch solve? On the simpler style side of things, don't do crazy casts. Setup local 'rbd' variable the places where you need to access it.

liupan1111:
Sorry for the confuse:
1) In fio_rbd_setup, it will call _fio_rbd_connect first, and call rbd_stat to get image size, and disconnect no matter get image successful or not. And call _fio_rbd_connect again, in fio_rbd_init.
    If client is already connect to cluster and get image size successful in fio_rbd_setup, we don't need to disconnect and connect again. For each connect operation, it will create nearly 20 work threads of librbd, and destroy them when disconnect.   When fio jobs are many, it is time consuming.

2) I will remove unreasonable casts, Thanks!

liupan1111:
@axboe removed casts, please take a look.

engines/rbd.c
```diff
@@ -529,6 +532,11 @@ static int fio_rbd_queue(struct thread_data *td, struct io_u *io_u)
 static int fio_rbd_init(struct thread_data *td)
 {
 	int r;
+	struct rbd_data *rbd;
+	rbd = (struct rbd_data*)(td->io_ops_data);
```
axboe: You don't have to cast it, it's a void pointer. Just do:

struct rbd_data *rbd = td->io_ops->data;

like everywhere else in the code.

engines/rbd.c
```diff
@@ -529,6 +532,11 @@ static int fio_rbd_queue(struct thread_data *td, struct io_u *io_u)
...
+	if(rbd && rbd->connected) {
```
axboe: Can 'rbd' ever be NULL here?

Also style - please follow the general style in the file. In this csae, space after the 'if'.

axboe:
Added some review comments.

engines/rbd.c
liupan1111: cool.. right

engines/rbd.c
liupan1111: seems impossible to be NULL. Add if(rbd &&...) just want to make code analysis tool clean. Do you prefer to remove this check？

liupan1111:
@axboe  modified as your suggestion. Thanks!

axboe:
I pulled this, but I still had to fix up some of your styling. No braces for single line if's, and always leave an empty line between variables and code.

Really, this goes for ANY project you want to contribute to. Follow the local style, it's important. It makes the code more easily readable and maintainable.

liupan1111:
er... yes, I'm so sorry always leaves you coding style issue...

PR #283, opened 2017-01-05
Add arm64 hardware assisted crc32c support

weixiaowilliam:
HW assisted crc32c get ~ x39.6 speedups on ARM Cortex-A57@2GHz

Auto-detect whether to use hw assisted crc32c.
If the hardware assisted crypto is available, always use it.
Otherwise, fallback to software.


axboe:
Awesome! Can we drop crc32c-arm64? The intel variant only exists because of history, crc32c should be use hardware assist, if available.

I'm going to pull this, but would be great if you could send a follow to fix up that part.

PR #285, opened 2017-01-17
Drop crc32c-arm64 option

weixiaowilliam:
Crc32c will use ARM64 hardware assist, if available.

Jira: ENTLLT-359
Change-Id: Ibdf88f14bd4669ed68a93078c43eb5c8412da543


axboe:
Pulled, thanks.

PR #288, opened 2017-01-20
ignore documentation output directory

termim:


axboe:
Will pull, thanks. Can we also ensure that make clean gets rid of any auto-generated content?

PR #282, opened 2016-12-26
Convert documentation to Sphinx

termim:
Please consider this conversion documentation files to Sphinx.
The result could be seen here: http://fio.readthedocs.io/en/sphinx-doc/

PR #290, opened 2017-01-20
doc: fix man page headings

sitsofe:
Sphinx skips over the top level heading when generating man page output
so add a top level heading for it to skip over and skip showing version
information.

### DIFF
--- a/doc/fio_man.rst
+++ b/doc/fio_man.rst
@@ -1,9 +1,7 @@
 :orphan:
 
-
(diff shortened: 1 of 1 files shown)


sitsofe:
This tries to address the issue mentioned in https://github.com/axboe/fio/pull/282#issuecomment-273975318 . https://github.com/axboe/fio/pull/291 also addresses the same problem .

PR #302, opened 2017-02-02
iolog replay_align fix

dylanfairchild:
The previous code was immediately returning on a positive alignment value, and accepting only alignment to zero.

I believe this should be doing the opposite.


sitsofe:
Gulp. As I'm the one who put the code in that introduced the error this fix looks necessary.

axboe:
yeah, that looks like an obviously correct fix. Applied.

Issue #298, opened 2017-01-31
testing a specific portion of file / device


celesteking:
It would be wonderful to have an option to specify the range of testing. Right now, I have to resort to DM in order to carve the blockdev into segments. 

`--size` works only for 0..$size, and it seems it's impossible to specify start offset.

axboe:
There's an --offset option too.

celesteking:
Yes, my bad, it indeed exists. Haven't read the manual -- it's way too long :)

axboe:
If only there existed a mechanism by which someone, who's reading a document, could search for specific phrases or words. Then you might have found it :-)

celesteking:
I actually did that, I searched for "range" and "portion". I stumbled upon `size`, and kept looking for more. It might worth adding a note to `size` option with reference to `offset` and vice versa.

axboe:
I'm mostly kidding, you used the phrase 'offset' in your issue here and that would have found it. I'm perfectly happy to apply patches to improve the documentation, so please do send that in.

PR #304, opened 2017-02-04
doc: minor HOWTO fixes

sitsofe:
- Add a note to to size/offset to see the other for constraining I/O to
  a range (requested in #298 ).
- Minor grammar and typo fixes.
- Update example fio ETA status to match what modern fio would likely
  display.
- Explain f= is open files and note that trim may be displayed.


axboe:
Pulled, thanks Sitsofe.

Issue #305, opened 2017-02-04
Runtime failure on macOS 10.11 when built using Xcode 8


JCount:
Building on macOS 10.11 using Xcode 8 completes without error, however there is a runtime error of:
```
dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
  Referenced from: /usr/local/Cellar/fio/2.17/bin/fio
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _clock_gettime
  Referenced from: /usr/local/Cellar/fio/2.17/bin/fio
  Expected in: /usr/lib/libSystem.B.dylib
```
This occurs because the Xcode 8 SDK declares clock_gettime for both 10.11 and 10.12, even though it is only actually implemented on 10.12. This means that a simple test to see if the symbol is declared is not sufficient when building for macOS 10.11.

Currently, when building with/for this platform, one way to work around this problem seems to be to modify the configure script to ensure that the clock_gettime probe will fail.

I am happy to try and provide any other information you may need.

sitsofe:
@JCount Ugh, not sure I like what Apple introduced here. Looking around many projects are going to suffer this and homebrew seems to workaround it by using the -no_weak_imports linker flag (https://github.com/Homebrew/homebrew-core/issues/3727 ). I would suggest that you change the formula to modify the configure script to add this flag. Does that workaround the problem?

ilovezfs:
Nope, we don't and can't use that flag as a solution since it breaks Xcode 7.

sitsofe:
@ilovezfs  In https://github.com/Homebrew/brew/pull/721 it isonly set whe targeting Xcode 8 and it's injected via LDFLAGS. Perhaps the best thing is to teach the Makefile to inject existing LDFLAGS in the Makefile for the Darwin platform: https://github.com/axboe/fio/blob/master/Makefile#L178 (see what the BSDs do above).

ilovezfs:
@sitsofe that was subsequently reverted. Trust me. I've been the one doing most of the clock_gettime PRs for Homebrew.

ilovezfs:
@sitsofe these are what proper solutions look like:

https://github.com/pyca/cryptography/pull/3354
https://github.com/jsquyres/libfabric/commit/469779935f9995b88a6f486fe3c4c4496989af36
https://github.com/jemalloc/jemalloc/commit/eb29d7ec0e2c4994e10ec40d42265a86e569500c
[...]

PR #309, opened 2017-02-13
configure: try to disable weak linking on OSX

sitsofe:
macOS 10.12 introduced support for clock_gettime() but when compiling
with Xcode 8 (or later) and targeting 10.11 (or older) the clock_gettime
symbol is now found at configure/compile time but referenced weakly.
Running the created binary on 10.11 results in a "dyld: lazy symbol
binding failed: Symbol not found: _clock_gettime" error.

This drama has played out across other projects:
- https://github.com/Homebrew/homebrew-core/issues/2674
- https://github.com/curl/curl/issues/1069
- https://svn.filezilla-project.org/filezilla?view=revision&revision=7824OO
- https://github.com/mesonbuild/meson/pull/949
- https://github.com/Homebrew/homebrew-core/issues?utf8=%E2%9C%93&q=label%3Aclock_gettime%20

Options for working around this issue include:
1. Turn references to functions that might be weak into compilation
   errors (-Werror=partial-availability).
2. Disable visibility of functions that might be weak at compile link
   time (-Wl,-no_weak_imports).
3. Change configure tests to directly check the targeted OSX version and
   fail the test if it is too old to have the required function.
4. Make an empty function declaration marked with
   __attribute__((weak_import)) that is used if the symbol would
   otherwise be unavailable. It is then possible to check if the symbol
   is NULL at runtime to determine availability.

This commit does 2. when targeting the Darwin platform after checking if
the compiler/linker work with that option. It also changes the clockid_t
configure test to no longer use clock_gettime() because targeting OSX
10.11 when using the 10.12 SDK results in having the clockid_t typedef
without having the clock_gettime function. The combination of these
fixes should solve https://github.com/axboe/fio/issues/305 .

### DIFF
--- a/configure
+++ b/configure
@@ -268,6 +268,17 @@ Darwin)
   if test -z "$cpu" && test "$(sysctl -n hw.optional.x86_64)" = "1"; then
     cpu="x86_64"
   fi
+  # Error at compile time linking of weak/partial symbols if possible...
+cat > $TMPC <<EOF
+int main(void)
+{
+  return 0;
+}
+EOF
+  if compile_prog "" "-Wl,-no_weak_imports" "disable weak symbols"; then
+    echo "Disabling weak symbols"
+    LDFLAGS="$LDFLAGS -Wl,-no_weak_imports"
+  fi
   ;;
 SunOS)
   # `uname -m` returns i86pc even on an x86_64 box, so default based on isainfo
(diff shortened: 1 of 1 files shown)


axboe:
Nice work Sitsofe, and a most excellent commit log as well. Pulled, thanks!

PR #310, opened 2017-02-14
Add make test and add OSX builds to travis

sitsofe:
- Add a "test" target to the Makefile which runs two simultaneous null ioengine jobs for a short time.
- Make travis run two parallel make jobs letting it finish faster.
- Add two OSX builds to travis (with a third build commented out). This is slightly controversial because at peak time travis is heavily backed up with OSX builds from other projects. This can mean the travis results don't return in the usual less than 4 minutes but can instead take up to 2 hours(!) because while the Linux jobs were scheduled quickly the OSX ones weren't. If this sounds problematic, perhaps I can rework this commit to just comment out the travis OSX builds...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: c
+os:
+  - linux
 compiler:
   - clang
   - gcc
@@ -6,5 +8,5 @@ env:
   global:
     - MAKEFLAGS="-j 2"
 before_install:
-  - sudo apt-get -qq update
-  - sudo apt-get install -qq -y libaio-dev libnuma-dev libz-dev
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get -qq update; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install -qq -y libaio-dev libnuma-dev libz-dev; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: c
 compiler:
   - clang
   - gcc
+env:
+  global:
+    - MAKEFLAGS="-j 2"
 before_install:
   - sudo apt-get -qq update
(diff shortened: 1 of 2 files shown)


axboe:
Will the potential OSX delay also impact the other reports? Say I break the linux build for some reason, usually travis would let me know asap.

But I love where this is going in general, would be great to have more automated/unit testing done automatically.

sitsofe:
Do you mean via email or by ticks/colours on GitHub/on the Travis website? It's definitely possible to see the Linux results happily finished and saying their result long before the OS X ones if you drill down on the Travis site but I think the interim result is mentioned on GitHub for example...

sitsofe:
travis

Example screenshot of Travis with completed Linux builds but pending OSX builds.

axboe:
I mean email, will it email me immediately if one test fails, or wait until all have been run and send a summary of that? If the former, than that's fine.

sitsofe:
Sadly it looks like it waits until all the sub-builds have finished before sending a results email (the above build only just finished and I'm still waiting for an email).

sitsofe:
OK by default travis doesn't email you if your previous build passes and your current build also passes. So instead of the above I did a purposely failing build on another branch and despite the Linux builds finishing close to two hours before the OS X builds, an email wasn't sent until after all builds had completed. Did you want me to break these patches up a bit more and comment the OS X travis builds out by default?

axboe:
Yes, I think it'd be useful to break them up, so we have an OSX test enable commit by itself. It's not necessarily a show stopper for merging this, but it is a bit annoying if the OSX test ends up delaying the rest by hours.

sitsofe:
OK the adding of the OSX builds is now in a commit of its own which should make it easier to revert. The only other thing of note is that `make test` depends on the fio binary so it will force the fio binary to be rebuilt if it is out of date. Is this incorrect and should it not depend on anything?

axboe:
That dependency sounds fine, we'd want to have a matching binary built before running tests if it isn't already uptodate.

Issue #303, opened 2017-02-04
How to avoid error message "Multiple writers may overwrite blocks that belong to other jobs. This can cause verification failures."


ashokkumaralluri:
Fio version: fio-2.9
**FIO command:**
fio  --name=integrity_test  --nrfiles=64 --ioengine=libaio  --iodepth=50 --rw=randwrite  --bs=4k --filesize=16m  --numjobs=5 --runtime=1500  --group_reporting --do_verify=1  --directory=/tmp  --continue_on_error=none  --verify=sha1 --fallocate=0 

**Output**

```
Multiple writers may overwrite blocks that belong to other jobs. This can cause verification failures.
integrity_test: (g=0): rw=randwrite, bs=4K-4K/4K-4K/4K-4K, ioengine=libaio, iodepth=50
...
fio-2.9
Starting 5 processes
integrity_test: Laying out IO file(s) (64 file(s) / 64MB)
integrity_test: Laying out IO file(s) (64 file(s) / 64MB)
integrity_test: Laying out IO file(s) (64 file(s) / 64MB)
integrity_test: Laying out IO file(s) (64 file(s) / 64MB)
integrity_test: Laying out IO file(s) (64 file(s) / 64MB)
Jobs: 5 (f=320)
integrity_test: (groupid=0, jobs=5): err= 0: pid=21812: Fri Feb  3 22:41:18 2017
  read : io=327680KB, bw=231086KB/s, iops=57771, runt=  1418msec
    slat (usec): min=0, max=63373, avg=26.36, stdev=637.56
    clat (usec): min=6, max=653473, avg=2698.35, stdev=23139.37
     lat (usec): min=7, max=653654, avg=2724.74, stdev=23174.56
    clat percentiles (usec):
     |  1.00th=[  358],  5.00th=[  358], 10.00th=[  358], 20.00th=[  378],
     | 30.00th=[  418], 40.00th=[  636], 50.00th=[  732], 60.00th=[  788],
     | 70.00th=[ 1320], 80.00th=[ 1528], 90.00th=[ 1880], 95.00th=[ 1992],
     | 99.00th=[31872], 99.50th=[85504], 99.90th=[460800], 99.95th=[518144],
     | 99.99th=[593920]
  write: io=327680KB, bw=1103.5MB/s, iops=282482, runt=   290msec
    slat (usec): min=1, max=89, avg= 3.02, stdev= 1.40
    clat (usec): min=0, max=55136, avg=710.80, stdev=2115.96
[...]

Issue #187, opened 2016-06-01
Verify_Async threads not verifying


shivapahwa:
Hi All,

  I was trying to use fio to verify asyncronously, ran a simple test but do not see any reads going to the device. 

./fio --name=Async_verify --filename=/dev/nvme0n1 --rw=randwrite --ioengine=libaio --iodepth=32 --bs=128k --numjobs=1 --verify_pattern=0xdead --do_verify=1 --buffered=0 --runtime=300 --time_based=1 --verify_async=64 --offset=0 --size=83748938934B --verify=pattern

Even if i set the number of threads to verify_async to be 64 i dont see any reads going to the device. I am using blktrace and blkparse to verify the same.

Please let me know if i am missing any of the parameters?

axboe:
Don't think you are missing anything. What version of fio are you running? Just tried your job file here, and it writes and verifies just fine:

`Async_verify: (groupid=0, jobs=1): err= 0: pid=31465: Wed Jun  1 08:08:22 2016
  read : io=27769MB, bw=2307.6MB/s, iops=18460, runt= 12034msec
[...]
  write: io=32368MB, bw=4986.7MB/s, iops=39892, runt=  6491msec
[...]
Disk stats (read/write):
  nvme2n1: ios=219564/258944, merge=0/0, ticks=16320/126500, in_queue=142750, util=99.31%
`

shivapahwa:
I downloded the latest code from github, version 2.11. The command output does not show Async_verify job details.

axboe:
I just copied some of it. It shows reads and writes, and while it was running, I saw state change between 'w' and 'V'. I also looked at iostat, and it was doing both reads and writes.

My device is smaller than yours, though. You might want to consider setting verify_backlog to force the verifies to be run more frequently. Also see verify_backlog_batch.

shivapahwa:
For now i am using  verify_backlog to force the verifies, I ran the same workload by increasing the duration to 1 hr, even then i did not see any reads to the device. let me run it all over again and send you the output.

shivapahwa:
Output from fio:

[...]

PR #313, opened 2017-02-17
Verify warn

sitsofe:
Try to make the "multiple jobs overwriting same file" warning more specific to address https://github.com/axboe/fio/issues/303 and add a new warning about the verify phase not running when using runtime + time_based (see https://github.com/axboe/fio/issues/187#issuecomment-223633831 and https://www.spinics.net/lists/fio/msg04990.html ).

These commits are kind of for review: for example if the user is using rw=rw / rw=randrw with their verify they may also get inline verifies because if at the point an area is read if it matches an area previously written by the same job it is verified. I wasn't sure if the skipping the warning in this situation when the user has asked for do_verify was worth it...

### DIFF
--- a/init.c
+++ b/init.c
@@ -731,9 +731,14 @@ static int fixup_options(struct thread_data *td)
 		o->size = -1ULL;
 
 	if (o->verify != VERIFY_NONE) {
-		if (td_write(td) && o->do_verify && o->numjobs > 1) {
-			log_info("Multiple writers may overwrite blocks that "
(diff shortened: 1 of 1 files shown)


sitsofe:
In the end I changed the patch to not fire the "time_based writes use all of runtime so there is no verify" warning when any type of verify (e.g. due to reads or verify_backlog) takes place.

@axboe any objections to these two commits?

sitsofe:
@axboe just a gentle prod for review of this pull request...

PR #316, opened 2017-02-20
appveyor: add CI building of the Windows version of fio

sitsofe:
Add AppVeyor (https://www.appveyor.com/ )  continuous integration build of the Windows port (you'll need an AppVeyor account to get going) . The 64 bit and 32 bit versions of the Window's binary are currently built. This also builds the Windows installer and the resulting MSI can be downloaded. Example output can be seen on https://github.com/sitsofe/fio/commits/appveyor and https://ci.appveyor.com/project/sitsofe/fio/build/1.0.37/job/c3i3jw3ug8bdt5pq . At the moment builds seem to be scheduled in under a minute.

Note that AppVeyor limits you to one build at a time but each build seems to finish after about one minute so the total time to get a result is about 3 minutes.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,27 @@
+clone_depth: 50
+environment:
+  MAKEFLAGS: -j 2
+  matrix:
+    - platform: x86_64
+      BUILD_ARCH: x64
+      CYG_ROOT: C:\cygwin64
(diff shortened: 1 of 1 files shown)


axboe:
Very cool! Thanks for doing this, merged.

PR #320, opened 2017-02-24
fio: fix overflow trying to use 'd' suffix

sitsofe:
Fix overflow that happened when using the 'd' time suffix on platforms
with 32 bit longs which led to a time of less than 10 minutes being
used.


axboe:
Thanks, pulled.

PR #321, opened 2017-02-26
configure: add generic pshared mutex test

sitsofe:
Add a feature test to configure to determine if a platform has
interprocess synchronization primitives that work with threads and
remove the hardcoded enablement in the individual os/os-* headers. This
should work better for platforms where support was overlooked or didn't
exist at some point in the past but has since been added.

This is based off work by the Boost project
(https://github.com/boostorg/interprocess/blob/boost-1.63.0/include/boost/interprocess/detail/workaround.hpp#L45
) that found that OSX 10.6 and Cygwin are buggy even though
_POSIX_THREAD_PROCESS_SHARED says this feature should be work on them.

Note: I've tested this in OSX, Linux and WIndows where it does the right thing but I don't know if it works on other OSes where it used to be enabled (Solaris, HPUX, Android) or those where it was never enabled (e.g. I gather support of pshared was recently added to FreeBSD - https://svnweb.freebsd.org/ports/head/benchmarks/fio/files/patch-os_os-freebsd.h?revision=428489&view=markup&pathrev=428489 ). @kusumi does it at least work for you ?


kusumi:
It compiled on FreeBSD, but I'm suspecting if there's any reason fio normally prefers to make compile-time based decisions (which is more practical in some ways) rather than using any sort of macros/etc that are _supposed_ to be portable.

Why did you changed the one you originally had ? That one was closer to what fio generally does.

sitsofe:
I think it was out of desire to try and fix this problem in a more general way rather than hardcoding a fix just for OSX (https://github.com/sitsofe/fio/commit/4e26e28fafc545a6df3733e003de42e0223fabce is what you're referring to) after I saw that FreeBSD might need something similar too. Which version of FreeBSD are you using and did you end up getting process job support or does `make test` still warn about needing to use `--thread`?

kusumi:
[root@]/tmp/fio# uname -r
11.0-RELEASE-p1
[root@]/tmp/fio# git log -2 --oneline
f8a9e362 configure: add generic pshared mutex test
c9057434 Fio 2.18
[root@]/tmp/fio# gmake clean
FIO_VERSION = fio-2.18-1-gf8a9
[root@]/tmp/fio# ./configure | grep pshared
POSIX pshared support         yes
[root@]/tmp/fio# gmake -j8 > /dev/null
FIO_VERSION = fio-2.18-1-gf8a9
[root@]/tmp/fio# gmake test
./fio --minimal --ioengine=null --runtime=1s --name=nulltest --rw=randrw --iodepth=2 --norandommap --random_generator=tausworthe64 --size=16T --name=verifynulltest --rw=write --verify=crc32c --verify_state_save=0 --size=100M
3;fio-2.18-1-gf8a9;nulltest;0;0;4667724;4667724;1166931;1000;0;0;0.000000;0.000000;0;336;0.270265;0.952134;1.000000%=0;5.000000%=0;10.000000%=0;20.000000%=0;30.000000%=0;40.000000%=0;50.000000%=0;60.000000%=0;70.000000%=0;80.000000%=1;90.000000%=1;95.000000%=1;99.000000%=1;99.500000%=1;99.900000%=17;99.950000%=19;99.990000%=26;0%=0;0%=0;0%=0;0;336;0.293853;0.982926;0;0;0.000000%;0.000000;0.000000;4668204;4668204;1167051;1000;0;0;0.000000;0.000000;0;387;0.272427;1.054234;1.000000%=0;5.000000%=0;10.000000%=0;20.000000%=0;30.000000%=0;40.000000%=0;50.000000%=0;60.000000%=0;70.000000%=0;80.000000%=1;90.000000%=1;95.000000%=1;99.000000%=1;99.500000%=1;99.900000%=17;99.950000%=19;99.990000%=26;0%=0;0%=0;0%=0;0;387;0.295769;1.080249;0;0;0.000000%;0.000000;0.000000;100.000000%;0.000000%;10;0;0;100.0%;0.0%;0.0%;0.0%;0.0%;0.0%;0.0%;99.80%;0.01%;0.02%;0.14%;0.03%;0.01%;0.00%;0.01%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%
3;fio-2.18-1-gf8a9;verifynulltest;0;0;102400;12800000;3200000;8;0;0;0.000000;0.000000;0;28;0.030078;0.348936;1.000000%=0;5.000000%=0;10.000000%=0;20.000000%=0;30.000000%=0;40.000000%=0;50.000000%=0;60.000000%=0;70.000000%=0;80.000000%=0;90.000000%=0;95.000000%=0;99.000000%=1;99.500000%=1;99.900000%=1;99.950000%=1;99.990000%=18;0%=0;0%=0;0%=0;0;28;0.048437;0.432115;0;0;0.000000%;0.000000;0.000000;102400;1600000;400000;64;0;0;0.000000;0.000000;0;36;0.053906;0.370586;1.000000%=0;5.000000%=0;10.000000%=0;20.000000%=0;30.000000%=0;40.000000%=0;50.000000%=0;60.000000%=0;70.000000%=0;80.000000%=0;90.000000%=0;95.000000%=0;99.000000%=1;99.500000%=1;99.900000%=1;99.950000%=1;99.990000%=16;0%=0;0%=0;0%=0;0;46;0.372187;0.975936;0;0;0.000000%;0.000000;0.000000;100.000000%;0.000000%;0;0;619;100.0%;0.0%;0.0%;0.0%;0.0%;0.0%;0.0%;99.97%;0.01%;0.01%;0.02%;0.01%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%;0.00%

sitsofe:
@kusumi Thanks. That seems to show it enabled and used pshared mutexs on FreeBSD 11 which was one of the desired effects.

As I don't know the impact on all the other platforms (although I can the OpenIndiana version of Solaris to the list I have tested) if Jens prefers I can close this pull request and submit just the OSX change instead - I was unaware that fio was trying to avoid so configure tests where possible (although I'll note that enabling FreeBSD support for this in a way that turns it off for 10 and below will most likely wind up still testing _POSIX_THREAD_PROCESS_SHARED).

sitsofe:
@axboe Since https://github.com/axboe/fio/commit/1df28a3960734e1e00cb2e5fe0e261fcba30f7c7 (Conditionally enable FIO_HAVE_PSHARED_MUTEX on FreeBSD) just went in that complicates this patch (but that commit doesn't fix the problem for OS X). The question is: do you want to me to rework this commit which does a generic configure test for PSHARED_MUTEX support for all platforms or do you prefer platform header hard coded configuration (as @kusumi suggested in https://github.com/axboe/fio/pull/321#issuecomment-282555658 but note that 1df28a3960734e1e00cb2e5fe0e261fcba30f7c7 tests against a macro too)?

kusumi:
If I'm seeing your main() function and ifdefs correctly, wouldn't this work differently from what FreeBSD ports committers had originally added to their package (which is what I submitted yesterday) ?

FreeBSD says it's conditionally on,

+#if _POSIX_THREAD_PROCESS_SHARED > 0
+#define FIO_HAVE_PSHARED_MUTEX
+#endif

but your compile-time switch seems to say CONFIG_PSHARED is always on for FreeBSD (or NetBSD or DragonFly or whatever).

(By the way the one I submitted is someway similar to this one as it relies on _POSIX_XXX macros which is supposed to be portable (not sure if it really is in practice), but I still submitted it anyway because that's what they say who are supposed to know it's correct on their platform.)

sitsofe:
@kusumi You're right - the compile test should have errored if _POSIX_THREAD_PROCESS_SHARED wasn't greater than 0 - thanks for the heads up. Unfortunately all the platforms I tested on either had _POSIX_THREAD_PROCESS_SHARED > 1 or were OSX 10.6 so I didn't see the mistake.

I've fixed that up, rebased, removed the `_POSIX_THREAD_PROCESS_SHARED > 0` test from os/freebsd and tried it out on a FreeBSD 10.3 machine where POSIX pshared support now returns "no" as it should have done. Travis and AppVeyor are still showing the right results for Linux/OSX 10.11+/Windows.

kusumi:
yeah, I really don't care if that FreeBSD outbox stuff is in fio's FreeBSD header or if your diff contains all that in ./configure as long as Jens is okay with whichever one, but in case of the latter one (your diff) I guess FreeBSD ports maintainer will just keep their local patch still, since they won't know what has happened in fio upstream within just a few days (bring in their local patch -> removed -> merged to configure).

But whatever it is outbox packages are on their own, so we really don't (need to) care.

sitsofe:
@axboe Jens, over to you! Let me know what you think of this pull request and if you don't like it I'll just send a rebased https://github.com/sitsofe/fio/commit/4e26e28fafc545a6df3733e003de42e0223fabce for OSX instead.

axboe:
I like the configure variant, when you can detect it, it's always a better approach than defines. I personally don't care if FreeBSD don't notice this and don't drop their patch for shared mutexes, it won't harm their side. At some point someone will notice and drop it.

PR #326, opened 2017-03-17
fixed the error=invalid argument when the lower bound of bsrange is n…

liupan1111:
…ot power of 2.

Signed-off-by: Pan Liu <liupan1111@gmail.com>

The case is: when I set bsrange=24k-32k, fio quit, and reports error=invalid argument.

### DIFF
--- a/io_u.c
+++ b/io_u.c
@@ -533,6 +533,7 @@ static unsigned int __get_next_buflen(struct thread_data *td, struct io_u *io_u,
 	unsigned int buflen = 0;
 	unsigned int minbs, maxbs;
 	uint64_t frand_max, r;
+	bool power_2 = false;
 
 	assert(ddir_rw(ddir));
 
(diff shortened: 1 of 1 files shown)


liupan1111:
I traced the code, and found one io_u with buflen 26260 was generated. I found in the code, if the minbs is power of 2, alignment will be done, but not for other values(such as 24k in the case).

liupan1111:
@axboe , please help take a look. Thanks!

io_u.c
```diff
@@ -577,9 +577,11 @@ static unsigned int __get_next_buflen(struct thread_data *td, struct io_u *io_u,
 			}
 		}
 
-		if (!td->o.bs_unaligned && is_power_of_2(minbs))
+		bool power_2 = is_power_of_2(minbs);
```
axboe: Variable declarations go at the top, please.

axboe:
I think the idea is sound, but there's a minor issue with variable declarations after code. Can you fix that up and resubmit?

liupan1111:
Sure. There is some issue with my VPN, so I will update it tomorrow. Thanks.

liupan1111:
@axboe done, thanks.

axboe:
Pulled, thanks. I removed the init of power_2 to false, we don't have to do that.

liupan1111:
@axboe sure.

PR #330, opened 2017-04-04
Rename README to README.rst

adamjstewart:
Compare the homepage of https://github.com/axboe/fio and https://github.com/adamjstewart/fio/tree/fixes/readme to see the difference. GitHub won't detect that your file is a reStructuredText file unless the filename ends in `.rst`.


sitsofe:
Sadly I've pushed for this before and the result wasn't good because Github doesn't understand sphinx. See https://github.com/axboe/fio/pull/282 for details. Additionally this patch breaks the Windows installer...

sitsofe:
@adamjstewart I admit the directives are less key for the README (whereas they are critical for the HOWTO) but the README is included when creating the complete documentation (you can see the result up on http://fio.readthedocs.io/en/latest/fio_doc.html ).

adamjstewart:
I see. That does indeed look much better than the GitHub rendering. Well, I'll leave it up to someone else to find a way to reconcile the two. Good luck!

sitsofe:
@adamjstewart I think you've raised the question as to whether it would be good to link to the readthedocs rendering of the docs - they certainly are easier to read than a raw rendering.

Doing some digging, there are old (5 years+) documents recommending that you don't use the .rst extension on GitHub (http://documentation-style-guide-sphinx.readthedocs.io/en/latest/style-guide.html#filenames ) because of the poor handling of directives. Having said that fio doesn't follow that guide's other recommendations for restructured text documents...

Poking around further shows that GitHub use https://github.com/github/markup/blob/master/lib/github/commands/rest2html for rendering .rst documents . It recently gained a magic directive `.. github display [on | off]` which can allow the display of text only on github and/or control the verbatim display of directives GitHub doesn't understand. My hope is that one day GitHub grows to understand a few key Sphinx directives and that will solve the rendering problem on GitHub once and for all but the suggestions appears to be that Sphinx directives won't be supported (https://github.com/github/markup/issues/554#issuecomment-288581453 ).

axboe:
For my personal preference, I don't care about github or their guidelines. The only reason fio exists on here is to have a an extra backup and to allow people to easily fork it and work on it. So as far as documentation goes, we should do what works the best in general, and not worry about github at all. If it does work well on github that's a bonus, but it's definitely not the primary objective.

adamjstewart:
To some extent, GitHub serves as your homepage (unless you have another homepage that I haven't found yet), and it's the first thing that many users (including myself) will see. I also didn't notice an immediate link to your documentation. It might be better to have a simple Markdown README.md that points to your nicely formatted documentation.

PR #331, opened 2017-04-07
Add new ioengine for sheepdog

sheepdog-tester:
This ioengine is used for a virtual disk image (VDI) of the distributed block strage "sheepdog".
It supports direct access to VDI and runs as sync-io.
Users don't have to build quem or tgt to test io-benchmark of VDI anymore.

https://sheepdog.github.io/sheepdog/

Signed-off-by: sheepdog-tester test.sheepdog@gmail.com


sheepdog-tester:
This is the result or test run with examples/sheepdog.fio

***********************************************************************
sheepdog_test: (g=0): rw=randread, bs=(R) 1024KiB-1024KiB, (W) 1024KiB-1024KiB, (T) 1024KiB-1024KiB, ioengine=sheepdog, iodepth=1
...
fio-2.19-1-gd98cd8
Starting 4 processes
Job No.3, connected to target: tcp:localhost:7000:testvdi
Job No.2, connected to target: tcp:localhost:7000:testvdi
Job No.1, connected to target: tcp:localhost:7000:testvdi
Job No.4, connected to target: tcp:localhost:7000:testvdi
Jobs: 4 (f=4): [r(4)][100.0%][r=432MiB/s,w=0KiB/s][r=432,w=0 IOPS][eta 00m:00s]
sheepdog_test: (groupid=0, jobs=4): err= 0: pid=6684: Fri Apr  7 16:36:59 2017
   read: IOPS=442, BW=442MiB/s (464MB/s)(4000MiB/9043msec)
    clat (usec): min=4188, max=16478, avg=9152.73, stdev=450.66
     lat (usec): min=4188, max=16479, avg=9153.14, stdev=450.66
    clat percentiles (usec):
     |  1.00th=[ 8512],  5.00th=[ 8768], 10.00th=[ 8896], 20.00th=[ 9024],
     | 30.00th=[ 9024], 40.00th=[ 9152], 50.00th=[ 9152], 60.00th=[ 9280],
     | 70.00th=[ 9280], 80.00th=[ 9408], 90.00th=[ 9408], 95.00th=[ 9536],
     | 99.00th=[ 9536], 99.50th=[ 9792], 99.90th=[11328], 99.95th=[13504],
     | 99.99th=[16512]
    lat (msec) : 10=99.62%, 20=0.38%
  cpu          : usr=4.65%, sys=7.22%, ctx=6371, majf=0, minf=3655
     issued rwt: total=4000,0,0, short=0,0,0, dropped=0,0,0

   READ: bw=442MiB/s (464MB/s), 442MiB/s-442MiB/s (464MB/s-464MB/s), io=4000MiB (4194MB), run=9043-9043msec
***********************************************************************

sheepdog-tester:
Currently, it works on Linux because sheepdog basically supports Linux.

axboe:
Some strange stuff going on here:

- The meat of your code is in a header file, that should be a C file.
- Same header file duplicates a lot of the basic fio provided functionality, like list management etc. That has to be deleted, use the fio files instead.
- Your configure test doesn't seem to do anything. Are you just testing for various headers being there?
- Your sheepdog file open function seems to read the file fully. Why?! On top of that, you fill the IO buffers before reading into them?
- Your fio_sheepdog_queue() should not return -errno, it should return the proper FIO_Q_*. Look at the other IO engines to see how they handle erroring IO.

sheepdog-tester:
Thank you for the review.
I modified the codes.

- The meat of your code is in a header file, that should be a C file.
> I moved my code to a C file "sheepdog.c" and removed a header file.

- Same header file duplicates a lot of the basic fio provided functionality, like list management etc. That has to be deleted, use the fio files instead.
> I tried to removed the duplication. I added some macros about flist operation for this engine in flist.h.

- Your configure test doesn't seem to do anything. Are you just testing for various headers being there?
> This engine doesn't use anything specific about sheepdog except for this code and it independents from sheepdog process. That means users don't have to check whether sheepdog is built before building fio. Therefore, I think it's enough to test for various headers.

- Your sheepdog file open function seems to read the file fully. Why?! On top of that, you fill the IO buffers before reading into them?
> Sheepdog is not a pure block storage. User can use it as a block storage. However, in the sheepdog architecture, a filesystem (e.g. ext4, xfs) is running and the data is stored as "object file" in a certain directory. Therefore, it's necessary to fill the IO buffers before reading into a directory related to the target VDI. Otherwise, this engine can't read anything because there are no object files. This file open function does that.

- Your fio_sheepdog_queue() should not return -errno, it should return the proper FIO_Q_*. Look at the other IO engines to see how they handle erroring IO.
> I looked at other IO engines, e.g. engines/glusterfs-sync.c, engines/fusion-aw.c. They use "return -EINVAL" in the same situation. So, I used it. If it's not good, which FIO_Q_* should I use?

sitsofe:
@sheepdog-tester Could you squash the new changes and the old into one commit - it makes it easier to comment on things?

> Therefore, it's necessary to fill the IO buffers before reading into a directory related to the target VDI. Otherwise, this engine can't read anything because there are no object files. 

Are you saying that the VDI you ask the ioengine to work with might not exist? If so shouldn't the user create it beforehand? What if I want to work with an existing VDI - won't this always prewarm it (thus potentially skewing results)?

> I looked at other IO engines, e.g. engines/glusterfs-sync.c, engines/fusion-aw.c. They use "return -EINVAL" in the same situation. So, I used it. If it's not good, which FIO_Q_* should I use?

You could set `io_u->error = EINVAL` and return FIO_Q_COMPLETED - this is what the pmemblk ioengine does (https://github.com/axboe/fio/blob/master/engines/pmemblk.c#L397 ).

On another note I'm curious as to why you need the pthread_rwlocks - Is this to try and stop other fio sheepdog jobs who happen to be doing I/O to the same "file" from interfering? What happens if those jobs are other processes?

Could all the `#define`s (e.g. https://github.com/sheepdog-tester/fio/blob/0dfaac2a0a7ce084c96dbf53fbb90e6f6d69f51b/engines/sheepdog.c#L1354 ) be moved to the top of the C file?

sitsofe:
(Moving comment down here)

There are a large amount of defines in engines/sheepdog.c. I can't help wondering whether sheepdog itself has these defines in a header file (e.g. in https://github.com/sheepdog/sheepdog/blob/48eb114ffb39a14e82aaed3c983f373dc9757674/include/sheepdog_proto.h ) and what should be done is that a dependency should be being added against the sheepdog include files to get them. This might be a bit onerous given that you're only after data structures but it means you're aware of API changes that might affect you at build time and keeps the engine code less cluttered.

sitsofe:
https://github.com/sheepdog-tester/fio/blob/0dfaac2a0a7ce084c96dbf53fbb90e6f6d69f51b/HOWTO#L1900 you probably need to say that you can use a unix prefix too to use a unix domain socket (is this section complete?). You may also want to say you don't use the contents of the filename option at all - I think this might be the first engine outside of null not to use its contents in any form.

The odd self reference to tgtd/tgt in comments from the copied code now seems strange in the the new context e.g. https://github.com/sheepdog-tester/fio/blob/0dfaac2a0a7ce084c96dbf53fbb90e6f6d69f51b/engines/sheepdog.c#L1169 . Perhaps they should be updated to say ioengine?

sheepdog-tester:
@sitsofe Thank you for the comments.
I modified my codes and squashed commits.

Here is my reply:
> Are you saying that the VDI you ask the ioengine to work with might not exist? If so shouldn't the user create it beforehand? What if I want to work with an existing VDI - won't this always prewarm it (thus potentially skewing results)?

No. Anytime, the targeted VDI has to be created before testing with fio. In fio process with this engine, any VDI is not created. The init function of this engine just tries to connect the existed VDI. 
What I mean before is that when io coming to the VDI, sheepdog process accesses to "object files" in a directory related to the VDI. When read process in sheepdog, there has to be object files to read. That's why I fill the data before read-io.

>On another note I'm curious as to why you need the pthread_rwlocks - Is this to try and stop other fio sheepdog jobs who happen to be doing I/O to the same "file" from interfering? What happens if those jobs are other processes?

I'm not sure the details.
I tested with and without the pthread_rwlocks and the results are not different.
I want to respect the original codes of tgt. So, I'd like to keep the current code.

>There are a large amount of defines in engines/sheepdog.c. I can't help wondering whether sheepdog itself has these defines in a header file (e.g. in https://github.com/sheepdog/sheepdog/blob/48eb114ffb39a14e82aaed3c983f373dc9757674/include/sheepdog_proto.h ) and what should be done is that a dependency should be being added against the sheepdog include files to get them. This might be a bit onerous given that you're only after data structures but it means you're aware of API changes that might affect you at build time and keeps the engine code less cluttered.

Sheepdog doesn't have libraries for clients and I want to make fio stand-alone from sheepdog's codes.
Usually, sheepdog's client does not have to have sheepdog's code and sheepdog process.
In sheepdog's engine of other program like qemu, defines  are also written directly. 
https://github.com/qemu/qemu/blob/master/block/sheepdog.c
I know It's not definitely clean. However, if ok, I want to keep the current state.

>you probably need to say that you can use a unix prefix too to use a unix domain socket (is this section complete?). 

Yes, I think this section is complete.
I skipped to write an example of a unix domain socket because sheepdog is often used not  a unix domain socket but tcp.  I think it's better that only well-known users use a unix domin socket in this engine without any explanation.

> You may also want to say you don't use the contents of the filename option at all - I think this might be the first engine outside of null not to use its contents in any form.

Sorry, I don't understand the meaning of this. Could you tell me this in a little bit concrete terms?

sitsofe:
> No. Anytime, the targeted VDI has to be created before testing with fio. In fio process with this engine, any VDI is not created. The init function of this engine just tries to connect the existed VDI.
>
> What I mean before is that when io coming to the VDI, sheepdog process accesses to "object files" in a directory related to the VDI. When read process in sheepdog, there has to be object files to read. That's why I fill the data before read-io. If there is a better way, could you tell me that?

So it's not so much that the first ever read to a VDI's LBA is disallowed but rather you want to guarantee that it has really been allocated? What happens if I have a brand new VDI hooked up to a VM and I immediately issue a read down before writing anything does it just return extra quickly because the storage is essentially thin?

What might be better is to consult the [`fallocate`](http://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-fallocate) option and if the user is asking for preallocation only then do you go on and force it by writing to every single block. Perhaps if you don't want to overload the meaning of an existing fio option it should just be an engine option ("disable_sparse, if 1 ensure that the VDI is fully allocated so reads are never to sparse areas). It's a bit of a bodge but you might have cases where the user really does want the VDI to remain thin/sparse...

> I'm not sure the details.
I tested with and without the pthread_rwlocks and the results are not different.
I want to respect the original codes of tgt. So, I'd like to keep the current code.

Perhaps you can ask @mitake (original author of the sheepdog tgt code) for comment?

> Sheepdog doesn't have libraries for clients and I want to make fio stand-alone from sheepdog's codes.

OK but bear in mind it would only be a build time requirement but I do see the other projects you linked took the "copy the source into our tree" approach.

>  I think it's better that only well-known users use a unix dom[a]in socket in this engine without any explanation.

Hmm, seems a bit rough that I can only discover it by accident or by reading the source. Still I'd recommend changing the text "How to describe" (which sounds a bit like a "TODO: Describe this further") to "Format of this argument".

> Sorry, I don't understand the meaning of this [non-usage of filename]. Could you tell me this in a little bit concrete terms?

I'm going to withdraw this comment. I thought all diskless I/O engines still used fio's `filename` in some shape or form but looking at the rbd engine shows it doesn't make use of it at all.

mitake:
@sitsofe 

sheepdog has a strong assumption about how clients issue requests, especially write requests. It is because sheepdog is a very simplified distributed file system built on key-value store. Some objects are used as metadata (e.g. inode) and other objects are used for storing ordinary data. These objects are distributed equally on members of sheepdog cluster, so there are no bottlenecks that limit throughput (unlike NameNode of HDFS).

However, not having metadata server means that sheepdog cannot provide metadata transaction basically. So the system cannot serialize update of metadata and it results corrupted inode objects. Fortunately sheepdog isn't a general purpose file system so having the assumption about client requests is realistic (virtual disks are accessed by only one client at a time).

So the reader writer lock of protecting the inode is used for providing the assumption. Clients (tgt, fio, QEMU, etc) needs to keep the assumption voluntary and the lock is required for the purpose.

sitsofe:
@mitake OK I see. So given that (as currently written) the fio sheepdog ioengine is synchronous these locks would essentially never be contended? The only case I can think of where they might be contended is if you were to run two fio jobs against the same sheepdog storage. In that case you could have two independent sheepdog ioengines writing to the same sheepdog region but if the ioengines were using the separate processes model then these pthread locks would be ineffective right?

mitake:
@sitsofe if the single ioengine doesn't use multiple threads, yes, the lock is needless (tgt uses multiple threads for issuing requests in a parallel manner so the original code has the lock).

sitsofe:
@sheepdog-tester more minor comments:

I think you can get rid of some includes:
```
#include <fcntl.h>
#include <arpa/inet.h>
#include <sys/epoll.h>
#include <sys/un.h>
```

You might need to convert those pthread locks over to fio's wrappers (as seen in mutex.c) and rework how the lock is found otherwise simultaneous jobs going to the same backend storage may not voluntarily serialize metadata updates between each other.

Might be nicer to shuffle sheepdog to appear below "NVML dev-dax engine" in configure.

sitsofe:
@sheepdog-tester a few more comments:

I think size might not be initialized when first assigned to orig_size in sd_io:
```c
 926         for (; idx < max; idx++) {
 927                 orig_size = size;
```

In the same function the 0 value stored via this particular create assignment will never be read because you either exit or go round the loop again (which resets create)...
```c
 974                                         need_update_inode = 1;
 975                                         create = 0;
```

Running
```
valgrind ./fio --thread --ioengine=sheepdog --size=100M --name=go 
```

results in the following warning:
```
==30996== Thread 3:
==30996== Invalid read of size 1
==30996==    at 0x4C30F62: strlen (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==30996==    by 0x5C128CD: strdup (strdup.c:41)
==30996==    by 0x483B6F: sd_open.isra.2 (sheepdog.c:1091)
==30996==    by 0x485167: fio_sheepdog_init (sheepdog.c:1308)
==30996==    by 0x41C72C: td_io_init (ioengines.c:370)
==30996==    by 0x4645E7: thread_main (backend.c:1645)
==30996==    by 0x576E6B9: start_thread (pthread_create.c:333)
==30996==    by 0x5C8E82C: clone (clone.S:109)
==30996==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
==30996== 
==30996== 
==30996== Process terminating with default action of signal 11 (SIGSEGV)
==30996==  Access not within mapped region at address 0x0
==30996==    at 0x4C30F62: strlen (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==30996==    by 0x5C128CD: strdup (strdup.c:41)
==30996==    by 0x483B6F: sd_open.isra.2 (sheepdog.c:1091)
==30996==    by 0x485167: fio_sheepdog_init (sheepdog.c:1308)
==30996==    by 0x41C72C: td_io_init (ioengines.c:370)
==30996==    by 0x4645E7: thread_main (backend.c:1645)
==30996==    by 0x576E6B9: start_thread (pthread_create.c:333)
==30996==    by 0x5C8E82C: clone (clone.S:109)
```
Perhaps there needs to be some error checking on values to the ioengine at init time or some of those operations could be moved to the open callback? 

There's a similar out of bounds read if connecting fails:

```
valgrind ./fio --thread --ioengine=sheepdog --size=100M --name=go --target=tcp:127.0.0.1:0
[...]
failed connect to 127.0.0.1:0=0KiB/s,w=0KiB/s][r=0,w=0 IOPS][eta 00m:00s]
connecting to sheep process failed.
fio: io engine sheepdog init failed.
fio: pid=31022, err=-1/
==31012== Thread 3:
==31012== Invalid read of size 8
==31012==    at 0x484F24: fio_sheepdog_cleanup (sheepdog.c:1331)
==31012==    by 0x41C355: close_ioengine (ioengines.c:184)
==31012==    by 0x46342E: thread_main (backend.c:1814)
==31012==    by 0x576E6B9: start_thread (pthread_create.c:333)
==31012==    by 0x5C8E82C: clone (clone.S:109)
==31012==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
==31012== 
==31012== 
==31012== Process terminating with default action of signal 11 (SIGSEGV)
==31012==  Access not within mapped region at address 0x0
==31012==    at 0x484F24: fio_sheepdog_cleanup (sheepdog.c:1331)
==31012==    by 0x41C355: close_ioengine (ioengines.c:184)
==31012==    by 0x46342E: thread_main (backend.c:1814)
==31012==    by 0x576E6B9: start_thread (pthread_create.c:333)
==31012==    by 0x5C8E82C: clone (clone.S:109)
```

I wonder if many of the `log_info()` calls would be better off as `dprint(FD_IO ,)` calls so they only appear when actively debugging?

sitsofe:
@sheepdog-tester Clang seems to be producing the odd warning about unused functions: https://travis-ci.org/axboe/fio/jobs/221984440 .

I forgot to say my comments are just that - comments. Jens has the final say so feel free to pushback on any of the points I raise that you object to.

Thanks for taking the time to submit a new I/O engine!

sheepdog-tester:
@sitsofe Thank you too for making the codes better.
I hope that this engine will be merged for sheepdog's users.

sitsofe:
@sheepdog-tester You're welcome!

When you're ready add a comment asking for a re-review and mention any feedback comments you're not planning to address etc.

sheepdog-tester:
@sitsofe 
About making a check with valgrind,
if sheepdog is build and sheepdog-daemon is active,  it's running normally as follows.
This engine should be used on the assumption that sheepdog is running.

```
valgrind fio --thread --ioengine=sheepdog --size=100M --name=go --target=tcp:localhost:7000:testvdi
==9033== Memcheck, a memory error detector
==9033== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==9033== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==9033== Command: fio --thread --ioengine=sheepdog --size=100M --name=go --target=tcp:localhost:7000:testvdi
==9033== 
go: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=sheepdog, iodepth=1
fio-2.19-1-g224d-dirty
Starting 1 thread
Job No.1, connected to target: tcp:localhost:7000:testvdi
Jobs: 1 (f=1): [R(1)][100.0%][r=24.4MiB/s,w=0KiB/s][r=6235,w=0 IOPS][eta 00m:00s]
go: (groupid=0, jobs=1): err= 0: pid=9041: Tue Apr 18 17:28:49 2017
   read: IOPS=6639, BW=25.1MiB/s (27.2MB/s)(100MiB/3856msec)
    clat (usec): min=104, max=33259, avg=138.96, stdev=213.13
     lat (usec): min=105, max=33262, avg=141.31, stdev=215.57
    clat percentiles (usec):
     |  1.00th=[  108],  5.00th=[  110], 10.00th=[  113], 20.00th=[  120],
     | 30.00th=[  125], 40.00th=[  131], 50.00th=[  135], 60.00th=[  141],
     | 70.00th=[  147], 80.00th=[  151], 90.00th=[  161], 95.00th=[  169],
     | 99.00th=[  217], 99.50th=[  233], 99.90th=[  266], 99.95th=[  354],
     | 99.99th=[ 2512]
    lat (usec) : 250=99.85%, 500=0.13%, 1000=0.01%
    lat (msec) : 2=0.01%, 4=0.01%, 10=0.01%, 50=0.01%
  cpu          : usr=55.54%, sys=20.31%, ctx=26101, majf=0, minf=1367
     issued rwt: total=25600,0,0, short=0,0,0, dropped=0,0,0

   READ: bw=25.1MiB/s (27.2MB/s), 25.1MiB/s-25.1MiB/s (27.2MB/s-27.2MB/s), io=100MiB (105MB), run=3856-3856msec
==9033== 
==9033== HEAP SUMMARY:
==9033==     in use at exit: 275 bytes in 16 blocks
==9033==   total heap usage: 391 allocs, 375 frees, 14,455,070 bytes allocated
==9033== 
==9033== LEAK SUMMARY:
==9033==    definitely lost: 70 bytes in 3 blocks
==9033==    indirectly lost: 74 bytes in 5 blocks
==9033==      possibly lost: 0 bytes in 0 blocks
==9033==    still reachable: 131 bytes in 8 blocks
==9033==         suppressed: 0 bytes in 0 blocks
==9033== Rerun with --leak-check=full to see details of leaked memory
==9033== 
==9033== For counts of detected and suppressed errors, rerun with: -v
==9033== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

sheepdog-tester:
@axboe Could you make a re-check to merge this engine? I think It's cleaner than before.

PR #342, opened 2017-04-26
GPUDirect RDMA support

yufeiren:


axboe:
I pulled this in, but I had to do a number of fixups on top. See git repo for details.

PR #343, opened 2017-04-26
Add serialize_overlap option

sitsofe:
If this isn't set (the default) fio can submit asynchronous I/Os that
overlap leading to potential data races. When this option is set, the
race is avoided by having fio serialize the overlapping submissions.

Thanks to Rachel Lunnon (StorMagic) for helping me debug the initial
version of this!

### DIFF
--- a/cconv.c
+++ b/cconv.c
@@ -96,6 +96,7 @@ void convert_thread_options_to_cpu(struct thread_options *o,
 	o->iodepth_batch = le32_to_cpu(top->iodepth_batch);
 	o->iodepth_batch_complete_min = le32_to_cpu(top->iodepth_batch_complete_min);
 	o->iodepth_batch_complete_max = le32_to_cpu(top->iodepth_batch_complete_max);
+	o->serialize_overlap = le32_to_cpu(top->serialize_overlap);
 	o->size = le64_to_cpu(top->size);
 	o->io_size = le64_to_cpu(top->io_size);
 	o->size_percent = le32_to_cpu(top->size_percent);
@@ -346,6 +347,7 @@ void convert_thread_options_to_net(struct thread_options_pack *top,
 	top->iodepth_batch = cpu_to_le32(o->iodepth_batch);
 	top->iodepth_batch_complete_min = cpu_to_le32(o->iodepth_batch_complete_min);
 	top->iodepth_batch_complete_max = cpu_to_le32(o->iodepth_batch_complete_max);
+	top->serialize_overlap = cpu_to_le32(o->serialize_overlap);
 	top->size_percent = cpu_to_le32(o->size_percent);
 	top->fill_device = cpu_to_le32(o->fill_device);
 	top->file_append = cpu_to_le32(o->file_append);
--- a/init.c
+++ b/init.c
@@ -698,6 +698,23 @@ static int fixup_options(struct thread_data *td)
 	if (o->iodepth_batch_complete_min > o->iodepth_batch_complete_max)
 		o->iodepth_batch_complete_max = o->iodepth_batch_complete_min;
 
+	/*
+	 * There's no need to check for in-flight overlapping IOs if the job
+	 * isn't changing data or the maximum iodepth is guaranteed to be 1
+	 */
+	if (o->serialize_overlap && !(td->flags & TD_F_READ_IOLOG) &&
+	    (!(td_write(td) || td_trim(td)) || o->iodepth == 1))
+		o->serialize_overlap = 0;
+	/*
+	 * Currently can't check for overlaps in offload mode
+	 */
+	if (o->serialize_overlap && o->io_submit_mode == IO_MODE_OFFLOAD) {
+		log_err("fio: checking for in-flight overlaps when the "
+			"io_submit_mode is offload is not supported\n");
+		o->serialize_overlap = 0;
+		ret = warnings_fatal;
+	}
+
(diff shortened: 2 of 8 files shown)


axboe:
I don't understand why we have a need for this? If we are making these kinds of guarantees, then the random map should prevent this from happening. If it doesn't, it's a bug. If the job file uses options that excludes the use of a random map, then that's the issue.

What am I missing?

sitsofe:
This patch isn't just about when fio does its own verification it's also to cope with external verification.

I'm testing iSCSI storage that really doesn't cope when two overlapping I/Os are submitted together when an I/O log is replayed at full speed. It's been noted that submitting I/O in this way isn't done by programs that have to read it back but it's difficult to prevent this from happening as the logs being replayed don't have synchronization information within them. The storage goes on to create regions where its own (internal) verification points out that one path has one set of data and the other path to it has a different set of data for the same LBA!

A similar issue is seen when trying to verify a random workload that is using a zipfian distribution - the last I/O that fio believes should be at a particular point is not actually the write that wound up winning so a verification mismatch occurs.

axboe:
That's a good point, there are workloads where we cannot (or don't want to) control blocks, like your zipf example. However, we should limit the checking for the cases where we absolutely HAVE to check. I'll need to think about this for a bit. It seems like a very heavy handed approach, would be nice if we could handle this more elegantly (and with much less overhead).

cconv.c
```diff
@@ -96,6 +96,7 @@ void convert_thread_options_to_cpu(struct thread_options *o,
 	o->iodepth_batch = le32_to_cpu(top->iodepth_batch);
 	o->iodepth_batch_complete_min = le32_to_cpu(top->iodepth_batch_complete_min);
 	o->iodepth_batch_complete_max = le32_to_cpu(top->iodepth_batch_complete_max);
+	o->serialize_overlap = le32_to_cpu(top->serialize_overlap);
```
axboe: This is missing the conversions the other way

init.c
```diff
@@ -698,6 +698,23 @@ static int fixup_options(struct thread_data *td)
 	if (o->iodepth_batch_complete_min > o->iodepth_batch_complete_max)
 		o->iodepth_batch_complete_max = o->iodepth_batch_complete_min;
 
+	/*
+	 * There's no need to check for in-flight overlapping IOs if the job
+	 * isn't changing data or the maximum iodepth is guaranteed to be 1
+	 */
+	if (o->serialize_overlap && !(td->flags & TD_F_READ_IOLOG) &&
```
axboe: I like this, disabling when we don't need it.

init.c
```diff
@@ -698,6 +698,23 @@ static int fixup_options(struct thread_data *td)
...
+	    (!td_trimwrite(td) || o->iodepth == 1))
+		o->serialize_overlap = 0;
+	/*
+	 * Currently can't check for overlaps in offload mode
+	 */
+	if (o->serialize_overlap && o->io_submit_mode == IO_MODE_OFFLOAD) {
```
axboe: What's preventing it from being enabled for offloaded submission?

init.c
sitsofe: Yeah - my hope was to try and turn it off whenever possible just in case people were in the habit of turning it on. Technically if you're replaying an iolog at depth 1 you don't need to go through the checks either so perhaps I need to tidy this further.

init.c
sitsofe: My ingenuity. I couldn't see a way to say "wait for completion" in the offloaded submission case so I took a punt. Can I just `goto reap` as other parts of the code do?

cconv.c
sitsofe: Should be addressed in v2.

sitsofe:
I admit I took the easy way out on this one and essentially force quadratic behaviour on the iodepth and the number of I/Os (but it worked surprisingly well on my depth of 32). Perhaps the better thing to do would be to stuff each in flight I/O into a tree then at least the searching would be sped up. I would say a sparse bitmap is good enough but I originally had code that avoided waiting when both overlapping I/Os were reads (I've since removed) but if you've got multiple overlapping reads in the bitmap and you only want to remove one of them. It also meant if statements at submission AND completion.

sitsofe:
The branch has been updated to add a man page entry for `serialize_overlap`, fix a bug where `serialize_overlap` would disable itself wrongly (because I thought `td_trimwrite` meant the job did a trim or write which turned out to be incorrect), and it also add the patches in the pull request https://github.com/axboe/fio/pull/345 (which supersedes the "iolog: sort verify entries when using dynamic blocksizes" commit) and the pull request https://github.com/axboe/fio/pull/359 .

sitsofe:
Awaiting a response from Jeff Furlong on these because he was wondering if the combination of these patches (and using the `serialize_overlap=1` option they introduce) might solve the issue mentioned over on http://www.spinics.net/lists/fio/msg06139.html .

sitsofe:
Jeff has come back and says he hasn't seen any issues with this patch so I've updated the commit messages to mention him in a Tested-by line.

axboe:
I pulled it it, and added an extra commit to move the overlap checking into a helper, and only call the overlap check if we have > 1 IO in flight, instead of > 0. The latter is because the io_u in question is already accounted. By definition, we need >= 2 IOs to be worried about overlap.

sitsofe:
Good catch - thanks Jens!

PR #345, opened 2017-04-27
iolog: fix spurious verify mismatch when random I/O overlaps

sitsofe:
Two fixes for issues that turn up when verifying overlapping I/O.

### DIFF
--- a/iolog.c
+++ b/iolog.c
@@ -227,20 +227,16 @@ void log_io_piece(struct thread_data *td, struct io_u *io_u)
 	}
 
 	/*
-	 * We don't need to sort the entries, if:
-	 *
-	 *	Sequential writes, or
-	 *	Random writes that lay out the file as it goes along
-	 *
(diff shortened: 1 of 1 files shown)


sitsofe:
@axboe I've updated this branch to only address the spurious mismatch issue by removing overwrite == 0 check so I think it's good enough to pull. I think this is the better way to go as there could be many other random jobs that have a randommap but generate overlapping I/Os and there seem to be many cases where the optimisation won't actually speed things up (file was pre allocated via fallocate, file already exists and was made sequentially in the past etc). I wasn't sure about removing the `file_randommap(td, ipo->file)` test because I couldn't tell if it was trying to still guard against the case where the user sets verifysort=0 .

sitsofe:
@axboe any thoughts on this one?

sitsofe:
This commit was merged via https://github.com/axboe/fio/pull/343 so closing this pull request.

PR #347, opened 2017-04-29
Warn when we know an iodepth > 1 can never be reached

sitsofe:
This pull request fixes up a small bug where bits weren't cleared in td_set_ioengine_flags and adds a warning that the requested iodepth can't be reached by the specified I/O engine e.g.:

```
fio: iodepth=32 but psync IO engine can't submit more than one I/O at a time
```

### DIFF
--- a/engines/null.c
+++ b/engines/null.c
@@ -149,7 +149,7 @@ void get_ioengine(struct ioengine_ops **ioengine_ptr)
 	ioengine.init           = fio_null_init;
 	ioengine.cleanup        = fio_null_cleanup;
 	ioengine.open_file      = fio_null_open;
-	ioengine.flags          = FIO_DISKLESSIO | FIO_FAKEIO;
+	ioengine.flags          = FIO_DISKLESSIO | FIO_FAKEIO | FIO_SUBMITMANY;
 }
 }
 #endif /* FIO_EXTERNAL_ENGINE */
--- a/engines/sg.c
+++ b/engines/sg.c
@@ -821,7 +821,7 @@ static struct ioengine_ops ioengine = {
 	.open_file	= fio_sgio_open,
 	.close_file	= generic_close_file,
 	.get_file_size	= fio_sgio_get_file_size,
-	.flags		= FIO_SYNCIO | FIO_RAWIO,
+	.flags		= FIO_SYNCIO | FIO_RAWIO | FIO_SUBMITMANY,
 };
 
 #else /* FIO_HAVE_SGIO */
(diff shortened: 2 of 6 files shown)


axboe:
The achieved depths are in the actual output of fio, so I'm not convinced we need this. This won't cover all the bases anyway. One example is sg. If you are operating on a block device, then it'll still be sync. And the sync engine will be QD=1 for almost all variants.

If we are going to do this, I think it should be bullet proof. If not, we should stick to ensuring that the documentation adequately describes what to look for to verify depths.

sitsofe:
Oh wow - I hadn't twigged the FIO_Q_QUEUED was because the sg engine worked on character devices too. I'm not convinced it's possible to make the detection of maximum depth for sg more bullet proof with this generic code given that one path (writing to a character device) can submit multiple I/Os even though the other path (writing to a block device) can't and you can't tell which case you're in until after the engine has opened the file.

With regard to the sync engines the only varient that was marked as FIO_SUBMITMANY was the vsync version (which fakes queueing for contiguous I/O).

I suppose you could have a "deferred warnings" callback (normally set to NULL meaning we do the current checks). You could call it just after open with the caveat that you would potentially see warnings every open. If an engine participates it would allow the tightening of checks and open the door for things like async engines saying they can only achieve a depth of 1 etc. Hard to justify the work to do it (and maintain it) though...

One part of this that might be easier to accept is the https://github.com/axboe/fio/pull/347/commits/3d92b849e59c9caa28a2207560fb3bd21da74003 commit ("fio.h: zero old flag bits when setting new ioengine flags"). Would it possible to cherry pick that or would you like a separate pull request?

axboe:
But what about the case where the user does:

```
ioengine=libaio
iodepth=32
direct=0

```
For that, fio will not warn, yet it won't achieve more than QD=1 at runtime.

My point is that there IS a way to make the detection fool proof, and that's to do it after the job has run. That's the only way. And I don't like half measures, since if we can't get 100% coverage, then it's almost worthless. Users can't rely on it anyway, and if they can't depend on it, then it almost makes the problem worse, since they think there's some protection against this.

sitsofe:
OK I see your point - this isn't reliable (and gives a false sense of security) but another approach is totally reliable.

Perhaps in the future there will be a depth_warn=1 option that checks whether the requested depth was ever achieved by the job and if not spits out a warning after the job finishes (I suppose this could be the default if there's a way not to pollute JSON output). For now let's close this pull request and I'll split commit 3d92b84 out.

PR #349, opened 2017-04-30
fio.h: zero old flag bits when setting new ioengine flags

sitsofe:
Fix a small bug in td_set_ioengine_flags() where it wouldn't zero old
bits in td->flags before setting new flag bits.


axboe:
Pulled, thanks.

PR #351, opened 2017-04-30
os/os-{linux, android}.h: fix broken byte swap definitions

### LINKED ISSUE #348: fio_swapN on Android
#### Description
`os/os-android.h` defines `fio_swapN` as `__bswap_N` (where `N` is 16, 32 or 64). However, there is no such function defined in the Android libc. There _is_ `bswap_N` (sans double underscore) defined in `<byteswap.h>`. This naturally causes compilation to fail, as it can't link properly.

`os/os-android.h` includes `<asm/byteorder.h>`, which in turn includes `<linux/byteorder/little_endian.h>` (at least for arm). This does define various byte swapping facilities, but does _not_ define `__bswap_N`. I wonder whether it used to define such functions, but at some point (perhaps with the change to unified headers) it removed them. Is the header being included for some other purpose, or can it be safely removed from the list of `#include`s?

#### Environment
Targeting Android API level 23 for arm-linux-androideabi with a standalone toolchain from ndk-r14, using clang, trying to compile fio 2.19.

#### Reproduction
Compile for Android using clang with unified headers.

#### Possible Solution

`os/os-linux.h` uses `__builtin_bswapN` if it's available, else it uses `__bswap_N`, which _is_ defined in glibc (but technically it is also supposed to be used as `bswap_N`, as per the `#define`s in the glibc `<byteswap.h>`).

`__builtin_bswapN` should be available for Android as well as a compiler intrinsic (I believe that clang implements most of the GCC builtins), so similar logic could be used in `os/os-android.h`, preferring the compiler builtin and dropping to the libc function if it's not available.

`__bswap_N` in `os/os-linux.h` should probably be changed to `bswap_N` as well (though the functions aren't usually used, as the builtins should be selected instead).

omor1:
Fixes #348

Currently this either uses byte swap intrinsics for ALL three byte swap operations (`fio_swapN` for `N = 16, 32, 64`). This can be changed, but at the cost of more complex checking. @sitsofe do you have any input on this?

Waiting on information about the `#include <asm/byteorder.h>` in `os/os-android.h`—I suspect it was used for the previous byte swap definitions, but it's broken in at least ndk-r14+ (when using unified headers).

### DIFF
--- a/os/os-android.h
+++ b/os/os-android.h
@@ -20,6 +20,10 @@
 #include "binject.h"
 #include "../file.h"
 
+#ifndef __has_builtin         // Optional of course.
+  #define __has_builtin(x) 0  // Compatibility with non-clang compilers.
+#endif
+
 #define FIO_HAVE_DISK_UTIL
 #define FIO_HAVE_IOSCHED_SWITCH
 #define FIO_HAVE_IOPRIO
@@ -219,9 +223,19 @@ static inline long os_random_long(os_random_state_t *rs)
 #define FIO_O_NOATIME	0
 #endif
 
-#define fio_swap16(x)	__bswap_16(x)
-#define fio_swap32(x)	__bswap_32(x)
-#define fio_swap64(x)	__bswap_64(x)
+/* Check for GCC or Clang byte swap intrinsics */
+#if (__has_builtin(__builtin_bswap16) && __has_builtin(__builtin_bswap32) \
+     && __has_builtin(__builtin_bswap64)) || (__GNUC__ > 4 \
+     || (__GNUC__ == 4 && __GNUC_MINOR__ >= 8)) /* fio_swapN */
+#define fio_swap16(x)	__builtin_bswap16(x)
+#define fio_swap32(x)	__builtin_bswap32(x)
+#define fio_swap64(x)	__builtin_bswap64(x)
+#else
+#include <byteswap.h>
+#define fio_swap16(x)	bswap_16(x)
+#define fio_swap32(x)	bswap_32(x)
+#define fio_swap64(x)	bswap_64(x)
+#endif /* fio_swapN */
 
 #define CACHE_LINE_FILE	\
 	"/sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size"
--- a/os/os-linux.h
+++ b/os/os-linux.h
@@ -16,11 +16,14 @@
 #include <linux/unistd.h>
 #include <linux/raw.h>
 #include <linux/major.h>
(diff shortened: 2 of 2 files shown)


axboe:
Pulled, thanks.

omor1:
According to da52c582b76a8752ad3dd113696c9debcb3449e8 it seems that `<asm/byteorder.h>` was indeed used for the byte swap routines, but `__bswap_N` aren't there in newer versions of Bionic with unified headers.

sitsofe:
@omor1 What you did is as simple as it gets until GCC grows a way to test for intrinsics especially given we weren't directly using the intrinsics before even though they were available.

omor1:
@sitsofe I meant that if only some of the intrinsics are available, it won't use them. It is possible to check separately, but it's much more complex:

```C
#if __has_builtin(__builtin_bswap16) || __GNUC__ > 4 \
    || (__GNUC__ == 4 && __GNUC_MINOR__ >= 8) /* fio_swap16 */
#define fio_swap16(x)	__builtin_bswap16(x)
#elif __GNUC__ == 4 && __GNUC_MINOR__ >= 3
#define fio_swap16(x)	__builtin_bswap32((x) << 16)
#else
#include <byteswap.h>
#define fio_swap16(x)	bswap_16(x)
#endif /* fio_swap16 */

#if __has_builtin(__builtin_bswap32) || __GNUC__ > 4 \
    || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3) /* fio_swap32 */
#define fio_swap32(x)	__builtin_bswap32(x)
#else
#include <byteswap.h>
#define fio_swap32(x)	bswap_32(x)
#endif /* fio_swap32 */

#if __has_builtin(__builtin_bswap32) || __GNUC__ > 4 \
    || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3) /* fio_swap64 */
#define fio_swap64(x)	__builtin_bswap64(x)
#else
#include <byteswap.h>
#define fio_swap64(x)	bswap_64(x)
#endif /* fio_swap64 */
```

This would be most efficient for every possible compiler version, but is a lot more complex.

sitsofe:
@omor1 I noticed, but I think you made a reasonable tradeoff between complexity and going fast. Besides, we never know what the libc being used will turn `bswap_N` into anyway - it might try and map it to the builtin anyway: https://github.com/bminor/glibc/blob/glibc-2.25/bits/byteswap.h#L46 (even if there are corner cases: https://sourceware.org/bugzilla/show_bug.cgi?id=20530 ).

Issue #352, opened 2017-05-01
Android and shared memory


omor1:
#### Description
`os/os-android.h` currently does not import `<sys/shm.h>`, which _does_ however exist in recent versions of the ndk/Bionic.

Without compiling with `--disable-shm`, there is a compile error: libc does not provide `struct shmid_ds`, but rather `struct shmid64_ds` and simply has `#define shmid_ds shmid64_ds` in `<sys/shm.h>`. Thus the header is in fact necessary to use shared memory facilities (though standard `shmat()`, `shmctl()`, `shmdt()`, and `shmget()` are still not provided).

#### Environment
Targeting Android API level 23 for arm-linux-androideabi with a standalone toolchain with unified headers from ndk-r14, using Clang, compiling fio git master (and 2.19).

#### Reproduction
Compile for Android using ndk-r14 using unified headers (may work with older versions of the NDK).

#### Solution
The entire POSIX shared memory compatibility section in `os/os-android.h` should be surrounded with `#ifndef CONFIG_NO_SHM` and also have `#include <sys/shm.h>` instead of `<linux/shm.h>` (or in addition to, but the former `#include`s the latter).
I'll submit a pull.

enh:
that's not quite right. to build with NDK r14/r15 with unified headers targeting API level 26 or above, you need something more like this:

```
--- /tmp/fio-2.20/os/os-android.h	2017-05-19 07:25:27.000000000 -0700
+++ os/os-android.h	2017-05-19 18:52:04.447953599 -0700
@@ -63,7 +63,18 @@
  * necessary stuff here.
  */
 
+#if __ANDROID_API__ >= 26
+#define shmat bionic_shmat
+#define shmctl bionic_shmctl
+#define shmdt bionic_shmdt
+#define shmget bionic_shmget
+#endif
 #include <sys/shm.h>
+#undef shmat
+#undef shmctl
+#undef shmdt
+#undef shmget
+
 #define SHM_HUGETLB    04000
 
 #include <stdio.h>
```

otherwise you'll hit errors about the static redefinitions of shm* conflicting with the non-static declarations in sys/shm.h...

enh:
(if you're thinking it's a wee bit early for any sane person to be targeting O or above, https://android-review.googlesource.com/400614 is me upgrading AOSP's copy of fio to 2.20 :-) )

omor1:
[...]

PR #353, opened 2017-05-01
os/os-android.h: fix shared memory support

omor1:
Fixes: #352 ("Android and shared memory")
Signed-off-by: Omri Mor <omor1@asu.edu>


axboe:
Pulled, thanks.

PR #354, opened 2017-05-03
remove redundant _fio_rbd_disconnect, which is already called in fio_rbd_cleanup

liupan1111:
Signed-off-by: Pan Liu <wanjun.lp@alibaba-inc.com>


liupan1111:
@axboe Please take a look, thanks.

axboe:
Looks good to me. Applied, thanks!

PR #357, opened 2017-05-04
Adjustments to support C++ engines

dyniusz:
I successfully built a cpp io engine after a few small code changes. This patch reflects what I had to do and additionally wraps null engine with dummy C++ class as a form of test. I was able to built it with g++4.8 and clang 3.4.

### DIFF
--- a/engines/null.c
+++ b/engines/null.c
@@ -139,8 +139,7 @@ static void fio_null_cleanup(struct thread_data *td)
 
 static int fio_null_init(struct thread_data *td)
 {
-	struct null_data *nd = (struct null_data *)td->io_ops_data;
-	return null_init(td, &nd);
+	return null_init(td, (struct null_data **)&td->io_ops_data);
 }
 
 static struct ioengine_ops ioengine = {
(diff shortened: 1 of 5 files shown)


sitsofe:
Hi @dyniusz ! I'll note this change makes the null engine quite complicated as an example of an fio C ioengine - perhaps it would be better to move all the C++ bits into their own file?

(Also any ideas why only one Travis build segfaulted rather than all of them?)

dyniusz:
> Hi @dyniusz ! I'll note this change makes the null engine quite complicated as an example of an fio C ioengine - perhaps it would be better to move all the C++ bits into their own file?

Note that there's zero change to the logic, it's all boilerplate to execute the same code from C and C++ functions.

Moving means I either need to duplicate that logic or make it accessible from the cpp_null context. That should be easy as long as null engine is always linked statically (is it?). Otherwise null would need to built and load every time you want to use cpp_null which sounds problematic. I don't have strong opinion.

> (Also any ideas why only one Travis build segfaulted rather than all of them?)

Good question. This is why I missed it in the first place. Kind of hard to believe that stack address wasn't overwritten during the execution but I see no other explanation.

dyniusz:
@axboe any chance of merging this?

Issue #336, opened 2017-04-13
Read of freed memory when verifying overlapping I/O


sitsofe:
Description of the problem:
When verifying overlapping I/O a read of freed memory can occur.

Steps to reproduce:
1. Run a job like
```
valgrind fio --thread --random_distribution=zipf:1.4 --filename=fiofile --io_limit=32G --verify=crc32c --name=verify --iodepth=32 --bsrange=512-1M --rw=randrw
```

Expected result:
Job to run without valgrind warnings

Actual result:
Something like
```
==18399== Invalid read of size 4
==18399==    at 0x443FEC: io_completed (io_u.c:1833)
==18399==    by 0x443FEC: ios_completed (io_u.c:1922)
==18399==    by 0x443FEC: io_u_queued_complete (io_u.c:1981)
==18399==    by 0x40DAEE: wait_for_completions (backend.c:455)
==18399==    by 0x4661D1: do_io (backend.c:1003)
==18399==    by 0x4661D1: thread_main (backend.c:1703)
==18399==    by 0x599A6B9: start_thread (pthread_create.c:333)
==18399==    by 0x5EBA82C: clone (clone.S:109)
==18399==  Address 0x64e1778 is 72 bytes inside a block of size 88 free'd
==18399==    at 0x4C2EDEB: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==18399==    by 0x4605F1: log_io_piece (iolog.c:286)
==18399==    by 0x46638D: do_io (backend.c:964)
==18399==    by 0x46638D: thread_main (backend.c:1703)
==18399==    by 0x599A6B9: start_thread (pthread_create.c:333)
==18399==    by 0x5EBA82C: clone (clone.S:109)
==18399==  Block was alloc'd at
==18399==    at 0x4C2FB55: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==18399==    by 0x46047E: log_io_piece (iolog.c:213)
==18399==    by 0x46638D: do_io (backend.c:964)
==18399==    by 0x46638D: thread_main (backend.c:1703)
==18399==    by 0x599A6B9: start_thread (pthread_create.c:333)
==18399==    by 0x5EBA82C: clone (clone.S:109)
```

So it looks like ipo is freed in iolog.c:
```c
 263                 __ipo = rb_entry(parent, struct io_piece, rb_node);
[...]
 285                         remove_trim_entry(td, __ipo);
[...]

PR #359, opened 2017-05-10
iolog: fix use after free when two in-flight writes overlap

sitsofe:
(From the comment in https://github.com/axboe/fio/issues/336 )

This is an attempt to try and fix a use after free when an overlap is detected on an in-flight I/O in a way that doesn't just leak memory. It would be simpler to just have completion do the free but in the `io_submit_mode=offload` case completion can happen asynchronously in a different thread to the one that is running `log_io_piece()` and would be racy...

Even with this updated patch I think we actually need locking in `log_io_piece()` and other functions such as `unlog_io_piece()` in the  `io_submit_mode=offload` case to prevent an existing `__ipo` (and any structures it is in) being fiddled with while you're using them (`unlog_io_piece()` can change the rb tree and free an ipo when it is called due to I/O failure).

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -1142,6 +1142,17 @@ static void cleanup_io_u(struct thread_data *td)
 	free_file_completion_logging(td);
 }
 
+static void cleanup_ipo_orphans(struct thread_data *td)
+{
+	struct io_piece *ipo;
+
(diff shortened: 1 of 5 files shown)


sitsofe:
This commit was merged via https://github.com/axboe/fio/pull/343 so closing this pull request.

PR #363, opened 2017-05-16
Allow a percent value for the offset parameter

### LINKED ISSUE #89: Is it possible to write to the second half of the device using --size flag?
I am trying to write one pattern to the half of the device and a different pattern to the other half. This is working fine with `--offset` option. However, I need to calculate the offset explicitly depending on my disk size.
To avoid this wondering if I can use `--size` flag on fio. 
I noticed that `--size=50%` is writing to the first 50% of the device. Since FIO already seem to be calculating the disk size and calculating 50% of it, Is it possible that we could tell the tool to write to the second half of the disk or start writing the size backwards ? Say something like `--size=-50%` or something?  
Please suggest.

### LINKED ISSUE #362: Accept percent value for offset parameter
Similar to how the size parameter accepts a percent value, allow offset to accept a percent value.  This would allow for uniform workloads across raw devices, regardless of the device's size.

cbwest3:
This change would allow the `offset` parameter to have a percent value.

Closes #362, Closes #89

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1072,10 +1072,11 @@ I/O type
 
 .. option:: offset=int
 
-	Start I/O at the given offset in the file. The data before the given offset
-	will not be touched. This effectively caps the file size at `real_size -
-	offset`. Can be combined with :option:`size` to constrain the start and
-	end range that I/O will be done within.
+	Start I/O at the provided offset in the file, given as either a fixed size or
+	a percentage. If a percentage is given, the next ``blockalign``-ed offset
+	will be used. Data before the given offset will not be touched. This
+	effectively caps the file size at `real_size - offset`. Can be combined with
+	:option:`size` to constrain the start and end range of the I/O workload.
 
 .. option:: offset_increment=int
 
--- a/filesetup.c
+++ b/filesetup.c
@@ -833,12 +833,35 @@ static unsigned long long get_fs_free_counts(struct thread_data *td)
 uint64_t get_start_offset(struct thread_data *td, struct fio_file *f)
 {
 	struct thread_options *o = &td->o;
+	unsigned long long align_bs;  /* align the offset to this block size */
+	unsigned long long offset;  /* align_bs-aligned offset */
 
 	if (o->file_append && f->filetype == FIO_TYPE_FILE)
 		return f->real_file_size;
 
-	return td->o.start_offset +
-		td->subjob_number * td->o.offset_increment;
+	if (o->start_offset_percent > 0) {
+
+		/* if blockalign is provided, find the min across read, write, and trim */
+		if (fio_option_is_set(o, ba)) {
+			align_bs = (unsigned long long) min(o->ba[DDIR_READ], o->ba[DDIR_WRITE]);
+			align_bs = min((unsigned long long) o->ba[DDIR_TRIM], align_bs);
+		} else {  /* else take the minimum block size */
+			align_bs = td_min_bs(td);
+		}
+
+		/* calculate the raw offset */
+		offset = (f->real_file_size * o->start_offset_percent / 100) +
+			(td->subjob_number * o->offset_increment);
+
+		/* block align the offset at the next available boundary at
+		   ceiling(offset / align_bs) * align_bs */
+		offset = (offset / align_bs + (offset % align_bs != 0)) * align_bs;
+
+	} else {  /* start_offset_percent not set */
+		offset = o->start_offset + o->start_offset +
+			td->subjob_number * o->offset_increment;
+	}
+	return offset;
 }
 
 /*
--- a/options.c
+++ b/options.c
@@ -1381,6 +1381,21 @@ static int str_gtod_reduce_cb(void *data, int *il)
 	return 0;
 }
 
+static int str_offset_cb(void *data, unsigned long long *__val)
+{
+	struct thread_data *td = cb_data_to_td(data);
+	unsigned long long v = *__val;
+
+	if (parse_is_percent(v)) {
+		td->o.start_offset = 0;
+		td->o.start_offset_percent = -1ULL - v;
+		dprint(FD_PARSE, "SET start_offset_percent %d\n", td->o.start_offset_percent);
+	} else
+		td->o.start_offset = v;
+
+	return 0;
+}
+
 static int str_size_cb(void *data, unsigned long long *__val)
 {
 	struct thread_data *td = cb_data_to_td(data);
@@ -1938,6 +1953,7 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
 		.lname	= "IO offset",
 		.alias	= "fileoffset",
 		.type	= FIO_OPT_STR_VAL,
+		.cb	= str_offset_cb,
 		.off1	= offsetof(struct thread_options, start_offset),
 		.help	= "Start IO from this offset",
 		.def	= "0",
--- a/thread_options.h
+++ b/thread_options.h
@@ -200,6 +200,7 @@ struct thread_options {
 	unsigned int numa_mem_prefer_node;
 	char *numa_memnodes;
 	unsigned int gpu_dev_id;
+	unsigned int start_offset_percent;
 
 	unsigned int iolog;
 	unsigned int rwmixcycle;
@@ -469,7 +470,7 @@ struct thread_options_pack {
 	uint8_t log_gz_cpumask[FIO_TOP_STR_MAX];
 #endif
 	uint32_t gpu_dev_id;
-	uint32_t pad;
+	uint32_t start_offset_percent;
 	uint32_t cpus_allowed_policy;
 	uint32_t iolog;
 	uint32_t rwmixcycle;
--- a/cconv.c
+++ b/cconv.c
@@ -104,6 +104,7 @@ void convert_thread_options_to_cpu(struct thread_options *o,
 	o->file_size_low = le64_to_cpu(top->file_size_low);
 	o->file_size_high = le64_to_cpu(top->file_size_high);
 	o->start_offset = le64_to_cpu(top->start_offset);
+	o->start_offset_percent = le32_to_cpu(top->start_offset_percent);
 
 	for (i = 0; i < DDIR_RWDIR_CNT; i++) {
 		o->bs[i] = le32_to_cpu(top->bs[i]);
@@ -543,6 +544,7 @@ void convert_thread_options_to_net(struct thread_options_pack *top,
 	top->file_size_low = __cpu_to_le64(o->file_size_low);
 	top->file_size_high = __cpu_to_le64(o->file_size_high);
 	top->start_offset = __cpu_to_le64(o->start_offset);
+	top->start_offset_percent = __cpu_to_le32(o->start_offset_percent);
 	top->trim_backlog = __cpu_to_le64(o->trim_backlog);
 	top->offset_increment = __cpu_to_le64(o->offset_increment);
 	top->number_ios = __cpu_to_le64(o->number_ios);
--- a/fio.1
+++ b/fio.1
@@ -904,7 +904,11 @@ If true, use buffered I/O.  This is the opposite of the \fBdirect\fR parameter.
 Default: true.
 .TP
 .BI offset \fR=\fPint
-Offset in the file to start I/O. Data before the offset will not be touched.
+Start I/O at the provided offset in the file, given as either a fixed size or a
+percentage. If a percentage is given, the next \fBblockalign\fR-ed offset will
+be used. Data before the given offset will not be touched. This effectively
+caps the file size at (real_size - offset). Can be combined with \fBsize\fR to
+constrain the start and end range of the I/O workload.
 .TP
 .BI offset_increment \fR=\fPint
 If this is provided, then the real offset becomes the
--- a/server.h
+++ b/server.h
@@ -49,7 +49,7 @@ struct fio_net_cmd_reply {
 };
 
 enum {
-	FIO_SERVER_VER			= 61,
+	FIO_SERVER_VER			= 62,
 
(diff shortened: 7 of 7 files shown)


axboe:
This looks good. Like Sitsofe said, cconv.c needs an update, and so does fio.1 to add the parameter to the man page too.

WRT packing, put the variable in near the other uint32_t variables, and adjust the padding. There just search for 'pad' in that structure, adding new options either need a new pad element, or an old one removed since it now packs correctly. When you do that, you should also bump the fio server version in server.h, since the protocol will now be slightly different due to the options struct being different.

sitsofe:
@cbwest3 any chance you could rebase these changes and perhaps squash your two real changes together?

axboe:
Looks like you're making good progress. Sitsofe, submitters should never have to rebase. If there's a conflict, I can fix it up. But the changes do need to be squashed.

sitsofe:
@axboe my bad - my apologies . I agree the changes look good.

I have a feeling the committer and author ids are different in https://github.com/axboe/fio/pull/363/commits/bf9d176df0c62f8abda85244e54d742a6d9fb214 . You can see this more clearly with `git log --format=full` .

I have one question though - what should happen with regard to alignment? Normally if I specify offset=2k bs=4k it's because I'm purposely trying to generate unaligned I/O but if I use offset=3% this might happen by accident. I'm not sure you can use `blockalign` to fix it back up...

axboe:
We have to round it to the minimum block size when used with percentages, it's too coarse to be a useful means of specifically unaligning IO. Or blockalign, if that has been set.

options.c
```diff
@@ -1938,6 +1953,7 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
 		.lname	= "IO offset",
 		.alias	= "fileoffset",
 		.type	= FIO_OPT_STR_VAL,
+		.cb	= str_offset_cb,
```
cbwest3: Thanks for the comment.  I believe this is addressed by reducing the starting_offset_percent to a `uint8_t` in a subsequent commit.

cbwest3:
@axboe and @sitsofe I don't see a way to reply specifically to your comments, so I will try to summarize:

1) "any chance you could rebase these changes and perhaps squash your two real changes together?" BW: Yes, I will work on this, still getting used to the github workflow.
2) "I have a feeling the committer and author ids are different in bf9d176 . You can see this more clearly with git log --format=full ." BW: Although commits may be authored by brantley@$JOB, cbwest is the preferred email (new commits should reflect this).
3) Regarding alignment "We have to round it to the minimum block size when used with percentages, it's too coarse to be a useful means of specifically unaligning IO. Or blockalign, if that has been set." BW: Would it be better to round **UP** to the next `blockalign-ed` value?  The docs state that fio will not write below a given offset value.  I think we would want to honor this in the case of a percent value.

Please let me know if I've missed something.  Thank you both for the review and helpful comments.

sitsofe:
@cbwest3:

2. No problem - just so long as you're happy with either one.
3. I guess I view this as a use case:

    ```
    fio --stonewall --rw=read --name=firstquarter --size=25% --offset=0% --name=secondquarter --size=25% --offset=25% --name=thirdquarter --size=25% --offset=50%
    ```

    In the example job I'd say most people will expect the following offset to be as close to end of the previous size as possible without causing an overlap. Does this force any particular alignment direction?

options.c
axboe: You need cconv.c conversions regardless of the data type used.

cbwest3: cconv.c conversions have been added.

filesetup.c
```diff
@@ -837,8 +837,13 @@ uint64_t get_start_offset(struct thread_data *td, struct fio_file *f)
 	if (o->file_append && f->filetype == FIO_TYPE_FILE)
 		return f->real_file_size;
 
-	return td->o.start_offset +
-		td->subjob_number * td->o.offset_increment;
+	if (o->start_offset_percent > 0) {
+		return (f->real_file_size * o->start_offset_percent / 100) +
+			(td->subjob_number * td->o.offset_increment);
```
sitsofe: @cbwest3 Can you round this to blockalign if it is set or to the minimum block size if it is not?

thread_options.h
```diff
@@ -328,6 +329,7 @@ struct thread_options_pack {
 	uint8_t opendir[FIO_TOP_STR_MAX];
 	uint8_t ioengine[FIO_TOP_STR_MAX];
 	uint8_t mmapfile[FIO_TOP_STR_MAX];
+	uint16_t start_offset_percent;
```
sitsofe: It might be easier to add this to the end of the struct and then you won't have to add the pads...

sitsofe:
@axboe Do you have opinion on the direction of rounding for percentage offsets? As @cbwest3 said the [documentation says this](http://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-offset):
> The data before the given offset will not be touched.

Which hints you have to round offset percentages **up** to the nearest 4k. Doing this will cause a "hole" not to be read if you have a 1Mbyte file and do:

```
fio --filename=onembyte --bs=4k --stonewall --rw=read --name=first --size=33% --offset=0% --name=last --offset=33%
```

I guess if it turned out to be a problem something like `offset_perc_rounding=down` could be introduced in the future though...

wyysoft:
My FIO parameter file is:
```
[global]
ioengine=libaio
buffered=0
filename=/dev/nvme0n1

[file1]
rw=write
bs=1m
size=1%
offset=0%
write_iolog=test1

[file2]
stonewall
rw=write
bs=1m
size=1%
offset=50%
write_iolog=test2
```


I was hoping [file2] to write from the beginning of half-size of the disk and write 1% size of total disk capacity, but this turned out only 0.5% size of total disk capacity is written in [file 2]:
```
  WRITE: bw=936MiB/s (981MB/s), 936MiB/s-936MiB/s (981MB/s-981MB/s), io=2441MiB (2560MB), run=2608-2608msec

Run status group 1 (all jobs):
  WRITE: bw=943MiB/s (989MB/s), 943MiB/s-943MiB/s (989MB/s-989MB/s), io=1220MiB (1279MB), run=1294-1294msec
```


Then I changed offset setting in my fio parameter file to
`offset=30%`

FIO prompted 
`fio: io_u error on file /dev/nvme0n1: Invalid argument: write offset=76818154291, buflen=1048576
file1: No I/O performed by libaio, perhaps try --debug=io option for details?
`

Anything wrong?

sitsofe:
@wyysoft Which version of fio are you are using?

Mainline fio doesn't support percentages for offsets - see https://github.com/axboe/fio/issues/362 . This pull request is about adding support for it but it hasn't been added yet.

wyysoft:
@sitsofe  I am using [https://github.com/axboe/fio/tree/d44a749fb789d3d08fb7ede39bd3cced433a80da](url)

sitsofe:
@wyysoft

```
fio: io_u error on file /dev/nvme0n1: Invalid argument: write offset=76818154291, buflen=1048576
```
76818154291 isn't a multiple of 1Mbyte but more importantly is it isn't a multiple of 4096 (I'm guessing that's your logical block size because you seem to be using an NVMe device but you'll have to check for yourself) and you're using a synonym of `direct=1` (`buffered=0`) which requires you do I/O which is aligned to the device's logical block (see `O_DIRECT` under http://man7.org/linux/man-pages/man2/open.2.html#NOTES ).

See the comments above in this pull request about the need to do alignment.

wyysoft:
@sitsofe 
Got it. Thanks!
'offset=percentage' is not making life easier, for now~

filesetup.c
sitsofe: (I guess `blocksize_unaligned` will also need to be checked)

sitsofe:
@wyysoft No problem. It's nice to know people are using this stuff though :-)

filesetup.c
```diff
@@ -833,12 +833,26 @@ static unsigned long long get_fs_free_counts(struct thread_data *td)
...
-	return td->o.start_offset +
-		td->subjob_number * td->o.offset_increment;
+	if (o->start_offset_percent > 0) {
+		// calculate the raw offset
+		offset = (f->real_file_size * o->start_offset_percent / 100) +
+			(td->subjob_number * td->o.offset_increment);
+
+		// block align the offset at the next available boundary
+		offset = (offset / o->ba[o->td_ddir] +
+			(offset % o->ba[o->td_ddir] != 0))  // ceiling(offset/blockalignment)
```
sitsofe: I'm not sure you can use `o->td_ddir` to index into `o->ba`. Isn't o->ba[] only 3 big (DDIR_RWDIR_CNT 
 comes from https://github.com/axboe/fio/blob/c1f50f765a0a51931605c1fb223d166e3b3a93c6/io_ddir.h#L8 and ba is defined in https://github.com/axboe/fio/blob/master/thread_options.h#L81 ) but o->td_ddir is kind of a bitfield (see https://github.com/axboe/fio/blob/c1f50f765a0a51931605c1fb223d166e3b3a93c6/io_ddir.h#L29 ).

HOWTO
```diff
@@ -1072,10 +1072,13 @@ I/O type
...
-	will not be touched. This effectively caps the file size at `real_size -
-	offset`. Can be combined with :option:`size` to constrain the start and
-	end range that I/O will be done within.
+	Start I/O at the provided offset in the file, given as either a fixed size or
+	a percentage. The data before the given offset will not be touched. This
+	effectively caps the file size at `real_size - offset`. Can be combined with
+	:option:`size` to constrain the start and end range that I/O will be done
+	within.  It is also possible to give offset as a percentage between 1 and
+	100. If ``offset=20%`` is given, fio will skip the first 20% of the file and
+	begin I/O.
```
sitsofe: Perhaps mention "(rounding up to the nearest block if necessary)"? What do you think?

filesetup.c
axboe: Yeah, o->td_ddir is not the way to go. You want something ala:

`if (fio_option_is_set(o, ba)) {
    align_bs = min(td->o.ba[DDIR_READ], td->o.ba[DDIR_WRITE]);
    align_bs = min(td->o.ba[DDIR_TRIM], align_bs);
} else
    align_bs = td_min_bs(td);
`

thread_options.h
cbwest3: Fixed, and removed some padding in the process.

HOWTO
cbwest3: I added a note about block alignment with an offset percentage, further suggestions are welcomed.

filesetup.c
cbwest3: Yeah, I got off in the weeds on this.  @axboe I incorporated your suggestion for calculating `align_bs` and now use that for aligning offset percentage values. Currently the offset is only aligned when provided as a percentage.  Do we want to apply the same treatment to fixed-value offsets?  Thanks all for the comments.

sitsofe: @cbwest3 I don't think you do want it for fixed value offsets because people use those for purposely misaligning I/O (because you have exact granularity).

sitsofe: @cbwest3 also: thank YOU for making these improvements!

filesetup.c
```diff
@@ -833,12 +833,37 @@ static unsigned long long get_fs_free_counts(struct thread_data *td)
...
+	unsigned long long offset;  // align_bs-aligned offset
 
 	if (o->file_append && f->filetype == FIO_TYPE_FILE)
 		return f->real_file_size;
 
-	return td->o.start_offset +
-		td->subjob_number * td->o.offset_increment;
+	if (o->start_offset_percent > 0) {
+
+		// if blockalign is provided, use take the min across read, write, and trim
```
sitsofe: Small thinko? - use take the min

thread_options.h
```diff
@@ -428,6 +429,7 @@ struct thread_options_pack {
 
 	uint32_t perc_rand[DDIR_RWDIR_CNT];
 
+	uint16_t start_offset_percent;
```
axboe: Just make this uint32_t, please. Ideally we don't want anything unaligned in the packed structure, that means that 32-bit entries should be 32-bit aligne. If you add a 16-bit value, then that shifts everything 16-bits.

Fix that up, and the typo Sitsofe noticed, and I think we're ready to get this merged!

sitsofe:
@axboe I was wondering about the misalignment too but as I looked I started wondering about going from `ddir_seq_add` to `size` wrt to alignment:
```c
[...]
	uint64_t ddir_seq_add;
	uint32_t iodepth;
	uint32_t iodepth_low;
	uint32_t iodepth_batch;
	uint32_t iodepth_batch_complete_min;
	uint32_t iodepth_batch_complete_max;

	uint64_t size;
```
Is it the case that we are saying `size` only has to be 32 bit aligned assuming `ddir_seq_add` was 64 bit aligned?

axboe:
Ideally, everything should be aligned. Much better to fix things up, rather than add a new 16 bit type to the mix.

filesetup.c
```diff
@@ -833,12 +833,37 @@ static unsigned long long get_fs_free_counts(struct thread_data *td)
...
+		if (fio_option_is_set(o, ba)) {
+			align_bs = min((unsigned long long) o->ba[DDIR_READ],
```
cbwest3: Done, reduced casting to minimum necessary when calculating align_bs.

thread_options.h
cbwest3: Done!

filesetup.c
cbwest3: I clarified the inline comment, let me know if there is that needs addressing.

axboe:
Looks good to me. One small ask - can you replace the c++ comments with normal c style comments? That follows the style used elsewhere in fio. Code wise, I'm happy with it now, thanks a lot of persevering and pulling this to completion!

cbwest3:
Comments are fixed. Thanks for the review!

axboe:
Pulled! Thanks.

PR #368, opened 2017-05-20
flist.h: replace offsetof macros by stddef.h include

omor1:
`stddef.h` is provided by the compiler and has a builtin `offsetof` implementation if it is available. For instance, both GCC and Clang have the following (or very similar):
```C
#define offsetof(t, d) __builtin_offsetof(t, d)
```

### DIFF
--- a/configure
+++ b/configure
@@ -1930,16 +1930,7 @@ fi
 cat > $TMPC << EOF
 #include <assert.h>
 #include <stdlib.h>
-#undef offsetof
-#ifdef __compiler_offsetof
-#define offsetof(TYPE,MEMBER) __compiler_offsetof(TYPE,MEMBER)
-#else
-#define offsetof(TYPE, MEMBER) ((size_t) &((TYPE *)0)->MEMBER)
-#endif
-
-#define container_of(ptr, type, member) ({			\
-	const typeof( ((type *)0)->member ) *__mptr = (ptr);	\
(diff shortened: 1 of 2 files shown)


sitsofe:
I've wondered about this one too - https://www.spinics.net/lists/fio/msg05514.html . It could be that stddef.h is already referenced elsewhere so you might not need to reference it here...

omor1:
@sitsofe I actually submitted this because of one of your comments. Since _technically_ `offsetof` is defined in `stddef.h`, it's good practice to include it, but it probably is included in other headers or by the compiler automatically.

omor1:
I actually just ran into a bug caused by the hand-rolled `offsetof`, albeit in the configure script, which leads to incorrect identification that the compiler does not support `_Static_assert`. I'll be adding fixes for that to this pull.

sitsofe:
@omor1 Good catch!

axboe:
I'll pull this in. I checked as far back as gcc-4.4, and it does provide offsetof() in stddef.h.

omor1:
@axboe offsetof() is part of ANSI C, so it should exist in all compilers, barring some ancient ones that are unlikely to work for this codebase in any case.

axboe:
I'm old enough to have used plenty of compilers that didn't support it, hence I'm wary or just removing it without first checking. But at this point it does look like we have support in everything we care about, so I'm fine with killing it. Heck, just checked now, and even 2.95 has it.

PR #369, opened 2017-05-20
Android: add support for cgroups

omor1:
Android is Linux too; there's no point in denying cgroups support. It appears that the AOSP copy of fio has been enabling it for a while anyway. Copying @enh on this as he does the AOSP fio stuff.


axboe:
Pulled, thanks.

Issue #356, opened 2017-05-04
Android: SIGBUS due to unaligned access


omor1:
#### Description
Running `fio` on Android (on an ARM device) with any job results in `SIGBUS` being sent to the process. By using `ndk-stack`, I have determined that it is due to `BUS_ADRALN`, or an unaligned memory access.
I have tracked this down to [line 1354 in init.c](../blob/f52e919826839eaba90903b67fe02042159a0023/init.c#L1354). Removing the access still results in `SIGBUS`, but this time from [line 1615 in stat.c](../blob/f52e919826839eaba90903b67fe02042159a0023/stat.c#L1615). Removing *that* access results in a working executable.

However, ensuring alignment of the structure by using the `aligned(8)` attribute did not change anything—it still crashed due to `SIGBUS`.

Furthermore, accessing the fields separately also works. That is, instead of `td->ts.clat_percentiles = o->clat_percentiles;`, using the following:
```C
/* arithmetic to prevent `myvar' from being optimized out */
uint64_t myvar = o->clat_percentiles+1;
td->ts.clat_percentiles = myvar-1;
```
Without the arithmetic, it fails—I assume that the variable is simply optimized out of existence (even at `-O0`).

I am truly mystified as to what's going on here.

#### Environment
Android 6.0.1, Linux 3.4.0 (with numerous device-specific patches, of course), ARMv7a
ndk-r14, Clang 3.8
fio git master

omor1:
Since `struct thread_options *o = &td->o;`, I wonder if this somehow has to do with race on reading and writing from the same pointer? Since it's equivalent to `td->ts.clat_percentiles = td->o.clat_percentiles;`.

sitsofe:
@omor1 With work (e.g. you may need to use a define to quiet down some of the false positives due to https://www.spinics.net/lists/fio/msg05514.html ) you might be able to get fio to compile under clang's undefined behaviour sanitizer. That might in turn may give some clues...

omor1:
@sitsofe I'll take a look at trying that, maybe it will help figure out what exactly is going on.

In any case, I also noticed that while `struct thread_stat` is packed, `struct thread_data` is not. Declaring it with `__attribute__((packed))` as well somehow fixes the problem too. This just gets more and more bizarre (`struct thread_options` is not packed, but `struct thread_options_pack` is, as its name suggests).

sitsofe:
[...]

PR #371, opened 2017-05-22
os/os-android.h: fix alignment problems in shared memory functions

omor1:
Fixes: #356 ("Android: SIGBUS due to unaligned access")

Signed-off-by: Omri Mor <omor1@asu.edu>


axboe:
Pulled, thanks.

PR #372, opened 2017-05-22
os/os-android.h: fix compilation for Android O

omor1:
As noted by @enh on #352, the current fix doesn't work for Android O (API level 26). This allows things to work again (by not including `<sys/shm.h>` and defining the single thing needed from it manually).


axboe:
Pulled, thanks.

PR #373, opened 2017-05-22
fio: fix some struct alignment issues

sitsofe:
Fix unaligned/misaligned accesses related to struct thread_stat and
struct jobs_eta seen when running a build produced by
CC=~/clang-3.9/build/bin/clang ./configure --disable-optimizations \
   --extra-cflags="-D__compiler_offsetof=__builtin_offsetof \
   -fsanitize=undefined"

and add to the compile time asserts to make these problems more visible.

### DIFF
--- a/libfio.c
+++ b/libfio.c
@@ -353,14 +353,17 @@ int initialize_fio(char *envp[])
 	 * can run into problems on archs that fault on unaligned fp
 	 * access (ARM).
 	 */
+	compiletime_assert((offsetof(struct thread_data, ts) % sizeof(void *)) == 0, "ts");
 	compiletime_assert((offsetof(struct thread_stat, percentile_list) % 8) == 0, "stat percentile_list");
 	compiletime_assert((offsetof(struct thread_stat, total_run_time) % 8) == 0, "total_run_time");
 	compiletime_assert((offsetof(struct thread_stat, total_err_count) % 8) == 0, "total_err_count");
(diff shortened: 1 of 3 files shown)


libfio.c
```diff
@@ -353,14 +353,17 @@ int initialize_fio(char *envp[])
 	 * can run into problems on archs that fault on unaligned fp
 	 * access (ARM).
 	 */
+	compiletime_assert((offsetof(struct thread_data, ts) % 8) == 0, "ts");
```
axboe: Should that be sizeof a pointer instead of 8 here? The others are all explicit 64-bit types, hence the 8.

sitsofe: You're right Jens - I'll push an update now...

axboe: Actually, why isn't this aligned to begin with? We don't pack struct thread_data, so I'm puzzled why this would ever be unaligned?

sitsofe: Because we told the compiler "the ts structure is packed" which means even though ts in within the thread_data struct the compiler doesn't have to care about aligning it. See https://github.com/axboe/fio/issues/356#issuecomment-302919781 for a worked example (alternatively keep the asserts but don't add `__attribute__ ((aligned))` and see if you trigger something.

axboe:
Looks good to me. Not sure why we need the align, but you are right, it aligns badly (32-bit here, should be 64) and the jobs_eta issue is explicitly badly aligned.

This should bump the server version, though.

libfio.c
axboe: Ah right, it all makes sense now.

PR #374, opened 2017-05-22
server: bump protocol version

sitsofe:


axboe:
The changelog should have included why this is done. I did that in the merge commit instead.

sitsofe:
Thanks Jens!

PR #380, opened 2017-06-05
Added information about minimal output to documentation

ianchakeres:
Added a single line that contains a single word short-name for each of the fields included in the terse minimal output.

This information can be used in conjunction with minimal output to more easily identify which results are associated with which field.


ianchakeres:
@axboe - here's the PR adding some additional information about the terse/minimal v3 output format.

axboe:
Pulled, thanks.

PR #386, opened 2017-06-17
<README/pkg: Add reference to Arch Linux package>

bobsaintcool:
A package is provided into Arch Linux repos

### DIFF
--- a/README
+++ b/README
@@ -108,6 +108,10 @@ Mandriva:
 	Mandriva has integrated fio into their package repository, so installing
 	on that distro should be as easy as typing ``urpmi fio``.
 
+Arch Linux:
+        An Arch Linux package is provided under the Community sub-repository:
+        https://www.archlinux.org/packages/?sort=&q=fio
+
 Solaris:
 	Packages for Solaris are available from OpenCSW. Install their pkgutil
 	tool (http://www.opencsw.org/get-it/pkgutil/) and then install fio via


README
```diff
@@ -108,6 +108,10 @@ Mandriva:
 	Mandriva has integrated fio into their package repository, so installing
 	on that distro should be as easy as typing ``urpmi fio``.
 
+Arch Linux:
+        Arch Linux package is provided under community.
```
sitsofe: Perhaps saying "An Arch Linux package is provided by the community" rather than "Arch Linux package is provided under community" sounds a little smoother?

bobsaintcool: This was more to emphase that the package under the "sub-repository" named community.

Eventually we can skip the community thing because it could eventually be understandable as the repository maintained by standard user (AUR).

axboe: Can we make it "community sub-repository" then? I agree that if you aren't familiar with Arch, then the current version reads a bit funny.

axboe: Just checked, and Arch already has version 2.21. Those guys move quick! I wish other distros were better at keeping uptodate.

bobsaintcool: I've provided an update with your feedback, I think its a good trade-off for anyone who is not familiar with Arch too.

Indeed, lot of thing go really quick, the maintainer do a good job!

README
```diff
@@ -109,8 +109,8 @@ Mandriva:
 	on that distro should be as easy as typing ``urpmi fio``.
 
 Arch Linux:
-        Arch Linux package is provided under community.
-        https://www.archlinux.org/packages/?sort=&q=fio&maintainer=&flagged=
+        Arch Linux package is provided under community sub-repository:
```
sitsofe: I'd still recommend a minor change like "An Arch Linux package is provided under the Community sub-repository" to help guide the reader to your intent...

axboe:
Final result looks good. Could you collapse those three commits into one single proper one? Then I can pull it in.

bobsaintcool:
not familliar with rebase, but I thing its okay

axboe:
Looks good to me! Pulled, thanks.

Issue #324, opened 2017-03-06
nanosecond timing for fio


vincentkfu:
I have been carrying out some research on providing nanosecond accuracy for timing in fio and would appreciate feedback on my efforts so far.

# Background
C has two structures for recording time values
fio currently uses 
```
struct timeval
  {
    __time_t tv_sec;        /* Seconds.  */
    __suseconds_t tv_usec;  /* Microseconds.  */
  };
```
for nanosecond timing we need 
```
struct timespec
  {
    __time_t tv_sec;        /* Seconds.  */
    __syscall_slong_t tv_nsec;  /* Nanoseconds.  */
  };
```

# Design Goals

* minimize the performance impact
  * avoid a command-line option to enable/disable nanosecond timing because this will require an if statement in the hot timing code
  * instead, choose whether to enable or disable nanosecond timing at compile time
* keep the changes simple
  * avoid introducing bugs
  * code will be easier to maintain
* avoid the appearance of nanosecond accuracy when it is actually not available

# Proposed Implementation
* at compile time choose to support ns timing on platforms where the CPU timer is available
```
#ifdef NANOSECOND_TIMING
typedef struct timespec fio_time;
#else
typedef struct timeval fio_time;
#endif
``` 
* with the patch, all fio time functions will use the fio_time type instead of struct timeval
* at run time fio determines whether the CPU timer is reliable
  * if it is unreliable, fio will use OS-provided timing facilities
  * this will provide the appearance of ns accuracy when it is actually not available
  * modern Intel CPUs all have reliable time stamp counters
* this requires much more extensive code changes because timing is all over the place in fio
* design goals: meets 1 and 3, but requires extensive code changes (2)

# Current challenges
* Thoroughly testing fio to check if changes introduced any bugs
* Integer arithmetic
  * If you have a 3 GHz CPU, then ticks per ns = 3 and ns = (clock ticks) / 3
  * What if you have a 3.66 GHz CPU? You will have poor accuracy because you will still end up with ns = (clock ticks) / 3 
  * Partial solution
    * ns = (clock ticks) * FACTOR / (3.66 * FACTOR) = (clock ticks) * (FACTOR / 3.66) / FACTOR
    * If FACTOR is a power of 2 this division is fast
    * This is what fio does now for microseconds: fio does this not for the sake of accuracy but because division by a power of 2 is a relatively fast operation
    * However, (clock ticks * FACTOR) can overflow
    * this overflow in fio's timing code led to regular stuck jobs
[...]

PR #387, opened 2017-06-22
Nanosecond timing for fio

vincentkfu:
Jens, please consider this pull request that adds nanosecond timing to fio (#324).

The changes include the following:

- Change most of the places that used struct timeval to instead use struct timespec
-  Add nsec bins for the latency counters: io_u_plat[][] in thread_stat is much larger now to accommodate the increased precision while retaining the top end ~9sec max latency value
- Change the normal and JSON output to accommodate nsec latencies. This will break scripts that parse fio output.
- Add t/time-test to carry out experiments with various ways of doing the clock ticks to nsec conversion
- For the clock ticks to nsec conversion I chose a two-stage approach to resolve the 64-bit overflow problem
-- First, consume hours' worth of clock ticks at a time
-- Then, carry out the standard multiply and shift conversion with the remaining clock ticks
-- This provides the means to accurately track small differences in time while simultaneously dealing with long periods of time
-- I was never able to get the seqlock to work; the time values would always be corrupted as if the lock was not really being held
-- This two-stage approach does a little bit more work for each call compared to using a seqlock, but it avoids the latency spike that occurs when the CPU clock must be read a second time because the lock is held
-- Here is a comparison of the compiler output for different clock ticks to nsec conversion strategies: https://godbolt.org/g/uN99Gb

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -3189,7 +3189,7 @@ Split up, the format is as follows:
 
         Total IO (KiB), bandwidth (KiB/sec), IOPS, runtime (msec)
         Submission latency: min, max, mean, stdev (usec)
-        Completion latency: min, max, mean, stdev(usec)
+        Completion latency: min, max, mean, stdev (usec)
         Completion latency percentiles: 20 fields (see below)
         Total latency: min, max, mean, stdev (usec)
         Bw (KiB/s): min, max, aggregate percentage of total, mean, stdev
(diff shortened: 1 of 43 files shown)


axboe:
Stellar work, Vincent! Thanks for following through on this. I looked over the code, and it looks mostly straight forward to me. I made some style changes, that is all. It's committed. Time to celebrate, now you don't have to carry this work forward anymore :-)

PR #388, opened 2017-06-26
Update HOWTO updates

sitsofe:
A chunk of HOWTO updates ranging from spelling/grammar fixes to updating example fio output and adding missing fields. There are lot of changes and there's always more to do but I figure it's more useful to get these out there sooner...


axboe:
Thanks, pulled!

PR #389, opened 2017-06-26
Fix group percent

sitsofe:
When samples was added it started generating warning when compiling on OSX. Further when the kilobytes/kibibytes change was introduced it broke the group percentage calculation. These patches should fix both of these up.

Sadly I've noticed the group percentage sometimes gets close to 100% when using one thread in the group but without actually being 100%. This might be because the aggregate bandwidth is exact whereas the mean used is from samples which introduces a "average of averages" scenario...


axboe:
Thanks, pulled.

PR #393, opened 2017-06-28
stats and documentation fixes

sitsofe:
The first commit finishes off the group percentage fix from a previous commit. The second changes the documentation to state that random_distribution's parameter for a Gaussian distribution is called
"normal" (but curiously file_service_type uses "gauss" for its Gaussian distribution parameter)

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1204,7 +1204,7 @@ I/O type
 		**pareto**
 				Pareto distribution
 
-		**gauss**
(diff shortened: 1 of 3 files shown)


axboe:
Thanks, I'll pull this in. Feel free to add an alias or duplicate parameter for file_service_type, to keep things in sync. Less confusing that way.

PR #394, opened 2017-06-28
fio: make gauss a duplicate of normal for file_service_type

sitsofe:
To be consistent with random_distribution add a normal parameter to
file_service_type and state that gauss is an alias of it.


axboe:
Yep, looks nice. Pulled.

Issue #376, opened 2017-05-31
fio sending writes of  "1 Byte" while writing 10G file 


aspandey:
We have used following command and options of fio tool to test performance of glusterfs - 

fio --randrepeat=1 --ioengine=libaio --direct=0 --gtod_reduce=1 --name=test --filename=fiotest.fio --iodepth=16 --size=10G --readwrite=write
test: (g=0): rw=write, bs=4K-4K/4K-4K/4K-4K, ioengine=libaio, iodepth=16

We have observed a very low performance. So when we captured and analyzed the packets over the network between client  and server we saw that it is sending a lot of writes of length "1 Bytes"

We have used this on nfs client and also glusetrfs client.

We don't know if it is expected form fio or it is a valid bug in fio tool. If it is expected, please explain the reason behind it.

[fio-nfs-without-truncate-pacp.txt](https://github.com/axboe/fio/files/1040997/fio-nfs-without-truncate-pacp.txt)

axboe:
Fio is writing in 4k chunks, if you don't give the bs= setting. So I'm inclined to blame gluster here. One oddity is that you are doing buffered writes, but you use libaio. This won't give you queue depth, as the writes will be sync (in the sense that they are completed on the OS side when returning, they could just be in the page cache). It's also possible that gluster does something weird with fallocate. You could try and add --fallocate=none to your options.

sitsofe:
@aspandey also just to verify: you're I/O you are doing is going to go into a file called "fiotest.fio" within the directory you are standing in when you run fio. If that isn't actually in your gluster filesystem that may explain the results you see...

aspandey:
Actually, gluster does not even comes into picture. 
The file which I have attached, contains packets between nfs client and nfs server.
I think this issue could be reproduced easily if we just write this fio  on a nfs mount point.

Niels (ndevos@redhat.com) has also faced the same issue and his findings are even more interesting. 
His findings () he did directio- 
It seems that fio does a single byte write at the last byte of each page (a page is 4096 bytes. stracing fio shows:
part of the trace of fio command - 
2314  pwrite(3, "\0", 1, 4095 <unfinished ...>
2314  <... pwrite resumed> )            = 1
2314  pwrite(3, "\0", 1, 8191)          = 1
2314  pwrite(3, "\0", 1, 12287)         = 1
2314  pwrite(3, "\0", 1, 16383)         = 1
2314  pwrite(3, "\0", 1, 20479)         = 1
[...]

Issue #22, opened 2014-10-21
Data size always 64 percent when use fio produce data on the nfs storage


MyCoXiao:
Hi,
    There is a problem need your help,  I wrote something by FIO but in certain point something weird happens. Fio version is 2.1.13.
    My command as below:
    fio --directory=./ --direct=1 --rw=randwrite --refill_buffers --norandommap --randrepeat=0 --ioengine=libaio --bs=4k --rwmixread=100 --iodepth=1 --numjobs=1  --group_reporting --name=4ktestwrite --size=100M
    If the directory is local disk, the data actual size(use du -h) is 100M, but if the directory is the nfs share storage the data actual size is 64M. Could you please help to solve this?

This is nfs storage.
Perofrmance-Node ~ # mount
10.121.190.9:/root/lvm on /root/centOS type nfs (rw,vers=4,addr=10.121.190.9,clientaddr=10.121.147.10)
Perofrmance-Node ~ # cd -
/root/centOS
Perofrmance-Node centOS # pwd
/root/centOS
Perofrmance-Node centOS # du -h
16K     ./lost+found
64M     .
Perofrmance-Node centOS # ls -lh
total 64M
-rw-r--r-- 1 nobody nogroup 100M Oct 20 12:02 4ktestwrite.0.0
drwx------ 2 nobody nogroup  16K Aug 13 16:25 lost+found

This is local disk
Perofrmance-Node test # pwd
/mnt/disk/test
Perofrmance-Node test # du -h
101M    .
Perofrmance-Node test # ls -lh
total 101M
-rw-r--r-- 1 root root 100M Oct 17 03:38 4ktestwrite.0.0

   Any assistance will be greatly appreciated!

Thanks

sitsofe:
@MyCoXiao You asked fio to make a 100MByte file and then potentially didn't write to all of it (`--norandommap`) so areas of the file could still be sparse.

On a local filesystem like ext4 fio will be able to issue a `fallocate` operation at layout time to tell the filesystem it needs to preallocate all of the blocks of the file. When fio is writing to your NFS mount the `fallocate` call will fail so the blocks aren't allocated up front - only when written to by the job. Subsequently, parts of the file remain sparse after fio has finished leading to du vs ls returning different file sizes (size of the file in terms of blocks that had to be allocated vs size the file would like to be).

If you want to observe the same sparse file behaviour on the local filesystem try adding `--fallocate=none` to your options.

PR #395, opened 2017-07-03
Fallocate native

sitsofe:
Add a native fallocate method to address https://github.com/axboe/fio/issues/376 (fio sending writes of "1 Byte" while writing 10G file). Also kind of addresses https://github.com/axboe/fio/issues/22 (Data size always 64 percent when use fio produce data on the nfs storage).

Additional patch on top to silence Sphinx warnings when building documentation.


sitsofe:
Groan, just noticed this: https://ci.appveyor.com/project/axboe/fio/build/1.0.209/job/q9n3ckmbmg8awooo#L122 ...

sitsofe:
OK sorted that one too with a bit of code rearranging. What do you think Jens?

axboe:
Pulled, thanks!

PR #397, opened 2017-07-05
filesetup: output more info if test file not exist when use direct io

dublio:
if test file not exist, and user set direct=1, modify error log:

before:
fio: looks like your file system does not support direct=1/buffered=0

after:
fio: /dev/nvme0n1p5 is not exist or your file system does not support direct=1/buffered=0

Signed-off-by: weiping zhang <zhangweiping@didichuxing.com>

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -54,51 +54,51 @@ Command line options
 
 .. option:: --debug=type
 
-    Enable verbose tracing of various fio actions.  May be ``all`` for all types
-    or individual types separated by a comma (e.g. ``--debug=file,mem`` will
-    enable file and memory debugging).  Currently, additional logging is
-    available for:
+	Enable verbose tracing of various fio actions.  May be ``all`` for all types
+	or individual types separated by a comma (e.g. ``--debug=file,mem`` will
+	enable file and memory debugging).  Currently, additional logging is
+	available for:
 
-    *process*
+	*process*
 			Dump info related to processes.
-    *file*
+	*file*
 			Dump info related to file actions.
-    *io*
+	*io*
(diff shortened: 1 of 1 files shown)


szaydel:
I would suggesting changing the message from `is not exist` to `does not exist`, so as to be more grammatically correct.

sitsofe:
I wonder if you already get a different message if you set `allow_file_create=0` (http://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-allow-file-create )...

axboe:
Add a stat() call to differentiate between the two. The error is technically correct, since:

1) The file does not exist
2) Fio tries to open O_DIRECT, and it fails because the file system hosting dev special files does not support O_DIRECT.

However, it is confusing for users. I'd suggest, in the EINVAL section, doing a stat() check to see if the file exists. That information could then be used to improve the error message.

dublio:
"allow_file_create" default value is true, so a regular file named "nvme0n1p5" was created before enter EINVAL section. Then I try to add stat() call , and it return 0. 

`
[root@localhost dev]# ls -lh nvme*
crw------- 1 root root 247, 0 Jul  5 16:31 nvme0
brw-rw---- 1 root disk 259, 0 Jul  5 17:02 nvme0n1
brw-rw---- 1 root disk 259, 7 Jul  5 17:02 nvme0n1p1
-rw-r--r-- 1 root root    10G Jul  6 01:18 nvme0n1p5
`
when I use direct=1 testing block device, there is no error or warning info,
1. seems O_DIRECT can be used for block device
2. I think /dev (devtmpfs) should disable creating regular file.

Issue #408, opened 2017-07-19
Presence of any command line job-option prevents ioengine-specific options in configuration files


benlwalker:
The presence of any command line job-option causes fio to treat all ioengine-specific options in configuration files as invalid. For example, take this configuration file:

```
[global]
ioengine=libaio
userspace_reap=1
...

[job0]
...
```

Running fio with a command line like such:

```
fio <my_config.fio> --iodepth=8
```

results in the following error:

```
Bad option <userspace_reap=1>
```

When fio encounters a command line option for a job (parse_cmd_line - init.c line 2504), if a job does not currently exist it creates a new one with default settings. It then loads the ioengine for that job, which on Linux will be psync. Later, the ioengine option may be encountered. If the ioengine option is on the command line, the code to handle it is in init.c line 2532 and it will unload psync and load the specified ioengine.

When later parsing the configuration file, when an unknown option (i.e. an engine-specific option) is encountered, a list is saved and reprocessed after all known options. However, when the ioengine option (a known option) is encountered, it just updates the string name of the ioengine - it doesn't unload the default engine like it does when the option is on the command line. In the loop to process unknown options it does correctly attempt to load the ioengine, but that call is short-circuited because td->io_ops is already set (because a different ioengine is already loaded).

To recap:

1) All options in configuration file works.
2) All options in command line, if ioengine appears before engine-specific options, works.
3) Engine-specific options that appear before the ioengine option will not work. This is minor - most people will put the ioengine first if they're putting everything on the command line.
4) Engine-specific options on the command line will fail if ioengine is specified in the configuration file. This is a real annoyance for users that have a template config file and then vary engine-specific options on the command line.
5) The presence of any job-option on the command line (like --iodepth) breaks all engine-specific options in configuration files. This is the primary issue.

sitsofe:
1, 2 describe ok behaviour. 5 has been fixed by #409. I'd argue that 3 is kind of a non-issue which leaves 4 as the remaining issue...

redlicha:
Commit 800334db17a22029553488b41a5ede8af909c66d breaks engine-specific options for external engines that are configured within a job section. Example:
```
$ git diff engines/skeleton_external.c
diff --git a/engines/skeleton_external.c b/engines/skeleton_external.c
index 56f89f95..920a25be 100644
--- a/engines/skeleton_external.c
+++ b/engines/skeleton_external.c
[...]

PR #409, opened 2017-07-19
Correctly detect whether ioengine_load can exit early

benlwalker:
Just the presence of io_ops isn't enough - the requested ioengine may have changed.

This fixes number 5 from issue #408. The other problems identified in that issue still remain and would probably take a much bigger set of changes.

Signed-off-by: Ben Walker <benjamin.walker@intel.com>


axboe:
Looks fine, pulled, thanks.

Issue #413, opened 2017-07-26
Where can i get the Json+ format for FIO result?


Cnexlabzhengwang:
hi alex,
I am runing with FIO2.99 for json+ output_format, but i want to know the meanings of "Bins", where can i get the detail info for json/Json+ format for fio output?
See the attachment, please help.
[TestRand_w30_bs4096_threads1_iodepth1_nvme0n1.zip](https://github.com/axboe/fio/files/1175667/TestRand_w30_bs4096_threads1_iodepth1_nvme0n1.zip)

vincentkfu:
In fio 2.99, the bins objects contains key, value pairs where the key is a latency duration and the value is the count of the number of IOs that required that latency for completion.

From your example::
`          "bins" : {
            "87552" : 1,
            "89600" : 1,
            "94720" : 1,
            "96768" : 1,
            "97792" : 1,
            "99840" : 1,
            "100864" : 2,
            "103936" : 6,
            "104960" : 534,
            "105984" : 5995,
            "107008" : 7529,
...
}
   `

One IO required 87552 ns to complete, 2 IOs required 100864ns to complete, 7529 IOs required 107008ns to complete.

The latency duration keys actually represent the midpoint of a variable width latency bin.  Full documentation is in stat.h

sitsofe:
@Cnexlabzhengwang who is Alex?

At any rate @vincentkfu has added brief documentation to http://fio.readthedocs.io/en/latest/fio_doc.html#json-output and has also added a note that full documentation can be read inside `stat.h`. @Cnexlabzhengwang given this, would you agree this issue is resolved and if so can you close this issue? Thanks!

Cnexlabzhengwang:
Hi wheeler,
Sorry, I means Axboe…

发件人: Sitsofe Wheeler [mailto:notifications@github.com]
发送时间: 2017年8月14日 15:37
收件人: axboe/fio <fio@noreply.github.com>
抄送: Zheng Wang <zwang@cnexlabs.com>; Mention <mention@noreply.github.com>
主题: Re: [axboe/fio] Where can i get the Json+ format for FIO result? (#413)


@Cnexlabzhengwang<https://github.com/cnexlabzhengwang> who is Alex?

[...]

PR #415, opened 2017-07-27
Latency output and HOWTO changes

vincentkfu:
Jens, please consider this set of patches.

1. Add a script to generate CSV data from the json+ latency bins
2. Change formatting and labeling of io_u_lat_n/u/m completion latency data in the normal output
3. Change the labelling of the io_u_lat_n/u/m completion latency data in the JSON output
4. Clarify description of io_u_lat_n/u/m and latency target output in the HOWTO
5. Provide details regarding the json+ output format in the HOWTO, addressing #413


axboe:
Most of this looks good, but I don't necessarily like the lat -> clat changes. The definitions, as far as fio and I are concerned, are:

slat: time to submit an IO
clat: time the IO took to complete
lat: clat+slat

It's confusing to talk about clat for sync engines, since the system call involves both a submission and completion latency. Hence it's lat.

vincentkfu:
The same values are used for both lat and clat. So I thought the lat -> clat change was appropriate because the values used are differences between issue time and completion time.

[io_u.c:1786 ](https://github.com/axboe/fio/blob/master/io_u.c#L1786)has
```
			llnsec = ntime_since(&io_u->issue_time, &icd->time);
```

Then starting at [io_u.c:1811 ](https://github.com/axboe/fio/blob/master/io_u.c#L1811) are:

```
			add_clat_sample(td, idx, llnsec, bytes, io_u->offset);
			io_u_mark_latency(td, llnsec);
```
The data structures modified by add_clat_sample are used for

    clat percentiles (nsec):
     |  1.00th=[   18],  5.00th=[   19], 10.00th=[   20], 20.00th=[   21],
     | 30.00th=[   22], 40.00th=[   23], 50.00th=[   23], 60.00th=[   24],
     | 70.00th=[   26], 80.00th=[   27], 90.00th=[   27], 95.00th=[   28],
     | 99.00th=[   42], 99.50th=[   45], 99.90th=[  101], 99.95th=[  165],
     | 99.99th=[ 6368]

The data structures modified by io_u_mark_latency are used for 

    lat (nsec) : 20=15.47%, 50=83.90%, 100=0.49%, 250=0.11%, 500=0.02%
    lat (nsec) : 750=0.01%, 1000=0.01%
    lat (usec) : 2=0.01%, 4=0.01%, 10=0.01%, 20=0.01%, 50=0.01%
    lat (usec) : 100=0.01%, 250=0.01%, 500=0.01%

It's the same latency measurement but in one place they are called 'clat' and in another place they are called 'lat'.

szaydel:
Consider changing the interpreter line in the python script to not explicitly call out python location, instead maybe use `!#/usr/bin/env python`, which I believe is the more accepted standard.

vincentkfu:
Perhaps a better way to resolve this would be to retain the 'lat' label but have io_u_lat_n/u/m actually measure total latency by moving the call to io_u_mark_latency up a few lines and having something like this around [io_u.c:1791](https://github.com/axboe/fio/blob/master/io_u.c#L1791):

```
		tnsec = ntime_since(&io_u->start_time, &icd->time);
		add_lat_sample(td, idx, tnsec, bytes, io_u->offset);
+		io_u_mark_latency(td, tnsec);
```

sitsofe:
@szaydel Whether you use env or not seems to be a matter of preference. There are cases where the env path searching is better and there are other cases where it leads to things breaking (https://www.debian.org/doc/packaging-manuals/python-policy/ch-python.html#s-interpreter ). I'd be tempted to leave it as the full path.

szaydel:
@sitsofe Fair point. I think it depends upon how portable you want this script to be. My sense is that on non-Linux OS's the path to python executable varies quite a bit more than it does on Linux.

PR #416, opened 2017-07-29
filesetup: keep O_DIRECT flag same as test  when pre-allocte file

dublio:
if set direct=1, but fs not support direct io, should use O_DIRECT when
pre-allocte test file, otherwise file open success and then fallocate
will alloc memory for it.

Signed-off-by: weiping zhang <zhangweiping@didichuxing.com>

### DIFF
--- a/filesetup.c
+++ b/filesetup.c
@@ -146,6 +146,8 @@ static int extend_file(struct thread_data *td, struct fio_file *f)
 		flags |= O_CREAT;
 	if (new_layout)
 		flags |= O_TRUNC;
+	if (td->o.odirect)
+		flags |= OS_O_DIRECT;
 
 #ifdef WIN32
 	flags |= _O_BINARY;


sitsofe:
It might be worth rewording the commit message:
```
filesetup: keep O_DIRECT flag when pre-allocating file

If direct=1 then the file open for pre-allocation should also use O_DIRECT otherwise
the initial file open may return success even though the filesystem cannot O_DIRECT
operations on it it.
```

Thoughts?

axboe:
Yeah that's better, honestly I did not understand the first version without looking at the code. @dublio would you mind rewording the commit message? Then I can pull it in.

dublio:
@sitsofe @axboe 
Thanks your review, I update commit message now. 

The first version, use O_DIRECT,  build error occur, then I change O_DIRECT to OS_O_DIRECT at version 2.

dublio:
V3 => V4: only correct a bit of commit message.

sitsofe:
@dublio I've got a minor rewording (sorry I was in a hurry with my previous suggestion):

```
filesetup: keep OS_O_DIRECT flag when pre-allocating file

When direct=1 the file open for pre-allocation should also set OS_O_DIRECT so 
we fail earlier if the filesystem/file doesn't support direct operations. 
```

What do you think?

dublio:
@sitsofe thanks you comment ^_^, I have update commit mssage.

axboe:
Pulled, thanks.

PR #417, opened 2017-07-31
Fio jsonplus patches

vincentkfu:
Jens, I have separated the json+ patches from #415 and put them into this pull request.

Please consider these two patches:

1) Add a script to generate CSV data from the json+ latency bins
2) Provide details regarding the json+ output format in the HOWTO, addressing #413

@szaydel and @sitsofe, thank you for the feedback in #415. Not knowing any better, I just copied the first line in the other python scripts distributed with fio. Since no fio users have complained about the path specified therein, I will leave it as is.

Thanks.


vincentkfu:
I committed another patch because I initially forgot to add the new script to the Makefile.

HOWTO
```diff
@@ -3363,6 +3363,27 @@ minimal output v3, separated by semicolons::
...
+JSON+ output
+------------
+
+The json+ output format is identical to the json output format except that it
+adds a full dump of the completion latency bins. Each `bins` object contains a
+set of (key, value) pairs where keys are latency durations and values count how
+many IOs had completion latencies of the corresponding duration. For example,
+consider:
+
+``"bins" : { "87552" : 1, "89600" : 1, "94720" : 1, "96768" : 1, "97792" : 1, "99840" : 1, "100864" : 2, "103936" : 6, "104960" : 534, "105984" : 5995, "107008" : 7529, ... }``
```
sitsofe: You may want to indent the bins like like https://github.com/axboe/fio/blob/fio-2.99/HOWTO#L3339 . That way it is rendered similarly to the lines above http://fio.readthedocs.io/en/latest/fio_doc.html#trace-file-format ...

HOWTO
```diff
@@ -3363,6 +3363,27 @@ minimal output v3, separated by semicolons::
...
 
 
+JSON+ output
+------------
+
+The json+ output format is identical to the json output format except that it
+adds a full dump of the completion latency bins. Each `bins` object contains a
+set of (key, value) pairs where keys are latency durations and values count how
+many IOs had completion latencies of the corresponding duration. For example,
```
sitsofe: The HOWTO seems to prefer I/O to IO.

HOWTO
vincentkfu: Thanks. I did spend some time studying the markup language but gave up after a while. Would I remove the single quotes too?

vincentkfu: I found an online reStructuredText editor http://rst.ninjs.org/ and it seems like I can omit the single quotes.

vincentkfu:
It seems like deleting the branch and pushing a replacement with the same name is not the right way to get the review changes squashed into the original commits.

sitsofe:
@vincentkfu I think GIthub just expects you to (force) push an updated branch...

PR #424, opened 2017-08-13
Enhance continuous integration builds

sitsofe:
Add a commits that ensure `make test` includes something that tries to use a file on a real filesystem and make CI builds abort on compiler warnings.


axboe:
Pulled, thanks.

PR #425, opened 2017-08-16
Fixes

sitsofe:
A small collection of fixes that fix a format specifier for a `uint64_t` in the rbd ioengine, correct a man page heading and changes that enable more ioengine on travis Linux builds. If travis eventually switches the Ubuntu distribution used from precise to trusty we should automatically start building the rbd ioengine but until then it should disable itself even when an old librbd is installed.


axboe:
Pulled, thanks.

Issue #433, opened 2017-08-22
About  "If a sample's MSB is greater than 23, it will be counted as 23"


football1222:
I'm confused with the setting in the title. As explained in stat.h:107 ** If a sample's MSB is greater than 23, it will be counted as 23, anything >8838608 will be count in the last group 18, which corresponds to [8838608,+inf] . 
However, I don't see any if-statement to ensure that. In **plat_val_to_idx()**,  if the "val"(nsec) is big enough, it will be mapped to group 19,20,21,... . This results goes into **add_clat_percentile_sample()**. Then there is check if IDX <= FIO_IO_U_PLAT_NR, which is 29X64, but not 18X64.
If 29 is the choice, then there's potential risk that in **plat_idx_to_val()**, "base" could exceed unsigned int.
So is there anything to ensure that which I didn't notice or something else? Thank you!

vincentkfu:
The maximum return value of plat_val_to_idx() is (FIO_IO_U_PLAT_NR - 1). So any extremely large values will always be put in the final bucket: https://github.com/axboe/fio/blob/master/stat.c#L91

vincentkfu:
FIO_IO_U_PLAT_GROUP_NR was originally 19 but this was increased to 29 when the unit was changed from usec to nsec:

https://github.com/axboe/fio/commit/d6bb626ef37d3905221ade2887b422717a07af09#diff-b1ee69fb5e2f7a77b82e6adfb0aa6492L111

football1222:
If 29 is the new case, then I think there are two issues:
1. [](https://github.com/axboe/fio/blob/master/stat.h#L102) should be modified.
2. in **plat_idx_to_val()**, base was defined as unsigned int, but 
[](https://github.com/axboe/fio/blob/master/stat.c#L112)
/* Find the group and compute the minimum value of that group */
	error_bits = (idx >> FIO_IO_U_PLAT_BITS) - 1;
	base = 1 << (error_bits + FIO_IO_U_PLAT_BITS);
base would easily exceed 32bits when error_bits reaches 26(when Group Id >= 27), and it should be.
'cause Group 27 should cover [4294967296, 8589934591] if I'm not getting this wrong.
So in my opinion, base should be unsigned long long. 
Am I right about this? Thank you.

PR #434, opened 2017-08-22
base should be unisgned long long

football1222:
In function **plat_idx_to_val()** in stat.c, base should be modified to be unsigned long long.

Variable base was defined as unsigned int, but it was designed to fit the old settings where **FIO_IO_U_PLAT_GROUP_NR** = 18.
However, with the alteration from usec to nsec, this value was set to 29 and the range of latency changed from uint32_t to uint64_t.
At https://github.com/axboe/fio/blob/master/stat.c#L112 in **plat_idx_to_val()**
```
/* Find the group and compute the minimum value of that group */
error_bits = (idx >> FIO_IO_U_PLAT_BITS) - 1;
base = 1 << (error_bits + FIO_IO_U_PLAT_BITS);
```
Now base would easily exceed 32bits range and cause an overflow after error_bits reaches 26(when idx >=27 * 64 and Group Id >= 27 ), and it could happen. 'Cause    
Group 27 should cover [4294967296, 8589934591] , with error_bits = 26
Group 28 [8589934591, 17179869183],  with error_bits = 27
Group 29 [17179869184, +inf], with error_bits = 28
if I'm not getting this wrong.
also see https://github.com/axboe/fio/issues/433

### DIFF
--- a/stat.c
+++ b/stat.c
@@ -100,8 +100,8 @@ static unsigned int plat_val_to_idx(unsigned long long val)
  */
 static unsigned long long plat_idx_to_val(unsigned int idx)
 {
-    unsigned int error_bits;
-    unsigned long long k, base;
+	unsigned int error_bits;
+	unsigned long long k, base;
 
 	assert(idx < FIO_IO_U_PLAT_NR);
 


sitsofe:
I'd recommend against using unsigned long long and suggest uint64_t if possible...

football1222:
Agreed. I did that just to be in accordance with the old style.

axboe:
unsigned long long vs uint64_t aside, the commits are also woefully lacking. A good commit should include a one line description of the fix as the first line, then a blank line, and then a full description of what the issue was and how it's fixed.

Your one-line description says what is being done. I can just read the code for that. It needs to be a description of WHY the change is made.

football1222:
Very good points! Thx~

sitsofe:
@football1222 even more than the description here at github we need the commits themselves to be amended. At the moment we see this: https://github.com/axboe/fio/pull/434/commits . We need the commit messages to be following this type of format: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html .

For example your first commit only says

> plat_idx_to_val:base change to int64

An example of improving its style would be the following:

> stat: increase the size of base to avoid overflow
>
> plat_idx_to_val() in stat.c has a variable base that is defined as an unsigned int but it was designed to fit old settings where FIO_IO_U_PLAT_GROUP_NR = 18. With the alteration from usec to nsec in commit 98789 ("summary line of commit 98789 here") FIO_IO_U_PLAT_GROUP_NR is now 29 and the range of latency changed from uint32_t to uint64_t. This means base can easily exceed a 32 bit range and cause an overflow after error_bits reaches 26.
>
> [further explanation]

For example see the following commit messages
* https://github.com/axboe/fio/commit/d19251c4a191dd024bec82be876f57f7770e32f6
* https://github.com/axboe/fio/commit/e1933299ed9f1525e010e0489f0185c063d6d129
* https://github.com/axboe/fio/commit/90e2cd1e1cfc6c2bb4c5643b4864347445f9571e

stat.c
```diff
@@ -100,8 +100,8 @@ static unsigned int plat_val_to_idx(unsigned long long val)
  */
 static unsigned long long plat_idx_to_val(unsigned int idx)
 {
-    unsigned int error_bits;
-    unsigned long long k, base;
+	unsigned int error_bits;
+	unsigned long long k, base;
```
sitsofe: Is this commit only a whitespace change? If so it would be better for the commit message say that rather than "stat.c base change to int64"...

football1222: I'll try to squash them and write a new commit message.

axboe:
Sorry if I wasn't clear - what Sitsofe says is true. I personally don't care AT ALL about what is on github. I care about what ends up in the git history. So it's the actual commits I care about, not any of the irrelevant meta data on this site.

sitsofe:
@football1222 Thank you for addressing the suggestions!

Issue #426, opened 2017-08-17
FIO IOPS Log and BW Log Issue


jrgruher:
Hi, I'm using FIO 3.0 on Ubuntu 16.10 with 4.12.7 kernel.  I'm running a 4KB random IO pattern against an Intel P3700 400GB NVMe SSD.  I'm trying to view the IOPS/BW over time using the write_bw_log and write_iops_log parameters.  FIO is writing the logs but the values in the logs don't align with the summary for the run.

Here is my INI:
[global]
bs=4k
rw=randread
numjobs=4
iodepth=32
thread=1
ioengine=libaio
direct=1
time_based=0
group_reporting=1
write_bw_log=joe
write_iops_log=joe
write_lat_log=joe
log_avg_msec=25000
[job00]
filename=/dev/nvme0n1

Some general data:

nvmf@nvmfjt01:~/jrg9/chandler$ lsblk|grep nvm
nvme0n1              259:0    0 372.6G  0 disk

nvmf@nvmfjt01:~/jrg9/chandler$ fio --version
fio-3.0

nvmf@nvmfjt01:~/jrg9/chandler$ uname -a
Linux nvmfjt01 4.12.7-041207-generic #201708160856 SMP Wed Aug 16 12:59:25 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

I let this workload run until FIO stopped after about 18m15s.  Looked like a normal completion, no errors.

FIO reported 358,733 IOPS for the run, which is a bit below the drive spec but not unreasonable.  However, when I look at the four IOPS log output files, they report an average of about 60K IOPS over the run (that's total for all four - each individual log reports an average around 15K IOPS).  The first value in each log looks good, then all the remaining values seem way too low.  BW log shows similar behavior.

I also tried this in FIO 2.10 and had a variation of the same problem - the first value in each IOPS/BW log looked reasonable, and then all remaining IOPS/BW values in the logs were zero, despite lots of IO running through the whole run.

FIO result:
[...]

PR #435, opened 2017-08-22
time: fix overflow in timespec_add_msec

sitsofe:
Commit 8b6a404cdd2c40715885e562416c3db039912773 ("nanosecond: initial
commit changing timeval to timespec") introduced a change to
timespec_add_msec that decreased the time to (unsigned) overflow of
adj_nsec leading to strange values in the write_*_log files when using
log_avg_msec. The following fio job reproduces the problem:

./fio --ioengine=null --size=2G --time_based --runtime=11s \
 --ramp_time=5s --write_iops_log=overflow --log_avg_msec=5000 \
 --name=overflow

cat overflow_iops.1.log shows the following:
5000, 997853, 0, 0
10640, 141323, 0, 0

Compiling fio with clang's undefined behaviour
sanitizer with unsigned wraparound detection turned on by using

CC=clang ./configure --extra-cflags="-fsanitize=undefined,integer"

and running the previous job produces a lot of false positives but also
flags up this:

time.c:11:35: runtime error: unsigned integer overflow: 1000000 * 5000 cannot be represented in type 'unsigned int'

Fix this by forcing the type of the integer literal to ensure that a 64
bit multiplication takes place. Further, force the result variables to
be uint64_t allowing us to remove UL from other literals. Finally adjust
some spacing while we're here.

Fixes https://github.com/axboe/fio/issues/426 (which was also reported
on the mailing list - http://www.spinics.net/lists/fio/msg06190.html ).

Reported-by: Joseph R Gruher <joseph.r.gruher@intel.com>
Suggested-by: Vincent Fu <vincent.fu@wdc.com>
Suggested-by: Jeff Furlong <jeff.furlong@wdc.com>
Reviewed-by: Vincent Fu <vincent.fu@wdc.com>
Tested-by: Joseph R Gruher <joseph.r.gruher@intel.com>

### DIFF
--- a/time.c
+++ b/time.c
@@ -8,17 +8,17 @@ static unsigned long ns_granularity;
 
 void timespec_add_msec(struct timespec *ts, unsigned int msec)
 {
-	unsigned long adj_nsec = 1000000 * msec;
+	uint64_t adj_nsec = 1000000ULL * msec;
 
 	ts->tv_nsec += adj_nsec;
 	if (adj_nsec >= 1000000000) {
-		unsigned long adj_sec = adj_nsec / 1000000000UL;
+		uint64_t adj_sec = adj_nsec / 1000000000;
 
-		ts->tv_nsec -=  adj_sec * 1000000000UL;
+		ts->tv_nsec -= adj_sec * 1000000000;
 		ts->tv_sec += adj_sec;
 	}
-	if (ts->tv_nsec >= 1000000000UL){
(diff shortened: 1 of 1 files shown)


axboe:
Pulled, thanks Sitsofe.

sitsofe:
No problem Jens. I know the detail is a bit over the top this time but hopefully it will act as a guide on how to find and fix other overflow issues in the future...

axboe:
Better over the top than not enough!

Issue #48, opened 2015-02-22
OSX fcntl(fd, F_NOCACHE, 1) not equivalent to O_DIRECT on Linux


aktau:
This is probably already well known, but it bit me while planning to do some comparative benchmarks between Linux and OSX. I started with OSX and `direct=1, rw=randread`, but noticed that OSX was reading at 1200MB/s from a 1GB file. This was unexpected as I only have one spinning rust HDD in the MBP. 

So I looked up the ways to do direct I/O on OSX. On stack overflow and the Apple mailing lists, `fcntl(fd, F_NOCACHE, 1)` looked to be the canonical solution. This was also implemented in fio(1) in 2011 in commit 7e8ad197. It seems that `F_NOCACHE` disables the page cache from that point on, but the file in quesition was _already_ in the page cache, it will not be purged and the pages will be used.

I also commented on [stack overflow: clear buffer cache on OSX]( http://stackoverflow.com/questions/459537/clear-buffer-cache-on-mac-os-x) with my observations. I'll copy it here as well:

> It's my impression that even when turning off the cache like this (with F_NOCACHE or F_GLOBAL_NOCACHE), if there are pages of the file already in the page cache, those will still be used. I tried to test this by using fio(1) with direct=1. It seems to confirm my suspicions (I get ~1200MB/s throughput on a random read, on a spindle HDD in my MBP, not an SSD). I've confirmed with dtruss that fio(1) actually calls fcntl correctly.

> After running "sudo purge" and trying the same fio(1) invocation, it's much slower. So yes, it appears that F_NOCACHE is no direct equivalent of O_DIRECT on Linux.

I'm not advocating running `sudo purge` as part of fio, but perhaps it could be added to the documentation of `direct` that the behaviour is quite different from `O_DIRECT` but they're both fio's way of doing direct IO. Running `sudo purge` is both slow, has an adverse effect on the rest of the system while (and after) it runs for obvious reasons.

Another idea I had was to forcibly re-write the file each time for reading, while having the file open with `F_NOCACHE`, which makes the written pages not enter the page cache (UPC). That would also be slow (possibly) but hopefully it wouldn't evict other, unrelated files.

[...]

PR #439, opened 2017-08-27
fio: refactor explicit directio calls

sitsofe:
Create an fio_directio() helper for the explicit directio setting call
and refactor its usage to be entirely within filesetup.c by calling it
from generic_open_file(). Also make initial layout in extend_file() call
it thus keeping up with the change introduced by
6e344dc3445bfec0583072e82bea728ab8d54d58 ("filesetup: keep OS_O_DIRECT
flag when pre-allocating file").

A positive side effect of this change means the following job

rm -f /tmp/fiofile; ./fio --loops=10 --filename /tmp/fiofile --bs=4k \
 --size=1M --direct=1 --name=go

that creates a brand new file will no longer report cached I/O speeds
(such as 2000MiB/s) on macOS. This was happening because macOS fio is
unable to invalidate already cached data (see
https://github.com/axboe/fio/issues/48 ) and data was being cached
during the layout phase.

### DIFF
--- a/filesetup.c
+++ b/filesetup.c
@@ -103,6 +103,37 @@ static void fallocate_file(struct thread_data *td, struct fio_file *f)
 }
 
 /*
+ * Helper for OSes (e.g. macOS, Solaris) that have a distinct call to mark the
+ * file non-buffered instead of using the O_DIRECT flag with open.
+ */
+static int fio_directio(struct thread_data *td, struct fio_file *f)
+{
+#ifdef FIO_OS_DIRECTIO
+	int ret;
+
+	if (!td->o.odirect) {
+		return 0;
+	}
+
+	dprint(FD_FILE, "set directio on file %s\n", f->file_name);
+	ret = fio_set_odirect(f->fd);
+	if (ret) {
+		td_verror(td, ret, "fio_set_odirect");
+		if (ret == ENOTTY) { /* ENOTTY suggests RAW device or ZFS */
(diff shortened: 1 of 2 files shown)


filesetup.c
```diff
@@ -164,7 +195,8 @@ static int extend_file(struct thread_data *td, struct fio_file *f)
 		else
 			td_verror(td, err, "open");
 		return 1;
-	}
+	} else if (fio_directio(td, f) != 0)
```
axboe: We had this on td_io_open_fie() before, which is used by anyone wishing to open a file. You moved it to generic_open_file(), which is just one possible implementation that engines can use. This changes the scope of it, have you verified this is actually OK?

sitsofe: @axboe I verified it to a limited extent - I tested it did the right thing on macOS (which is was one of the two platforms where it is implemented would affect). How would it have interacted on such platforms previously with engines that implemented their own open that wasn't a regular file?

axboe: It's the IO engines that will set it, it's not an architecture/platform thing. Grep for ".open_file" in engines/*.c and there are quite a few that do not directly use generic_file_open(). Some of them do call it, so maybe it's fine, but the intent is really for it to be standalone. This means that you cannot put generic code in generic_file_open() that you expect all engines to use.

sitsofe: @kusumi got here first in https://github.com/axboe/fio/commit/47534cda051d15767f175967b91ef56771adb0e0 so let's drop this pull request.

Issue #429, opened 2017-08-21
Error 22 testfile does not work


C4rst3n:
I just tried to run a fio test. The SSD and HDD are both mounted.
The test is finished too quick (faster then ls) and in the log files is only one entry.

Here the console output from my test (md made some things strikethrough)
>raid@raid-PowerEdge-T630:~/Dokumente$ fio randrw.fio --output-format=json --output=randrw_noRaid.json
>raid@raid-PowerEdge-T630:~/Dokumente$ fio --version
>fio-3.0-5-g168b
>raid@raid-PowerEdge-T630:~/Dokumente$ cat randrw.fio
>[global]
>ioengine=sync
>direct=1
>rw=randrw
>size=2g
>runtime=300
>time_based
>refill_buffers=1
>log_avg_msec=1000
>[HDD_randrw_2g_5m]
>write_bw_log=HDD_rand
>write_iops_log=HDD_rand
>write_lat_log=HDD_rand
>filename=/mnt/HDD/randrw
>[SSD_randrw_2g_5m]
>stonewall
>write_bw_log=SSD_rand
>write_iops_log=SSD_rand
>write_lat_log=SSD_rand
>filename=/mnt/SSD/randrw

The json file says something about Error 22

>{
>  "fio version" : "fio-3.0-5-g168b",
>  "timestamp" : 1503302509,
>  "timestamp_ms" : 1503302509228,
>  "time" : "Mon Aug 21 10:01:49 2017",
>  "global options" : {
>    "ioengine" : "sync",
>    "direct" : "1",
>    "rw" : "randrw",
>    "size" : "2g",
>    "runtime" : "300",
>    "refill_buffers" : "1",
>    "log_avg_msec" : "1000"
>  },
>  "jobs" : [
>    {
>      "jobname" : "HDD_randrw_2g_5m",
>      "groupid" : 0,
>      "error" : 22,
>      "eta" : 2147483647,
>      "elapsed" : 1,
>      "job options" : {
>        "write_bw_log" : "HDD_rand",
>        "write_iops_log" : "HDD_rand",
>        "write_lat_log" : "HDD_rand",
>        "filename" : "/mnt/HDD/randrw"
>      },
> ...

When i don't use the json format it is more detailed:
>HDD_randrw_2g_5m: (groupid=0, jobs=1): err=22 (file:filesetup.c:227, func=write, error=Invalid argument): pid=3630: Mon Aug 21 10:10:18 2017

[...]

PR #440, opened 2017-08-28
Direct layout fix

sitsofe:
Fix for layout not working when it has to make a file with `direct=1` on Linux due to memory not being aligned and an additional warning about `direct=1` not being liked if direct layout fails with EINVAL.

Still waiting for feedback via https://github.com/axboe/fio/issues/429 but creating the pull request now so I don't forget about it after feedback arrives.


axboe:
Pulled, thanks.

Issue #410, opened 2017-07-20
Averaged IOps Log File write results appear shifted compared to the reads


shimrot:
We noticed that when performing mixed read/write I/O, the averaged IOps log file results do not accurately reflect the average I/O rates compared to results from bus traces (other logs might be the same, but didn't verify).
The cause appears to be that the "write" sample results are shifted respect to the read results.

That is, given an example job:  
```
[global]
bs=4KB
direct=1
filename=/dev/sdb
ioengine=libaio
log_avg_msec=1000
norandommap=1
ramp_time=5s
randrepeat=0
runtime=10s

[RW-Job]
rw=randrw
write_iops_log=RW-Job
write_lat_log=RW-Job
```

The IOPs log is supposed to show the IOPs average over 1 second intervals. The IOPs log file produced: 
```
1000, 5089, 0, 0
2000, 5305, 0, 0
2000, 5182, 1, 0
3000, 5249, 0, 0
3000, 5255, 1, 0
4000, 5324, 0, 0
4000, 5218, 1, 0
5000, 5181, 0, 0
5000, 5180, 1, 0
6001, 5285, 0, 0
6001, 4938, 1, 0
7000, 5221, 0, 0
7000, 5273, 1, 0
8000, 5382, 0, 0
8000, 5397, 1, 0
9000, 5051, 0, 0
9000, 5261, 1, 0
10028, 5144, 1, 0
```

The write entries appear to be shifted, thus entry "2000, 5182, 1, 0" belongs at timestamp 1000 rather than 2000, and so on.

We discovered the issue during some product testing, comparing some IOMeter 1 second samples to the fio average IOps samples.  IOMeter shows a much larger peak and valley swing than the FIO log file results, when looking at the sum of the read and write per interval.

Below is a chart showing the issue.  The IOPs log (sum of read and write at each time value) at 1 second is smoother (green line) than when taking results from a bus trace and counting the number of ops which occur each second (orange line).   If we have fio record every op, and count the ops over 1 second intervals (blue line), the data matches our bus trace.
![image](https://user-images.githubusercontent.com/8331559/28440892-db13ed82-6d6d-11e7-8afa-837cf6fd8ce3.png)

[...]

PR #443, opened 2017-09-01
stat: fix shifted windowed logging when using multiple directions

sitsofe:
When using log_avg_msec and multiple I/O directions, the results
recorded for the directions other than the first are attributed to the
wrong time. This is demonstrated by the following:

sudo modprobe null_blk completion_nsec=100000000 irqmode=2
sudo ./fio --bs=4k --direct=1 --ioengine=libaio --log_avg_msec=1000 \
 --runtime=2s --disable_lat=1 --disable_bw_measurement=1 \
 --filename=/dev/nullb0 --write_iops_log=shiftedlog --rw=rw \
 --name=shiftedlog

results in this:
cat shiftedlog_iops.1.log
1000, 5, 0, 0
2000, 6, 0, 0
2000, 4, 1, 0
2256, 4, 1, 0

shiftedlog_iops.1.log is missing an entry at 1000ms for for the write
direction (1) because the sample has been misattributed to 2000ms (the
third row) due to iolog->avg_last being shared by all directions.

Fix the shifting by turning iolog->avg_last into an array indexed by
direction and only update the windowed stats for the current direction
from add_log_sample().

Fixes https://github.com/axboe/fio/issues/410 .

Reported-by: Kris Davis <shimrot@gmail.com>

(This issue was very time consuming and moderately difficult to track down so it's hard for me to give strong correctness guarantees. Additionally I wonder if windowed logging should be made to record "0" samples entries when time expires and there's a question about whether "samples" is being incorrectly reported when there are multiple directions (it looks like the sum total of all directions is given rather than the individual direction's total samples)...)

### DIFF
--- a/iolog.h
+++ b/iolog.h
@@ -117,7 +117,7 @@ struct io_log {
 	 */
 	struct io_stat avg_window[DDIR_RWDIR_CNT];
 	unsigned long avg_msec;
-	unsigned long avg_last;
+	unsigned long avg_last[DDIR_RWDIR_CNT];
 
 	/*
 	 * Windowed latency histograms, for keeping track of when we need to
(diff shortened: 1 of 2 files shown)


axboe:
Pulled, thanks.

PR #444, opened 2017-09-03
glusterfs: silence printf specifier warnings

sitsofe:
On 32 bit systems compiling the gluterfs ioengine would generate some printf specifier warning so try and silence with macros and casts.


axboe:
Pulled, thanks.

PR #445, opened 2017-09-03
travis: add 32 bit build, minor updates and cleanups

sitsofe:
Get Travis to build a 32 bit fio (although sadly the one engine that would have benefitted the most from this (glusterfs) has a dependency that is not multilib compatible), minor Travis file reformatting and up the version of Xcode the OSX build targets.


axboe:
Pulled, thanks.

PR #446, opened 2017-09-04
windowsaio: obey sequential/random I/O hinting

sitsofe:
The existing code for selecting whether sequential or random I/O is used seems a bit strange in engines/windowsaio.c. This commit makes it mirror what's done over in td_io_open_file() of ioengines.c. I passed up the chance to stick an `assert(0)` on the default: case to again mirror what is done elsewhere - let me know if this was wrong.

### DIFF
--- a/engines/windowsaio.c
+++ b/engines/windowsaio.c
@@ -169,13 +169,26 @@ static int fio_windowsaio_open_file(struct thread_data *td, struct fio_file *f)
 
 	/*
 	 * Inform Windows whether we're going to be doing sequential or
-	 * random io so it can tune the Cache Manager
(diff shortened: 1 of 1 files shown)


axboe:
Looks good to me, but I'd really like someone with some Windows knowledge to review it. Can you check with Rebecca Cran if she likes it?

sitsofe:
@bcran What do you think of these changes?

bcran:
Looks good to me too. It makes sense to only tell the kernel what type of workload will be used when the user asks us to, instead of always setting it based on the workload type read/write/randread/etc.

axboe:
Merged, thanks.

Issue #431, opened 2017-08-21
fio intermittently fails to perform IO when --trigger-timeout, --trigger is specified


saurabhbpl:
We have observed that fio intermittently fails to perform IO when --trigger-timeout, --trigger is specified. It works fine when these options are removed.
[saurabha@devvm ~/temp/fio-fio-3.0] ./fio --client=ip6:client ~/temp/data_integrity.fio --verify_state_save=1 --verify_state_load=1 --verify=crc32c  --rw="write"
hostname=<hostname>, be=0, 64-bit, os=Linux, arch=x86-64, fio=fio-3.0, flags=1
<hostname> job0: (g=0): rw=read, bs=(R) 1024KiB-1024KiB, (W) 1024KiB-1024KiB, (T) 1024KiB-1024KiB, ioengine=libaio, iodepth=128
<hostname> Starting 1 process
job0: (groupid=0, jobs=1): err= 0: pid=3009069: Fri Aug 18 17:25:04 2017
   read: IOPS=329, BW=342MiB/s (359MB/s)(3454MiB/10092msec)
    slat (usec): min=61, max=41362, avg=3004.19, stdev=1839.61
    clat (msec): min=91, max=546, avg=378.15, stdev=37.44
     lat (msec): min=95, max=549, avg=381.15, stdev=37.53
    clat percentiles (msec):
     |  1.00th=[  194],  5.00th=[  321], 10.00th=[  363], 20.00th=[  380],
     | 30.00th=[  380], 40.00th=[  380], 50.00th=[  384], 60.00th=[  384],
     | 70.00th=[  384], 80.00th=[  384], 90.00th=[  393], 95.00th=[  422],
     | 99.00th=[  472], 99.50th=[  502], 99.90th=[  531], 99.95th=[  542],
     | 99.99th=[  550]
   bw (  KiB/s): min=313344, max=374784, per=97.25%, avg=340823.60, stdev=13581.11, samples=20
   iops        : min=  306, max=  366, avg=332.60, stdev=13.19, samples=20
  lat (msec)   : 100=0.09%, 250=1.62%, 500=101.56%, 750=0.54%
  cpu          : usr=0.09%, sys=2.57%, ctx=3159, majf=0, minf=2
  IO depths    : 1=0.1%, 2=0.1%, 4=0.1%, 8=0.2%, 16=0.5%, 32=1.0%, >=64=238.1%
[...]

PR #454, opened 2017-09-10
backend: verify-trigger fixes

sitsofe:
Clear the trigger_timeout after we have found it has been exceeded in
trigger_timedout() to stop us repeatedly firing the trigger again and
again in the future.

When --trigger-remote is unset, the server winds up with the remote
trigger command represented by buf[] being zero length rather than buf
being NULL in handle_trigger_cmd(). Cope with this by returning right
away if we find the cmd string is NULL or empty in exec_trigger().

Make server mode fio terminate after sending the verify state back to
the client so it goes on behave the same way as standalone fio. This
stops the while loop in thread_main() restating and in turn prevents
state being cleared away by clear_io_state().

Make keep_running() return false if td->terminate has been set Just In
Case.

The below would trigger the above problem before this patch:

cat <<EOF > trigger.fio
[global]
ioengine=null
time_based=1
runtime=10s
size=1M
bs=4k
rate=4k

[trigtest]
EOF
./fio --server &
./fio --trigger-timeout=2s --trigger='hostname' --client=localhost \
 --debug=net trigger.fio

Fixes https://github.com/axboe/fio/issues/431 ("fio intermittently fails
to perform IO when --trigger-timeout, --trigger is specified").

Reported-by: saurabhbpl <saurabh.shiva@gmail.com>

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -1391,6 +1391,8 @@ static bool keep_running(struct thread_data *td)
 
 	if (td->done)
 		return false;
+	if (td->terminate)
+		return false;
 	if (td->o.time_based)
 		return true;
 	if (td->o.loops) {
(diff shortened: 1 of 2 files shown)


axboe:
Looks good, good work @sitsofe

Issue #450, opened 2017-09-05
zone_size > zone_range leads to incorrect zoning


gvkovai:
Tried to do 3x amount of IO per zone and hence had the following config. Expectation was 9M worth of IO done in each 3M size of the zone. I.e., each zone would be read thrice before moving to next zone. The actual result shown in pic. See pic attached. IO continues (increments offset) until end of zone_size instead of zone_range. zone_skip=1M became ineffective as skipped area also read.

> $ ./fio --version
fio-3.0

> [global]
ioengine=libaio
direct=1
time_based
disk_util=0
continue_on_error=all
rate_process=poisson
write_iolog=offsetlog
[db-dss1]
bs=8K
filesize=524288M
zonesize=9M
zonerange=3M
zoneskip=1M
filename=/dev/sdb
rw=read
iodepth=1
rate_iops=100

> sudo ./fio --runtime 120 --debug=file,io,blktrace --write_iops_log=/tmp/IOPS --write_lat_log=/tmp/LAT --status-interval=10 --output=/tmp/fio.out --output-format=json <above.fio>

Y Axis = Nth IO

![image](https://user-images.githubusercontent.com/27024474/30082842-3293e3bc-925a-11e7-9df5-5d1bccbffe57.png)

![image](https://user-images.githubusercontent.com/27024474/30082890-69c6b008-925a-11e7-9ad8-7ec6adf6b022.png)

Affecting code is in io_u.c: static int get_next_seq_offset...
It's using io_size instead of zone_range.

sitsofe:
I admit I've never known the difference between zonesize and zonerange. The HOWTO says this

> zonesize=int
> Divide a file into zones of the specified size. See zoneskip.
>
> zonerange=int
> Give size of an I/O zone. See zoneskip.
>
> zoneskip=int
> Skip the specified number of bytes when zonesize data has been read. The two zone options can  be used to only do I/O on zones of a file.

It seems weird how only zonerange and zoneskip refer to each other. However digging down on zonerange gets to https://github.com/axboe/fio/commit/ed33585509a97c4a65e46967cb49419b6a4f8870 whose commit message says:

> This also adds a 'zonerange' configuration option which may be too ambiguous
and is subject to potential renaming in the future.
>
> When doing random I/O, it is beneficial to be able to specify how large the
addressible space is in the zone, while specifying a different metric for
how much data to read from that zone (i.e., how many samples to take from each
zone). When 'zonerange' is not specified, it defaults to be equal to the
[...]

PR #457, opened 2017-09-12
Fix zoning issue with seq-io and randommap issue

gvkovai:
The case of zonerange < zonesize scenario was not handled
correctly earlier. When zonesize > zonerange, IO must continue
in the same zonerange of the size zonesize for seq-io.

For random io, zonesize > zonerange leads to sequential io
after first zonerange size of io is done when 'norandommap'
is not set. In this case, map needs to be reset for every
zonerange size of IO on a zone.

### DIFF
--- a/io_u.c
+++ b/io_u.c
@@ -850,6 +850,45 @@ void requeue_io_u(struct thread_data *td, struct io_u **io_u)
 	*io_u = NULL;
 }
 
+static void __fill_io_u_zone(struct thread_data *td, struct io_u *io_u)
+{
+	struct fio_file *f = io_u->file;
+
+	/*
+	 * See if it's time to switch to a new zone
+	 */
+	if (td->zone_bytes >= td->o.zone_size && td->o.zone_skip) {
+		td->zone_bytes = 0;
+		f->file_offset += td->o.zone_range + td->o.zone_skip;
+
+		/*
+		 * Wrap from the beginning, if we exceed the file size
+		 */
+		if (f->file_offset >= f->real_file_size)
+			f->file_offset = f->real_file_size - f->file_offset;
+		f->last_pos[io_u->ddir] = f->file_offset;
+		td->io_skip_bytes += td->o.zone_skip;
+	}
+
+	/*
+ 	 * If zone_size > zone_range, then maintain the same zone until
+ 	 * zone_bytes >= zone_size.
+ 	 */
+	if (f->last_pos[io_u->ddir] >= (f->file_offset + td->o.zone_range)) {
+		dprint(FD_IO, "io_u maintain zone offset=%" PRIu64 "/last_pos=%" PRIu64 "\n",
+				f->file_offset, f->last_pos[io_u->ddir]);
+		f->last_pos[io_u->ddir] = f->file_offset;
+	}
+
+	/*
+	 * For random: if 'norandommap' is not set and zone_size > zone_range,
+	 * map needs to be reset as it's done with zone_range everytime.
+	 */
+	if ((td->zone_bytes % td->o.zone_range) == 0) {
+		fio_file_reset(td, f);
+	}
+}
+
 static int fill_io_u(struct thread_data *td, struct io_u *io_u)
 {
 	unsigned int is_random;
@@ -866,21 +905,10 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
 		goto out;
 
 	/*
-	 * See if it's time to switch to a new zone
+	 * When file is zoned zone_range is always positive
 	 */
-	if (td->zone_bytes >= td->o.zone_size && td->o.zone_skip) {
-		struct fio_file *f = io_u->file;
-
-		td->zone_bytes = 0;
-		f->file_offset += td->o.zone_range + td->o.zone_skip;
-
-		/*
-		 * Wrap from the beginning, if we exceed the file size
-		 */
-		if (f->file_offset >= f->real_file_size)
-			f->file_offset = f->real_file_size - f->file_offset;
-		f->last_pos[io_u->ddir] = f->file_offset;
-		td->io_skip_bytes += td->o.zone_skip;
+	if (td->o.zone_range) {
+		__fill_io_u_zone(td, io_u);
 	}
 
 	/*


io_u.c
```diff
@@ -884,6 +883,23 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
...
+		dprint(FD_IO, "io_u maintain zone offset=%" PRIu64 "/last_pos=%" PRIu64 "\n",
+				f->file_offset, f->last_pos[io_u->ddir]);
+		f->last_pos[io_u->ddir] = f->file_offset;
+	}
+
+	/*
+	 * For random: if 'norandommap' is not set and zone_size > zone_range,
+	 * map needs to be reset as it's done with zone_range everytime.
+	 */
+	if (td->zone_bytes % td->o.zone_range == 0)
```
axboe: What happens if td->o.zone_range == 0?

You also might want to consider using parenthesis to make this clearer to read:

```if ((td->zone_bytes % td->o.zone_range) == 0)```

but the mod-by-0 is the biggest issue here.

Comment (gvkovai) [reply io_u.c]: makes sense. I have made changes and updated the tree with it. Please let me know. Thanks.

**Comment by sitsofe** [comment]
@gvkovai Any chance you can squash this new commit into the previous one? Also it might be handy to include the fio command that generated the problem in the commit message for posterity:

```
./fio --name=badzoning --bs=8k --size=256k --zonesize=256k --zonerange=128k \
  --filename=/tmp/fiofile
```

Also a trailing line like
```
Fixes https://github.com/axboe/fio/issues/450
```

Would also be handy...

**Comment by sitsofe** [comment]
(see https://github.com/axboe/fio/commit/8355119b7ef7007b0a251654602584eec000f958 for an example commit message with an example line and a Fixes line)

**Comment by sitsofe** [comment]
Latest commit LGTM.

File: io_u.c [hunk 2]
```diff
@@ -884,6 +883,24 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
...
+				f->file_offset, f->last_pos[io_u->ddir]);
+		f->last_pos[io_u->ddir] = f->file_offset;
+	}
+
+	/*
+	 * For random: if 'norandommap' is not set and zone_size > zone_range,
+	 * map needs to be reset as it's done with zone_range everytime.
+	 */
+	if (td->o.zone_range && (td->zone_bytes % td->o.zone_range == 0)) {
+		fio_file_reset(td, f);
```
sitsofe: You might want to add the extra parenthesis Jens mentioned here - it might not be obvious that % will bind tighter than == .

sitsofe:
@gvkovai Since this commit hasn't been accepted yet I've got a question:

```
./fio --debug=io --name=seqzoneread --zonesize=16k --zonerange=8k --zoneskip=4k --size=20k --filename=/tmp/fiofile | grep "io complete"
```

Doesn't act as I would expect - I only get 20k's worth of I/O rather than 32k's worth (within a 20k region). If I do

```
./fio --debug=io --name=seqzoneread --zonesize=16k --zonerange=8k --zoneskip=4k --size=20k --filename=/tmp/fiofile --io_limit=32k | grep "io complete"
```

Then I see the pattern I would have expected. I guess the question is - how do we expect zone repetition to interact with size?

sitsofe:
(This might be a difficult question to answer because it's hard to know what the total I/O will be if the blocksizes are a range and the sizes used don't fit exactly into a zone)

sitsofe:
OK another problem is that if the file is 12MBytes but size=256k fio won't wraparound to 0 at the 256k mark.

sitsofe:
In fact the current patch seems to break the following:
```
$ ./fio --name=loops2 --size=20k --filename=/tmp/fiofile --rw=read --loops=2
[...]
fio-3.0-53-g117996
Starting 1 process
loops2: No I/O performed by psync, perhaps try --debug=io option for details?
```

gvkovai:
@sitsofe 

I have wrapped zone_* specific code by checking for positive zone_range value which is always set if at least zone_size is set. (Default: zone_range = zone_size). This should avoid mangling non zone related use cases.

I will test some more cases you mentioned and update here. some cases do make sense at the current form and the bugs you pointed out should not be there now with this change.

io_u.c
```diff
@@ -866,21 +867,42 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
 		goto out;
 
 	/*
-	 * See if it's time to switch to a new zone
+	 * if file is zoned, zone_range is set to positive value always.
```
sitsofe: "When file is zoned zone_range is always positive"

sitsofe:
@gvkovai I see your rationale for the big if - perhaps you're very close to refactoring it into a helper function.

Sorry for the false positive earlier - it can take multiple reviews before spotting some of these points.

io_u.c
gvkovai: Done

gvkovai:
The loops reset issue does not look like related to the zone change #457 . master version also fails while fio-2.8 does not show this issue.

==> ./fio --debug=io --name=seqzoneread --rw=read --filename=/tmp/fiofile --size=256k --filesize=12M --loops=1 | grep "io complete" | wc -l
64

==> **./fio** --debug=io --name=seqzoneread --rw=read --filename=/tmp/fiofile --size=256k --filesize=12M --loops=2 | grep "io complete" | wc -l
**seqzoneread: No I/O performed by psync, perhaps try --debug=io option for details?**
128

==> **./fio_2** --debug=io --name=seqzoneread --rw=read --filename=/tmp/fiofile --size=256k --filesize=12M --loops=2 | grep "io complete" | wc -l
128

==> **./fio_3fix** --debug=io --name=seqzoneread --rw=read --filename=/tmp/fiofile --size=256k --filesize=12M --loops=2 | grep "io complete" | wc -l
**seqzoneread: No I/O performed by psync, perhaps try --debug=io option for details?**
128

==> ./fio_2 -v **(succeeds)**
fio-2.8-1-gbac6c

==> ./fio_3fix -v **(fails)**
fio-3.0-54-g2d84-dirty
==> ./fio -v **(fails)**
fio-3.0-61-g9acb0

gvkovai:
`./fio --name=loops2 --size=20k --filename=/tmp/fiofile --rw=read --loops=2`

This works with latest fix

gvkovai:
`./fio --debug=io --name=seqzoneread --zonesize=16k --zonerange=8k --zoneskip=4k --size=20k --filename=/tmp/fiofile | grep "io complete"`

This works as expected. Here zone area is 8k and zone size is 16k. first zone is read twice (total = 16k). And does 1 4k on the next zone.  `size` is the total size and hence IO is complete when io-completed >= size. This makes sense to me.

`./fio --debug=io --name=seqzoneread --zonesize=16k --zonerange=8k --zoneskip=4k --size=20k --filename=/tmp/fiofile --io_limit=32k | grep "io complete"`

This looks like `io_limit` overrides `size`. 

`./fio_3fix --debug=io --name=seqzoneread --zonesize=16k --zonerange=8k --zoneskip=4k --size=64k --filename=/tmp/fiofile --filesize=24k --time_based --runtime 1 | grep "io complete`

The above command makes `size` not effective due to `time_based` option. As expected, this got two zones each of 12k (8k ws + 4k skip) on a file of size  filesize=24k. Each zone is read twice and repeats all zones (2) in the file until time is elapsed.

sitsofe:
@gvkovai I admit multiple files does sound like it will be a problem with zones and I think offset will need handling. Taking your example though:

You have `--filename=/tmp/testfile`, `--filesize=1024M` and `--size=4M`. Each file is 1GByte big and in this case you only have one file. Due to `size` you'll only do I/O between 0-4Mbytes of it (see http://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-size ). How do you fit four zones of 1M+(255M gap) into a range of only 4 MBytes? Zonesize+zoneskip=256M so after the first zone is done you'll stop doing I/O as you'll have exceeded size...

gvkovai:
Yes - I kept using old meaning `size` meaning. I intended to use `io_size` here which works. Thanks for the clarification.

On zonesize comment though - needed bit more clarification. The way it works now is:
number of zones = filesize (or size) / (zonerange + zoneskip)
`zonesize` is actually the `io_size` on a `zonerange`.

sitsofe:
@gvkovai You're right about `zonerange`/`zonesize` - I've updated my comment.

Any ideas on how we should handle zones and multiple files and zones when using offset (and size)?

gvkovai:
@sitsofe Looks like I will not be able to look into multi file issues immediately. I will when I get a chance as soon as possible.

meanwhile your zone_calc and this one can be an improvement from the current version?

sitsofe:
I agree that the current patches look like an improvement. I doubt I'm going to work on multi-file any time soon either but I'd like to see if I get some time this week to polish the branch a touch before submitting a PR.

sitsofe:
@axboe I suggest we take this pull request as-is for now (I never did get round to that polishing) as it can always be improved later and others are hitting issues this would fix..

axboe:
Pulled, thanks.

PR #461, opened 2017-09-14
Small documentation changes

vincentkfu:
Jens, please consider these two patches that add a brief description of the JSON output format and emphasize that --status-interval does not provide per-period output.

Thanks to @sitsofe for feedback on the initial version of these patches.


axboe:
Looks good, thanks Vincent.

PR #465, opened 2017-09-24
CI and configure changes

sitsofe:
The first change makes Appveyor install zlib for the Windows CI builds and does some cleanups. The good news is that it is has always been possible to download the builds that Appveyor makes (e.g. here a 3.0 MSI https://ci.appveyor.com/project/axboe/fio/build/1.0.263/job/304hphqrbcos4tk0/artifacts ) so this makes those builds more useful (as well as serving as an example of how to zlib fio building on Windows) if someone wants more pre-built fio Windows builds. The bad news is that it slows the combined time for all the Windows builds to complete down by about 30 seconds and it hard codes a mirror site (but that seems to be the site that other Appveyor Cygwin projects use).

The second commit is more finicky and showed up with the change to only use Cygwin64. Basically it changes how we probe for the fallback compiler and makes things a bit consistent when we try to override the compiler we want to use. I tested out the logic on Windows, Linux and on macOS so it seems fine to me. @kusumi could you cast your eye over this one and see if it looks good for FreeBSD and other platforms you use?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,15 +1,20 @@
-clone_depth: 50
+clone_depth: 1
 environment:
+  CYG_MIRROR: http://cygwin.mirror.constant.com
+  CYG_ROOT: C:\cygwin64
   MAKEFLAGS: -j 2
   matrix:
     - platform: x86_64
       BUILD_ARCH: x64
-      CYG_ROOT: C:\cygwin64
+      PACKAGE_ARCH: x86_64
       CONFIGURE_OPTIONS:
     - platform: x86
       BUILD_ARCH: x86
-      CYG_ROOT: C:\cygwin
+      PACKAGE_ARCH: i686
       CONFIGURE_OPTIONS: --build-32bit-win
+
+install:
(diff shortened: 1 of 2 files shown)


sitsofe:
@kusumi any thoughts on b73af7389a47bb04d841d749eb8a4e996eb5a271 ?

kusumi:
@sitsofe Not sure what you're asking, but it compiles on FreeBSD (so most likely on DragonFlyBSD, and likely on Net/OpenBSD).

BTW, the default compiler on recent FreeBSD are clang. Others still use gcc by default, but generally speaking *BSD prefer non-GPL (i.e. BSDL or other similar ones) alternatives if any.

sitsofe:
@kusumi In https://github.com/axboe/fio/commit/b73af7389a47bb04d841d749eb8a4e996eb5a271 I'd changed configure's probing so the request was more to check that it hadn't broken things on a platform I don't have - thanks for checking.

I am aware that FreeBSD prefers clang - I've also changed configure to make macOS to prefer clang too.

@axboe I think this one is ready for merging.

sitsofe:
@axboe just a quick ping on this one.

axboe:
Pulled, thanks.

PR #467, opened 2017-09-26
windowsaio: add best effort invalidation

sitsofe:
https://stackoverflow.com/questions/478340/clear-file-cache-to-repeat-perf`ormance-testing/7113153#7113153
mentions that opening a file without buffering on Windows has the side
effect of flushing cached parts of the file out of RAM. Use this side
effect to do cache invalidation on non-direct jobs that are using the
windowsaio engine. Note this invalidation will fail if the file is also
open elsewhere and we can't check whether we have exclusive access
without introducing a race where other jobs fail while we temporarily
have exclusive access to the file.

Preliminary pull request to gather suggestions and feedback.

### DIFF
--- a/engines/windowsaio.c
+++ b/engines/windowsaio.c
@@ -142,6 +142,44 @@ static void fio_windowsaio_cleanup(struct thread_data *td)
 	}
 }
 
+static int windowsaio_invalidate_cache(struct fio_file *f)
+{
+	DWORD error;
+	DWORD isharemode = (FILE_SHARE_DELETE | FILE_SHARE_READ |
+			FILE_SHARE_WRITE);
+	HANDLE ihFile;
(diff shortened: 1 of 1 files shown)


sitsofe:
@bcran any thoughts on this one?

axboe:
Looks useful! I'd probably put the invalidation code into a separate function instead of inlining it, however.

sitsofe:
OK update with invalidation has been moved to its own function and the commit message has been made longer.

sitsofe:
@axboe just a quick ping on this one.

axboe:
Pulled, thanks. Just a minor note not to use braces for single line if blocks, fio generally doesn't do that.

sitsofe:
Oops, sorry about that one.

PR #469, opened 2017-10-03
Several fixes

ErwanAliasr1:
After running the scan-build from clang, I produce the following fixes.

### DIFF
--- a/client.c
+++ b/client.c
@@ -1864,10 +1864,12 @@ static void request_client_etas(struct client_ops *ops)
 static int handle_cmd_timeout(struct fio_client *client,
 			      struct fio_net_cmd_reply *reply)
 {
+	uint16_t reply_opcode = reply->opcode;
+
 	flist_del(&reply->list);
 	free(reply);
 
-	if (reply->opcode != FIO_NET_CMD_SEND_ETA)
+	if (reply_opcode != FIO_NET_CMD_SEND_ETA)
 		return 1;
 
 	log_info("client <%s>: timeout on SEND_ETA\n", client->hostname);
--- a/client.c
+++ b/client.c
@@ -1312,14 +1312,16 @@ static void client_flush_hist_samples(FILE *f, int hist_coarseness, void *sample
 static int fio_client_handle_iolog(struct fio_client *client,
 				   struct fio_net_cmd *cmd)
 {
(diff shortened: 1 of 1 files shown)


client.c
```diff
@@ -1372,8 +1378,17 @@ static int fio_client_handle_iolog(struct fio_client *client,
 					pdu->nr_samples * sizeof(struct io_sample));
 		}
 		fclose(f);
-		return 0;
+		ret = 0;
 	}
+
+out:
+	if (pdu)
```
axboe: Should this be:

if (pdu && pdu != cmd->payload)
        free(pdu);

since pdu is only allocated for the compressed case.

client.c
```diff
@@ -1864,12 +1864,12 @@ static void request_client_etas(struct client_ops *ops)
 static int handle_cmd_timeout(struct fio_client *client,
 			      struct fio_net_cmd_reply *reply)
 {
-	flist_del(&reply->list);
-	free(reply);
-
 	if (reply->opcode != FIO_NET_CMD_SEND_ETA)
 		return 1;
 
+	flist_del(&reply->list);
```
axboe: I'm not sure this is correct. Before we always killed a reply, which means pruning from the list and freeing it. Now we leave the non-ETA commands on there. How about something ala:

```
@@ -1849,10 +1849,12 @@ static void request_client_etas(struct client_ops *ops)
 static int handle_cmd_timeout(struct fio_client *client,
                              struct fio_net_cmd_reply *reply)
 {
+       uint16_t op = reply->opcode;
+
        flist_del(&reply->list);
        free(reply);
 
-       if (reply->opcode != FIO_NET_CMD_SEND_ETA)
+       if (op != FIO_NET_CMD_SEND_ETA)
                return 1;

```

axboe:
I picked 5 of the 7, the other 2 I commented on.

Additionally, can you ensure that you properly sign-off on your changes? Each commit should have a:

Signed-off-by: Erwan Velu <erwan@redhat.com>

at the end of the commit message. Thanks!

client.c
ErwanAliasr1: Updated this way.

client.c
ErwanAliasr1: Updated this way.

axboe:
Pulled, thanks.

PR #476, opened 2017-10-09
Adding support for multiple jobs for fio test

Venutiwa:
Adding support for multiple jobs for fio test, adding 6 job files.

Signed-off-by: Venu <venutiwa@in.ibm.com>

### DIFF
--- a/examples/fio-rand-RW.job
+++ b/examples/fio-rand-RW.job
@@ -1,0 +1,18 @@
+; fio-rand-RW.job for fiotest
+
+[global]
+name=fio-rand-RW
+filename=fio-rand-RW
+rw=randrw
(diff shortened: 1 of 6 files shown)


sitsofe:
Hi @Venutiwa!

Can you tell us what these additional jobs are adding that is unique or demonstrating something special? A quick glance shows they are varying the `rw` parameter in comparison to each other but does that justify adding them?

sitsofe:
(I'd also point out that the job name will be used as the file's name when filename is unset)

Venutiwa:
Adding the files in https://github.com/avocado-framework-tests/avocado-misc-tests/pull/644 for KVM performance CI

PR #477, opened 2017-10-10
Add note for increasing shmmax if necessary

rnsanchez:
While trying out a large-ish fio test in a machine with 4 GB of RAM, `fio` ran out of shared memory by a small amount.  After checking with `strace` for clues, I've noticed the probing for shmem.  The system could accomodate using 64 MB instead of ~33 MB (auto-tuned by the kernel).  `fio` ran happily.


axboe:
Pulled, thanks.

sitsofe:
Hang on - what about the HOWTO?

sitsofe:
Ah that's already been taken care of too :-)

rnsanchez:
Sorry about that, I didn't remember about the HOWTO.

PR #478, opened 2017-10-12
gettime: fix cycles_per_msec overflow when using 32 bit longs

sitsofe:
Compiling fio with clang's undefined behaviour sanitizer and unsigned
wraparound detection enabled on a 32 bit Linux build turned up the
following:

gettime.c:313:28: runtime error: unsigned integer overflow: 3600 * 2600730 cannot be represented in type 'unsigned long'

Fix this by make cycles_per_msec a long long.


axboe:
LGTM, pulled.

PR #482, opened 2017-10-20
doc: minor formatting fixes

sitsofe:
- Sphinx doesn't like non-punctuation characters directly after closing
  backticks unless they are escaped so add some backslashes
  appropriately.
- Markup /proc/sys/kernel/shmmax in --max-jobs like other /proc/
  references.


axboe:
Pulled, thanks.

PR #484, opened 2017-10-27
Proposal: Make name of job part of log filename

szaydel:
This is mainly a proposal and I did not intend this commit to be merged in, it is quite obviously not exactly baked. :)

I found that often I have several jobs in a file, and usually each job is separated with a stonewall. But, each said job also typically will run more than one thread, so I usually will use group_reporting.

Of course if I don't want per job logs, which in my case would be per-thread logs really, it is hard for me to later know which part of the log belongs to which job in the file.

So for sake of argument I have say 3 jobs defined, like so:
```
[cXYZ#foo]
...
sonewall

[cXYZ#bar]
...
stonewall

[cXYZ#baz]
...
stonewall
```

What I would like is to have a series of logs, when group reporting is used that look like the following, when things are said and done.
```
./something_bw_cXYZ#foo.log
./something_bw_cXYZ#bar.log
./something_bw_cXYZ#baz.log

./something_lat_cXYZ#foo.log
./something_lat_cXYZ#bar.log
./something_lat_cXYZ#baz.log

... etc ...
```

I have no idea if anyone is observing the need for this, or if I am way off doing something silly because I did not read the docs well enough and don't know that I can achieve same some other way, or maybe because I am "special". Anyway, this is meant to raise a question about usefulness. I think I am going to refine this a bit for my own purposes. I wanted to make sure I had an example handy of what I was looking for and so figured a diff might be a good thing.


sitsofe:
If you check out the HOWTO you will find that log file naming following job names should actually already be there if you don't pass a name: http://fio.readthedocs.io/en/latest/fio_man.html#cmdoption-arg-write-lat-log . But

```
$ fio --name=logtest --ioengine=null --size=4k --bs=4k --write_lat_log
fio: option `--write_lat_log' requires an argument
fio: unrecognized option '--write_lat_log'
Did you mean write_lat_log?
```

If you dig through the code you see the message comes from here in init.c:

```c
2365 int parse_cmd_line(int argc, char *argv[], int client_type)
2366 {
[...]
2561                         if ((!val || !strlen(val)) &&
2562                             l_opts[lidx].has_arg == required_argument) {
2563                                 log_err("fio: option %s requires an argument\n", opt);
2564                                 ret = 1;
```

and working out how an option gets its `has_arg` set leads to here in options.c:
```c
4506 static void add_to_lopt(struct option *lopt, struct fio_option *o,
4507                         const char *name, int val)
4508 {
4509         lopt->name = (char *) name;
4510         lopt->val = val;
4511         if (o->type == FIO_OPT_STR_SET)
4512                 lopt->has_arg = optional_argument;
4513         else
4514                 lopt->has_arg = required_argument;
4515 }
```

elsewhere in parse.c you find this:
```
 472 static int __handle_option(struct fio_option *o, const char *ptr, void *data,
 473                            int first, int more, int curr)
 474 {
 475         int il=0, *ilp;
 476         fio_fp64_t *flp;
 477         long long ull, *ullp;
 478         long ul1, ul2;
 479         double uf;
 480         char **cp = NULL;
 481         int ret = 0, is_time = 0;
 482         const struct value_pair *vp;
 483         struct value_pair posval[PARSE_MAX_VP];
 484         int i, all_skipped = 1;
 485 
 486         dprint(FD_PARSE, "__handle_option=%s, type=%d, ptr=%s\n", o->name,
 487                                                         o->type, ptr);
 488 
 489         if (!ptr && o->type != FIO_OPT_STR_SET && o->type != FIO_OPT_STR) {
 490                 log_err("Option %s requires an argument\n", o->name);
 491                 return 1;
 492         }
```

So it looks like at some point in the past FIO_OPT_STR parameters were optional argument but this flexibility isn't enforced everywhere so in many cases the parameter to FIO_OPT_STR is actually mandatory. Perhaps fio needs a `FIO_OPT_STR_OPTVAL` which can be used to clean up this problem?

szaydel:
Hey @sitsofe! So, I think one thing I should have also mentioned is that I have been using either no value, which as you pointed out may be a problem or using a value like this: `write_lat_log=name_of_job`, and of course, as a result, if I ask it to group jobs, all jobs are grouped into a single file, which is not helpful, for my purposes anyway, and if I don't group jobs then I basically have a bunch of files with same name, all maintaining this prefix: `write_lat_log=name_of_job`, and distinguished by the index of the job. I still don't find that useful.

I did in fact read that part of the doc before doing this, and as it points out all logs basically have a static name, akin to `foo_slat.x.log`, where x is effectively replaced by index of the job. That's not really what I am after. My hope is to have that x instead be a symbolic representation of the job, if I am doing group reporting, or a combination of symbolic and index if I am doing per job logs.

Hopefully that's making sense.

PR #485, opened 2017-10-29
Miscellaneous changes 

sitsofe:
Fixes to appease stricter packaging guidelines and some documentation changes. The big one is updating the COPYING file in fio to use the one from https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt . It's worth double checking that it was a like-for-like replacement given the significance of the file...

PS: I thought that fio was under the GPL v2 only?


axboe:
Looks good to me, the COPYING change looks fine too as it's word-for-word the same and just formatting.

Fio is gpl v2 only, what makes you think/say otherwise?

sitsofe:
I think I saw 'and "any later version"' in the COPYING file and thought that applied across the whole repo actually that was just an example there and you also need to use/read the shorter to understand the detail.

Issue #422, opened 2017-08-11
Change json bw key to be in bytes?


sitsofe:
Over in https://github.com/axboe/fio/issues/378#issuecomment-306380285 it was noted that `bw` is in bytes even though `io_bytes` has now been converted from being in kilobytes to bytes. To prevent this being lost I'm creating a new issue. @cbwest3 suggested the following:

> In the interest of not breaking current users and enabling fio use in a programmatic manner, can we:
>
> 1. Leave the current bw calculation as is and document it as KBps.
> 2. Create a new bytes_per_sec, and document it as bytes per second? Like io_bytes, this would allow users to perform their own unit conversion (base 10 or base 2), and eliminates (possible) undesired rounding and reduced precision.

Issue #378, opened 2017-06-02
io_bytes reported in JSON output is off by factor of 1024


cbwest3:
Sending a single 1M op to disk results in io_bytes=1024, per JSON output. This issue was mentioned in #233, among other things.  Reproducing the issue is simple, as shown below.

    $ sudo fio --name=test --filename=/dev/sdb --bs=1m --size=1m --output-format=json
    {
      "fio version" : "fio-2.19-66-gd44a",
      "timestamp" : 1496428095,
      "timestamp_ms" : 1496428095820,
      "time" : "Fri Jun  2 14:28:15 2017",
      "jobs" : [
        {
          "jobname" : "test",
          "groupid" : 0,
          "error" : 0,
          "eta" : 2147483647,
          "elapsed" : 1,
          "job options" : {
            "name" : "test",
            "filename" : "/dev/sdb",
            "bs" : "1m",
            "size" : "1m"
          },
          "read" : {
            "io_bytes" : 1024,
            "bw" : 256000,
            "iops" : 250.000000,
            "runtime" : 4,
            "total_ios" : 1,
            "short_ios" : 0,
            "drop_ios" : 0,
            "slat" : {
              "min" : 0,
              "max" : 0,
              "mean" : 0.000000,
              "stddev" : 0.000000
            },
            "clat" : {
              "min" : 3495,
              "max" : 3495,
              "mean" : 3495.000000,
              "stddev" : 0.000000,
              "percentile" : {
                "1.000000" : 3504,
                "5.000000" : 3504,
                "10.000000" : 3504,
                "20.000000" : 3504,
                "30.000000" : 3504,
                "40.000000" : 3504,
                "50.000000" : 3504,
                "60.000000" : 3504,
                "70.000000" : 3504,
                "80.000000" : 3504,
                "90.000000" : 3504,
                "95.000000" : 3504,
                "99.000000" : 3504,
                "99.500000" : 3504,
                "99.900000" : 3504,
                "99.950000" : 3504,
                "99.990000" : 3504,
                "0.00" : 0,
                "0.00" : 0,
                "0.00" : 0
              }
            },
            "lat" : {
              "min" : 3496,
              "max" : 3496,
              "mean" : 3496.000000,
              "stddev" : 0.000000
            },
            "bw_min" : 0,
            "bw_max" : 0,
            "bw_agg" : 0.000000,
            "bw_mean" : 0.000000,
            "bw_dev" : 0.000000
          },
          "write" : {
[...]

PR #487, opened 2017-11-03
stat: add bw_bytes JSON key

sitsofe:
Does what it says on the tin. It's pretty much too late to fix the JSON `bw` to be in bytes (see https://github.com/axboe/fio/issues/378#issuecomment-306380285 ) so add a `bw_bytes`  key to help the attentive.

Fixes: https://github.com/axboe/fio/issues/422


axboe:
Pulled, thanks.

PR #488, opened 2017-11-06
Makefile: tiny fix of libhdfs

follitude:
It seems that we need to change the order of linked libraries to make it work by using gcc.


axboe:
Pulled, thanks.

PR #489, opened 2017-11-08
Compressed log filename probably should have had a log.gz suffix

szaydel:
I am guessing this was a type made worse by copy/paste.


axboe:
It's not a typo. The files aren't compatible with gzip, the 'fz' is for 'fio zip'.

szaydel:
OK, well that explains that. I guess best to close this. Thanks for explaining.

PR #493, opened 2017-11-17
Add libpmem ioengine

tishizaki:
Adding an ioengine that access with the libpmem as memory
through a memory mmaped file on DAX filesystem.

It's very similar to the mmap engine and the dev-dax engine.

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1820,6 +1820,11 @@ I/O engine
 			set  `filesize` so that all the accounting still occurs, but no
 			actual I/O will be done other than creating the file.
 
+		**libpmem**
+			Read and write using mmap I/O to a file on a filesystem
+			mounted with DAX on a persistent memory device through the NVML
+			libpmem library.
+
 I/O engine specific parameters
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,9 @@ endif
 ifdef CONFIG_LINUX_DEVDAX
   SOURCE += engines/dev-dax.c
 endif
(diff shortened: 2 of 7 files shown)


axboe:
I've pulled this in, but made a number of changes on top to both code style, but also fixing up some redundant code (like Pagesize). Please verify that it works as intended for you.

axboe:
Actually, looking a bit closer, what makes this any different than the pmemblk engine? It does support windows/freebsd, which is nice, but apart from that?

tishizaki:
Thanks for fixing and merging.

I think the libpmem engine's purpose is testing most basic operation(memcpy, cpu cache flush, page fault) for persistent memory.
On the other hand the pmemblk engine's perpose is testing the user-space BTT feature.

sitsofe:
@tishizaki it might be worth spelling out the difference in the help/HOWTO/man page (e.g. http://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-ioengine ). At the moment it just sounds like one uses libpmem and the other uses libpmemblk...

tishizaki:
@bpkroth gettimeofday() is executed for each I/O operation if you enable latency measurement options. So I think it should turn off those options when you measure peak throughput.
Certainly, it will be better to use gtod_reduce as axboe mentions.

sitsofe:
@bpkroth Just for the record: [`gettimeofday()` calls on Linux can be implemented by vDSO](http://man7.org/linux/man-pages/man7/vdso.7.html) and may not even need to make a "real" syscall thus remaining entirely userspace and becoming cheap (but this depends on the platform being used and the parameters passed to the call).

PR #494, opened 2017-11-22
Minor cleanup and documentation tweaks

sitsofe:
- Fix up some broken formatting introduced with 074f08170af15ca8ebd469c3a1fa09c38f29904e ("doc: rewrite write_*_log sections")
- Reword buffer_compress_percentage, buffer_compress_chunk and dedupe_percentage sections.

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -2748,8 +2748,8 @@ Measurements and reporting
 .. option:: write_lat_log=str
 
 	Same as :option:`write_bw_log`, except this option creates I/O
-	submission (e.g., `file:`name_slat.x.log`), completion (e.g.,
-	`file:`name_clat.x.log`), and total (e.g., `file:`name_lat.x.log`)
+	submission (e.g., :file:`name_slat.x.log`), completion (e.g.,
+	:file:`name_clat.x.log`), and total (e.g., :file:`name_lat.x.log`)
 	latency files instead. See :option:`write_bw_log` for details about
 	the filename format and `Log File Formats`_ for how data is structured
 	within the files.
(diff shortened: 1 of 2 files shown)


init.c
```diff
@@ -772,9 +772,6 @@ static int fixup_options(struct thread_data *td)
 			ret = warnings_fatal;
 		}
 
-		if (!fio_option_is_set(o, refill_buffers))
```
axboe: What am I missing here, I don't see where else it's being set for this?

sitsofe: I think it gets set over in `init_flags()` https://github.com/sitsofe/fio/blob/9164932ecbb3aa4b1538080f97452a2fdb3f02f1/init.c#L1112 and `init_flags()` is called before `fixup_options()` over in `add_job()` (see https://github.com/sitsofe/fio/blob/9164932ecbb3aa4b1538080f97452a2fdb3f02f1/init.c#L1359 ).

axboe: init_flags() may set the REFILL_BUFFERS flag, but not o->refill_buffers.

sitsofe: Ah you're right (e.g. in any case where you're not doing `compress_percentage`). OK let me drop that commit.

sitsofe:
OK updated.

sitsofe:
Hmm I wonder, should `if (!fio_option_is_set(o, refill_buffers))` needs to have `TD_F_REFILL_BUFFERS` inside the braces too...

axboe:
I think that init_flags() should just be moved after we've fiddled with the options.

axboe:
In any case, I've pulled this branch, since you dropped the bad commit.

Issue #500, opened 2017-12-19
Large number of threads result in Seg faults


yurzo:
The output:
```
Starting 1920 threads
Jobs: 640 (f=639)
...
),P(2),W(1),P(2),W(1),Segmentation fault (core dumped)
```

From gdb:

```[New LWP 69851]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/root/hw_qual_suite/hw-site-pkgs/wperfmark/fio/foed --output-format=normal --cr'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  sprintf (__fmt=<optimized out>, __s=0x7f03cef511d8 "") at /usr/include/x86_64-linux-gnu/bits/stdio2.h:33
33	  return __builtin___sprintf_chk (__s, __USE_FORTIFY_LEVEL - 1,
(gdb)
```

Thoughts? ideas?

sitsofe:
@yurzo:
* What does your fio jobfile and fio command line look like?
* Which version of fio are you using and does the problem go away if you are using fio 3.3 (https://github.com/axboe/fio/releases ) or later?
* Which version of Linux are you using?
* Did the information in http://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-max-jobs make a difference?
* What does the full stack trace look like (`thread apply all bt`), does your system have a lot RAM and is it all available to fio?

In general we need issues to be supplied with at least the information requested in https://github.com/axboe/fio/blob/fio-3.3/REPORTING-BUGS ...

yurzo:
@sitsofe:
> Which version of fio are you using and does the problem go away if you are using fio 3.3 (https://github.com/axboe/fio/releases ) or later? 
Cloned from trunk
> Which version of Linux are you using?
Ubuntu
> Did the information in http://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-max-jobs make a difference?
Nope
> It does have a lot of RAM. I need to figure out how to fetch the full stack trace.

What I do know now is that the problem goes away if --output-format=json.
i.e. the bug is in the code that periodically prints the thread state.

I'll do more debugging in the new year.

sitsofe:
@yurzo:

> I need to figure out how to fetch the full stack trace.

Try
```
thread apply all bt
```
at your gdb prompt.
.

[...]

PR #502, opened 2017-12-24
Eta display fixes

sitsofe:
A bunch of commits around the eta display that

* Fix various off-by-one errors
* Skip the clearing the stale parts of a line when we are starting a new line anyway
* A change how we work out how much of a line is stale
* An increase of the output buffer (as requested in https://github.com/axboe/fio/issues/500#issuecomment-353374981 )

The new stale line clearing is inexact - if the eta line had a newline and then printed 40 characters and the current eta line just prints 40 characters then we will end up thinking we need to clear one character. If you think that is necessary it could be fixed by introducing a variable to track where the "true" start of the line is within the buffer.

### DIFF
--- a/eta.c
+++ b/eta.c
@@ -531,6 +531,7 @@ void display_thread_status(struct jobs_eta *je)
 
 	if (eta_new_line_pending) {
 		eta_new_line_pending = 0;
+		linelen_last = 0;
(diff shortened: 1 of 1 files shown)


axboe:
Pulled, thanks Sitsofe.

PR #504, opened 2017-12-28
mutex: down_timeout: check against the base time

yashi:
When the commit 8b6a404cdd2c40715885e562416c3db039912773 changed
timeval to timespec, it accsidentally picked timeout time `t' instead
of the current `base' time for checking timewait() really timed out.

t is base + msecs as follows:

      base = gettimeofday()
      t = base + msecs
      pthread_cond_timedwait()
      while (!done)
         if (timedout)
	    if (mutex_timed_out(t, msecs))
	       done = true

Thus, after a good timedout, it burns the cpu for the given msecs
seconds doing tight loop in the while shown above.  This causes 30 to
40 % CPU load on my i5-3360M 2.80GHz while fio server is running.

### DIFF
--- a/mutex.c
+++ b/mutex.c
@@ -156,14 +156,15 @@ static bool mutex_timed_out(struct timespec *t, unsigned int msecs)
 int fio_mutex_down_timeout(struct fio_mutex *mutex, unsigned int msecs)
 {
 	struct timeval tv_s;
+	struct timespec base;
 	struct timespec t;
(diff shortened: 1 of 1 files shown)


vincentkfu:
This patch looks correct to me. This bug was indeed introduced in 8b6a404.

axboe:
Thanks, good catch - pulled.

PR #505, opened 2017-12-31
Percentile fixes

sitsofe:
First fix address total latency percentiles being 0 when only completion latency is disabled.

The second disables percentile gathering when the corresponding latency option has been disabled. This later change might be more contentious as it might make some outputs (e.g. percentiles with JSON) no longer appear to be present when using conflicting options...


axboe:
Looks good to me, pulled, thanks.

PR #524, opened 2018-02-06
Make fiologparser_hist.py work for python3

brycepg:


brycepg:
I definitely agree @sitsofe. I decided to make the problem `map` functions list comprehensions since they are much more readable anyway.

sitsofe:
@brycepg Assuming it works with both Python 2 and Python 3 this looks good. Any chance you could add your "Signed-off-by:" (e.g. like https://github.com/axboe/fio/commit/f5ec81235eaf21bd3a97556427d4a84e48a87e54 ) to the commit?

sitsofe:
I can confirm that this commit makes the script python3 compatible while remaining usable on with python2. With the previously mentioned change and possibly a comment saying the script is python3 compatible I think this should be good to go.

axboe:
Pulled, thanks everyone.

PR #529, opened 2018-02-09
Add support for Ceph Rados benchmarking.

ifed01:
Signed-off-by: Igor Fedotov <ifedotov@suse.com>

### DIFF
--- a/engines/rados.c
+++ b/engines/rados.c
@@ -1,0 +1,479 @@
+/*
+ *  Ceph Rados engine
+ *
+ * IO engine using Ceph's RADOS interface to test low-level performance of
+ * Ceph OSDs.
+ *
+ */
+
+#include <rados/librados.h>
+#include <pthread.h>
+#include "fio.h"
+#include "../optgroup.h"
+
+struct fio_rados_iou {
+	struct thread_data *td;
+	struct io_u *io_u;
+	rados_completion_t completion;
+	rados_write_op_t write_op;
+};
+
+struct rados_data {
+	rados_t cluster;
+	rados_ioctx_t io_ctx;
+	char **objects;
+	size_t object_count;
+	struct io_u **aio_events;
+	bool connected;
+};
+
+/* fio configuration options read from the job file */
+struct rados_options {
+	void *pad;
+	char *cluster_name;
+	char *pool_name;
+	char *client_name;
+	int busy_poll;
+};
+
+static struct fio_option options[] = {
+	{
+		.name     = "clustername",
+		.lname    = "ceph cluster name",
+		.type     = FIO_OPT_STR_STORE,
+		.help     = "Cluster name for ceph",
+		.off1     = offsetof(struct rados_options, cluster_name),
+		.category = FIO_OPT_C_ENGINE,
+		.group    = FIO_OPT_G_RBD,
+	},
+	{
+		.name     = "pool",
+		.lname    = "pool name to use",
+		.type     = FIO_OPT_STR_STORE,
+		.help     = "Ceph pool name to benchmark against",
+		.off1     = offsetof(struct rados_options, pool_name),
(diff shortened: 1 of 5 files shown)


axboe:
Looks fine to me. Can you add HOWTO and fio.1 additions as well to explain the new engine type?

ifed01:
@axboe - thanks for looking into that. Will do next week.

engines/rados.c
```diff
@@ -0,0 +1,501 @@
...
+	rados_completion_t completion;
+	rados_write_op_t write_op;
+	int io_seen;
+	int io_complete;
+};
+
+struct rados_data {
+	rados_t cluster;
+	rados_ioctx_t io_ctx;
+	char** objects;
```
axboe: Please follow the fio style, which would make this:

```
char **objects;
```
like you actually do for other members.

engines/rados.c
```diff
@@ -0,0 +1,501 @@
...
+	size_t object_count;
+	struct io_u **aio_events;
+	bool connected;
+};
+
+/// fio configuration options read from the job file
```
axboe: No C++ style comments, use /**/

engines/rados.c
```diff
@@ -0,0 +1,501 @@
...
+struct Options {
```
axboe: struct rados_options

engines/rados.c
```diff
@@ -0,0 +1,501 @@
...
+	{
+		.name		= "pool",
+		.lname		= "pool name to use",
+		.type		= FIO_OPT_STR_STORE,
+		.help		= "Ceph pool name to benchmark against",
+		.off1		= offsetof(struct Options, pool_name),
+		.category	= FIO_OPT_C_ENGINE,
+		.group		= FIO_OPT_G_RBD,
+	},
+	{
```
axboe: space vs tabs confusion here

engines/rados.c
```diff
@@ -0,0 +1,501 @@
...
+		r = rados_aio_create_completion(io_u,
+		on_complete_cb,
+		NULL /*cb_safe*/,
+		&fri->completion);
+		if (r < 0) {
+			log_err("rados_aio_create_completion failed.\n");
+			goto failed;
+		}
+
+		r = rados_aio_write(
```
axboe: Indentation. The fio standard is to use up 80 chars, then spill into the next line if you have more arguments.

engines/rados.c
```diff
@@ -0,0 +1,501 @@
...
+			fri = u->engine_data;
+			if (fri->io_complete) {
+				fri->io_complete = 0;
+				rados->aio_events[events] = u;
+				events++;
+			}
+		}
+		if (events >= min)
+			break;
+		usleep(100);
```
axboe: You should not sleep unless ```t``` is set and indicates that the callers wants to wait for completions.

ifed01:
@axboe - updated.
I decided to introduce an additional parameter 'heavy_load' to control whether straightforward rados poll or a wait on conditional variable are used. The rationale is a trade-off between CPU utilization and resulting throughput. In direct poll mode I'm getting ~10-15% better throughput at cost of 100% CPU utilization. Don't see any other ways to get such a performance without that CPU load. And not sure that high utilization is always good enough. Hence adding an option to be able to control this behavior.

sitsofe:
@ifed01 re direct poll - is there any chance you can the option busy_poll ? There's precedent insofar as that's what the RBD engine calls it: (https://github.com/axboe/fio/blob/1a7bf7536e6cb4eda8354b527a639ca41f91d767/engines/rbd.c#L89 )...

ifed01:
@sitsofe - renamed to busy_poll.

axboe:
I've pulled this, but I'm hoping you'll also get the HOWTO updated for the engine and parameters.

ifed01:
@axboe - thanks, will do.

PR #543, opened 2018-02-27
Refer to zlib-static package in zlib static build warning

### LINKED ISSUE #542: zlib not found when building statically
Reproduction:

    git checkout fio-3.5
    make clean
    yum install -y zlib-devel
    ./configure --build-static

Configure notes that zlib isn't available:

    bguinta  ➜  fio git:(cd174b90) ✗ ./configure --build-static | grep zlib
    zlib                          no
    Consider installing zlib-dev (zlib-devel), some fio features depend on it.

But without --build-static this works fine:

    bguinta  ➜  fio git:(cd174b90) ✗ ./configure | grep zlib       
    zlib                          yes

brycepg:
Close #542

### DIFF
--- a/configure
+++ b/configure
@@ -2376,7 +2376,7 @@ if test "$disable_opt" = "yes" ; then
   output_sym "CONFIG_DISABLE_OPTIMIZATIONS"
 fi
 if test "$zlib" = "no" ; then
-  echo "Consider installing zlib-dev (zlib-devel), some fio features depend on it."
+  echo "Consider installing zlib-dev (zlib-devel or zlib-static), some fio features depend on it."
 fi
 if test "$cuda" = "yes" ; then
   output_sym "CONFIG_CUDA"


axboe:
Pulled, thanks.

configure
```diff
@@ -2376,7 +2376,7 @@ if test "$disable_opt" = "yes" ; then
   output_sym "CONFIG_DISABLE_OPTIMIZATIONS"
 fi
 if test "$zlib" = "no" ; then
-  echo "Consider installing zlib-dev (zlib-devel), some fio features depend on it."
+  echo "Consider installing zlib-dev (zlib-devel or zlib-static), some fio features depend on it."
```
sitsofe: @brycepg I'm guessing to use zlib-static you would have to have zlib-devel too (-devel still provides the headers) so zlib-static isn't really an alternative but more of a complement...

axboe: It is a bit unclear and, again, RH centric. I've pushed an improvement.

PR #544, opened 2018-02-28
Add documentation about statically compiling fio

brycepg:
Add a section about statically linking the IO Engine as well (took me a long time to figure out how).

### DIFF
--- a/README
+++ b/README
@@ -200,6 +200,38 @@ output formats are supported run ``make -C doc help``.
 .. _reStructuredText: http://www.sphinx-doc.org/rest.html
 .. _Sphinx: http://www.sphinx-doc.org
 
+Static Compilation
+~~~~~~~~~~~~~~~~~~
+
+Fio can be statically compiled using ``./configure --build-static``. Make sure to install static packages of desired libraries such as ``zlib-static``. Desired IO Engines such as ``libaio`` will also need to be statically included.
+
+static compilation with libaio
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -200,6 +200,38 @@ output formats are supported run ``make -C doc help``.
...
 
+Static Compilation
+~~~~~~~~~~~~~~~~~~
+
+Fio can be statically compiled using ``./configure --build-static``. Make sure to install static packages of desired libraries such as ``zlib-static``. Desired IO Engines such as ``libaio`` will also need to be statically included.
+
+static compilation with libaio
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+libaio can be statically compiled into fio. To do so, clone the libaio source and compile libaio:
```
sitsofe: Some distros such as Debian (https://packages.debian.org/stretch/amd64/libaio-dev/filelist ) actually ship a libaio.a in their development package making this far easier. In fact Debian often ships static binaries with its -dev packages - https://packages.debian.org/stretch/amd64/zlib1g-dev/filelist so your instructions are probably aimed at Red Hat style distros?

axboe: Yeah, this is largely distro dependent. The instructions should mention that devel packages sometimes ship with static versions, and that sometimes you need to install specific static library versions. This isn't really specific to building fio statically though, it's generic advice. Which is fine, but it should not be lob sided to RH style distros. I don't run any of them :-)

brycepg: I do see what you mean. I'm only running Centos/Fedora right now. I think I'll just put this text on my blog with a new title. Thanks for the feedback

Issue #545, opened 2018-02-28
Overflow of latency counters during long NVMe test running


ifke:
The fio utility uses uint32_t arrays for individual latency counters and uint64_t for total ones. In stat.h:
```
    uint32_t io_u_map[FIO_IO_U_MAP_NR];
    uint32_t io_u_submit[FIO_IO_U_MAP_NR];
    uint32_t io_u_complete[FIO_IO_U_MAP_NR];
    uint32_t io_u_lat_n[FIO_IO_U_LAT_N_NR];
    uint32_t io_u_lat_u[FIO_IO_U_LAT_U_NR];
    uint32_t io_u_lat_m[FIO_IO_U_LAT_M_NR];
    uint32_t io_u_plat[DDIR_RWDIR_CNT][FIO_IO_U_PLAT_NR];
    uint32_t io_u_sync_plat[FIO_IO_U_PLAT_NR];
    uint32_t pad;

    uint64_t total_io_u[DDIR_RWDIR_SYNC_CNT];
    uint64_t short_io_u[DDIR_RWDIR_CNT];
    uint64_t drop_io_u[DDIR_RWDIR_CNT];
    uint64_t total_submit;
    uint64_t total_complete;
```

We got an issue after eight hours NVMe testing with the libaio engine. Total iops value was 6323529453, and most of iops had latency 250 us. So there was overflow of corresponding io_u_lat_u counter and we saw incorrect latency distribution: the sum of values of fields latency_ns, latency_us, latency_ms was significantly less than 100 percent. The fio output in json format is here.
[fio-3.5.eighthours.txt](https://github.com/axboe/fio/files/1766154/fio-3.5.eighthours.txt)

Is it possible to change latency counters to uint64_t type and maybe do so for other counters incremented on each IO operation?

sitsofe:
@ifke This would make sense and was suggested by Jens over in https://www.spinics.net/lists/fio/msg06407.html . Would you like to whip up a patch for this?

ifke:
I can try. I have already patched fio of version 2.2-10 used in our company. I saw new uint32_t counters clat_percentiles and lat_percentiles in version 3.5. I'm not sure whether I have to update them too.

sitsofe:
Yes, I think they would need updating too. Basically anything that is storing total counts of I/Os in 32 bits would need bumping to 64 bits. It's probably worth giving the patched fio's null ioengine a spin using the the sanitizer's `-fsanitize=integer` flag...

[...]

PR #547, opened 2018-03-01
Fix counter overflow

ifke:
Change sizes of counters incremented on each I/O operation from 32-bit to 64-bit. It allows to avoid overflow during several hours test of NVMe devices.

That's related to issue https://github.com/axboe/fio/issues/545

### DIFF
--- a/client.c
+++ b/client.c
@@ -905,21 +905,21 @@ static void convert_ts(struct thread_stat *dst, struct thread_stat *src)
 	}
 
 	for (i = 0; i < FIO_IO_U_MAP_NR; i++) {
-		dst->io_u_map[i]	= le32_to_cpu(src->io_u_map[i]);
-		dst->io_u_submit[i]	= le32_to_cpu(src->io_u_submit[i]);
-		dst->io_u_complete[i]	= le32_to_cpu(src->io_u_complete[i]);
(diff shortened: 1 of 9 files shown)


sitsofe:
Unfortunately your changes are breaking the CI builds. Visit this page in GitHub and click the red crosses to see more...

sitsofe:
Ok that looks good. Could you up the FIO_SERVER_VER as in https://github.com/axboe/fio/commit/52830b23d88149617d553a98fe6cb81e2fe1d743 , squash all these commits into one and add your signed off by?

ifke:
Done

client.c
```diff
@@ -1283,7 +1283,7 @@ static void client_flush_hist_samples(FILE *f, int hist_coarseness, void *sample
 	int log_offset;
 	uint64_t i, j, nr_samples;
 	struct io_u_plat_entry *entry;
-	unsigned int *io_u_plat;
+	unsigned long *io_u_plat;
```
axboe: This should probably be uint64_t as well, unsigned long is only 32-bit on 32-bit archs.

ifke:
Done

axboe:
Thanks, pulled.

PR #526, opened 2018-02-06
Correct slave device dev path

bharatpotnuri:
Running fio over NVMEoF multipath slave devices is failing with "No such
file or directory" error.
Corrected the dev path for slave devices.
Supposed fix for issue #525

Signed-off-by: Potnuri Bharat Teja <bharat@chelsio.com>

PR #549, opened 2018-03-02
Fix Linux NVMe disk stats

sitsofe:
The main commit builds on the work of @bharatpotnuri in https://github.com/axboe/fio/pull/526. As the the main NVMe disk doesn't record any stats when NVMe multipath is in use and virtual NVMe slave devices aren't real block devices it is necessary to try a different sysfs path when the usual one fails.

The second commit just changes some spacing and remove a dot in logging.


axboe:
Thanks for getting this sorted! Pulled.

PR #552, opened 2018-03-05
doc: fix typos in fio man page

dirtyharrycallahan:


fio.1
```diff
@@ -1552,7 +1552,7 @@ single CPU at the desired rate. A job never finishes unless there is
 at least one non\-cpuio job.
 .TP
 .B guasi
-The GUASI I/O engine is the Generic Userspace Asyncronous Syscall
+The GUASI I/O engine is the Generic Userspace Asynchronous Syscall
```
sitsofe: This one is oddly consistent - if you visit http://www.xmailserver.org/guasi-lib.html it's misspelled there too!

axboe: Pretty sure I just copy/pasted the description way back when, which would explain why they are consistently wrong :-)

axboe:
Pulled - would you mind giving the HOWTO the same treatment? Thanks!

PR #557, opened 2018-03-07
Disable march optimisations in a couple of scenarios

sitsofe:
* If the user is specifying `--disable-optimizations` turn off `march=native` setting too because the chances are they aren't interested. This should keep some distros that already specify `--disable-optimizations` when building their fio packages from getting surprised...
* Disable `march=native` setting in AppVeyor builds. While this technically means we aren't running with the same defaults users get out of the box it's safer to prevent the downloadable MSI binaries from gaining instructions that might be specific to the cloud build host. I suppose if we care enough we can introduce a "native" CI build too but it doesn't seem worth it for now.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,7 +17,7 @@ install:
   - SET PATH=%CYG_ROOT%\bin;%PATH% # NB: Changed env variables persist to later sections
 
 build_script:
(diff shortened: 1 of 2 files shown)


axboe:
Looks fine to me. But why are distros using --disable-optimizations?

sitsofe:
> But why are distros using --disable-optimizations?

So they can only and exactly their own compilation flags. For example see https://src.fedoraproject.org/rpms/fio/blob/master/f/fio.spec#_36 .

axboe:
Gotcha, that actually makes sense.

PR #558, opened 2018-03-12
Make it possible to build fio on NetBSD

bvanassche:
To verify whether or not my Linux-specific changes did not cause any portability problems, I tried to build fio on a NetBSD system. The patches in this series make it possible to build fio on NetBSD. The only verification I have done of these changes is to verify that "make test" passes on both Linux and NetBSD.

### DIFF
--- a/configure
+++ b/configure
@@ -252,6 +252,8 @@ elif check_define __linux__ ; then
   targetos="Linux"
 elif check_define __OpenBSD__ ; then
   targetos='OpenBSD'
+elif check_define __NetBSD__ ; then
+  targetos='NetBSD'
 elif check_define __sun__ ; then
   targetos='SunOS'
   CFLAGS="$CFLAGS -D_REENTRANT"
@@ -280,7 +282,7 @@ fi
 # cross-compiling to one of these OSes then you'll need to specify
 # the correct CPU with the --cpu option.
 case $targetos in
-AIX|OpenBSD)
+AIX|*BSD)
   # Unless explicitly enabled, turn off lex.
   # OpenBSD will hit syntax error when enabled.
   if test -z "$disable_lex" ; then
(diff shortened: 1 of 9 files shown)


axboe:
Pulled, thanks Bart.

krytarowski:
Out of curiosity, are there differences between NetBSD and Linux in terms of benchmarking results?

configure
```diff
@@ -280,7 +282,7 @@ fi
 # cross-compiling to one of these OSes then you'll need to specify
 # the correct CPU with the --cpu option.
 case $targetos in
-AIX|OpenBSD)
+AIX|*BSD)
```
sitsofe: @bvanassche Won't this also disable it on FreeBSD?

bvanassche: Is FreeBSD already supported? From the configure script:

    $ grep targetos= configure 
    targetos=""
      targetos="Android"
      targetos="Linux"
      targetos='OpenBSD'
      targetos='NetBSD'
      targetos='SunOS'
      targetos='CYGWIN'
      targetos=`uname -s`

bvanassche:
On 03/12/18 19:11, Kamil Rytarowski wrote:
> Out of curiosity, are there differences between NetBSD and Linux in 
> terms of benchmarking results?

That's a great question. Anyone who wants to know this can run the 
necessary measurements.

configure
axboe: FreeBSD is supported and works, we just don't check for it specifically. As such, the change would not have disabled lex on FreeBSD, but I still think the targetos check should just target netbsd and openbsd specifically.

sitsofe: @bvanassche it is - it just falls into the `uname -s` case.

axboe: Oh yes, good point, it is actually broken as-is.

sitsofe: @axboe Thanks for fixing that one up (I'm not a FreeBSD user, just a curious bystander)!

axboe: Thanks for the review, good catch!

PR #560, opened 2018-03-13
Improvements for analyzing fio with Valgrind

bvanassche:
The patches in this pull request inform Valgrind about the memory allocations performed by the smalloc allocator and also suppress two drd complaints. drd is a thread error detector. More information about drd is available at http://valgrind.org/docs/manual/drd-manual.html.


smalloc.c
```diff
@@ -12,6 +12,14 @@
 #include <sys/types.h>
 #include <limits.h>
 #include <fcntl.h>
+#include "config-host.h"
```
axboe: Not needed, config-host.h is included automatically.

gettime.c
```diff
@@ -564,7 +564,6 @@ struct clock_thread {
 	int cpu;
 	int debug;
 	pthread_mutex_t lock;
-	pthread_mutex_t started;
```
axboe: I think we should change this to be a fio_mutex instead. The mutex is used to synchronize startup of the clock threads.

bvanassche: What should the behavior be of the actions on the 'started' synchronization object? The unlock from the clock threads on t->started succeeds immediately. I don't see how the actions on 'started' impose any ordering constraints.

smalloc.c
bvanassche: OK, I will remove that #include directive.

gettime.c
axboe: By the time we end up looping around and unlocking t->lock to kick off the threads at the same time, we know they are all waiting there. It makes us independent of thread startup time.

bvanassche:
This pull request has been updated as follows:
- #include "config-host.h" has been removed from the first patch.
- The second and third patches have been replaced by a new patch that should make the delay between the thread start times even smaller than it is today.

gettime.c
```diff
@@ -559,12 +559,14 @@ struct clock_entry {
 	uint64_t tsc;
 };
 
+static pthread_mutex_t clock_lock;
```
axboe: This might be miserable on large CPU counts, where the previous approach used private mutexes. I still think the best solution, if the cross-thread lock/unlock is an issue, is to switch them to fio_mutex'es.

sitsofe:
I have another approach for the clock thread problem see https://github.com/sitsofe/fio/commit/7471f3fabaf2998e687801a4db9391025a993b0d . Potentially the __sync_synchronize is still needed...

sitsofe:
Well I say that it's different but actually the broadcast to wake them all up at the same time is identical so ignore that bit :-)

configure
```diff
@@ -2050,6 +2050,23 @@ fi
...
+cat > $TMPC << EOF
+#include <valgrind/valgrind.h>
+int main(int argc, char **argv)
+{
+  return 0;
+}
+EOF
+if compile_prog "" "" "valgrind_valgrind_h"; then
+  valgrind_valgrind_h="yes"
+  output_sym "HAVE_VALGRIND_VALGRIND_H"
```
sitsofe: Don't we normally save the result into a variable and then towards the end of the makefile output_sym a CONFIG_* variables (e.g. CONFIG_VALGRIND) that we then act on later?

bvanassche:
This pull request has been updated again. All review comments should have been addressed. Please let me know if I missed something.

configure
axboe: Additionally, it's named differently than the others. We should make it CONFIG_VALGRIND_DEV or something.

bvanassche: I had missed this comment. It should now have been addressed.

gettime.c
```diff
@@ -693,6 +698,12 @@ int fio_monotonic_clocktest(int debug)
 	if (debug)
 		log_info("cs: Testing %u CPUs\n", nr_cpus);
 
+	/*
+	 * Since multiple threads modify the variable seq on purpose, avoid
+	 * that thread checking tools complain about these accesses.
+	 */
+	DRD_IGNORE_VAR(seq);
```
sitsofe: It is probably right to warn on this one. Here's what one of the author's of ThreadSanitizer said: https://groups.google.com/d/msg/thread-sanitizer/XzNS5pRTaE8/ivMDq7ybAgAJ

> > Doesn't the fact that the write happens atomically mean the fact the read 
> > might be non-atomic is safe? 
>
> Hi, 
>
> C standard is pretty clear on this -- no, it's not safe, both accesses 
> must be atomic.

configure
```diff
@@ -2354,6 +2370,9 @@ fi
 if test "$disable_opt" = "yes" ; then
   output_sym "CONFIG_DISABLE_OPTIMIZATIONS"
 fi
+if test "$valgrind_valgrind_h" = "yes"; then
+  output_sym "HAVE_VALGRIND_VALGRIND_H"
```
sitsofe: Perhaps this should be CONFIG_VALGRIND_DEV as mentioned by Jens. I'm also wondering if we should just roll the drd header check into it too as it seems likely you would have both and it saves another compile test...

bvanassche: I will rename HAVE_VALGRIND_VALGRIND_H into CONFIG_VALGRIND_DEV and I will also combine the two header checks into a single check.

bvanassche:
Yet another update of this pull request. As usual, feedback is welcome.

configure
```diff
@@ -2050,6 +2050,24 @@ fi
...
+#include <valgrind/drd.h>
+int main(int argc, char **argv)
+{
+  return 0;
+}
+EOF
+if compile_prog "" "" "valgrind_dev"; then
+  valgrind_dev="yes"
+fi
+print_config "<valgrind/drd.h>" "$valgrind_dev"
```
sitsofe: Perhaps just say "valgrind support" rather than "<valgrind/drd.h>" here.

init.c
```diff
@@ -376,6 +383,8 @@ static int setup_thread_area(void)
 #endif
 
 	memset(threads, 0, max_jobs * sizeof(struct thread_data));
+	for (i = 0; i < max_jobs; i++)
+		DRD_IGNORE_VAR(threads[i]);
```
sitsofe: I think if we teach fio to stop detaching threads and join on them when they're complete a lot of the issues related to this one will go away...

sitsofe:
I've got to admit Bart has a point with the "Rename fio_mutex into fio_sema" - fio_mutex was really acting like a semaphore even if pieces of it were implemented by using mutexes because even a thread who didn't lock it could unlock it. My only thoughts are:
1. I prefer sem to sema :-)
2. The filename synchr.c didn't initially mean much to me (I prefer synchronization.c)
3. The seqlocks aren't inside it

3 doesn't seem to matter, and 1 and 2 are just a matter of personal taste from a passerby.

gettime.c
```diff
@@ -731,6 +722,7 @@ int fio_monotonic_clocktest(int debug)
 		pthread_join(t->thread, &ret);
 		if (ret)
 			failed++;
+		__fio_sema_remove(&t->lock);
 	}
```
sitsofe: Why not `fio_mutex_remove()` so it get unmapped for safety reasons?

gettime.c
```diff
@@ -731,6 +722,7 @@ int fio_monotonic_clocktest(int debug)
 		pthread_join(t->thread, &ret);
 		if (ret)
 			failed++;
+		__fio_sema_remove(&t->lock);
```
sitsofe: Why not `fio_mutex_remove()` so it get unmapped for safety reasons?

gettime.c
```diff
@@ -702,9 +701,7 @@ int fio_monotonic_clocktest(int debug)
 		t->seq = &seq;
 		t->nr_entries = nr_entries;
 		t->entries = &entries[i * nr_entries];
-		pthread_mutex_init(&t->lock, NULL);
-		pthread_mutex_init(&t->started, NULL);
-		pthread_mutex_lock(&t->lock);
+		__fio_sema_init(&t->lock, FIO_SEMA_LOCKED);
```
sitsofe: As below why not `fio_sema_init()` so it get mapped for safety reasons? Do we just not the need the safety checks?

sitsofe:
OK the rest looks fine to me (also interesting that commit 
6583b89 didn't get CI tested for some reason).

sitsofe:
Ah 6583b89 did get tested it's just github displays the commits out of order on the pull page.

gettime.c
bvanassche: What safety checks are you referring to? __fio_sema_remove() checks against memory corruption by verifying the value that is present in the 'magic' member.

bvanassche:
I will rename the file synchr.c into synchronization.c. But I'm not sure whether it's really that useful to rename "sema" into "sem". I think that's a matter of personal taste ...

gettime.c
sitsofe: I think I meant the fact that munmap() will prevent the lock being used once it's destroyed.

init.c
sitsofe: ...but there's an argument that perhaps we should keep this ignore until that happens.

gettime.c
bvanassche: Isn't that using mmap() for a purpose other than what it was introduced for? It can e.g. happen that another thread maps other data at the same address so calling munmap() does not guarantee that use-after-free will trigger a crash. Personally I prefer to use Valgrind for detection of use-after-free.

bvanassche:
On 03/14/18 23:03, Sitsofe Wheeler wrote:
> *@sitsofe* commented on this pull request.
> 
> In init.c <https://github.com/axboe/fio/pull/560#discussion_r174683912>:
> 
>> @@ -376,6 +383,8 @@ static int setup_thread_area(void)
>   #endif
>   
>   	memset(threads, 0, max_jobs * sizeof(struct thread_data));
> +	for (i = 0; i < max_jobs; i++)
> +		DRD_IGNORE_VAR(threads[i]);
> 
> I think if we teach fio to stop detaching threads and join on them when 
> they're complete a lot of the issues related to this one will go away...

That's a very interesting thought. However, I consider modifying the fio 
threading model as something that falls outside the scope of this patch 
series.

bvanassche:
A new version of this patch series has been pushed and all review comments should have been addressed. As usual, feedback is welcome.

sitsofe:
@bvanassche you're very diligent! At any rate the latest pull request looks good to me - @axboe?

axboe:
Pulled, thanks Bart.

PR #561, opened 2018-03-15
Add an asprintf() implementation

bvanassche:
Since I would like to use the asprintf() function in the ZBC code and since that function is not available on every platform supported by fio, add an asprintf() implementation. Use asprintf() where appropriate.


axboe:
Merged, thanks Bart.

PR #562, opened 2018-03-16
Call pthread_cond_signal() with the associated mutex held

bvanassche:
Sitsofe's comment about deadlocks on MinGW made me have a look at the pthread_cond_signal() and pthread_cond_broadcast() calls in fio. Hence this pull request.


axboe:
Pulled, thanks. Sucks that we need it, though.

PR #563, opened 2018-03-20
Expand fiologparser_hist operations with new options

shimrot:
New options are: 
* noweight: don't weight samples across intervals
* percentiles: allow print of more percentiles than just 50,90,95%s
* Fix so works with newer 'ns' results, but add --usbin option
  so can still be used with older results.
* directions=rwtm option to allow print of independent direction results


axboe:
Pulled, but you have a ton of trailing whitespace in various lines. I killed them, but please mind those in the future.

shimrot:
Interesting.   I spent some effort to removing trailing whitespace.  Not
sure whats creating it.  I'll double check further.
Thx

PR #564, opened 2018-03-21
Include refactoring

sitsofe:
Adjust the #includes across fio with an aim to make files more self dependent:
- Aim to remove more includes than we add
- Try to make files more standalone
- Try to ensure that .c files include matching .h files so the compiler can check declarations match definitions
- Make some of the oslib/str.* files more uniform
- Use get rid of `/sys/` when referencing poll.h and fcntl.h .
- Add some comment markup to make future analysis a bit more accurate.


sitsofe:
I should note there's a risk due to the amount of code motion but these changes compile on recent Linux, macOS, Illumos and FreeBSD.

axboe:
Pulled, thanks.

PR #565, opened 2018-03-22
compiler: set minimum compiler version to GCC 4.1.0

sitsofe:
fio hasn't compiled under GCC 3 for a while due to unguarded calls to
GCC's __sync intrinsics in critical paths. Since those intrinsics didn't
come along until GCC 4.1.0 change fio to explicitly error out on older
compilers.


axboe:
Looks fine, pulled, thanks.

PR #567, opened 2018-03-23
smalloc: Remove Valgrind instrumentation

bvanassche:
Since it can happen that memory allocated with smalloc() is freed by
a child process and since Valgrind's memory tracking is per process
the smalloc()/sfree() instrumentation leads to incorrect Valgrind
reports. Hence remove that instrumentation again.

Fixes: 0ffccc21fcd6 ("Improve Valgrind instrumentation of memory allocations")

### DIFF
--- a/smalloc.c
+++ b/smalloc.c
@@ -5,13 +5,6 @@
 #include <sys/mman.h>
 #include <assert.h>
 #include <string.h>
-#ifdef CONFIG_VALGRIND_DEV
-#include <valgrind/valgrind.h>
-#else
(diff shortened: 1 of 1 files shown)


axboe:
Pulled, thanks. BTW, something funky is going on with that commit, it's got me as the author for some reason.

bvanassche:
Sorry, that's my bad. Due to a git rebase the topmost commit ended up with you as the author. I noticed that and tried to reset the author information but apparently resetting the author information failed.

PR #569, opened 2018-03-29
Make switch_ioscheduler() easier to analyze

bvanassche:
The two patches in this pull request do not change the behavior of switch_ioscheduler() but make this function easier to analyze for static analyzers like Coverity.


axboe:
Both look good, pulled, thanks.

PR #570, opened 2018-03-30
Aarch64 crc32c + minor style fixes

sitsofe:
It turns out FreeBSD 12 has added a `sys/auxv.h` header but does not implement `getauxval()` (they have added a slightly incompatible alternative called `elf_aux_info()`) causing issues so the first commit makes the configure test more robust by checking we also have `getauxval()` (but only on Linux) and erroring in other cases. Also move Linux specific runtime check into `os/os.h` and reduce some defines.

(I've a follow on commit in https://github.com/sitsofe/fio/tree/aarch64-crc32c-freebsd that adds runtime CRC32c detection for FreeBSD/arm64 it but FreeBSD haven't gotten around to hooking up the code in their kernel that fills out `AT_HWCAP2` on aarch64 so my commit couldn't be tested in the case where the feature should be detected successfully at runtime...)

The second commit here just adds some #endif comments and makes it clearer what the second posixaio test is doing when you look in `config.log`.

### DIFF
--- a/arch/arch-aarch64.h
+++ b/arch/arch-aarch64.h
@@ -27,8 +27,4 @@ static inline int arch_ffz(unsigned long bitmask)
 
 #define ARCH_HAVE_FFZ
 
-#ifdef ARCH_HAVE_CRC_CRYPTO
-#define ARCH_HAVE_ARM64_CRC_CRYPTO
-#endif
-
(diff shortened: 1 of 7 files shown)


axboe:
Pulled, thanks.

PR #571, opened 2018-04-02
fix bug, when the iodepth is greater then 1, the io_u->engine_data is not be initialized.

simon-rock:
85    -union {
 86=>      unsigned int index;
 87        unsigned int seen;
 88        void *engine_data;
 89    };
engine_data and index is a union.
so when the index > 0,  engine_data != NULL and the engine_data will not be initialized.


sitsofe:
@simon-rock Hi!

Any chance you could change the title to say `glusterfs: ...` (e.g. see https://github.com/axboe/fio/commit/93782cbbd8c8ab46bdc1f18b8a99c740ae6954e5#diff-cacabac5e2939d58047b72953f8dc7bc ) and add your Signed-off-by?

sitsofe:
Actually while that structure is a union can you show where that instance of it is set to > 0?

sitsofe:
OK scratch the last comment I see where it happens over in https://github.com/axboe/fio/blob/d6d74886759e3f268a6a3b12a47872865b867023/backend.c#L1296 .

axboe:
Looks fine, but please put the explanation in the actual commit message. Nobody will ever look at this pull request again, whereas the git log is often used to figure out why a change was made. A good commit message should explain the reason for why a change was made.

simon-rock:
yes. I have changed the commit message.

sitsofe:
@simon-rock The Signed-off-by needs to be among the trailing lines of the commit rather than first (the first line is treated like a title) and just for the record git has a `-s` option. Also any chance you can format the lines in your commit message according to the various guides (e.g. https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html )? You may also want to format your name nicely (e.g. title caps and a surname) in lines that have your email :-)

simon-rock:
yes. thanks .I commit message with "git commit -a -s" last time.

sitsofe:
@simon-rock OK I see where you're going with your commit message. What do you think of this:

```
glusterfs: always allocate io_u->engine_data 

When the iodepth is greater then 1 some io_us in fio_gf_io_u_init() can 
be fooled into thinking their .engine_data has already been set when it
hasn't. This happens because .index and .engine_data are part of the
same union and .index can be set to non-zero values:

static int init_io_u(struct thread_data *td)
...
        for (i = 0; i < max_units; i++) {
            ...
            io_u->index = i;
            ...
            if (td->io_ops->io_u_init) {
                int ret = td->io_ops->io_u_init(td, io_u);

which means io_u->engine_data != 0 in fio_gf_io_u_init().

Fix the issue by removing initial check of io_u->engine_data.

Signed-off-by: Simon Gao <simon29rock@gmail.com>
```

If you like the rewording just respin your commit with it.

simon-rock:
yes. thanks. :-)

PR #573, opened 2018-04-04
Various fixes and code refactoring

bvanassche:
Hello Jens,

As you probably know I'm still working on adding zoned block device support to fio. Although I hope to finish that work soon I already want to send you a pull request with various changes I came up with while working on zoned block device support. Please consider these patches for the official fio repository.

Thanks,

Bart.


axboe:
Looks good to me, thanks Bart.

parse.h
```diff
@@ -117,7 +117,7 @@ static inline void *td_var(void *to, struct fio_option *o, unsigned int offset)
 	else
 		ret = to;
 
-	return (char *) ret + offset;
+	return ret + offset;
```
sitsofe: Void pointer arithmetic is technically illegal in C (https://stackoverflow.com/a/3524270/9109338 ) and is a GCC extension...

parse.c
```diff
@@ -71,10 +71,11 @@ static void show_option_range(const struct fio_option *o,
 			      size_t (*logger)(const char *format, ...))
 {
 	if (o->type == FIO_OPT_FLOAT_LIST) {
-		if (o->minfp == DBL_MIN && o->maxfp == DBL_MAX)
+		if (!o->minfp && !o->maxfp)
 			return;
 
-		logger("%20s: min=%f", "range", o->minfp);
+		if (o->minfp != DBL_MIN)
+			logger("%20s: min=%f", "range", o->minfp);
```
sitsofe: If you were to only set o->maxfp the logger output might look strange

parse.c
```diff
@@ -668,15 +669,17 @@ static int __handle_option(const struct fio_option *o, const char *ptr,
...
-		if (uf > o->maxfp) {
-			log_err("value out of range: %f"
-				" (range max: %f)\n", uf, o->maxfp);
-			return 1;
-		}
-		if (uf < o->minfp) {
-			log_err("value out of range: %f"
-				" (range min: %f)\n", uf, o->minfp);
-			return 1;
+		if (o->minfp || o->maxfp) {
```
sitsofe: So if either of the fp limits is set you check both of them - is this correct or should the guard be rolled into the sub conditions?

parse.c
bvanassche: Can you clarify your comment? Either both minfp and maxfp should be set or none of them so I don't see the issue.

parse.c
bvanassche: Yes, if any of the fp limits has been set both are checked. So if someone only wants to set one limit the other should be set to DBL_MIN or DBL_MAX as appropriate.

parse.c
sitsofe: Imagine I want a range from 0 - DBL_MAX then won't I only see ", max=..." in the logs?

parse.c
sitsofe: OK that makes sense.

parse.h
axboe: Yeah I do know that, but it's supported everywhere. I did ponder dropping this patch, but I'd be very surprised if I don't already do pointer math in a lot of other places in fio.

axboe: Adding -Wpointer-arith gives me 103 warnings :-)

parse.c
bvanassche: With the following patch:

diff --git a/options.c b/options.c
index 17d7245f5490..3cd84bfe0608 100644
--- a/options.c
+++ b/options.c
@@ -3504,7 +3511,7 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
 		.def	= "100",
 		.maxlen	= 1,
 		.minfp	= 0.0,
-		.maxfp	= 100.0,
+		.maxfp	= DBL_MAX,
 		.category = FIO_OPT_C_IO,
 		.group	= FIO_OPT_G_LATPROF,
 	},

fio shows the following output:

./fio --cmdhelp=latency_percentile
  latency_percentile: Percentile of IOs must be below latency_target
                type: list of floating point values separated by ':' (opt=5.9:7.8)
             default: 100
               range: min=0.000000

I think that's the output we all expect?

sitsofe: I had it the wrong way round - o->minfp would have to be DBL_MIN and o->maxfp would have to be something other than DBL_MAX:

patch:
```patch
diff --git a/options.c b/options.c
index 17d7245f..0b7c3076 100644
--- a/options.c
+++ b/options.c
@@ -6,6 +6,7 @@
 #include <assert.h>
 #include <sys/stat.h>
 #include <netinet/in.h>
+#include <float.h>
 
 #include "fio.h"
 #include "verify.h"
@@ -3503,7 +3504,7 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
                .help   = "Percentile of IOs must be below latency_target",
                .def    = "100",
                .maxlen = 1,
-               .minfp  = 0.0,
+               .minfp  = DBL_MIN,
                .maxfp  = 100.0,
                .category = FIO_OPT_C_IO,
                .group  = FIO_OPT_G_LATPROF,
```
example:
```
$ ./fio --cmdhelp=latency_percentile
  latency_percentile: Percentile of IOs must be below latency_target
                type: list of floating point values separated by ':' (opt=5.9:7.8)
             default: 100
, max=100.000000
```

bvanassche: See also https://github.com/axboe/fio/pull/574

sitsofe: Yup that got it.

PR #574, opened 2018-04-06
Fix floating point option range formatting

bvanassche:
Ensure that floating point option ranges are properly formatted if
only one of the two boundaries is specified. A few examples of how
option ranges are formatted with this patch applied:

               range: max=100.000000
               range: min=0.000000, max=100.000000
               range: min=0.000000

Reported-by: Sitsofe Wheeler <sitsofe@gmail.com>


axboe:
Pulled, thanks.

PR #583, opened 2018-04-18
Several code refactoring patches

bvanassche:
Hello Jens,

This series consists of five patches that rework the fio code slightly and that do not change the behavior of fio. I came up with these patches while working on adding FMR support to fio.

Thanks,

Bart.


backend.c
```diff
@@ -283,7 +283,7 @@ static bool fio_io_sync(struct thread_data *td, struct fio_file *f)
 
 requeue:
 	ret = td_io_queue(td, io_u);
-	if (ret < 0) {
+	if ((int)ret < 0) {
```
axboe: How can this ever be true, it should only return fio_q_status at this point?

bvanassche: I was wondering whether there would be any out-of-tree I/O engines for which that code should be kept? If not, I think the code under "if (ret < 0)" can be removed.

axboe: We version the API anyway, so they would not load. You should bump that version number.

axboe:
Pulled, thanks Bart. Note that I did a fixup patch to fix the split lines in the engine queue routines.

PR #585, opened 2018-04-20
Add zoned block device support

bvanassche:
This patch series adds zoned block device support to fio. Zoned block devices consist of multiple zones. Writes to sequential zones have to be serialized for host managed devices. Reads should occur below the write pointer. The function zbd_adjust_block() ensures that this is the case.

All zoned block device specific code exists in a new source file, zbd.c. The number of calls from the fio core to ZBD-code is minimal.

Examples of what has been tested and which new command line options have been added are available in t/zbc/test-zbd-support. All the tests in this shell script pass against the devices I used in my tests.

### DIFF
--- a/parse.c
+++ b/parse.c
@@ -120,19 +120,22 @@ static void show_option_values(const struct fio_option *o)
 static void show_option_help(const struct fio_option *o, int is_err)
 {
 	const char *typehelp[] = {
-		"invalid",
-		"string (opt=bla)",
-		"string (opt=bla)",
-		"string with possible k/m/g postfix (opt=4k)",
-		"string with time postfix (opt=10s)",
-		"string (opt=bla)",
-		"string with dual range (opt=1k-4k,4k-8k)",
-		"integer value (opt=100)",
-		"boolean value (opt=1)",
-		"list of floating point values separated by ':' (opt=5.9:7.8)",
-		"no argument (opt)",
-		"deprecated",
-		"unsupported",
+		[FIO_OPT_INVALID]	  = "invalid",
+		[FIO_OPT_STR]		  = "string (opt=bla)",
+		[FIO_OPT_STR_ULL]	  = "string (opt=bla)",
+		[FIO_OPT_STR_MULTI]	  = "string with possible k/m/g postfix (opt=4k)",
+		[FIO_OPT_STR_VAL]	  = "string (opt=bla)",
+		[FIO_OPT_STR_VAL_TIME]	  = "string with time postfix (opt=10s)",
+		[FIO_OPT_STR_STORE]	  = "string (opt=bla)",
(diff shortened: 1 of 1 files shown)


stat.c
```diff
@@ -2265,45 +2265,8 @@ static inline void reset_io_stat(struct io_stat *ios)
...
-	for (i = 0; i < FIO_IO_U_LAT_N_NR; i++)
-		ts->io_u_lat_n[i] = 0;
-	for (i = 0; i < FIO_IO_U_LAT_U_NR; i++)
-		ts->io_u_lat_u[i] = 0;
-	for (i = 0; i < FIO_IO_U_LAT_M_NR; i++)
-		ts->io_u_lat_m[i] = 0;
-
-	ts->total_submit = 0;
-	ts->total_complete = 0;
+	memset(ts, 0, sizeof(*ts));
```
axboe: It's currently not clearing anything near all of it, however. Might be fine, I'd have to double check, but the commit makes it seems like we're just replacing an all-field-clear with a memset, which would just be a cleanup.

bvanassche: I'm not sure whether I need this patch. I will see whether it can be removed.

bvanassche: That patch has been removed.

configure
```diff
@@ -2429,6 +2447,9 @@ fi
 if test "$valgrind_dev" = "yes"; then
   output_sym "CONFIG_VALGRIND_DEV"
 fi
+if test "$linux_blkzoned_h" = "yes" ; then
+  output_sym "HAVE_LINUX_BLKZONED_H"
```
axboe: Can you make this CONFIG_LINUX_BLKZONED? I noticed that slipped through in the asprintf/vasprintf patches, and it doesn't match what any of the other configure tests name things.

bvanassche: OK, done.

filesetup.c
```diff
@@ -1866,6 +1874,9 @@ void fio_file_reset(struct thread_data *td, struct fio_file *f)
 		axmap_reset(f->io_axmap);
 	else if (fio_file_lfsr(f))
 		lfsr_reset(&f->lfsr, td->rand_seeds[FIO_RAND_BLOCK_OFF]);
+
```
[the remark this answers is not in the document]
bvanassche: That extra blank line has been removed.

zbc.c
```diff
@@ -0,0 +1,773 @@
...
+
+	if (z->verify_block * ba >= f->zbd_info->zone_size)
+		log_err("%s: %d * %d >= %ld\n", f->file_name, z->verify_block,
+			ba, f->zbd_info->zone_size);
+	io_u->offset = (z->start << 9) + z->verify_block++ * ba;
+}
+
+static int zbc_drain_queue(struct thread_data *td)
+{
+#if 0
```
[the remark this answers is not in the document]
bvanassche: Thanks, I replaced that code with a call to io_u_quiesce().

PR #589, opened 2018-05-14
This partly revert 97bb54c9606c(add __load_ioengine() to separate ioe…

majianpeng:
…ngine loading from td context)

In fact, if __load_ioengine failed, in load_ioengine it will call
dlopen_ioengine to open lib in LD_LIBRARY_PATH.

Signed-off-by: Jianpeng Ma <jianpeng.ma@intel.com>

### DIFF
--- a/ioengines.c
+++ b/ioengines.c
@@ -574,6 +574,7 @@ int td_io_get_file_size(struct thread_data *td, struct fio_file *f)
 int fio_show_ioengine_help(const char *engine)
 {
 	struct flist_head *entry;
+	struct thread_data td;
 	struct ioengine_ops *io_ops;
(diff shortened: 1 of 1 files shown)


majianpeng:
In ceph. https://github.com/ceph/ceph/tree/master/src/test/fio is a ioengine of fio.

axboe:
At Sitsofe alludes to, a good commit message should explain why you are making the change. Your commit message is also leaving me confused, I don't understand what you are trying to fix. Can you expand?

majianpeng:
update comments by your suggestions. please review again.

axboe:
That's better, thanks.

PR #590, opened 2018-05-16
Add sgunmap ioengine to submit multiple discard ranges

vincentkfu:

These patches create a new sgunmap ioengine that submits multiple discard ranges with a single system call using the SCSI generic sg interface.

This has enabled us to improve trim IOPS from 6K to 39K on a SAS device.

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1749,6 +1749,18 @@ I/O engine
 			character devices.
 			The sg engine includes engine specific options.
 
+		**sgunmap**
+			Send UNMAP commands via SCSI generic sg v3 I/O.  This ioengine
+			provides a way to send multiple trim operations with a single
+			system call using the SG_IO ioctl.  Use :option:`iodepth_batch_submit' to
+			control how many operations are submitted each time.  This ioengine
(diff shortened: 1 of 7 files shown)


axboe:
Seems wrong to add an engine for this, trim is really just an IO type. Why not just update the sg engine to both support unmaps, and add the queue depth for it too?

axboe:
Adding a new IO engine to support a read/write/trim doesn't make any sense. "Easier to do" is really not a valid reason, since it's "harder to use and maintain" which are much more important criteria. Just implement for sg, and ensure that the char device is used if you want queued operation. The sg engine supports async/queued IO just fine, as long as you use the char device. All you need to do is add UNMAP support, which would be handy to have, and then your regular read/write/trim workloads should work with that already.

vincentkfu:
Ok. I will work on that and create a new pull request when it's ready.

axboe:
Thanks Vincent.

sitsofe:
(@vincentkfu just a thought: when you implement it could you consider adding some means of switching between trim via UNMAP and via WRITE SAME?)

PR #592, opened 2018-05-18
Add support for WRITE AND VERIFY and WRITE SAME to the sg ioengine

vincentkfu:
Jens, please consider this pull request that adds to the sg ioengine support for the WRITE AND VERIFY and WRITE SAME commands. We have collected bus traces and the commands are working as expected.


axboe:
Looks good, pulled, thanks.

PR #593, opened 2018-05-20
rados: ddd reuse support to rados engine

FacelessManipulator:
This patch adds reuse support to Rados engine. There are two main reasons to add the new options "reuse" and "cleanup".

- Since creating and removing objects linearly cost too much time, especially with a lot of small files, the different jobs could reuse the objects with the option "reuse=[same id]", and purge those objects in the last job to save time.
- In Ceph Mimic with bluestore, when fio want to read an unallocated data, the read ops would return immediately in OSD without DISK IOs, which cause unreal high read speed. With options "reuse", users could write full data to objects first and then perform the reading bench.

Any idea?

Signed-off-by: Hanjie Zhou <miller_maxwell@outlook.com>

### DIFF
--- a/engines/rados.c
+++ b/engines/rados.c
@@ -34,6 +34,8 @@ struct rados_options {
 	char *pool_name;
 	char *client_name;
 	int busy_poll;
+	int cleanup;
+	int reuse;
 };
 
 static struct fio_option options[] = {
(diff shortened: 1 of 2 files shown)


sitsofe:
@FacelessManipulator would you like to resubmit?

PR #599, opened 2018-06-05
Rename appveyor.yml into .appveyor.yml

bvanassche:
Make the appveyor.yml file less visible because it is not a fio source code file.


axboe:
I'm assuming the CI still knows how to find it?

axboe:
I guess so, just checked and it says:

```
You can also put project configuration in .appveyor.yml dot-file in the root of your repo. AppVeyor searches repo for appveyor.yml first and then for .appveyor.yml.
```

PR #600, opened 2018-06-05
Use the monotonic clock for condition variables if available

bvanassche:


axboe:
Should this one just be discarded?

PR #601, opened 2018-06-05
Suppress gcc 8 compiler warnings

bvanassche:
Introduce a new function nowarn_snprintf() that behaves identically
to snprintf() except that it does not make gcc complain about potential
truncation of the output buffer.


axboe:
Pulled, thanks Bart!

PR #602, opened 2018-06-05
Make nowarn_snprintf() call va_end()

bvanassche:
This patch suppresses the following Coverity complaint:

CID 175867:  API usage errors  (VARARGS)
    va_end was not called for "args".


axboe:
Hah, was just looking at that and missed that it does the return directly. Will pull, thanks.

PR #609, opened 2018-06-11
rate_process=skew implementation

shimrot:
Requesting this change to add a "skew" option to the "rate_process" value.  When "numjobs" is greater than 1 with a rate, the operation arrival times from each subjob actually occur synchronized.  That is, each job sends the op at effectively the same time.   What I'd like, is to do a "linear" process, but have each subjob delay the operation slightly - spread the arrival times across the rate time (time it takes to perform one operation).  

The point is to have multiple streams of equal activity (numjobs), but to not have them issue ops at the same time.  Poisson process works, but the arrival variation is too extreme, I really wanted more of a linear rate.  Using "start_delay" introduced other side effects, which I'll address later.  My intent is also to have some randomization in the number of blocks between thinktime, and in thinktime itself, but again - that's a different consideration.

The "skew" implementation adds to the thread_options a "rate_skew" value for each subjob.  However, in order to calculate, I needed each subjob to also know what was the original "numjobs" value.  However,  the numjobs value is set to 1 for all subjobs.  There seem to be a number of dependencies on that value, so I created a "related_numjobs" value as well, which then could be used by calculation code.

I was not able to discover another place to get the original "numjobs" option value.  So if there is one, please let me know and I will change.

I really didn't intend my "push" to get recorded in the base tree yet - I thought it would occur only to my github clone so I could start a discussion.  However, it's done and I'll avoid that in the future.  I have 

Thanks much, 
Kris Davis

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -2252,13 +2252,21 @@ I/O rate
 
 .. option:: rate_process=str
 
-	This option controls how fio manages rated I/O submissions. The default is
-	`linear`, which submits I/O in a linear fashion with fixed delays between
-	I/Os that gets adjusted based on I/O completion rates. If this is set to
-	`poisson`, fio will submit I/O based on a more real world random request
-	flow, known as the Poisson process
-	(https://en.wikipedia.org/wiki/Poisson_point_process). The lambda will be
-	10^6 / IOPS for the given workload.
+	This option controls how fio manages rated I/O submissions:
+	
+	**linear** (default)
+		fio submits I/Os in a linear fashion with fixed delays between
+		I/Os that are adjusted based on I/O completion rates.
+	**poisson**
+		fio will submit I/O based on a more real world random request
+		flow, known as the Poisson process
+		(https://en.wikipedia.org/wiki/Poisson_point_process). 
+		The lambda will be 10^6 / IOPS for the given workload.
+	**skew**
+		when numjobs > 1, fio submits I/Os in a linear fashion, but
+		with I/Os from each subjob spread evenly within the calculated seconds
+		per I/O (skewed between each job). Thus, the operations will not be 
+		mostly synchronized between jobs (as occurs with the `linear` process).
 
 .. option:: rate_ignore_thinktime=bool
 
--- a/backend.c
+++ b/backend.c
@@ -845,7 +845,22 @@ static long long usec_for_io(struct thread_data *td, enum fio_ddir ddir)
 		uint64_t secs = bytes / bps;
 		uint64_t remainder = bytes % bps;
 
-		return remainder * 1000000 / bps + secs * 1000000;
+		uint64_t usskew = 0;
+		if (td->o.rate_process == RATE_PROCESS_SKEW) {
+			/*
+			 * set skew only when in ramp, or if no ramp applies, because ramp expiration is only
+			 * recognized when a new io is submitted, and counters are reset.
+			 * When ios are already skewed between jobs, the ramp expiration is also skewed.
+			 * Continued skew additions causes subsequent operations to effectively be in sync again.
+			 */
+			if (td->o.ramp_time == 0 || in_ramp_time(td)) {
+				usskew = td->rate_skew[ddir];
+			}
+			dprint(FD_RATE, "skew rate thd=%d, ddir=%d, subjob=%d, usskew=%llu\n",
+					td->thread_number, ddir, td->subjob_number,	(unsigned long long)usskew);
+		}
+
+		return remainder * 1000000 / bps + secs * 1000000 + usskew;
 	}
 
 	return 0;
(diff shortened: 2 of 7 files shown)


axboe:
Did you analyze why they appear to be synced? I think it would be better to figure that out. Especially if it's not fixable with start_delay, which would be the logical way to attempt to fix this imho.

shimrot:
Thanks Jens,

The sync occurs just because each job created from "--numjobs" ends up
issues the ops at the same time interval.  So, for example, if
--rate_iops=6, and numjobs=100 with iodepth=1, all the jobs send the
initial ops at mostly the same time, and it can stay in sync for quite a
long time.  Until something causes an operation response to be delayed,
they will stay in sync.

I tried using start_delay, but that had other issues:

   - Doesn't currently support a separate delay for each subjob (like some
   other options variabilities)
   - I found if I use a start_delay with a range, then the 1st job gets one
   delay, and then all the other jobs get another (seems must be a bug that I
   was going to look into, haven't got there yet, and the other start_delay
   issues below kind of excluded using start_delay anyway.
   - The log file timestamps appear to be based on the start of the job,
   not the start of the fio execution.  Thus, it's hard to combine results
   from multiple log files because the timestamps are skewed.  I did  actually
   added some preprocessing to my parser to adjust, but even if the range
   issue was fixed, I don't know of a way to know what the random delay was
   determined by the range in order to adjust the log timestamps. (I did see
   log epoch option which should obviate the issue, but have't tried working
   with it yet)
   - When jobs are delayed, they still run for the "runtime", so they start
   at different times, and end at different times.   Thus, I got the
   impression the "start_delay" is for something with a more gross difference
   between jobs, as the runtime doesn't end up synced, especially when using
   numjobs.

I choose to add to the rate_process because:

   - The delay caused by the "skew" is pretty small and based on the rate,
   it seemed related to the rate and just a new form of the "process" value.
   - It's really a slight modification of the "linear" process, and only
   varies things for subjobs created by "numjobs". I was thinking it might be
   nice to have some randomization that is not as extreme as poisson.  But,
   didn't really know how to define what would make sense. Might be something
   future - and again would seem "rate_process" related rather than
   start_delay.
   - It really just "starts" out the jobs skewed.  Nothing really prevents
   a device from delaying operations such that they get in sync again, however
   I think that would indicate something undesirable for a device anyway
   (that's not really any different than "start_delay" though).

To use "start_delay" to accomplish this, I think it would require:

   - a fix to the delay assigned to subjobs.
   - A linear delay option which would be spread across the number of jobs
   in some way.
   - An option to adjust the runtimes according to the actual start delay
   for the job (might be nice anyway)
   - probably an option to adjust log timestamps by the actual start delay
   (though might just require folks use the epoch time if want to sync
   separate logs).
      - might have an json value which indicates what the actual delay was,
      so tools could adjust

All that seemed fairly complicated, and a 'skew' process pretty simple.

Thanks
Kris

axboe:
OK, so sounds like we have a couple of issues here. I just fixed the startdelay option, see that fix here:

http://git.kernel.dk/cgit/fio/commit/?id=49050b56681b17968256702a7a3ec0f545c7dad8

with this, each thread will have a correct startdelay that's randomized within the range given.

I'd be happy to take fixes to support making the runtime the same regardless of startdelay, and for the logging as well. Ditto for the json output of startdelay, but that one is a bit more tricky since the intent is for that to be parseable and generate the same run again.

shimrot:
I don't really object to a "start_delay" based operation, but let's be clear...  You don't really want the rate_process based solution - but rather one based on "start_delay"  correct?

I haven't looked into what changes would be required for a start_delay based solution, but in order to achieve equivalent operation, I anticipate the need for the following new options (just guessing at names): 
- start_delay_process:  when a range is used, this indicates how to spread the delay across jobs:
-- random: (default), works as today
-- linear: evenly spreads job start times within indicated range (I really need this rather than random.  My test is latency sensitive and for repeatability I need something consistent, especially since the rates for each job are so low - and I'm actually ramping up the number of jobs.  I saw some issues like this when using poisson process - though that was before I figured out the ramp_time issue, so not sure how much related). Running for a longer time seemed to smooth it out (which seems to hint that the ramp_time was the issue).

- start_delay_sync_runtime:  adjust runtime based on the start_delay so each job ends at the same time.
- start_delay_sync_logtimes: adjust log timestamps to the same starttime regardless of start_time  (probably could be done at print time rather than continuous)

Or perhaps a single sync option is better: 
- start_delay_sync_times: adjust both runtime and log timestamps

What do you think?
Thanks again
Kris

axboe:
Right, I think what you describe is pretty close to what I was thinking. Now that startdelay actually works for a random range, adding a start_delay_process or similar to set the behavior makes sense. Ditto the start_delay_sync_times, that makes sense.

shimrot:
Ok - I'll work on the starttime based solution instead.  This discussion can probably be closed and the branch tossed.  Not sure what that involves.  
Thanks
Kris

PR #613, opened 2018-06-15
Fix read_ini_data() error path

bvanassche:


axboe:
I beat you to it, but I forgot to close() the dup'ed descriptor if fdopen fails though.

axboe:
You can't use 'errno' as close() will have reset it to 0, though. I'm applying a fixup.

PR #614, opened 2018-06-18
trim and readonly option

vincentkfu:
Jens, please consider these patches related to trim workloads and the readonly option. fio currently allows destructive trim workloads when the readonly option is enabled. These patches put write and trim workloads on equal footing with respect to the readonly option.

### DIFF
--- a/init.c
+++ b/init.c
@@ -594,6 +594,12 @@ static int fixup_options(struct thread_data *td)
 	struct thread_options *o = &td->o;
 	int ret = 0;
 
+	if (read_only && (td_write(td) || td_trim(td))) {
+		log_err("fio: trim and write operations are not allowed"
+			 " with the --readonly parameter.\n");
+		ret |= 1;
+	}
+
 #ifndef CONFIG_PSHARED
 	if (!o->use_thread) {
 		log_info("fio: this platform does not support process shared"
--- a/init.c
+++ b/init.c
@@ -600,7 +600,7 @@ static int fixup_options(struct thread_data *td)
 			 " mutexes, forcing use of threads. Use the 'thread'"
 			 " option to get rid of this warning.\n");
 		o->use_thread = 1;
-		ret = warnings_fatal;
+		ret |= warnings_fatal;
 	}
 #endif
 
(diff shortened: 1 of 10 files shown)


init.c
```diff
@@ -594,6 +594,12 @@ static int fixup_options(struct thread_data *td)
 	struct thread_options *o = &td->o;
 	int ret = 0;
 
+	if (read_only && (td_write(td) || td_trim(td))) {
+		log_err("fio: trim and write operations are not allowed"
+			 " with the --readonly parameter.\n");
+		ret += 1;
```
axboe: This is a bit odd, why isn't it using warnings_fatal for this? Also, += 1 reads weird, most people would do ++ here. In any case, this will be overwritten by other warnings. So just use warnings_fatal, and we expect this to be caught fatally later on any way.

init.c
```diff
@@ -662,7 +662,7 @@ static int fixup_options(struct thread_data *td)
 	    !o->norandommap) {
 		log_err("fio: Any use of blockalign= turns off randommap\n");
 		o->norandommap = 1;
-		ret = warnings_fatal;
+		ret += warnings_fatal;
```
axboe: This is better, just fold this with the previous patch and make your new user do the same.

axboe:
Looks good, just added a few comments.

init.c
vincentkfu: I'll make the change to ++.

However, fio --readonly --rw=write is caught as a fatal condition during option parsing via 
rw_verify() in options.c.  So, for the sake of consistency, it seems appropriate to make conditions this patch catches (fio --rw=write --readonly) fatal as well.

This problem will indeed be caught later on but the error message says to try running with --debug=io which isn't helpful.

axboe: Sure, but you need to fix the overwrite first then, since others can just stomp on it. So if you want to do the two separate patches, you need to flip the order of them. And then just have them all do:

```ret |= warnings_fatal```

instead.

vincentkfu: Jens, I updated the fixup_options() changes:

* everywhere that used to be
``ret = warnings_fatal``
is now
``ret |= warnings_fatal``
* everywhere that used to be
``ret = 1``
is now
``ret |= 1``
* if it detects --readonly and --rw=trim/write set
``ret |= 1``
* stomping on ret is fixed first and then --readonly change is added

axboe:
Perfect, looks good! Pull, thanks.

PR #615, opened 2018-06-20
capable to test glusterfs with one single glfs instance

zhanghuan:
Current multi-thread test creates one glfs instance per job. However,
there is requirement to run all jobs on one single instance, and that
all jobs would share the same set of underlying working threads.

Add a new option "single-instance" to control whether to create one
global instance or one instance per job. For testing with mutilple
gluster volumes, use host name and volume name to filter out same
volume, one specific volume will have one instance.


axboe:
Looks fine to me, pulled, thanks.

PR #617, opened 2018-06-21
doc: add text about possibly verify errors with norandommap

larrystevenwise:
In both the norandommap option section, and the verify option section,
add text warning that norandommap can cause data verification errors
when using async io engines and io depths > 1.

Signed-off-by: Steve Wise <swise@opengridcomputing.com>


axboe:
Perfect, thanks Steve!

PR #626, opened 2018-07-23
Add trim support to the sg ioengine

vincentkfu:
This is a follow-up to #590 

These patches add trim support to the sg ioengine via the SCSI unmap command. Following the established sg pattern, unmap commands for block devices are carried out synchronously. For character devices, however, it is possible to submit multiple ranges with a single unmap command by setting iodepth_batch.

@sitsofe, I considered adding an option to use WRITE SAME but using unmap was my first priority since that was what was requested by the group that this feature was developed for. Note that it would be pretty easy to add an option so that write workloads with sg_write_mode=same carry out trims.

### DIFF
--- a/engines/sg.c
+++ b/engines/sg.c
@@ -3,6 +3,51 @@
  *
  * IO engine that uses the Linux SG v3 interface to talk to SCSI devices
  *
+ * This ioengine can operate in two modes:
+ *	sync	with block devices (/dev/sdX) or
+ *		with character devices (/dev/sgY) with direct=1 or sync=1
+ *	async	with character devices with direct=0 and sync=0
+ *
+ * What value does queue() return for the different cases?
+ *				queue() return value
+ * In sync mode:
+ *  /dev/sdX		RWT	FIO_Q_COMPLETED
+ *  /dev/sgY		RWT	FIO_Q_COMPLETED
+ *   with direct=1 or sync=1
+ *
+ * In async mode:
+ *  /dev/sgY		RWT	FIO_Q_QUEUED
+ *   direct=0 and sync=0
+ *
+ * Because FIO_SYNCIO is set for this ioengine td_io_queue() will fill in
+ * issue_time *before* each IO is sent to queue()
+ *
+ * Where are the IO counting functions called for the different cases?
+ *
+ * In sync mode:
+ *  /dev/sdX (commit==NULL)
+ *   RWT
+ *    io_u_mark_depth()			called in td_io_queue()
+ *    io_u_mark_submit/complete()	called in td_io_queue()
+ *    issue_time			set in td_io_queue()
+ *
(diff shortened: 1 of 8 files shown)


engines/sg.c
```diff
@@ -3,6 +3,39 @@
...
+ *  /dev/sgY with direct=1 or sync=1:
+ *  commit does nothing	RWT	called in td_io_queue()		called in queue()		set in td_io_queue()
+ *
+ *
+ * In async mode:		read and write operations are submitted in queue()
+ *			RW	called in queue()		submit called in queue()	set in td_io_queue()
+ *
+ *				trim operations are queued in queue() and submitted in commit()
+ *			T	called in commit()		submit called in commit()	set in commit()
+ *
```
axboe: I like the detailed explanations, but please keep it within 80 chars. It's unreadable in my term.

engines/sg.c
```diff
@@ -21,6 +54,8 @@ enum {
 	FIO_SG_WRITE_SAME	= 3
 };
 
+static struct ioengine_ops ioengine;
```
axboe: Nope... See below.

engines/sg.c
```diff
@@ -383,65 +501,120 @@ static int fio_sgio_prep(struct thread_data *td, struct io_u *io_u)
...
+	}
+
 	ret = fio_sgio_doio(td, io_u, do_sync);
 
 	if (ret < 0)
 		io_u->error = errno;
 	else if (hdr->status) {
 		io_u->resid = hdr->resid;
 		io_u->error = EIO;
+	} else if (ioengine.commit != NULL) {
```
axboe: This is really backwards. td->io_ops->commit if you must check.

ioengines.c
```diff
@@ -350,7 +350,9 @@ enum fio_q_status td_io_queue(struct thread_data *td, struct io_u *io_u)
 			 "invalid block size. Try setting direct=0.\n");
 	}
 
-	if (!td->io_ops->commit || io_u->ddir == DDIR_TRIM) {
+	if (!td->io_ops->commit ||
+	    (io_u->ddir == DDIR_TRIM &&
+             !td_ioengine_flagged(td, FIO_ASYNCTRIM))) {
```
axboe: Instead of adding FIO_ASYNCTRIM, we should stop special casing trim here.

vincentkfu:
@axboe, I have updated the branch based on your comments. Here are the main changes:

- reformatted the io accounting comments
- now use td->io_ops->commit to check if commit is null
- eliminated FIO_ASYNCTRIM
- added patch with HOWTO and manpage changes
- changed sgunmap-test.py to be more informative when an assertion fails
- add more detail to the sg error string

axboe:
Pulled, thanks Vincent!

PR #637, opened 2018-08-03
platforms/windows: Add S_ISSOCK macro.

aclamk:


axboe:
I'm going to pull this, but in the future, please do provide a proper changelog. The commit message should explain why you are making this change. In this case, it should have explained how it's a prep patch for the iolog change.

PR #644, opened 2018-08-09
Realloc io_u buffers to adapt to operation size.

aclamk:
When iolog is read in chunked mode, some prefetching is done. Based on this size of buffers are determined.
Further operations may need larger buffers. If so, io_u buffers are reallocated.
Example that cause problem when read_iolog_chunked=1:
fio version 2 iolog
rbd_data.0 add
rbd_data.0 open
rbd_data.0 write 0 4096 (x20)
rbd_data.0 write 0 4096000 (x10)
rbd_data.0 close

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -1201,19 +1201,10 @@ static void cleanup_io_u(struct thread_data *td)
 static int init_io_u(struct thread_data *td)
 {
 	struct io_u *io_u;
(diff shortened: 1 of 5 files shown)


aclamk:
First commit forms #643.

axboe:
Looks good, pulled, thanks.

PR #646, opened 2018-08-12
arch: fix build breakage on armv6

sitsofe:
The FreeBSD folks found fio didn't build on their armv6 systems
(https://svnweb.freebsd.org/ports/head/benchmarks/fio/Makefile?revision=432897&view=markup
) but Mikaël Urankar fixed the issue in
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=228042 so copy the fix
to mainline.


axboe:
Pulled, thanks.

PR #648, opened 2018-08-13
engines/http: Add support for WebDAV and S3

l-mb:
The http engine allows fio to read/write objects to WebDAV and S3
storage systems via libcurl.

Writes are mapped to PUT, reads to GET, and TRIM to DELETE of individual
objects within the bucket/path.


sitsofe:
Minor points:
- Did you mean to change the indentation in the HOWTO file?
- Unfortunately you'll need to duplicate the information into the man page (fio.1)

engines/http.c
```diff
@@ -0,0 +1,566 @@
...
+	unsigned int hmac_len;
+
+	ctx = HMAC_CTX_new();
+	HMAC_Init_ex(ctx, key, key_len, EVP_sha256(), NULL);
+	HMAC_Update(ctx, (unsigned char*)data, strlen(data));
+	HMAC_Final(ctx, md, &hmac_len);
+	HMAC_CTX_free(ctx);
+}
+
+#if 0
```
sitsofe: Are you sure you want to ship disabled code?

engines/http.c
```diff
@@ -0,0 +1,566 @@
...
+	strftime (date_short, sizeof(date_short), "%Y%m%d", gtm);
+	strftime (date_iso, sizeof(date_iso), "%Y%m%dT%H%M%SZ", gtm);
+	uri_encoded = _aws_uriencode(uri);
+
+	if (op == DDIR_WRITE) {
+		dsha = _gen_hex_sha256(buf, len);
+		sprintf(method, "PUT");
+	} else {
+		/* DDIR_READ && DDIR_TRIM supply an empty body */
+		if (op == DDIR_READ) {
```
sitsofe: We generally skip single line braces...

engines/http.c
```diff
@@ -0,0 +1,566 @@
...
+	http = calloc(1, sizeof(*http));
+	if (!http) {
+		log_err("calloc failed.\n");
+		goto cleanup;
+	}
+
+	http->curl = curl_easy_init();
+	if (o->curl_verbose) {
+		curl_easy_setopt(http->curl, CURLOPT_VERBOSE, 1L);
+		// curl_easy_setopt(http->curl, CURLOPT_DEBUGFUNCTION, &_curl_trace);
```
sitsofe: Disabled code?

engines/http.c
l-mb: Good call out. I should make that a second tier of the verbosity. (Tracing S3 Authentication errors really isn't fun without massive amounts of logging.) Thanks, reworked.

engines/http.c
l-mb: Left over from when I had more lines here; no braces also make me personally twitchy, but I've changed the code accordingly. Thanks!

engines/http.c
l-mb: Side effect of the first; now configurable via an option.

l-mb:
@sitsofe On the fio.1 - isn't that generated from the howto? Is copying doc/output/man/fio.1  to fio.1 an acceptable change?

On the whitespace - @axboe had previously pointed out some whitespace issues (spaces instead of tabs) in my changes, so I just fixed the files I had touched. I can strip the other sections, but this seemed to be consistent.

sitsofe:
@l-mb it could be but but there were fidelity problems so it's not for now and is manually maintained. One more thing - if you could squash these down that would be great...

axboe:
The space issues you fixed were from the output examples, which are copied from the actual output. Those use spaces, not tabs. I was mostly referring to code and header files.

As for the man page, for now the HOWTO and man page are still separately maintained. Yes this sucks, and yes lots of work has already gone into synchronizing them and generating the man page. But it isn't quite there yet.

axboe:
Another thing that's missing is an update to options.c, the ioengine entry needs a:

```
#ifdef CONFIG_HTTP
                          { .ival = "http",                                    
                            .help = "HTTP WebDAV and S3",                         
                          },
#endif
```

or something like that.

l-mb:
All folded in and fio.1 updated. Sorry, the comment in README suggested that these were auto-generated and I thought fio.1 was an artifact.

axboe:
LGTM, pulled! Thanks Lars.

l-mb:
Thanks, excellent! My first actual code contribution in *years!* ;-)

PR #652, opened 2018-08-16
iolog replay: Fixing problem when using read_iolog= and directory= together.

aclamk:
add_file adds file with 'directory' prefix.
    get_fileno searches for file without 'directory'

    Problem occurs when processing any iolog file, eg:
    fio version 2 iolog
    rbd_data.0 add
    rbd_data.0 open
    rbd_data.0 write 0 4096
    rbd_data.0 close


filesetup.c
```diff
@@ -1839,9 +1839,16 @@ void dup_files(struct thread_data *td, struct thread_data *org)
  */
 int get_fileno(struct thread_data *td, const char *fname)
 {
+	char file_name[PATH_MAX];
 	struct fio_file *f;
 	unsigned int i;
 
+	if (td->o.directory) {
+		int len = set_name_idx(file_name, PATH_MAX, td->o.directory, 0, false);
+		sprintf(file_name + len, "%s", fname);
```
sitsofe: Any chance you can use snprintf? There are static analysis tools that start getting angsty when they can't prove you won't overflow the buffer...

filesetup.c
```diff
@@ -1551,15 +1551,17 @@ int add_file(struct thread_data *td, const char *fname, int numjob, int inc)
...
 	dprint(FD_FILE, "add file %s\n", fname);
 
 	if (td->o.directory)
 		len = set_name_idx(file_name, PATH_MAX, td->o.directory, numjob,
 					td->o.unique_filename);
 
-	sprintf(file_name + len, "%s", fname);
+	assert (len < PATH_MAX);
+	chars = snprintf(file_name + len, PATH_MAX - len, "%s", fname);
+	assert(chars < PATH_MAX);
```
sitsofe: Is it `chars <= PATH_MAX`? Technically if you copy everything and the NULL it could be equal to PATH_MAX without overflowing the buffer?

Further since you're in control of the parameters can't you get away with doing one check for length before snprintf and know that snprintf can't copy any more characters than you specified?

sitsofe: (also don't could you squash these latter commits once you're happy with them?)

aclamk: I made a check, and it seems to work according to man snprintf:
snprintf(buffer, 4, "%s", "0123") == 4
buffer == "012\0"

Since snprintf will never overflow buffer and assert is always fatal, I think checking result after is better place.

bvanassche: Please consider to use asprintf() instead of snprintf(). See also https://github.com/axboe/fio/blob/master/oslib/asprintf.c.

aclamk: @bvanassche 
I do like an idea of replacing snprintf with asprintf.
But this would require to change logic of `set_name_idx`, `get_max_name_idx` and `get_next_name`.

aclamk: @bvanassche I tried to switch snprintf -> asprintf and I failed.
Problem is that `set_name_idx` fills some part of file_name, but amount of it is unknown.
`set_name_idx` extracts directory prefix related to specified subjob, but this is encoded in a way that cannot be handled by printf format.

PR #653, opened 2018-08-16
engines/http: Add support for Swift storage backend

l-mb:
This is a simple extension to the engine and adds basic support for OpenStack Swift.

It does change a previous option (http_s3 is subsumed in http_mode), but since this was not released yet, I think this is the cleaner approach.


engines/http.c
```diff
@@ -122,12 +137,12 @@ static struct fio_option options[] = {
 		.group    = FIO_OPT_G_HTTP,
 	},
 	{
-		.name     = "http_s3",
-		.lname    = "S3 extensions",
-		.type     = FIO_OPT_BOOL,
-		.help     = "Whether to enable S3 specific headers",
-		.off1     = offsetof(struct http_options, s3),
-		.def	  = "0",
+		.name     = "http_mode",
```
axboe: This really wants to be a posval option type, that's much cleaner than just accepting a string and then checking if it matches one of the valid types. With posval, that comes automatically and no extra checking is needed. Might want to consider for https as as well, it could be "off", "on", "nossl" for instance.

l-mb:
posval, how did I miss that! Indeed, that's much nicer. Tidied up.

axboe:
Much better! Pulled.

l-mb:
Thanks!

PR #654, opened 2018-08-16
engines: Add support for DDN's Infinite Memory Engine

gbossu:
This PR aims at bringing support for DDN’s Infinite Memory Engine (IME) in FIO.

To that end, 3 engines were created to test the different functions provided by IME. These are called ime_psync, ime_psyncv, and ime_aio. You can find these engines in engines/ime.c.

Could you have a look and tell us if this patch can be merged ?
If you want to compile and run some tests, we can provide a wrapper library to mock IME function calls using POSIX calls.

Signed-off-by: Gaëtan Bossu <gbossu@ddn.com>

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1882,6 +1882,22 @@ I/O engine
 			mounted with DAX on a persistent memory device through the PMDK
 			libpmem library.
 
+		**ime_psync**
+			Synchronous read and write using DDN's Infinite Memory Engine (IME).
+			This engine is very basic and issues calls to IME whenever an IO is
+			queued.
+
+		**ime_psyncv**
+			Synchronous read and write using DDN's Infinite Memory Engine (IME).
+			This engine uses iovecs and will try to stack as much IOs as possible
+			(if the IOs are "contiguous" and the IO depth is not exceeded)
+			before issuing a call to IME.
+
+		**ime_aio**
+			Asynchronous read and write using DDN's Infinite Memory Engine (IME).
+			This engine will try to stack as much IOs as possible by creating
+			requests for IME. FIO will then decide when to commit these requests.
+
 I/O engine specific parameters
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,9 @@ endif
 ifdef CONFIG_LIBPMEM
   SOURCE += engines/libpmem.c
 endif
+ifdef CONFIG_IME
+  SOURCE += engines/ime.c
+endif
 
 ifeq ($(CONFIG_TARGET_OS), Linux)
   SOURCE += diskutil.c fifo.c blktrace.c cgroup.c trim.c engines/sg.c \
--- a/configure
+++ b/configure
@@ -199,6 +199,8 @@ for opt do
   ;;
   --disable-native) disable_native="yes"
   ;;
+  --with-ime=*) ime_path="$optarg"
+  ;;
   --help)
     show_help="yes"
     ;;
@@ -230,6 +232,7 @@ if test "$show_help" = "yes" ; then
   echo "--disable-optimizations Don't enable compiler optimizations"
   echo "--enable-cuda           Enable GPUDirect RDMA support"
   echo "--disable-native        Don't build for native host"
+  echo "--with-ime=             Install path for DDN's Infinite Memory Engine"
   exit $exit_val
 fi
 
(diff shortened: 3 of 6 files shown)


axboe:
The engine looks pretty clean to me, minor style issues that aren't of big concern. One concern is the use of fio_ime_is_initialized however, what is that supposed to do? This is multi-threaded code, static globals aren't usually a great idea.

gbossu:
To answer your question, fio_ime_is_initialized is used to determine whether or not a call to ime_native_init should be issued. The general rule could be this: one call to ime_native_init per process.

Consequently, in the case where threads are used for jobs, it is not a problem to share this global variable, because if one thread has already initialized IME, then the other ones don't need to. In the case where forks are used for jobs however, ime_native_init should be called for every new fork. This is what should happen because each fork has its own copy of fio_ime_is_initialized.

In cases where both forks and threads are used for fio jobs, there could be issues though. eg: if some threads are created and had time to initialize IME, then fio_ime_is_initialized will be 1. As a consequence, if forks are created afterwards, they will think IME is initialized and won't call ime_native_init, which is a problem. To that end, the fio_ime_engine_init function should throw a warning if such a scenario happens.

axboe:
Seems like you should track that in your engine data. What happens if I define workloads that has both threads and processes right now? If td->o.use_thread is true, then you need one call for that regardless of td->o.numjobs. If td->o.use_thread is false, then you need one call per td->o.numjobs.

gbossu:
I agree this kind of information should be saved in the engine data. However, if we remove the global variable, the current job being initialized will have no knowledge that threads might have already initialized IME. This can cause a problem because the behavior of IME is undefined when ime_native_init is called before forking. So the global variable also helps tracking this kind of behavior.

However, it seems that FIO creates all forks and threads before initializing the engines. Can you confirm this behavior ? If that is actually the case, then we won't have problems with mixing forks and threads, because ime_native_init will be called after fork() is issued. If that is not the case, then I'm afraid we won't be able to safely mix forks and threads with IME as of today.

axboe:
If you do whatever you need to do from your io_ops->init() handler, then all forks or thread creations have been done at that time. That hook is called from the actual process or thread when it's starting, so I think that should work fine for you.

gbossu:
That's good news then, because it means jobfiles can safely mix threads and forks for jobs. I kept the warning message just in case, but it should never be displayed.
I also updated the patch to add a comment above the fio_ime_is_initialized flag. It should explain the purpose of this flag :)

If this looks good to you, could the PR be merged ?

axboe:
There are a few things that need to be done first:

- Remove the fio_ime_is_initialized. If you don't need it, then you don't need it. Don't keep it around.
- Generally we add --disable-ime and similar options, so that things are on by default. You have a configure check for it anyway. So I'd flip that around.
- You've got some trailing whitespace in various places
- Don't use brackets for single line if/else.
- "else" goes on the same line as the closing bracket. fio_ime_aio_complete_cb() is a good example of both this one and the previous line.
- This isn't C++, don't declare your loop iterator inside the for(;;) statement. Variable declarations go at the top of the scope you need them in.
- You need documentation. Nobody is going to know that this engine is there. Doesn't need to be much, but at least the engines need a blurb in the HOWTO.
- You need to wire up the engine in options.c or nobody is going to see it (the ioengine entry).
- You need an example job file to show how to use it. Preferably one for each of the engines you defined.

The style issues aren't major at all, the documentation and options.c parts are a must though.

gbossu:
Hmmm yes, I should have been more careful with the style issues. I have removed the c99-like variable declaration, removed brackets when I could, and other things. It should be better now ;)

As for the "--disable-ime", the thing is that, IME isn't usually installed in a known location. For that reason, I thought it was best give the installation path when running the configure script. I can change this if this doesn't comply with the FIO guidelines though. Just let me know how I should modify this part of the configure script.

Also, the fio_ime_is_initialized flag is not useless. In the case where some jobs used threads and initialized IME, then the flag will tell FIO that IME should be finalized when unregistering the engines.

For the documentation, options.c and example jobfile, I had totally skipped that, my bad!

The patch should be cleaner now, can you have a look again and let me know if some other issues need to be addressed ?

axboe:
That looks a lot better. One final suggestion - change the --enable-ime= to --with-ime= instead. None of the other --enable-foo take arguments, I think it's important to make the distinction because of that.

axboe:
Also, can you tell me where to get the SDK? As it stands, I can't even compile it...

axboe:
Sounds good, and yes please do. Thanks.

axboe:
OK, pulled in, made a few fixups to the style, and also added the man page entry. Please check that my current master branch looks good.

gbossu:
Looks good to me, thank you!

PR #655, opened 2018-08-16
http: use pkg-config and support older openssl versions

ddiss:


axboe:
Reopening this - any chance you can make it work without pkg-config? If we don't have that, then the old test case makes sense to do. Currently we're failing if pkg-config isn't installed.

ddiss:
I can do that, but it gets quite ugly due to the need to handle HMAC_CTX differences in the compiler check.

ddiss:
I was under the impression that pkg-config was safe to use based on the existing GTK_CFLAGS/GTK_LIBS checks.

axboe:
The gtk checks aren't on by default. So I would not say that pkg-config is a build requirement. It is if you use --enable-gfio, but with your change, it's a requirement by default. And it throws a visible error:

```
./configure: 1576: ./configure: pkg-config: not found
http engine                   no
```

If it makes it easier, you could just have separate tests and defines.

ddiss:
Reworking it now, will resubmit when I have something I'm happy with. Thanks for the feedback.

axboe:
Great, thanks a lot.

axboe:
Also check failures like this one:

https://travis-ci.org/axboe/fio/jobs/416901259

ddiss:
I've staged the changes at https://github.com/ddiss/fio/commit/9ee669faa39003c2317e5df892314bcfcee069e3 .
Using two separate tmp files for the compile check comes out a little cleaner IMO. @l-mb any chance you could do a quick test of the openssl >= 1.1.0 case (I've tested the < 1.1.0 path).

ddiss:
I've gone ahead and submitted via https://github.com/axboe/fio/pull/656 (I need to take off for a few hours). But keep in mind that I've only tested the openssl < 1.1.0 fallback case.

axboe:
I'm going to go ahead and pull it, it looks better than what we had.

ddiss:
Thanks. FWIW I confirmed that the openssl 1.1.0+ case works alright.

axboe:
Thanks, also seems to work fine on my laptop, which is 1.1.0g.

PR #659, opened 2018-08-17
Reimplement axmap_next_free() to prevent distribution skew

kelleymh:
New algorithm starts the search at the specified bit, and only
looks forward (higher index) for a free bit, wrapping around the
end of the map if necessary.  This avoids a distribution skew in
the current algorithm where the low order 5 or 6 bits of the
replacement address tend toward zero.

The first_free cached value is not updated or used. But performance
of the new algorithm is marginally faster than the old algorithm
when measured on a 200 million entry map that is half full,
averaged across searches starting with each of the 200 million
entries.

A separate commit adds new tests to t/axmap.c to confirm the
correct behavior in a variety of situations and edge cases.

### DIFF
--- a/lib/axmap.c
+++ b/lib/axmap.c
@@ -336,99 +336,119 @@ bool axmap_isset(struct axmap *axmap, uint64_t bit_nr)
 	return false;
 }
 
-static uint64_t axmap_find_first_free(struct axmap *axmap, unsigned int level,
-				       uint64_t index)
+/*
+ * Find the first free bit that is at least as large as bit_nr.  Return
+ * -1 if no free bit is found before the end of the map.
+ */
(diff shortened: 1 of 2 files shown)


axboe:
This is awesome, thanks! Took a look at the changes, and looks fine to me. Also did comparison runs between the two. I'd say this is good, pulling it in.

kelleymh:
With this in now, I'll submit another small set of changes to remove all vestiges of the first_free cache.  There's still some code in axmap.c that sets the value, and it's no longer used anywhere.

Also, I noticed that axmap_clear() is never referenced.  Should it stick around for potential future use?  Or should I just remove it as well as part of the cleanup?

axboe:
Sounds good, thanks. If axmap_clear() isn't used, just kill it. No point in hanging on to dead code, it's always in the git history if folks have a need to resurrect it.

PR #660, opened 2018-08-17
Remove unused code in lib/axmap.c

kelleymh:
With the new axmap_next_free() implementation, the "first_free" cache
is no longer used.  Remove all references to it.  Also axmap_clear()
is never referenced, so remove it and related code.


axboe:
Thanks, pulled.

PR #661, opened 2018-08-22
Document and optimize the axmap implementation

bvanassche:
While I was verifying the axmap implementation I documented what is nontrivial and I also implemented a few small optimizations.

### DIFF
--- a/lib/axmap.c
+++ b/lib/axmap.c
@@ -128,6 +128,8 @@ struct axmap *axmap_new(unsigned long nr_bits)
 
 	axmap->nr_levels = levels;
 	axmap->levels = calloc(axmap->nr_levels, sizeof(struct axmap_level));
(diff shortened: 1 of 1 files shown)


axboe:
Looks good to me. I was surprised at patch #2, but you are right, once we get to __axmap_set() we know we only have a max of BLOCKS_PER_UNIT. Maybe that deserves a comment? Rest looks like good changes to me.

bvanassche:
Have you noticed the following comment that was added through patch #1 above __axmap_set()? Is this sufficient?

> It is guaranteed that all bits that have been requested to set fit in the same unsinged long word of
> level 0 of @axmap.

axboe:
I did, I tend to never read comments above functions :-). Anyway, looks good, pulled. Thanks Bart.

PR #663, opened 2018-08-23
20180820 terse remote fix

mcgrof:
Just a minor fix, as already posted on the mailing list.


fio.h
```diff
@@ -534,6 +535,13 @@ extern char *aux_path;
 
 extern struct thread_data *threads;
 
+static inline bool is_running_backend(void)
+{
+	if (is_backend || is_local_backend)
```
axboe: Just make that function be:

```return is_backend || is_local_backend```

PR #664, opened 2018-08-23
20180823 terse remote fix

mcgrof:
Addressed feedback from the first iteration. No other changes were made.


axboe:
Thanks Luis, pulled.

PR #665, opened 2018-08-23
lib/axmap: Remove an assert() statement that is no longer correct

bvanassche:
Due to patch "lib/axmap: Simplify axmap_set_fn()" it is now possible
that mask is zero. Avoid that this causes fio to abort.

Fixes: 15a4f4962fad ("lib/axmap: Simplify axmap_set_fn()")

### DIFF
--- a/lib/axmap.c
+++ b/lib/axmap.c
@@ -244,7 +244,6 @@ static bool axmap_set_fn(struct axmap_level *al, unsigned long offset,
 		mask = bit_masks[nr_bits] << bit;
 	}
 
-	assert(mask);
 	assert(!(al->map[offset] & mask));
 	al->map[offset] |= mask;
 


lib/axmap.c
```diff
@@ -244,7 +244,6 @@ static bool axmap_set_fn(struct axmap_level *al, unsigned long offset,
 		mask = bit_masks[nr_bits] << bit;
 	}
 
-	assert(mask);
```
axboe: I think this should be cleaner, ala:

```
if (mask) {                                                           
        assert(!(al->map[offset] & mask));
        al->map[offset] |= mask;
}
```

preferably with a comment above that stating when ```mask``` can be zero.

axboe:
Also, bonus points for adding a t/axmap.c test case that catches this.

PR #671, opened 2018-09-04
Log files names start with _ when write_XX_log= keys in config file have empty value(s)

szaydel:
This is a gist of the problem. According to documentation, if no job name is given, resulting name of the log should be: `jobname_type.x.log`. However, it is unspecified what happens when config file has for instance:
```
write_lat_log=
write_bw_log=
write_iops_log=
```
... as opposed to ...
```
write_lat_log
write_bw_log
write_iops_log
```
What ends-up happening, if the former is used in the config file are names of logs which start with `_`, and look like `_iops.log`, or `_bw.log`, etc.

While one might argue that it is not reasonable to expect that names will be correct if the value is in effect an empty string, in my mind this should be accounted for and jobname string used instead, just like what would happen if there was no `=some value`. Correctness of this change is debatable, but I feel like this is better than the alternative, at least maybe less unexpected.

### DIFF
--- a/init.c
+++ b/init.c
@@ -1542,7 +1542,9 @@ static int add_job(struct thread_data *td, const char *jobname, int job_add_num,
 			.log_gz = o->log_gz,
 			.log_gz_store = o->log_gz_store,
 		};
-		const char *pre = o->lat_log_file ? o->lat_log_file : o->name;
+		const char *pre =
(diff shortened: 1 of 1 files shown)


axboe:
I agree with the change of making an empty string the same as no string. I'll pull this, but then move it into a helper.

PR #675, opened 2018-09-07
remove dependency on unittest2 module

bengland2:
this dependency on unittest2 module was causing heartburn for pbench community so I removed the dependency.  If python-unittest2 package is installed and unittest2 module can be imported, that's fine, we'll use it if requested, but if it's not, that's fine too, you can still use the program.

### DIFF
--- a/tools/hist/fio-histo-log-pctiles.py
+++ b/tools/hist/fio-histo-log-pctiles.py
@@ -24,7 +24,12 @@
 import sys, os, math, copy
 from copy import deepcopy
 import argparse
-import unittest2
+
+unittest2_imported = True
+try:
+    import unittest2
+except ImportError:
+    unittest2_imported = False
 
 msec_per_sec = 1000
 nsec_per_usec = 1000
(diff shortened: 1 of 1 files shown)


axboe:
Hi Ben, the change looks fine, but please put all that relevant information into the actual commit. The git history is for prosperity, the github data isn't.

bengland2:
Sorry, not sure what you mean.  do you mean I should add the comment above into the code?   If so I can easily do that.

axboe:
No, I mean the reasoning for the commit that you put into a github conversation (first comment here) should have been the git commit message. When you browse the git repo, you should have all the info you need right there. Any information in comments here should be considered ephemeral.

Hope that makes sense.

axboe:
Here's an example of a good commit:

https://github.com/axboe/fio/commit/ee3696bdfd84bb2a0ef0757a6ce0140eea46228a

it has a title line, then an empty line, then a section (or more sections) explaining the reasoning for WHY the change is being made. If you run git log on the repo, all commits should follow that recipe.

bengland2:
ok, will have it to you soon, I didn't know about this convention but it
makes sense.  I'm just curious, is this part of some coding standard or is
this specific to fio?

axboe:
That is a great resource, hadn't seen that before. It explains both why it matters, and how to do it. I absolutely hate sloppy git commit messages. Gospel:

1) Separate subject from body with a blank line
2) Limit the subject line to 50 characters
3) Capitalize the subject line
4) Do not end the subject line with a period
5) Use the imperative mood in the subject line
6) Wrap the body at 72 characters
7) Use the body to explain what and why vs. how

Of these, 2-5 aren't super important (eg I don't mind a 72 char subject line). #1/#6/#7 are VERY important.

bengland2:
I changed the commit comment to this format 2 days ago, let me know if there is a problem.

axboe:
I've pulled it, but you're still way over the 72 char-per-line limit. This is important, as it helps it display properly when people run git log in a 80 char wide terminal.

PR #679, opened 2018-09-11
rate_submit: synchronize accesses to io_u_queue->nr

vincentkfu:

Another way to resolve this issue would be to set td = td->parent before calculating needs_lock in put_io_u(). If that is better, I can submit a patch for that change instead.

### DIFF
--- a/rate-submit.c
+++ b/rate-submit.c
@@ -126,7 +126,7 @@ static int io_workqueue_init_worker_fn(struct submit_worker *sw)
 	clear_io_state(td, 1);
 
 	td_set_runstate(td, TD_RUNNING);
-	td->flags |= TD_F_CHILD;
+	td->flags |= TD_F_CHILD | TD_F_NEED_LOCK;
 	td->parent = parent;
 	return 0;
 


axboe:
Thanks, that was pretty stupid. Another fix would be to have td_async_processing() return true for td->parent as well, since that's the case we need locking for. The same issue exists in requeue_io_u().

I'm not adverse to the flags inheriting, but it would need some careful audit to ensure that we can inherit any flag and it'd be fine. Might be cleaner to simply just set

```
td->flags |= TD_F_CHILD | TD_F_LOCKING;
```

when we initialize it.

axboe:
Oh and Vincent, when you respin with that change instead, can you ensure that the commit messages includes a:

Fixes: 26b3a1880d38 ("Make td_io_u_lock/unlock() explicit")

line at the bottom? That's the right way to indicate what commit introduced the bug. BTW, let this be a lesson to myself (and others), never add code to make a static checker happy. It's just stupid, and it's more likely to introduce bugs than you'd like.

vincentkfu:
It sounds like the best fix is to do it in td_async_processing() since it would only need to be done in one location and we wouldn't need to worry about issues introduced by flags inheriting.

axboe:
I'd just do the one-liner addition of TD_F_NEED_LOCK in the rate-submit init worker function. It's the only spot that needs it. I think that's cleaner than the implicit knowledge that we need locking, if td->parent is set (hence td is a child).

Issue #683, opened 2018-09-14
Fio fails to link libcurl on windows (undefined references to __imp_curl_easy_cleanup)


rustyscottweber:
Attempting to pull the latest changes for master branch on windows and attempting to build according to instructions on:
https://media.readthedocs.org/pdf/fio/latest/fio.pdf
Results in the following error.
```
engines/http.o: In function `fio_http_cleanup':
/home/RWeber/Ion_Projects/fio/engines/http.c:438: undefined reference to `__imp_curl_easy_cleanup'
engines/http.o: In function `fio_http_setup':
/home/RWeber/Ion_Projects/fio/engines/http.c:597: undefined reference to `__imp_curl_easy_init'
/home/RWeber/Ion_Projects/fio/engines/http.c:598: undefined reference to `__imp_curl_easy_setopt'
engines/http.o: In function `fio_http_cleanup':
/home/RWeber/Ion_Projects/fio/engines/http.c:438: undefined reference to `__imp_curl_easy_cleanup'
engines/http.o: In function `fio_http_queue':
/home/RWeber/Ion_Projects/fio/engines/http.c:503: undefined reference to `__imp_curl_easy_setopt'
/home/RWeber/Ion_Projects/fio/engines/http.c:568: undefined reference to `__imp_curl_slist_free_all'
/home/RWeber/Ion_Projects/fio/engines/http.c:532: undefined reference to `__imp_curl_easy_perform'
/home/RWeber/Ion_Projects/fio/engines/http.c:534: undefined reference to `__imp_curl_easy_getinfo'
/home/RWeber/Ion_Projects/fio/engines/http.c:520: undefined reference to `__imp_curl_easy_perform'
/home/RWeber/Ion_Projects/fio/engines/http.c:522: undefined reference to `__imp_curl_easy_getinfo'
engines/http.o: In function `_add_swift_header':
/home/RWeber/Ion_Projects/fio/engines/http.c:420: undefined reference to `__imp_curl_slist_append'
engines/http.o: In function `fio_http_queue':
/home/RWeber/Ion_Projects/fio/engines/http.c:552: undefined reference to `__imp_curl_easy_perform'
/home/RWeber/Ion_Projects/fio/engines/http.c:554: undefined reference to `__imp_curl_easy_getinfo'
engines/http.o: In function `_add_aws_auth_header':
/home/RWeber/Ion_Projects/fio/engines/http.c:389: undefined reference to `__imp_curl_slist_append'
[...]

PR #685, opened 2018-09-15
build: change where we set -static for Windows

sitsofe:
The Makefile forced static Windows builds by adding -static to the
CFLAGS but this is long after the configure tests had run. This could
lead to problems (such as https://github.com/axboe/fio/issues/683 "Fio
fails to link libcurl on windows (undefined references to
__imp_curl_easy_cleanup)") where the availability of a dynamic version
of library causes a feature to be enabled but when it comes to final
linking time we find that there's no static version of the library
available and fall over.

Fix the above by asking for a static build via the "Forcing some known
good options on Windows" section of the configure file which occurs
prior to configure tests. This also means configure will correctly
display that we are doing a static build.

Fixes https://github.com/axboe/fio/issues/683

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,7 +198,7 @@ endif
 ifneq (,$(findstring CYGWIN,$(CONFIG_TARGET_OS)))
   SOURCE += os/windows/posix.c
   LIBS	 += -lpthread -lpsapi -lws2_32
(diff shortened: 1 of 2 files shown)


axboe:
Thanks, I like this. My previous guess work was related to exactly that, ensuring that configure runs with the same options that the actual build does. As a longer term goal, we need to move even more of the ifdef PLATFORM stuff from the Makefile into configure.

PR #686, opened 2018-09-16
Add Musl support

neheb:
The *48_r functions are glibc extensions that also happen to be supported
by uclibc as well. Adapted from the Android solution.

__USE_FILE_OFFSET64 is glibc specific so use LARGEFILE_SOURCE.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


sitsofe:
Hmm how is this going to interact with things like BSD libc and Android's bionic?

sitsofe:
I also wonder if these are better off living in some sort of compat area that that activates them to paper  over the differences based on configure test results...

neheb:
The first part is from the Linux section.

The second one the Makefile defines anyway.

AFAIK most if not all of the BSDs use a 64-bit off_t as well as Android and musl.

oslib/libmtd_common.h
```diff
@@ -69,7 +69,7 @@ extern "C" {
 #endif
 
 /* define a print format specifier for off_t */
-#ifdef __USE_FILE_OFFSET64
+#ifdef _LARGEFILE_SOURCE
```
sitsofe: Is this correct? i.e. on a 32 bit systems will you definitely make all the file functions take 64 bits offsets?

os/os-linux.h
```diff
@@ -172,14 +176,27 @@ static inline unsigned long long os_phys_mem(void)
 
 static inline void os_random_seed(unsigned long seed, os_random_state_t *rs)
 {
+#if defined(__GLIBC__) | defined (__UCLIBC__)
```
sitsofe: Did you intend to use a single pipe symbol here?

neheb: Typo. Will fix.

oslib/libmtd_common.h
neheb: It is correct. The Makefile currently passes -D_LARGEFILE_SOURCE. This change merely fixes format strings.

sitsofe: Excellent point. Since the Makefile does so unconditionally does it make more sense to do away with the #ifdef branches?

sitsofe: In fact, where do `PRIxoff_t` and `PRIdoff_t` get used in fio?

sitsofe:
I still can't help wondering if this should be tested in configure and then used appropriately - I don't like the idea of having to code around each alternate libc (e.g. what happens to `dietlibc` users with this patch?)...

neheb:
From a quick look, it doesn't seem like dietlibc supports srand48_r either.

axboe:
That's part of old code that is not used anymore, I just killed it.

neheb:
Excellent. Now the only issue is the _LARGEFILE_SOURCE thing which on a second look seems wrong.

That macro makes the o functions available, which is interesting since those are completely unused.

_LARGEFILE64_SOURCE makes the 64 suffixed functions available. On Musl, off_t is 64 bits with no support for 32-bits. Not sure of a good way to deal with this...

PR #688, opened 2018-09-20
engines/rados: Moved initialization of ceph rados connection from setup to init.

aclamk:
This fixes problems with initialization when --read-iolog is used.

### DIFF
--- a/engines/rados.c
+++ b/engines/rados.c
@@ -381,7 +381,7 @@ int fio_rados_getevents(struct thread_data *td, unsigned int min,
   return events;
 }
 
-static int fio_rados_setup(struct thread_data *td)
+static int fio_rados_init(struct thread_data *td)
 {
 	struct rados_data *rados = NULL;
 	int r;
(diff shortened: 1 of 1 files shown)


axboe:
Some of the things that rados currently does in its ->setup() can't be done at init time - for instance, it's forcing the use of threads. That's somewhat questionable, but that aside, it won't do anything if called from ->init() as the thread or process is already launched at that point. It's buggy right now too though, as it'll depend on ```create_serialize``` what time it's called at. That should probably be the first cleanup done, just get rid of that.

Outside of that, don't see anything that prevents moving to ->init() instead of ->setup().

aclamk:
@axboe
The actual problem is that in case of reading from iolog, `setup()` is never called.
https://github.com/axboe/fio/blob/master/filesetup.c#L911-L912
Would it be ok, to move this code below calling to setup() ?

axboe:
Yeah that's a good point, and seems like a better fix.

PR #689, opened 2018-09-20
Mixed priority

trinof:
This adds the prio_percent option which allows a certain percentage of commands to be sent at an altered priority (controlled by the existing prioclass option). Since this only is relevant to reads on the drive side, only reads will be broken down into multiple priorities like this. The latencies of the two priority types will be reported separately in addition to the combined latencies. This will only affect normal and json reporting.


axboe:
I'm assuming you didn't intend to submit this pull request as-is?

sitsofe:
@trinof You broke the Windows build and we aren't all that interested in adding things like your SlickEdit project to mainline fio :-) We wouldn't merge your pull request it stands because we generally expect you to squash things down to the minimum commits before they go in (e.g. all your cleanups should not be visible in the pull request). You also seem to have merged later versions of fio into your branch and we generally prefer to avoid that and instead expect you to rebase your branch to avoid that pattern. Also we need you to add documentation to both the man page (`fio.1`) and the `HOWTO` explaining the new feature.

Could you change the above, add your signed-off-by to the commits? Then we can re-review and it will be easier to comment on the tweaks necessary for this to go in...

eta.c
```diff
@@ -494,9 +494,9 @@ bool calc_thread_status(struct jobs_eta *je, int force)
 		calc_rate(unified_rw_rep, rate_time, io_bytes, rate_io_bytes,
 				je->rate);
 		memcpy(&rate_prev_time, &now, sizeof(now));
-		add_agg_sample(sample_val(je->rate[DDIR_READ]), DDIR_READ, 0);
-		add_agg_sample(sample_val(je->rate[DDIR_WRITE]), DDIR_WRITE, 0);
-		add_agg_sample(sample_val(je->rate[DDIR_TRIM]), DDIR_TRIM, 0);
+		add_agg_sample(sample_val(je->rate[DDIR_READ]), DDIR_READ, 0, td->priorityBit);
```
sitsofe: We generally avoid studlycaps in fio - underscores between words are preferred.

engines/sync.c
```diff
@@ -149,20 +148,14 @@ static int fio_pvsyncio2_queue(struct thread_data *td, struct io_u *io_u)
...
-            {
+            if ((rand()%100 <= o->hipri_percentage) == 0) {
                 dprint(FD_IO, "Disable RWF_HIPRI\n");
-            }
-            else
-            {
+            } else {
                 dprint(FD_IO, "Enable RWF_HIPRI\n");
                 flags |= RWF_HIPRI;
             }
```
sitsofe: I think we would prefer this to use fio's own internal random number generator routines rather than `rand()` to make things deterministic...

axboe:
To clarify on the slickedit part, I actually don't mind adding aux support like that. The issue with this one, though, is that we now have to maintain the file list separately as well. I don't want to do that, it'll certainly bitrot since it's not in widespread use.

Your series also needs a lot of cleanup, it should be boiled down to a set of concise patches. Right now it's just a history of how it was developed.

trinof:
Thank you for being patient and pointing out all of the problems with this pull request. I haven't ever issued a pull request before, and only started using git this year, so I'm sorry for all of the mistakes. I'll work on all the problems mentioned, but I'm not sure I understand which version of FIO to base my pull request on. I merged with the latest available, should I have merged with the latest tagged release instead?

eta.c
trinof: Ok, I'll change all instances of studlycaps to be consistent.

axboe:
It should be a clean series on top of something recent, usually there's no reason NOT to just use current -git. I don't necessarily care of it's the last tag, as long as it merges cleanly, but new developments should always fork from the current tip of the tree.

Don't worry about the unclean series, we're happy to help you along and hopefully end up with something that is suitable for merging.

sitsofe:
@trinof Don't worry about this being your first pull request - we all start somewhere and it's a learning process and it looks like some of this was inherited by you too. We can go round this loop of submitting and reviewing the pull request until we get to where it needs to be :-)

You might find using git's interactive rebase is useful for squashing things down (or it might just be more confusing...).

trinof:
I believe I have addressed all of your initial concerns, could you look over the cleaned up pull request when you have the time? Thank you for being patient and understanding.

ioengines.c
```diff
@@ -311,6 +313,38 @@ enum fio_q_status td_io_queue(struct thread_data *td, struct io_u *io_u)
 					sizeof(io_u->issue_time));
 	}
 
+	if (fio_option_is_set(o, ioprio) ||
+	    fio_option_is_set(o, ioprio_class)) {
+
+	    if ( io_u->ddir == DDIR_READ) {
+	        if (o->prio_percent) {
+	            if ((rand()%100 >= o->prio_percent) == 0) {
```
axboe: rand() % 100 is NOT a good way to check for this. I think even the man page states that pretty clearly. This needs to be a dedicated random state so that things are manageable and repeatable (if desired). Also use rand_between() for this.

trinof: I changed the rand() call to rand_between() including adding a new random state to the thread data structure. I squashed my update into the existing commit, but I see doing so marked your comment as outdated. In the future, should I be waiting to squash commits until the patch is more or less finalized?

eta.c
```diff
@@ -494,9 +494,9 @@ bool calc_thread_status(struct jobs_eta *je, int force)
 		calc_rate(unified_rw_rep, rate_time, io_bytes, rate_io_bytes,
 				je->rate);
 		memcpy(&rate_prev_time, &now, sizeof(now));
-		add_agg_sample(sample_val(je->rate[DDIR_READ]), DDIR_READ, 0);
-		add_agg_sample(sample_val(je->rate[DDIR_WRITE]), DDIR_WRITE, 0);
-		add_agg_sample(sample_val(je->rate[DDIR_TRIM]), DDIR_TRIM, 0);
+		add_agg_sample(sample_val(je->rate[DDIR_READ]), DDIR_READ, 0, td->priority_bit);
```
axboe: Let's split the logging changes from the actual functional change.

io_u.c
```diff
@@ -1808,6 +1808,8 @@ static void account_io_completion(struct thread_data *td, struct io_u *io_u,
 	if (td->parent)
 		td = td->parent;
 
+	td->priority_bit = io_u->priority_bit;
```
axboe: I don't like tracking this in the td, that seems very random. Why track something that belongs to a specific IO in the thread data?

ioengines.c
```diff
@@ -311,6 +313,38 @@ enum fio_q_status td_io_queue(struct thread_data *td, struct io_u *io_u)
 					sizeof(io_u->issue_time));
 	}
 
+	if (fio_option_is_set(o, ioprio) ||
```
axboe: I'm not convinced this part is even workable. In practice, per-io priority should be passed in during submission, through libaio actually utilizing the priority fields. That would be a lot more straight forward, and also a lot faster. This setup relies on the fact that we don't look at the task io priority after the IO has been queued in the kernel, It seems fragile.

iolog.h
```diff
@@ -43,6 +43,7 @@ struct io_sample {
 	union io_sample_data data;
 	uint32_t __ddir;
 	uint64_t bs;
+	unsigned int priority_bit;
```
axboe: This enlarges the IO samples (obviously), which adds to the memory pressure on tracking these. It also means that the fio client/server needs a revision jump, since we pass these on the wire. This is also why it's explicitly sized right now, so you should not use unsigned ints here.

stat.h
```diff
@@ -237,6 +238,12 @@ struct thread_stat {
 	fio_fp64_t ss_deviation;
 	fio_fp64_t ss_criterion;
 
+	uint32_t priority_bit;
+	uint64_t io_u_plat_prio[FIO_IO_U_PLAT_NR];
+	uint64_t io_u_plat_low_prio[FIO_IO_U_PLAT_NR];
+	struct io_stat clat_prio_stat;
+	struct io_stat clat_low_prio_stat;
+
```
axboe: The stat changes would need client/server changes as well, see how we convert values across the wire in client.c and server.c

axboe:
I'm having some doubts on all this, unfortunately my first review got disrupted but I have now finished it. Would this be workable using multiple jobs, and flow control between them?

ioengines.c
sitsofe: I think what you did was fine - if you touch the area where a comment is then it's going to become marked as outdated which is the expected (and desired) behaviour.

eta.c
trinof: Ok, in my next update I will split the patch into a functional and a logging commit.

ioengines.c
trinof: I was hoping to make the prio_percent option workable on all IO engines rather than just libaio. Do you know if other IO engines can directly set the priority fields? Would it make sense to add a new function to the ioengine_ops struct called prio_prep and create a libaio specific prio_prep that sets the priority fields directly but leave other IO engines to use this current suboptimal implementation until they are updated with an engine specific implementation?

trinof:
Thanks, that fixes the flow counter behaviour. However, I think there are still reasons to want the priority reporting in one job. Our current target workload is a mixed random readwrite mixed priority workload where we are most interested in completion latency distribution. We want 99% reads and 1% writes, but if we try and split up our job into a readwrite and a read job there's no way that I can see with integer rwmix percentages to get the exact 99/1 rwmix we want for arbitrary priority percentages. And if we make both jobs random readwrite jobs, then we would have a hard time getting the aggregate write workload latency percentages short of getting the individual IO latencies and calculating write latencies ourselves. And if we split the jobs into two read jobs and one write job, it seems like using flow counters isn't sufficient to properly constrain the exact proportions between three jobs. So while we might be able to use flow counters to create the workload we are interested in, it seems to me there would be a great deal of acrobatics involved for just this relatively simple workload.

iolog.h
trinof: It looks to me like not all bits are used in the __ddir field. Would it make sense to avoid the additional memory footprint by rolling priority_bit into the existing uint32 with bit field declarations? I don't know if that would make the client/server updates easier or more complicated since I've never explored that code before.

io_u.c
trinof: I agree, it seems like this is unnecessary and could be more cleanly handled. I'll investigate a way to fix this.

trinof:
Would you be willing to take a look at this version? I've changed the prio_percent to a libaio specific option and changed the implementation to use the reqprio field. Unfortunately this only seems to work on kernel 4.18 and forward, and I wasn't sure if the Makefile should be changed to check for support. If so, I would appreciate advice on how to do so. I think I updated the client/server appropriately, and I've removed the priority bit from the thread data. Thank you again for your patience.

trinof:
@axboe @sitsofe Have either of you had a chance to review this version? I think it fixes the fundamental problems you were worried about, but if you have further concerns I would be interested to hear them.

sitsofe:
@trinof if you haven't had a review from me in a few day keep poking me...

trinof:
@sitsofe Thanks for volunteering to be poked, I'm excited to hear what you think.

engines/libaio.c
```diff
@@ -13,13 +13,28 @@
 #include "../lib/pow2.h"
 #include "../optgroup.h"
 
+// Should be defined in newest aio_abi.h
```
sitsofe: We prefer C style comments rather than C++ ones where possible.

engines/libaio.c
```diff
@@ -366,27 +444,26 @@ static int fio_libaio_init(struct thread_data *td)
...
 
 	td->io_ops_data = ld;
+	/*
+	 * Initialize prio_percent mechanisms if relevant
+	 */
+	if (!(fio_option_is_set(to, ioprio) ||
+			fio_option_is_set(to, ioprio_class) ||
+			(o->prio_percent == 0))) {
+		ioengine.prio_prep = fio_libaio_prio_prep;
+		ld->prio_seed = FIO_RANDSEED * td->thread_number
```
sitsofe: Shouldn't we be using fio's built-in random number generation for this - (see the functions in `lib/rand.h`)? Then `FIO_RANDSEED` wouldn't need duplicating...

engines/libaio.c
```diff
@@ -51,13 +67,36 @@ static struct fio_option options[] = {
...
+#error "FIO_HAVE_IOPRIO_CLASS requires FIO_HAVE_IOPRIO"
+#endif
+	{
+		.name	= "prio_percent",
+		.lname	= "prio percentage",
+		.type	= FIO_OPT_INT,
+		.off1	= offsetof(struct libaio_options, prio_percent),
+		.minval	= 1,
+		.maxval	= 100,
+		.help	= "Split the prioclass setting for IO",
```
sitsofe: There was kind of a rename a while ago to call it I/O rather than IO.

engines/libaio.c
```diff
@@ -79,6 +118,23 @@ static int fio_libaio_prep(struct thread_data fio_unused *td, struct io_u *io_u)
 	return 0;
 }
 
+static void fio_libaio_prio_prep(struct thread_data *td, struct io_u *io_u)
+{
+	struct libaio_options *o = td->eo;
+	struct libaio_data *ld = td->io_ops_data;
+	if (rand_between(&ld->prio_state, 0, 99) < o->prio_percent) {
+		dprint(FD_IO, "Enable PRIO \n");
```
sitsofe: Will these `dprint` statements have to be evaluated on every I/O?

engines/libaio.c
```diff
@@ -335,10 +393,30 @@ static void fio_libaio_cleanup(struct thread_data *td)
 	}
 }
 
+static struct ioengine_ops ioengine = {
+	.name			= "libaio",
+	.version		= FIO_IOOPS_VERSION,
+	.init			= fio_libaio_init,
+	.prep			= fio_libaio_prep,
+	.prio_prep		= NULL,
```
sitsofe: [idle speculation] I see what you're going for here (making it generic to other engines) but I can't help wondering if it should be a `libaio` only option for now and it can be done off its queue hook...

engines/libaio.c
```diff
@@ -366,27 +444,26 @@ static int fio_libaio_init(struct thread_data *td)
...
+		ioengine.prio_prep = fio_libaio_prio_prep;
+		ld->prio_seed = FIO_RANDSEED * td->thread_number
+				+ FIO_RAND_NR_OFFS;
+		init_rand_seed(&ld->prio_state, ld->prio_seed, false);
+	}
+	else if (o->prio_percent != 0) {
+		log_err("%s: prio_percent option and mutually exclusive "
+				"prio or prioclass option is set, exiting\n", to->name);
+		td_verror(td, EINVAL, "fio_libaio_init");
+		return 1;
```
sitsofe: Should this be mentioned in the help?

options.c
```diff
@@ -3255,6 +3255,8 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
 		.help	= "Your platform does not support IO scheduler switching",
 	},
 #endif
+
+	/* Parameters for zones defined in the fio job */
```
sitsofe: Maybe these should be queued in another commit?

server.c
```diff
@@ -1562,6 +1562,14 @@ void fio_server_send_ts(struct thread_stat *ts, struct group_run_stats *rs)
 	p.ts.ss_deviation.u.i	= cpu_to_le64(fio_double_to_uint64(ts->ss_deviation.u.f));
 	p.ts.ss_criterion.u.i	= cpu_to_le64(fio_double_to_uint64(ts->ss_criterion.u.f));
 
+	p.ts.priority_bit = ts->priority_bit;
+	for (i = 0; i < FIO_IO_U_PLAT_NR; i++) {
+		p.ts.io_u_plat_prio[i] = cpu_to_le64(ts->io_u_plat_prio[i]);
```
sitsofe: Isn't it technically high priority and regular priority rather than low priority and regular priority?

server.c
```diff
@@ -1988,7 +1996,7 @@ int fio_send_iolog(struct thread_data *td, struct io_log *log, const char *name)
 
 			s->time		= cpu_to_le64(s->time);
 			s->data.val	= cpu_to_le64(s->data.val);
-			s->__ddir	= cpu_to_le32(s->__ddir);
+			s->__ddir	= __cpu_to_le32(s->__ddir);
```
sitsofe: It seems a pity to lose the typechecking but as a bit ha been "stolen" there's probably not much that can be done...

stat.h
```diff
@@ -14,6 +14,7 @@ struct group_run_stats {
 	uint32_t sig_figs;
 	uint32_t groupid;
 	uint32_t unified_rw_rep;
+	uint64_t nr_zone_resets;
```
sitsofe: Not sure about this...

stat.c
```diff
@@ -472,9 +472,12 @@ static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
 		display_lat("clat", min, max, mean, dev, out);
 	if (calc_lat(&ts->lat_stat[ddir], &min, &max, &mean, &dev))
 		display_lat(" lat", min, max, mean, dev, out);
+	if (calc_lat(&ts->clat_prio_stat, &min, &max, &mean, &dev) && ddir == DDIR_READ)
+		display_lat("PRIO_clat", min, max, mean, dev, out);
```
sitsofe: I don't think the other fields use mixed word case like this. Also maybe swap the order of comparisons, i.e. check `DDIR_READ` first?

stat.c
```diff
@@ -486,6 +489,25 @@ static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
 					samples,
 					ts->percentile_list,
 					ts->percentile_precision, name, out);
+
+        // Only print this if some priority stats were collected
```
sitsofe: C++ comment and strange indentation...

stat.c
```diff
@@ -486,6 +489,25 @@ static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
...
+		if (ddir == DDIR_READ && ts->clat_prio_stat.samples > 0) {
+			sprintf(prio_name, "high prio (%.2f%%) %s",
```
sitsofe: This looks a bit free form to me. Won't this be clear from the percentiles?

stat.c
```diff
@@ -1600,6 +1740,10 @@ void sum_thread_stats(struct thread_stat *dst, struct thread_stat *src,
 				dst->runtime[l] = src->runtime[l];
 		} else {
 			sum_stat(&dst->clat_stat[0], &src->clat_stat[l], first);
+			if (l == DDIR_READ) {
+				sum_stat(&dst->clat_prio_stat, &src->clat_prio_stat, first);
```
sitsofe: How come these don't follow the array forms of the lines below?

stat.h
```diff
@@ -237,6 +238,12 @@ struct thread_stat {
 	fio_fp64_t ss_deviation;
 	fio_fp64_t ss_criterion;
 
+	bool priority_bit;
+	uint64_t io_u_plat_prio[FIO_IO_U_PLAT_NR];
+	uint64_t io_u_plat_low_prio[FIO_IO_U_PLAT_NR];
+	struct io_stat clat_prio_stat __attribute__((aligned(8)));
+	struct io_stat clat_low_prio_stat __attribute__((aligned(8)));
+
```
sitsofe: There's a lot of if `ddir == DDIR_READ...` in the sample printing code now. I wonder it wouldn't be better to just avoid the ifs and print zeros for those directions that priority wasn't gathered on.

sitsofe: OK I've done my review for now and asked a few questions.

engines/libaio.c
trinof: Yes. Since the decision is made on a per-I/O basis, I'm not sure if there's a less verbose way to include this debug information. Do you think this will produce too much output and should be taken out?

stat.c
trinof: Since we're only tracking data for reads, I didn't make the the prio stats use the normal DDIR array, since only the reads will ever be populated. Since the other I/O types won't be tracked I didn't think I should be creating an array for them.

engines/libaio.c
trinof: Ok, I'll convert the comments wherever I've made them

engines/libaio.c
trinof: Ok, I'll update this. I'll also update the help text to be accurate, since prioclass and prio_percent cannot be used together in this implementation.

engines/libaio.c
trinof: I'm happy to incorporate the prio selection into the queue hook.

options.c
trinof: I'll take them out. I think the comments give helpful structure, but they certainly aren't related to any priority changes.

stat.h
trinof: I'll take this out.

stat.h
trinof: We could print zeroes and add the other I/O types to the various priority specific structures, but this seemed like a cleaner implementation to me. But if you think it's better to print zeroes and allocate ddir arrays for the priority structures, then I'll be happy to change it.

engines/libaio.c
trinof: It isn't clear to me how to use fio's built-in random number generation to seed the random number generator. Wouldn't I need to set up another random state to use any of the random number generators, which would in turn need its own seed? I'm not particularly attached to using the FIO_RANDSEED value, but I was trying to follow the way random seeds generally seem to be chosen. Should I just pick an arbitrary constant value?

engines/libaio.c
trinof: Yes, I'll add that to the prio_percent description.

stat.c
trinof: Good idea, and I'll change the case.

stat.c
trinof: I'll fix this.

stat.c
trinof: Hopefully, this will be clear from the percentiles, but in case something unexpected happens it seems better to explicitly state which stats are from the high priority reads and which ones are the low priority (although as you mentioned before, these are really regular priority reads, but they are comparatively lower priority). Since this is outputting to the normal human readable format, I thought it was good to be a little bit more verbose. Is there another way to display the data that you'd like to use instead?

server.c
trinof: Yes, you are correct. I'll rename the variables and try not to confuse the division between the functional and the logging commits.

axboe:
What happened with this one?

trinof:
I don't remember precisely, but it appeared to me like the pull request was closed without merging because the changes seemed too risky to include. I tried to address all the concerns raised, so I was somewhat surprised when it got closed.

axboe:
Huh that's so weird, maybe I thought I had pulled it. Then someone asked me about IO priorities today and I couldn't understand why I could not find the mixed priority bits. Sorry about that! Can you rebase on current master and I'll take another look, hopefully we can get this pulled.

trinof:
I don't see any way to reopen this pull request, would you prefer a new pull request or a patch file to review?

axboe:
Yeah I can't reopen it either, which is weird. Just open a new one!

Issue #687, opened 2018-09-18
steady state detection function does not work


tyneliu:
All the three option ss_ramp, ss_dur and ss_ramp does not work. no matter how you change the three parameter it will not affect the fio steady state detection.
here is my fio options:
fio --ioengine=libaio --direct=1 --thread --norandommap --group_reporting --log_avg_msec=60000 --per_job_logs=0 --minimal --eta=always --name=RandRdPre --rw=randread --bs=4k --ss=iops_slope:0.1% --ss_dur=1800 --ss_ramp=3600 --write_iops_log=fioinst  --filename=/dev/sda --iodepth=32  --numjobs=1
and the fio instant log:
60000, 79617, 0, 0
120000, 79793, 0, 0
180000, 79661, 0, 0
240000, 79704, 0, 0
300000, 79483, 0, 0
360000, 79454, 0, 0
420001, 79556, 0, 0
480000, 79343, 0, 0
540000, 79049, 0, 0
600001, 78776, 0, 0
660000, 78029, 0, 0
720000, 78023, 0, 0
780001, 77554, 0, 0
840000, 77225, 0, 0
900000, 76945, 0, 0
960000, 67886, 0, 0
1020001, 65353, 0, 0
1080000, 64239, 0, 0
1140000, 62856, 0, 0
1200000, 61628, 0, 0
1260000, 60328, 0, 0
1320000, 58853, 0, 0
1380000, 72292, 0, 0
1440000, 79839, 0, 0
1500000, 79852, 0, 0
1560000, 79879, 0, 0
I set the ss_ramp to 3600s and fio should run 1 hour before start steady state detection but it only run 1560 seconds  including the steady state detection phase( I set the ss_dur as 1800s)

Here is another set of fio options
fio --ioengine=libaio --direct=1 --thread --norandommap --group_reporting --log_avg_msec=60000 --per_job_logs=0 --minimal --eta=always --name=RandRdPre --rw=randread --bs=4k --ss=iops:0.1% --ss_dur=3600 --ss_ramp=7200 --write_iops_log=fioinst  --filename=/dev/sda --iodepth=32  --numjobs=1
And the fio log:
60000, 81947, 0, 0
120000, 81955, 0, 0
180000, 81746, 0, 0
240000, 81715, 0, 0
300000, 81783, 0, 0
360000, 81891, 0, 0
420000, 81954, 0, 0
480000, 81855, 0, 0
540000, 81217, 0, 0
600000, 81226, 0, 0
[...]

PR #690, opened 2018-09-21
docs: clarify usage of steadystate detection feature

vincentkfu:
Jens, here is the documentation update requested in #687


axboe:
Thanks Vincent!

PR #693, opened 2018-09-24
iolog: Fix problem with setup() not invoked when read_iolog is used.

aclamk:


aclamk:
Overrides #688.

axboe:
Pulled, thanks.

PR #697, opened 2018-09-27
zbd: Fix zbd_zone_nr()

bvanassche:
Instead of returning sizeof(struct fio_zone_info) * (zone number),
return the zone number.


axboe:
Oops, pulled thanks Bart.

PR #698, opened 2018-09-28
Restore the check_swd() function

bvanassche:
This patch series restores the check_swd() function in such a way that the compiler can verify the syntactical correctness of the code. A few patches that improve the robustness of the test scripts have been included too.


axboe:
Merged manually

PR #699, opened 2018-10-01
engines/rados: Now rados engine uses fio provided file names.

aclamk:
Instead of internally generated ones.
               This is necessary when using multiple, consecutive --read-iolog.
               Added check if --pool parameter is set.

### DIFF
--- a/engines/rados.c
+++ b/engines/rados.c
@@ -21,8 +21,6 @@ struct fio_rados_iou {
 struct rados_data {
 	rados_t cluster;
 	rados_ioctx_t io_ctx;
-	char **objects;
-	size_t object_count;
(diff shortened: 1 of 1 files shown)


aclamk:
@ifed01 Igor, could you please take a look on that? Have internally generated names been somehow important for rados?

sitsofe:
Just out of interest will this mean new objects will be made if a user upgrades from an older fio without this?

axboe:
I pulled this to apply manually. Change looks fine to me, but please use proper formatting for the commit. Lines should be 72 chars wide, at most, with the first line being a "title", then a blank line, then the commit message.

PR #700, opened 2018-10-01
serialize_overlap with io_submit_mode=offload

vincentkfu:

We are interested in running workloads that avoid overlap among in-flight IOs across different jobs. Based on the mailing list discussion from a few months ago, we implemented this using offload submission mode.

### DIFF
--- a/init.c
+++ b/init.c
@@ -741,23 +741,6 @@ static int fixup_options(struct thread_data *td)
 	if (o->iodepth_batch_complete_min > o->iodepth_batch_complete_max)
 		o->iodepth_batch_complete_max = o->iodepth_batch_complete_min;
 
-	/*
-	 * There's no need to check for in-flight overlapping IOs if the job
-	 * isn't changing data or the maximum iodepth is guaranteed to be 1
-	 */
-	if (o->serialize_overlap && !(td->flags & TD_F_READ_IOLOG) &&
-	    (!(td_write(td) || td_trim(td)) || o->iodepth == 1))
-		o->serialize_overlap = 0;
-	/*
-	 * Currently can't check for overlaps in offload mode
-	 */
-	if (o->serialize_overlap && o->io_submit_mode == IO_MODE_OFFLOAD) {
-		log_err("fio: checking for in-flight overlaps when the "
-			"io_submit_mode is offload is not supported\n");
-		o->serialize_overlap = 0;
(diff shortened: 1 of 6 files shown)


init.c
```diff
@@ -741,23 +741,6 @@ static int fixup_options(struct thread_data *td)
 	if (o->iodepth_batch_complete_min > o->iodepth_batch_complete_max)
 		o->iodepth_batch_complete_max = o->iodepth_batch_complete_min;
 
-	/*
-	 * There's no need to check for in-flight overlapping IOs if the job
-	 * isn't changing data or the maximum iodepth is guaranteed to be 1
-	 */
-	if (o->serialize_overlap && !(td->flags & TD_F_READ_IOLOG) &&
```
axboe: This should be a separate change.

workqueue.c
```diff
@@ -20,6 +21,8 @@ enum {
 	SW_F_ERROR	= 1 << 4,
 };
 
+static pthread_mutex_t serialize_lock;
```
axboe: This should be per workqueue struct, not a global entity. I'm also suspecting that this type of alloc is why it doesn't work for processes, just threads?

workqueue.c
```diff
@@ -113,6 +116,79 @@ void workqueue_enqueue(struct workqueue *wq, struct workqueue_work *work)
...
+	struct workqueue *tdwq;
+	int i, j;
+	unsigned long long x1, x2, y1, y2;
+
+	x1 = io_u->offset;
+	x2 = io_u->offset + io_u->buflen;
+
+	pthread_mutex_lock(&serialize_lock);
+
+	for_each_td(td, i) {
```
axboe: This seems pretty nutty... Why are you iterating every td? Doesn't this for every IO is going to be super slow, depending on what job is being run. You're looping each work item, in each worker, in each td - for every IO.

axboe:
I added some review comments, but I can address the overall design as well. Why isn't this done by having a shared io_u queue for the threads+workers that we care about? If you have a job with X threads using offload, then each thread should have a pending list that you can check for overlap in. There should be no need to iterate threads, workers, or anything like that. Think about the big O runtime of doing what you are doing right now.

vincentkfu:
Ok. That sounds like a better approach. I will close this pull request and give it a go with a shared io_u_queue.

PR #702, opened 2018-10-05
Check if `MPOL_LOCAL` declared before defining

joaomlneto:
`MPOL_LOCAL` will be defined in `numaif.h` in the future.

Also, `MPOL_MAX` will be changed to match the kernel values.


axboe:
Pulled, thanks.

PR #705, opened 2018-10-15
Enable --serialize-overlap when --io_submit_mode=offload

vincentkfu:
Jens, please consider this follow-up to #700 

We used blktrace to collect data from runs with 4K r/w jobs combined with a 1G trim and found no overlapping IOs submitted.

Here are responses to your earlier comments and changes from the previous version:

- moved overlap check from main job thread to submit worker threads
- use original serialize_overlap check function to compare each io_u with io_u_all queues for each running thread of interest
- moved init changes to a separate patch
- there needs to be a single overlap check mutex for all threads (instead of one per workqueue) to prevent two threads with overlapping IOs from finding that the coast is clear and submitting IOs that overlap
- I could not think of a way to avoid iterating over threads. I could create an array of *io_u_queue and have relevant threads insert/remove their io_u_all as they start/stop running, but each check would still have to iterate over the entire array of size thread_number.
- added a patch to remove an impossible branch from rate-submit.c

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -2339,8 +2339,13 @@ I/O depth
 	``serialize_overlap`` tells fio to avoid provoking this behavior by explicitly
 	serializing in-flight I/Os that have a non-zero overlap. Note that setting
 	this option can reduce both performance and the :option:`iodepth` achieved.
-	Additionally this option does not work when :option:`io_submit_mode` is set to
-	offload. Default: false.
+
+	This option only applies to I/Os issued for a single job except when it is
+	enabled along with :option:`io_submit_mode`=offload. In offload mode, fio
+	will check for overlap among all I/Os submitted by offload jobs with :option:`serialize_overlap`
+	enabled. Threads must be used for all such jobs.
+
+	Default: false.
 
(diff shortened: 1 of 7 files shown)


vincentkfu:
Thanks, Jens. I've fixed the two issues you mentioned.

axboe:
I pulled it, but the locking logic in check_overlap() is weird. Why isn't it just:

```
pthread_mutex_lock(&overlap_check);
for_each_td(td, i) {
    ...
    overlap = in_flight_overlap(&td->io_u_all, io_u);
    if (overlap)
        break;
}
pthread_mutex_unlock(&overlap_check);
```

Let's say all threads are currently hitting the continue condition, you'll have a problem.

axboe:
And shouldn't we return ```overlap``` from check_overlap(), and handle the overlap if it occurs? What am I missing here?

axboe:
It's a little convoluted, it doesn't follow any of the usual patterns for things like this. I think it at least needs a comment in both spots that touch overlap_lock.

vincentkfu:
Ok. I will send you something tomorrow.

PR #706, opened 2018-10-16
add rsp. time samples as column 2, use meaningful pctile names

bengland2:
This change was triggered by integration with [pbench](https://github.com/distributed-system-analysis/pbench) but should be useful to others.  This calculation does not work for a small enough sample count.   The output format change is not backward-compatible change, if this is a problem let me know, but I think this should not be a big issue at this point where not many applications are parsing output of this program. Output now looks like this:
```
msec-since-start, samples, min, median, 90.0, 95.0, 99.0, max, 
       0,        0, , , , , , , 
    2000,      392, 8.0, 11.9214545455, 15.0646518519, 292.2496, 696.89344, 761.856
    4000,      399, 6.976, 10.9718158936, 14.0618981662, 15.0542178183, 17.0315329838, 761.856
    6000,      792, 6.976, 11.9043493415, 15.0058790567, 16.0169657658, 499.957657343, 761.856
    8000,      801, 2.0, 10.9904056226, 14.9847229429, 15.0715337336, 17.0011426747, 761.856
```

### DIFF
--- a/tools/hist/fio-histo-log-pctiles.py
+++ b/tools/hist/fio-histo-log-pctiles.py
@@ -272,6 +272,13 @@ def add_to_histo_from( target, source ):
     for b in range(0, len(source)):
         target[b] += source[b]
 
+
+# calculate total samples in the histogram buckets
+
+def get_samples(buckets):
+    return reduce( lambda x,y: x + y, buckets)
+
+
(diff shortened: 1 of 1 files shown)


axboe:
Pulled, thanks.

Issue #709, opened 2018-11-05
Not include IO offset and length in the verify failure trace


fengc5:
If verify_interval is not equal to the block size, FIO would verify the data one segment by one segment according to the verify_interval.  The offset and length in the verify failure trace are the offset and length of the corrupted segment, not the offset and block size of the IO.

On block storage system, we need the offset and length of the IO to investigate data corruption issue. For me, I usually search the offset of the IO in the storage system trace to figure out what happened in the system.

So, I recommend that add the offset and length of the IO to the verify failure trace.

fengc5:
@sitsofe I have changed the source code for this enhancement. Can I push my change to this repository?

When I tried to push my change, I got the following error message. Is it due to I have no permission or due to I have wrong git config?
fatal: unable to access 'https://github.com/axboe/fio.git/': The requested URL returned error: 403

sitsofe:
@fengc5 You will have to issue a pull request by forking a version of the repo into your github account, pushing your changes to your fork then issuing a pull request from that repo against the main fio repo :-)

fengc5:
@sitsofe Thank you very much for your help. I issued a pull request.

sitsofe:
@fengc5 forgot to say - thanks for submitting this!

PR #710, opened 2018-11-06
For enhancement issue #709:"Include length/offset of requested block when verify_interval is not equal to block size"

fengc5:
If verify_interval is not equal to the block size, FIO would verify the data one segment by one segment according to the verify_interval. The offset and length in the verify failure trace are the offset and length of the corrupted segment, not the offset and block size of the IO.

On block storage system, we need the offset and length of the IO to investigate data corruption issue. For me, I usually search the offset of the IO in the storage system trace to figure out what happened in the system.

So, I recommend that add the offset and length of the IO to the verify failure trace.

### DIFF
--- a/verify.c
+++ b/verify.c
@@ -345,8 +345,10 @@ static void log_verify_failure(struct verify_header *hdr, struct vcont *vc)
 
 	offset = vc->io_u->offset;
 	offset += vc->hdr_num * hdr->len;
-	log_err("%.8s: verify failed at file %s offset %llu, length %u\n",
-			vc->name, vc->io_u->file->file_name, offset, hdr->len);
+	log_err("%.8s: verify failed at file %s offset %llu, length %u"
+			" (requested block: offset=%llu, length=%llu)\n",
+			vc->name, vc->io_u->file->file_name, offset, hdr->len,
+			vc->io_u->offset, vc->io_u->buflen);
 
 	if (vc->good_crc && vc->bad_crc) {
 		log_err("       Expected CRC: ");
@@ -865,9 +867,11 @@ static int verify_header(struct io_u *io_u, struct thread_data *td,
 	return 0;
 
 err:
-	log_err(" at file %s offset %llu, length %u\n",
+	log_err(" at file %s offset %llu, length %u"
+		" (requested block: offset=%llu, length=%llu)\n",
 		io_u->file->file_name,
-		io_u->offset + hdr_num * hdr_len, hdr_len);
+		io_u->offset + hdr_num * hdr_len, hdr_len,
+		io_u->offset, io_u->buflen);
 
 	if (td->o.verify_dump)
 		dump_buf(p, hdr_len, io_u->offset + hdr_num * hdr_len,


fengc5:
**Test was done. jobs are as below**.
`[global]
thread=1
create_on_open=1
numjobs=1
ioengine=libaio
size=10k
direct=1
filename=/home/fio_test/write_file.txt

[write_correct_data]
rw=write
bs=2k
do_verify=1
verify=crc32c
verify_interval=1k
verify_dump=1
io_size=4k

[corrupt_header]
wait_for=write_correct_data
rw=write
bs=1k
io_size=1k
offset=1k

[read_and_verify_corrupted_header]
wait_for=corrupt_header
rw=read
bs=2k
do_verify=1
verify=crc32c
verify_interval=1k
verify_dump=1
io_size=2k

[corrupt_data]
wait_for=read_and_verify_corrupted_header
rw=write
bs=512
io_size=512
offset=3584

[read_and_verify_corrupted_data]
wait_for=corrupt_data
rw=read
bs=2k
do_verify=1
verify=crc32c
verify_interval=1k
verify_dump=1
io_size=2k
offset=2k`

**Test output are as below.**
verify: bad magic header 7178, wanted acca at file /home/fio_test/write_file.txt offset 1024, length 1024, **requested block(offset 0, length 2048)**
       hdr_fail data dumped as write_file.txt.1024.hdr_fail

crc32c: verify failed at file /home/fio_test/write_file.txt offset 3072, length 1024, **requested block(offset 2048, length 2048)**
       Expected CRC: e2077042
       Received CRC: e21b711e
       received data dumped as write_file.txt.3072.received
       expected data dumped as write_file.txt.3072.expected

fengc5:
#

verify.c
```diff
@@ -345,8 +345,9 @@ static void log_verify_failure(struct verify_header *hdr, struct vcont *vc)
 
 	offset = vc->io_u->offset;
 	offset += vc->hdr_num * hdr->len;
-	log_err("%.8s: verify failed at file %s offset %llu, length %u\n",
-			vc->name, vc->io_u->file->file_name, offset, hdr->len);
+	log_err("%.8s: verify failed at file %s offset %llu, length %u, requested block(offset %llu, length %llu)\n",
```
sitsofe: I'd re-wrap these lines at 79 characters (or thereabouts) and put the whole requested block message within the brackets...

sitsofe:
(not sure what happened to my extra comments so reposting)

- Could you put
`Fixes: https://github.com/axboe/fio/issues/709`
  into the body of the commit message?
- Could you change the commit title to something like: `verify: add requested block to failure information` ?

But generally this change looks fine!

verify.c
fengc5: Thank you for your comments.  Changed

sitsofe:
Looks promising - can you squash these two commits together and only keep the latter commit message?

fengc5:
@sitsofe Thank you very much. Commits have been squashed.

axboe:
I've committed this manually. There was nothing wrong with the patch itself, but the commit left a lot to be desired. A good commit message should include WHY you are making this change. As it so happens, you already wrote all of that, it's the first entry in this bug issue. See the resulting commit:

http://git.kernel.dk/cgit/fio/commit/?id=46ff70e8795e91acb2ffc041e6c1fdb4e157dff4

First you have a line that is the commit title, you already had that, but it was mixed in with the Fixes line. Then follows an empty line, then the commit message body. This is where the justification for the patch goes. Finally you include data like Fixes, signed-off-by, etc.

PR #720, opened 2018-11-29
rand: fix compressible data ratio per segment

bari-lb:
Fio should write (100 - compress_percentage) * segemnt of random data
followed by compress_percentage * segemnt of compressible
data.

Currently, at each itereation fio fills (100 - compress_percentage) * segemnt
data, followed by (100 - compress_percentage) * segemnt of compressible
data until a the segment is filled.

This commit fixes a bug that was mistakenly
created by a code change in:
811ac503a6196f13da22565eb165c278988da4da

Signed-off-by: Bari Antebi <bari@lightbitslabs.com>

### DIFF
--- a/lib/rand.c
+++ b/lib/rand.c
@@ -166,6 +166,7 @@ void __fill_random_buf_percentage(unsigned long seed, void *buf,
 		if (!len)
 			break;
 		buf += this_len;
+		this_len = segment - this_len;
 
 		if (this_len > len)
 			this_len = len;


bari-lb:
To Whom It May Concern,

I've noticed a bug in fio while testing it. I expected to receive
output similar to the expected output below for: "hexdump -n 4096
/dev/nvme0n1" (considering the configuration file I've used  that may
be found below).

Expected output:

0000000 fdc6 a8a8 7190 0219 1fb8 9632 d9dd 1e64
                /* random data */
00004c0 d8a3 13fe aeec 0fb6 5b14 162e 0000 0000
00004d0 0000 0000 0000 0000 0000 0000 0000 0000
*
0001000


However, the output I've received  contained data after address 4cc
(which should have been filled with zeros until 1000 as far as I
understand, but as you can see 99a contains data.


0000000 fdc6 a8a8 7190 0219 1fb8 9632 d9dd 1e64
/* random data */
00004c0 d8a3 13fe aeec 0fb6 5b14 162e 0000 0000
00004d0 0000 0000 0000 0000 0000 0000 0000 0000
*
0000990 0000 0000 0000 0000 fdc6 a8a8 7190 0219

Config file:

[global]
group_reporting=1
filename=/dev/nvme0n1
name=task_nvme0n1
rw=write
bs=4096
numjobs=1
iodepth=32
buffer_compress_chunk=4096
buffer_compress_percentage=70
cpus_allowed=0-8
cpus_allowed_policy=split
direct=1
ioengine=libaio
loops=1
refill_buffers=0

[job empty]
size=4096

The version I ran is  fio-3.11, but it seems the bug wasn't fixed in
the new version from a look at the code.

I solved the issue locally & would like to send a pr for the fix.
The problem in the code maybe found in "lib/rand.c", at the function
"__fill_random_buf_percentage".
After fio finishes to write the random part, it should write the
remainder of "((unsigned long long)segment * (100 - percentage)) /
100",
but currently it writes "this_len" zeros which seems wrong.

axboe:
Thanks, I've merged this one manually. Your description in here is so good, it should be in the commit message, so I included that.

Additionally, there's a format for specifying if this commit fixes an issue introduced by some other commit. For this one, I added:

Fixes: 811ac503a619 ("Compression fixes")

PR #758, opened 2019-04-01
arch: fix build breakage on armv6 again

neheb:
6K was missing from the defines.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


axboe:
Pulled, thanks.

PR #762, opened 2019-04-10
fio: add libiscsi engine

kylezh:
Adding a new engine that access iscsi lun with libiscsi.

Signed-off-by: Kyle Zhang <kyle@smartx.com>

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1991,6 +1991,8 @@ I/O engine
 			Asynchronous read and write using DDN's Infinite Memory Engine (IME).
 			This engine will try to stack as much IOs as possible by creating
 			requests for IME. FIO will then decide when to commit these requests.
+		**libiscsi**
+			Read and write iscsi lun with libiscsi.
 
 I/O engine specific parameters
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,12 @@ ifdef CONFIG_LIBHDFS
   SOURCE += engines/libhdfs.c
 endif
 
+ifdef CONFIG_LIBISCSI
+  CFLAGS += $(LIBISCSI_CFLAGS)
+  LIBS += $(LIBISCSI_LIBS)
+  SOURCE += engines/libiscsi.c
+endif
+
 ifdef CONFIG_64BIT
   CFLAGS += -DBITS_PER_LONG=64
 endif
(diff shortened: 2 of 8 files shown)


sitsofe:
You beat me to the punch :-) Any chance you can look at https://github.com/sitsofe/fio/tree/libiscsi and see if there's anything there you can take/steal?

kylezh:
@sitsofe haha~
I haven't know that there is already an implementation of fio iscsi engine. So I just wrote a quite simple one by myself.
Why not make a PR from https://github.com/sitsofe/fio/tree/libiscsi ? Seems that your branch has more features than mine.

axboe:
This looks pretty clean, good job. Any chance you can include a sample job file as well? That's usually pretty helpful for new users. Would also be great if you could add the same change you did to HOWTO to fio.1, unfortunately they aren't fully synced yet.

kylezh:
@axboe I'm updating the patch. And I will also update the configuration manner that uses a formal iSCSI URL format as `filename`, instead of configuring by three separate options.

kylezh:
@axboe 

I have found some problems with making filename as iSCSI URL.

The problem is, URL contains `/` which is assumed as an OS path separator. fio will try to create a directory when parsing filename that would fail of course.

I was trying to add a new flag to ioengine to indicate that creating a directory is not necessary. However, because ioengine is not loaded when the filename was parsing, so the flag makes no effect. I was able to add a fio option to do that. It works, but I think it would be better to be a ioengine flag, instead of a fio option.

Any suggestion for solving this problem?

kylezh:
Hi @axboe, I think I found a way of passing iscsi url by filename. This is much simpler and can support more features without adding new options. You could find an example file in examples/libiscsi.fio .

axboe:
Not sure I like having that logic in the open path, it should remain in the setup path. What happens if a file is opened multiple times during the run of the job?

kylezh:
How about move create_work_dirs() in seutp_files()?

BTW, I'm not quite sure the usage of TD_F_DIRS_CREATED. Why not call create_work_dirs for each file?

axboe:
You are right, for some reason I totally missed the TD_F_DIRS_CREATED flag. With that in mind, I think the change is fine.

axboe:
Actually, the problem is that it's a per-td thing, not per file. So putting it in generic_open_file() doesn't make a whole lot of sense. setup_files() would definitely be a better spot.

kylezh:
I have moved create_work_dirs() to seutp_files() and rebased to master.

axboe:
Merged, thanks.

PR #767, opened 2019-04-17
verify: fix the width of rand_seed

mingnus:
The long data type is 32-bit on LLP64 platforms


mingnus:
Oops, I sent wrong commit. I'll open another one later.

Issue #771, opened 2019-04-18
Replay of blktrace containing Trim operations causes assert


shimrot:
Replay of blktrace containing trim operations gets assert failure: 
```
> fio: fio.h:552: fio_ro_check: Assertion `!(io_u->ddir == DDIR_WRITE && !((td)->o.td_ddir & TD_DDIR_WRITE)) && !(io_u->ddir == DDIR_TRIM && !((td)->o.td_ddir & TD_DDIR_TRIM))' f 
```

To reproduce: 

1. Start blktrace:
    ```
    > blktrace -d /dev/nvme0n1
    ```
   
1. Run fio with some trim operations: 
    ```
    > fio --output="fio.out" --direct=1 --filename="/dev/nvme0n1" --ioengine=libaio --norandommap=1 --random_generator=tausworthe64 --randrepeat=0 --runtime=1m --time_based=1 --name="Measure_RW_256KB_QD1" --bs=256KB1 --rate=64000KB --rw=randwrite --name="Measure_Trim_256MB_QD1" --bs=256MB --rate=64000KB --rw=randtrim
     ```    
1. Create blkparse bin file: 
    ```
    > blkparse nvme0n1 -o /dev/null -d nvme0n1.blkparse.bin
    ```

1. Replay gives fio_ro_check assertion: 
    ```
    > fio  --direct=1 --filename="/dev/nvme0n1" --replay_redirect="/dev/nvme0n1" --ioengine=libaio --randrepeat=0 --name="Replay" --ramp_time=0s --read_iolog=nvme0n1.blkparse.bin Replay: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=libaio, iodepth=1
    fio-3.13-27-gef32d
    Starting 1 process
    fio: fio.h:552: fio_ro_check: Assertion `!(io_u->ddir == DDIR_WRITE && !((td)->o.td_ddir & TD_DDIR_WRITE)) && !(io_u->ddir == DDIR_TRIM && !((td)->o.td_ddir & TD_DDIR_TRIM))' failed.
    fio: pid=44785, got signal=6eta 18h:50m:20s]

    Replay: (groupid=0, jobs=1): err= 0: pid=44785: Thu Apr 18 13:19:07 2019
      lat (usec)   : 100=2.44%, 250=7.32%, 500=68.29%, 750=4.88%, 1000=2.44%
[...]

PR #777, opened 2019-05-03
fix trim blktrace replay

shimrot:
A fix for handling blktraces containing trims.  Please see github issue: https://github.com/axboe/fio/issues/771
Thanks

### DIFF
--- a/blktrace.c
+++ b/blktrace.c
@@ -544,16 +544,19 @@ bool load_blktrace(struct thread_data *td, const char *filename, int need_swap)
 	    !ios[DDIR_SYNC]) {
 		log_err("fio: found no ios in blktrace data\n");
 		return false;
-	} else if (ios[DDIR_READ] && !ios[DDIR_WRITE]) {
-		td->o.td_ddir = TD_DDIR_READ;
+	}
+
+	td->o.td_ddir = 0;
+	if (ios[DDIR_READ]) {
+		td->o.td_ddir |= TD_DDIR_READ;
 		td->o.max_bs[DDIR_READ] = rw_bs[DDIR_READ];
-	} else if (!ios[DDIR_READ] && ios[DDIR_WRITE]) {
-		td->o.td_ddir = TD_DDIR_WRITE;
+	}
+	if (ios[DDIR_WRITE]) {
+		td->o.td_ddir |= TD_DDIR_WRITE;
 		td->o.max_bs[DDIR_WRITE] = rw_bs[DDIR_WRITE];
-	} else {
-		td->o.td_ddir = TD_DDIR_RW;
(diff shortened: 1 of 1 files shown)


blktrace.c
```diff
@@ -544,16 +544,19 @@ bool load_blktrace(struct thread_data *td, const char *filename, int need_swap)
...
+	if (ios[DDIR_READ] && !ios[DDIR_WRITE]) {
+		td->o.td_ddir |= TD_DDIR_READ;
 		td->o.max_bs[DDIR_READ] = rw_bs[DDIR_READ];
+	}
+	if (ios[DDIR_WRITE]) {
+		td->o.td_ddir |= TD_DDIR_WRITE;
 		td->o.max_bs[DDIR_WRITE] = rw_bs[DDIR_WRITE];
+	}
+	if (ios[DDIR_TRIM]) {
+		td->o.td_ddir |= TD_DDIR_TRIM;
```
sitsofe: @shimrot Are you saying the data direction is something like WRITE and TRIM together?

sitsofe: In fact looking closer I'm not sure this change is correct. I think we are no longer handling the data direction read/write case...

shimrot:
Oh, I'm blind I think.
The first if was supposed to be 
    if (ios[DDIR_READ] )   // wo the write test

Does that work?

Im actually seeing other serious issues during playback.  I'm not how they could be related, but maybe.

If I fix the if, do you thinks its ok?  But suspect this change shouldnt be merged until my other issue is resolved - but that one is complicated. Ill update further when i get to a computer.
Thanks

sitsofe:
Fix looks sensible. Can you squash these into one?

shimrot:
(Sorry for the delay and forgetting to squash earlier).  While this change does appear to make sense to me, and also prevents the assert error, my blktrace replay containing trims still does not work correctly (Write operations are not performed correctly while trims included)  I will describe fully to the forum - have another debug experiment to perform yet.

shimrot:
I posted my further issue which is much more serious. [ https://www.spinics.net/lists/fio/msg07903.html]( https://www.spinics.net/lists/fio/msg07903.html)
Should merging this fix wait for resolution?  Or can we go ahead.  It's not possible to debug the referenced problem without this fix.
Thanks

sitsofe:
This fix looks sensible to me. @axboe I'd say this one is worth pulling...

khazhyk:
Is there any ideas on this & the related writes after trim issue? I'm running into the same assert failure while running blktrace containing trims, and assume would run into the writes after trim oddity, although I can test if it would help.

axboe:
Pulled, thanks. For next time, please follow the normal commit format for commit messages.

PR #779, opened 2019-05-06
io_uring: fix possible infinite loop

satoru-takeuchi:
When reaping some completed I/O before io_uring_enter(), it waits forever.

Signed-off-by: Satoru Takeuchi <sat@cybozu.co.jp>


axboe:
Looks good to me, thanks!

PR #785, opened 2019-05-31
configure: attempt to link against tcmalloc by default if available

dillaman:
librbd will run up to 20% faster under small IO workloads when fio is linked
against tcmalloc. By automatically linking against tcmalloc, it avoids
the need to use LD_PRELOAD to pull in a faster memory management toolset.

Signed-off-by: Jason Dillaman <dillaman@redhat.com>

### DIFF
--- a/configure
+++ b/configure
@@ -207,6 +207,8 @@ for opt do
   ;;
   --enable-libiscsi) libiscsi="yes"
   ;;
+  --disable-tcmalloc) disable_tcmalloc="yes"
+  ;;
   --help)
     show_help="yes"
     ;;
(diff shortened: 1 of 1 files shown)


axboe:
Looks fine, pulled, thanks.

sitsofe:
I think this should be opt in rather than opt out - if someone wants to use a non standard malloc it should be explicit...

axboe:
@sitsofe Normally I'd disagree, but I've yet to hear of cases where tcmalloc is worse. Most bigger shops actually run it by default.

axboe:
Plus tcmalloc isn't installed by default, it's a good guess to use it if it's actually there.

Issue #760, opened 2019-04-04
Impliment a no-cleanup config opt


brettmilford:
In testing a number scenarios it would be useful to disable the cleanup function some engines appear to run (e.g. rados). Is it possible to submit such a feature?

Glad to submit a PR myself with some guidance.

Thanks.

sitsofe:
@brettmilford If you can afford the time give it a go! It might take some back and for but most things can be worked out...

sitsofe:
Note" the whole "no-clean" idea may only be applicable to specific ioengines (e.g. I'm not quite sure how this idea works with libaio which just leaves the files it makes as is when it's done). If so, you may want to make code this feature as an additional ioengine option (see https://github.com/axboe/fio/blob/master/engines/rados.c#L66 for how a boolean was made and the lines leading to https://github.com/axboe/fio/blob/master/engines/rados.c#L317 for how it was retrieved).

sitsofe:
OK I found you've already started https://github.com/brettmilford/fio-rados-tools/blob/master/nocleanup-patch/rpmbuild/SOURCES/fio-3.7-nocleanup.patch . We probably don't want to `nop` it out like that and still leave the existing code... Maybe you could make rados respect unlink option if an additional option is specified (we may have to hide the new behaviour behind a flag to avoid catching out existing rados ioengine users)?

axboe:
This does sound like something that should just be covered with the ```unlink``` option. What specific parts of the cleanup/teardown do you want to avoid?

brettmilford:
@sitsofe Yeah that was a quick hack I put together so that I could keep working on a problem; I wouldn't expect it an implementation to work that way.

@axboe I'm specifically wanting to avoid deleting objects after a job run with the rados engine so that I can observe other parts of the system whilst at that state. I figured that, although the implementation would be different across different engines the concept/option would be the same.

I see `ioengines.h` appears to define a cleanup param?
https://github.com/axboe/fio/blob/master/ioengines.h#L36

And I've noticed a number of the engines (possibly all, I haven't checked) define a cleanup function e.g. https://github.com/axboe/fio/blob/master/engines/rados.c#L441
https://github.com/axboe/fio/blob/master/engines/sync.c#L428

However not all engines appear to call it, or at least call it in the same way as the rados engine does. Is there something I'm missing here? Is the cleanup function _suppose_ to clean up files/objects generated or is it for something else? It appears, at least, that the rados engine is using this for deleting objects.

Thanks for your feedback.

[...]

PR #790, opened 2019-06-11
implement cleanup option for rados engine

brettmilford:
implements a cleanup option for the rados engine.
This is to resolve issue #760

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -2157,6 +2157,10 @@ with the caveat that when used on the command line, they must come after the
         Poll store instead of waiting for completion. Usually this provides better
         throughput at cost of higher(up to 100%) CPU utilization.
 
+.. option:: cleanup=bool : [rados]
+
+        Clean up rados objects after run.
+
 .. option:: skip_bad=bool : [mtd]
 
 	Skip operations against known bad blocks.
--- a/engines/rados.c
+++ b/engines/rados.c
@@ -32,6 +32,7 @@ struct rados_options {
 	char *pool_name;
 	char *client_name;
 	int busy_poll;
+	int cleanup;
 };
 
 static struct fio_option options[] = {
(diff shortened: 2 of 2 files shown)


sitsofe:
@brettmilford Could you squash these commits?

sitsofe:
@brettmilford somehow a build breakage was triggered...

brettmilford:
Ah yeah. It appears to be in an unrelated/untouched piece of code; any ideas @sitsofe?

sitsofe:
@brettmilford Hmm, I can only see one call to it from `rbd_iter_events()`. Maybe it could be hidden inside a `#ifdef CONFIG_RBD_POLL` guard (or inlined directly at the sole call site)?

I've no idea why this wouldn't have triggered earlier though.

sitsofe:
@brettmilford I think this one got lost... did you want to try and finish it off?

brettmilford:
Hmm, looks like I chose the wrong commit to fork off. I think master at the time (971341ad) produces the same error - will try and rebase off a tagged commit.

sitsofe:
@brettmilford nice! Could you fix up the commit message to be in fio's usual style and add your signed off by?

sitsofe:
(I think that mainly means adding an rbd: prefix to the title)

sitsofe:
Oh and you might want to include a fixes https://github.com/axboe/fio/issues/760 line...

brettmilford:
@sitsofe `rbd:` or `rados:` - it should only change the rados engine.

sitsofe:
Let's go with rados.

sitsofe:
(@brettmilford Quick ping re last comment)

brettmilford:
@sitsofe thanks, not sure why I didn't get notified.
Um - I should probably default this to True to retain the expected behavior hey?

HOWTO
```diff
@@ -2157,6 +2157,10 @@ with the caveat that when used on the command line, they must come after the
         Poll store instead of waiting for completion. Usually this provides better
         throughput at cost of higher(up to 100%) CPU utilization.
 
+.. option:: cleanup=bool : [rados]
+
+        Clean up rados objects after run.
+
```
sitsofe: So just state the default and don't forget to update the fio.1 man page too...

sitsofe:
> Um - I should probably default this to True to retain the expected behavior hey?

You're entirely correct - my bad!

axboe:
Did this all get resolved?

brettmilford:
Hi @sitsofe @axboe, sorry for the delayed response, I needed a chance to test this properly. This patch doesn't appear to work as expected, but also I can't seem to get the same behavior in the version compiled from source as the version packaged in fedora 21 (that is the cleanup function doesn't work at all when I compile from source). In light of this, and a lack of time to delve deeper into why this is, I'm going to close of this PR, Thanks for your assistance.

Issue #794, opened 2019-07-01
filename argument seems limited to ~4k characters


powernap:
In trying to work around issue #631 I discovered what seems like a ~4k character limit on the filename argument. I see this on fio 3.5 and 3.14 on FreeBSD 11.2 and on fio 3.12 on Ubuntu 19.04.

```
# ~/fio-3.14-11.2 filename_ok.fio 
create: (g=0): rw=write, bs=(R) 128KiB-128KiB, (W) 128KiB-128KiB, (T) 128KiB-128KiB, ioengine=posixaio, iodepth=128
fio-3.14
Starting 1 process

create: (groupid=0, jobs=1): err= 0: pid=64798: Mon Jul  1 11:17:28 2019
  write: IOPS=29.4k, BW=3678MiB/s (3857MB/s)(1207MiB/328msec)
    slat (nsec): min=662, max=473876, avg=2666.92, stdev=14338.17
    clat (usec): min=314, max=44140, avg=4001.28, stdev=2581.62
     lat (usec): min=318, max=44173, avg=4003.95, stdev=2581.59
    clat percentiles (usec):
     |  1.00th=[ 2573],  5.00th=[ 2868], 10.00th=[ 2999], 20.00th=[ 3130],
     | 30.00th=[ 3294], 40.00th=[ 3458], 50.00th=[ 3654], 60.00th=[ 3851],
     | 70.00th=[ 4047], 80.00th=[ 4293], 90.00th=[ 4883], 95.00th=[ 5342],
     | 99.00th=[10552], 99.50th=[26084], 99.90th=[40109], 99.95th=[42206],
     | 99.99th=[44303]
  lat (usec)   : 500=0.06%, 750=0.07%, 1000=0.07%
  lat (msec)   : 2=0.40%, 4=67.73%, 10=30.64%, 20=0.34%, 50=0.68%
  cpu          : usr=12.54%, sys=85.93%, ctx=137, majf=0, minf=2
  IO depths    : 1=0.1%, 2=0.1%, 4=0.1%, 8=0.2%, 16=0.3%, 32=0.7%, >=64=98.7%
     submit    : 0=0.0%, 4=100.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
     complete  : 0=0.0%, 4=91.5%, 8=7.4%, 16=1.0%, 32=0.1%, 64=0.1%, >=64=0.1%
[...]

PR #795, opened 2019-07-01
Increase maximum length of line in jobs file to 8192

powernap:
Per issue #794, double the maximum length of a line in a fio job file to 8192 bytes. This changes the hard-coded malloc and fgets in init.c to use OPT_LEN_MAX instead of 4096 and increases OPT_LEN_MAX to 8192. I've also documented the maximum length of a line in the man page, since 8192 is a somewhat arbitrary increase and someone may need it increased further later on.

### DIFF
--- a/fio.1
+++ b/fio.1
@@ -201,6 +201,8 @@ argument, \fB\-\-cmdhelp\fR will detail the given \fIcommand\fR.
 See the `examples/' directory for inspiration on how to write job files. Note
 the copyright and license requirements currently apply to
 `examples/' files.
+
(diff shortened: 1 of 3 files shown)


axboe:
Looks fine to me, but can you please follow the standard commit format? Basically it's:

1) First line should be a title for the commit
2) Then an empty line
3) Then a commit message body
4) Then an empty line
5) Last line should be "Signed-off-by: " followed by your name and email.

powernap:
Sorry about that. Force-pushed amended commit message.

axboe:
Great, pulled, thanks.

PR #802, opened 2019-07-19
libiscsi: continue working when meets EINTR or EAGAIN

kylezh:
When poll meets EINTR, it should continue working rather than return
error.

Signed-off-by: Kyle Zhang <kyle@smartx.com>

### DIFF
--- a/engines/libiscsi.c
+++ b/engines/libiscsi.c
@@ -351,6 +351,9 @@ static int fio_iscsi_getevents(struct thread_data *td, unsigned int min,
 
 		ret = poll(iscsi_info->pfds, iscsi_info->nr_luns, -1);
 		if (ret < 0) {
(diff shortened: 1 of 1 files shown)


engines/libiscsi.c
```diff
@@ -351,6 +351,9 @@ static int fio_iscsi_getevents(struct thread_data *td, unsigned int min,
 
 		ret = poll(iscsi_info->pfds, iscsi_info->nr_luns, -1);
 		if (ret < 0) {
+			if (errno == EINTR) {
```
sitsofe: What about `EAGAIN` on other OSes?

kylezh: I missed this case. I will update soon. Thanks!

axboe:
Pulled, thanks.

PR #808, opened 2019-07-31
smalloc garbage collection

vincentkfu:
More and longer jobs in client/server mode with --status-interval=1 increase shared memory usage and may trigger an erroneous smalloc OOM error.

@axboe, please consider the patches in this pull request to resolve the issue.

The patches include a modification of t/stest.c to expose the problem and a fix for the problem.

### DIFF
--- a/smalloc.c
+++ b/smalloc.c
@@ -48,6 +48,12 @@ struct block_hdr {
 #endif
 };
 
+/*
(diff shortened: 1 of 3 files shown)


axboe:
LGTM, nice fix! Pulled.

PR #818, opened 2019-09-13
engines/io_uring: Fully clear out previous SQE contents.

anarazel:
Without this change SQEs can contain data set in previous
submissions. E.g. a WRITE following an fdatasync would have still have
IORING_FSYNC_DATASYNC set in sync_flags, which shares storage with the
WRITE's rw_flags. Which was not reset, causing all writes to be
synchronous. Similarly, an fsync following a READ/WRITE would not
reset off/addr/len, which causes errors, because the kernel's
io_prep_fsync returns an error if e.g. addr is not 0.

While this could also be fixed by resetting only the unused fields in
the respective branches, it seems less failure prone to start with a
zeroed out sqe.

### DIFF
--- a/engines/io_uring.c
+++ b/engines/io_uring.c
@@ -152,15 +152,16 @@ static int fio_ioring_prep(struct thread_data *td, struct io_u *io_u)
 	struct io_uring_sqe *sqe;
 
 	sqe = &ld->sqes[io_u->index];
+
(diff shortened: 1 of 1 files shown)


axboe:
Seems saner than attempting to be clever with clearing fields. Thanks, pulled.

Issue #834, opened 2019-10-05
Drop macOS support for fdatasync


sitsofe:
https://github.com/axboe/fio/blob/fio-3.16/os/os-mac.h#L101 is apparently incorrect because it is actually calling a kernel stub whose arguments are not what we expect them to be (see https://github.com/gbrault/picoc/issues/145#issuecomment-89734655 and https://gitlab.freedesktop.org/xdg/shared-mime-info/issues/7 ). We would need to make the `configure` test more robust (e.g. by adding `-Werror-implicit-function-declaration` around https://github.com/axboe/fio/blob/fio-3.16/configure#L91 and https://github.com/axboe/fio/blob/fio-3.16/configure#L98 ) as technically there is a symbol to link against but it's not defined in public header.

We will likely need to update the docs (e.g. https://github.com/axboe/fio/blob/fio-3.16/HOWTO#L1279 , https://github.com/axboe/fio/blob/fio-3.16/fio.1#L1051 ) too.

PR #835, opened 2019-10-06
configure: stop enabling fdatasync on OSX

sitsofe:
Change configure compile probes to fail on implicit declarations. This
(correctly) stops us from enabling fdatasync on OSX which was
problematic because we were actually calling a syscall stub with
arguments different to what was declared in its prototype (see
https://github.com/gbrault/picoc/issues/145#issuecomment-89734655 and
https://gitlab.freedesktop.org/xdg/shared-mime-info/issues/7 ).

Fixes: https://github.com/axboe/fio/issues/834 ("Drop macOS support for
fdatasync")

@axboe This patch is simplistic and basically adds erroring on `implicit-function-declaration` warnings to the two probe compilation functions. It might be better to just error on all warnings during compilation probes (if so let me know and I'll recreate the patch). I couldn't find a situation where compilation probes weren't clean against basic warnings but I haven't built with all possible libraries/options/platforms...

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1279,7 +1279,7 @@ I/O type
 .. option:: fdatasync=int
 
 	Like :option:`fsync` but uses :manpage:`fdatasync(2)` to only sync data and
-	not metadata blocks.  In Windows, FreeBSD, and DragonFlyBSD there is no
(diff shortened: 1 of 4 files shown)


axboe:
Christ, what a joke OSX is. Thanks, pulled.

Issue #831, opened 2019-10-02
Inflated throughput numbers with fsync_on_close


llamafilm:
Doing a sequential write of a single file with `fsync_on_close`, the bandwidth number reported at the bottom is way too high.  Here it shows 2576MiB/s.  If I do the same test with `end_fsync` instead, the number is accurate 731MiB/s.  Since I am writing a single file, I thing these numbers should be the same.

Sitsofe found that the runtime may be taken too soon
(https://github.com/axboe/fio/blob/fio-3.15/backend.c#L1878 )

I don't have the chops to fix this myself but wanted to raise the issue.  This is from fio 3.1 but I saw the same thing on FreeBSD with version 3.5.

```
$ time fio --name=larry --rw=write --bs=1M --size=10G --directory=/mnt/raid --fsync_on_close=1
larry: (g=0): rw=write, bs=(R) 1024KiB-1024KiB, (W) 1024KiB-1024KiB, (T) 1024KiB-1024KiB, ioengine=psync, iodepth=1
fio-3.1
Starting 1 process
Jobs: 1 (f=1): [f(1)][100.0%][r=0KiB/s,w=0KiB/s][r=0,w=0 IOPS][eta 00m:00s]      
larry: (groupid=0, jobs=1): err= 0: pid=49397: Wed Oct  2 16:43:40 2019
  write: IOPS=2576, BW=2576MiB/s (2701MB/s)(10.0GiB/3975msec)
    clat (usec): min=354, max=475, avg=373.87, stdev=10.59
     lat (usec): min=363, max=644, avg=387.39, stdev=11.89
    clat percentiles (usec):
     |  1.00th=[  359],  5.00th=[  363], 10.00th=[  363], 20.00th=[  363],
     | 30.00th=[  367], 40.00th=[  367], 50.00th=[  371], 60.00th=[  379],
     | 70.00th=[  379], 80.00th=[  383], 90.00th=[  388], 95.00th=[  392],
     | 99.00th=[  404], 99.50th=[  408], 99.90th=[  416], 99.95th=[  420],
     | 99.99th=[  465]
   bw (  MiB/s): min= 2517, max= 2655, per=100.00%, avg=2587.07, stdev=56.34, samples=7
   iops        : min= 2517, max= 2655, avg=2587.00, stdev=56.29, samples=7
  lat (usec)   : 500=100.00%
  cpu          : usr=4.10%, sys=95.65%, ctx=3, majf=0, minf=562
[...]

PR #837, opened 2019-10-08
backend: fix final fsync behaviour

sitsofe:
Previously, fsync_on_close's "final" fsync was done after fio's
accounting had finished thus causing the bandwidth reported to be too
high (due to the reported time being too low). An example job that
demonstrates the issue (on machines with a few gigabytes of RAM and
non-fast disks) is the following:

./fio --gtod_reduce=1 --filename=fio.tmp --size=5G --bs=2M --rw=write \
 --stonewall --name=end_fsync --end_fsync=1 --name=fsync_on_close \
 --fsync_on_close=1
end_fsync: (g=0): rw=write, bs=(R) 2048KiB-2048KiB, (W) 2048KiB-2048KiB, (T) 2048KiB-2048KiB, ioengine=psync, iodepth=1
fsync_on_close: (g=1): rw=write, bs=(R) 2048KiB-2048KiB, (W) 2048KiB-2048KiB, (T) 2048KiB-2048KiB, ioengine=psync, iodepth=1
[...]
  WRITE: bw=381MiB/s (400MB/s), 381MiB/s-381MiB/s (400MB/s-400MB/s), io=5120MiB (5369MB), run=13424-13424msec

Run status group 1 (all jobs):
  WRITE: bw=1726MiB/s (1810MB/s), 1726MiB/s-1726MiB/s (1810MB/s-1810MB/s), io=5120MiB (5369MB), run=2966-2966msec

This patch fixes the issue by doing an fsync for fsync_on_close at the
same point as it does for end_fsync. Technically fsync_on_close will go
and do another fsync after this point too but this should be harmless as
there is no new write data.

This patch also fixes the assert seen with the following job

./fio --size=8k --io_limit=32k --filename=fio.tmp --rw=write \
 --end_fsync=1 --name=extended_end_fsync
[...]
fio: filesetup.c:1682: void get_file(struct fio_file *): Assertion `fio_file_open(f)' failed.

by preventing auto-close on last I/O when it is being submitted as a
final fsync (due to end_fsync, fsync_on_close).

Finally it also fixes the strange IO depths seen with when using end_fsync:

./fio --size=4k --filename=fio.tmp --rw=write --end_fsync=1 \
 --name=end_fsync_test
end_fsync_test: (g=0): rw=write, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
[...]
  IO depths    : 1=200.0%, 2=0.0%, 4=0.0%, 8=0.0%, 16=0.0%, 32=0.0%, >=64=0.0%

by no longer including such fsyncs in accounting.

Fixes: https://github.com/axboe/fio/issues/831

(This is an experimental attempt at a fix for #831 to get some feedback so I don't think it's necessarily ready for commiting)

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -281,6 +281,7 @@ static bool fio_io_sync(struct thread_data *td, struct fio_file *f)
 
 	io_u->ddir = DDIR_SYNC;
 	io_u->file = f;
+	io_u_set(td, io_u, IO_U_F_NO_FILE_PUT);
 
 	if (td_io_prep(td, io_u)) {
 		put_io_u(td, io_u);
@@ -314,7 +315,7 @@ requeue:
 
 static int fio_file_fsync(struct thread_data *td, struct fio_file *f)
 {
-	int ret;
+	int ret, ret2;
 
 	if (fio_file_open(f))
 		return fio_io_sync(td, f);
@@ -323,8 +324,10 @@ static int fio_file_fsync(struct thread_data *td, struct fio_file *f)
 		return 1;
 
 	ret = fio_io_sync(td, f);
-	td_io_close_file(td, f);
-	return ret;
+	ret2 = 0;
+	if (fio_file_open(f))
+		ret2 = td_io_close_file(td, f);
+	return (ret || ret2);
 }
 
 static inline void __update_ts_cache(struct thread_data *td)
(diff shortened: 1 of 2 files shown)


sitsofe:
@axboe so I take a belt and braces approach here and both stop the fund being allowed to close the file as the last I/O and then also check if the file has been closed after we do the fsync. If you're ok with this you can pull as-is but I'm curious as to the thoughts :-)

PR #843, opened 2019-10-21
Link to the system logging library on Android

kdrag0n:
Android replaces syslog calls with calls to the Android system logging library, `liblog`. Link to `liblog` when compiling for Android to fix the following undefined reference errors:

```
aarch64-linux-android-ld: log.o: in function `android_polyfill_vsyslog':
/usr/include/syslog.h:185: undefined reference to `__android_log_print'
aarch64-linux-android-ld: /usr/include/syslog.h:182: undefined reference to `__android_log_vprint'
```


axboe:
Pulled, thanks.

PR #846, opened 2019-10-22
enforce strtoll string parsing

brchrisman:
From strtoll(3):
  In particular, if *nptr is not '\0' but **endptr is '\0' on return, the entire string is valid.

- check implemented as close to above spec as possible.


brchrisman:
wrong comparison in first branch of conditional.. will fix and push again

PR #848, opened 2019-10-22
enforce strtoll string parsing

brchrisman:
From strtoll(3):
  In particular, if *nptr is not '\0' but **endptr is '\0' on return, the entire string is valid.

- check implemented as close to above spec as possible.


brchrisman:
That was my fear, that the number involved isn't being parsed prior to calling strtoll().
Fix would require positive assertion of allowed formats which doesn't look to have an infrastructure built for it in fio right now.

PR #856, opened 2019-11-03
Enable io_uring engine on powerpc arch

aphreet:


sitsofe:
@aphreet Minor comments but could you:
- Change the title to start with `powerpc: `
- (Extra minor) Could you move `#define ARCH_HAVE_IOURING` closer to the syscall numbers?

PR #857, opened 2019-11-03
gettime: use CLOCK_MONOTONIC when CLOCK_MONOTONIC_RAW is slow

sitsofe:
This one doesn't do all that much clever (e.g. I assume `gettimeofday()` will be fast and accurate enough to time multiple `clock_gettime()` calls) and it introduces a conditional into the clock path. Even so it seems to give numbers similar to using the cpu clock and I think the cost in accuracy is made up for by the lower overhead.

What do you think Jens?

### DIFF
--- a/gettime.c
+++ b/gettime.c
@@ -23,6 +23,7 @@ static unsigned int cycles_wrap;
 #endif
 #endif
 bool tsc_reliable = false;
+static bool cg_prefer_monotonic = false;
 
 struct tv_valid {
 	int warned;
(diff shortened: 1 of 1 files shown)


gettime.c
```diff
@@ -127,10 +128,60 @@ static void fio_init gtod_init(void)
...
+{
+        int i;
+        struct timeval start, end;
+        struct timespec ts;
+
+        gettimeofday(&start, NULL);
+        for (i = 0; i < rounds; ++i)
+                clock_gettime(clkid, &ts);
+        gettimeofday(&end, NULL);
+        return ((end.tv_sec * 1e6 + end.tv_usec) - \
```
axboe: use utime_since_now()?

gettime.c
```diff
@@ -127,10 +128,60 @@ static void fio_init gtod_init(void)
...
+	/*
+	 * On Linux kernels prior to 5.3 there are systems where
+	 * clock_gettime(CLOCK_MONOTONIC, ...) is implemented via vDSO but
+	 * clock_gettime(CLOCK_MONOTONIC_RAW, ...) is not (and falls back to a
+	 * "slow" syscall) so check if for a dramatic difference in call time.
+	 */
+	int bench_monotonic_raw, bench_monotonic;
+	const int rounds = 1000;
+
+        bench_monotonic = bench_clock_gettime(CLOCK_MONOTONIC, rounds);
```
axboe: Something funky going on with indentation/tabs here

sitsofe:
@axboe does this look better?

gettime.c
```diff
@@ -127,10 +128,58 @@ static void fio_init gtod_init(void)
...
+	return false;
+}
+#endif /* defined(CONFIG_CLOCK_MONOTONIC_RAW) && defined(CONFIG_CLOCK_MONOTONIC) */
+
 #ifdef CONFIG_CLOCK_GETTIME
 static int fill_clock_gettime(struct timespec *ts)
 {
-#if defined(CONFIG_CLOCK_MONOTONIC_RAW)
+#if defined(CONFIG_CLOCK_MONOTONIC_RAW) && defined(CONFIG_CLOCK_MONOTONIC_RAW)
+	if (cg_prefer_monotonic)
```
axboe: I'm not convinced we need to do this, but if we do, there's a trivial way to avoid this branch - just have the tester set the preferred clocksource, and don't do the check here.

sitsofe:
Let's close this one for now.

PR #859, opened 2019-11-06
Add fallocate=truncate option.

### LINKED ISSUE #833: Poor windowsaio performance due to synchronization of writes unless --overwrite 1 is given
I'm using fio to try to benchmark high-throughput, pipelined writes to networked storage on Windows, and I'm running into a slightly annoying issue - in order to actually get good performance, I have to pass `--overwrite=1` -- which has a couple of small problems:

1. It causes the entire file to be written twice (and the first time is rather slow), needlessly consuming time;
2. It means that the test is measuring something slightly different from what is intended, since the write operations do not induce allocation.

[On Windows, all appends get synchronized](https://support.microsoft.com/en-us/help/156932/asynchronous-disk-i-o-appears-as-synchronous-on-windows); this is true even when using unbuffered, overlapped IO. The way to get around this issue is to set the length of the file before issuing IO, generally by using the Win32 SetEndOfFile function. (Note the MS documentation I linked says to use SetFileValidData; I suspect this may be an error; using SetEndOfFile to create a sparse file seems to be the "right" way to do this, and this is what Windows Explorer does when copying a file.)

So, the behavior I want is for fio to extend the file sparsely before doing the write workload. I read `extend_file` (filesetup.c), and it does not appear to support this behavior right now - any time it truncates to extend the size of the file, it appears it will also write data to the entire file.

I'd like to craft a small PR to support this, but I figured I'd float the idea first since there are a couple different ways one might do it. My first thought was to just add a flag to the thread options, something like `--extend-only-truncates` (perhaps with a better name), which just changes the behavior of `extend_file()` to do the truncate but not the writes. On the other hand, maybe this could be something specific to the windowsaio engine, since that's the only place that I'm aware of where you'll see this behavior.

Any thoughts? Is there perhaps a better way to achieve this?

tripped:
Fixes #833. Provides the ability to initially "layout" the file
by ftruncating it to the desired size before performing IO. This
is mainly useful on Windows, which serializes all writes that
extend the size of a file. Using this option with a suitable
iodepth allows fio to emulate the behavior of Windows Explorer
file copy, which always truncates to the expected size before
issuing writes for performance reasons.

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1173,6 +1173,10 @@ I/O type
 			Pre-allocate via :manpage:`fallocate(2)` with
 			FALLOC_FL_KEEP_SIZE set.
 
+		**truncate**
+			Extend file to final size via :manpage:`ftruncate(2)`
+			instead of allocating.
+
 		**0**
 			Backward-compatible alias for **none**.
 
@@ -1182,7 +1186,15 @@ I/O type
 	May not be available on all supported platforms. **keep** is only available
 	on Linux. If using ZFS on Solaris this cannot be set to **posix**
 	because ZFS doesn't support pre-allocation. Default: **native** if any
-	pre-allocation methods are available, **none** if not.
+	pre-allocation methods except **truncate** are available, **none** if not.
+
+	Note that using **truncate** on Windows will interact surprisingly
+	with non-sequential write patterns. When writing to a file that has
+	been extended by setting the end-of-file information, Windows will
+	backfill the unwritten portion of the file up to that offset with
+	zeroes before issuing the new write. This means that a single small
+	write to the end of an extended file will stall until the entire
+	file has been filled with zeroes.
 
 .. option:: fadvise_hint=str
 
--- a/file.h
+++ b/file.h
@@ -67,6 +67,7 @@ enum fio_fallocate_mode {
 	FIO_FALLOCATE_POSIX	= 2,
 	FIO_FALLOCATE_KEEP_SIZE	= 3,
 	FIO_FALLOCATE_NATIVE	= 4,
+	FIO_FALLOCATE_TRUNCATE	= 5,
 };
 
 /*
--- a/filesetup.c
+++ b/filesetup.c
@@ -95,6 +95,18 @@ static void fallocate_file(struct thread_data *td, struct fio_file *f)
 		break;
 		}
 #endif /* CONFIG_LINUX_FALLOCATE */
+	case FIO_FALLOCATE_TRUNCATE: {
+		int r;
+
+		dprint(FD_FILE, "ftruncate file %s size %llu\n",
+				f->file_name,
+				(unsigned long long) f->real_file_size);
+		r = ftruncate(f->fd, f->real_file_size);
+		if (r != 0)
+			td_verror(td, errno, "ftruncate");
+
+		break;
+	}
 	default:
 		log_err("fio: unknown fallocate mode: %d\n", td->o.fallocate_mode);
 		assert(0);
--- a/fio.1
+++ b/fio.1
@@ -943,6 +943,10 @@ Pre-allocate via \fBposix_fallocate\fR\|(3).
 Pre-allocate via \fBfallocate\fR\|(2) with
 FALLOC_FL_KEEP_SIZE set.
 .TP
+.B truncate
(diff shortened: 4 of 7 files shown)


axboe:
Looks good to me. Maybe missing the matching HOWTO addition to the fio.1 change? I'll wait for @sitsofe to comment before pulling this in, as he's been a lot more involved with this issue.

sitsofe:
I agree with Jens - we're missing the documentation in the HOWTO and manpage (fio.1) and maybe prefix the commit title with `filesetup: ` .

sitsofe:
@tripped We'll likely also need a warning in the help about Windows backfill behaviour if you do random I/O....

tripped:
> I agree with Jens - we're missing the documentation in the HOWTO and manpage (fio.1)

Ah, is there something I missed? I did add (brief) documentation of the new option to both HOWTO and fio.1 in that commit.

> @tripped We'll likely also need a warning in the help about Windows backfill behaviour if you do random I/O....

Good point. I'll add something to the fallocate documentation. Anywhere else that should be called out?

tripped:
Oh, and of course there's something else important that I missed - even with this patch, it's still not possible to use this option on Windows, since it fails the `FIO_HAVE_ANY_FALLOCATE` check. That's what I get for writing and testing the change on Linux and not Windows. :-)

I suppose one option is to remove that check and not disable the option, since "none" and "truncate" should be valid options on all platforms. @axboe / @sitsofe any thoughts?

sitsofe:
> Ah, is there something I missed? I did add (brief) documentation of the new option to both HOWTO and fio.1 in that commit.

My bad - it's there.

> Good point. I'll add something to the fallocate documentation. Anywhere else that should be called out?

As the option isn't on by default I think in `fallocate` will be enough.

Re: `FIO_HAVE_ANY_FALLOCATE` well it's there to say what the default `fallocate` should be. As it stands I'm not sure this is safe to use as a default due to the backfill behaviour. Maybe the define should be `FIO_HAVE_DEFAULT_FALLOCATE`?

tripped:
> Re: `FIO_HAVE_ANY_FALLOCATE` well it's there to say what the default `fallocate` should be. As it stands I'm not sure this is safe to use as a default due to the backfill behaviour/sparsity. Maybe the define should be `FIO_HAVE_DEFAULT_FALLOCATE`?

Yeah, I agree this definitely shouldn't be the default. That approach sounds reasonable - rename the macro, don't disable the option if it's not defined, but still default to none except on platforms where POSIX or native fallocate is available. I'll update.

sitsofe:
@tripped Tidy!

@axboe The pull request LGTM!

sitsofe:
@axboe brief reminder on this one...

tripped:
@sitsofe @axboe Ah, hold the presses! I just did some _actual_ testing on Windows again and there's one more fairly silly thing I missed. This still doesn't allow the option to really be used on Windows; apparently I misunderstood the point at which `fixup_options` is called; it is forcing the value of the fallocate flag to `FIO_FALLOCATE_NONE` even when `truncate` was explicitly passed.

The desired behavior per the documentation is that the default for fallocate is `native` when a native or POSIX implementation is available (i.e., when `FIO_HAVE_ANY_FALLOCATE` / `FIO_HAVE_DEFAULT_FALLOCATE` is defined), but none otherwise. Seems we could do this by just putting an ifdef around the `.def` for fallocate in options.c, and then remove the force-to-none in `fixup_options`. I'll update the PR again for your consideration. :-)

axboe:
Where are we with this one?

tripped:
@axboe I think this PR is complete, at least as far as I can tell. To summarize:

- Adds a new fallocate option, `truncate`, which just results in a call to `ftruncate`.
- Default fallocate behavior is not changed; it will still be `native` on platforms with a native `fallocate`, and `none` on platforms without (e.g., Windows).
- The potentially surprising interaction with non-sequential writes on Windows is called in out in the documentation.
- I've tested on Windows and verified that the `ftruncate` gets translated to a SetEndOfFile and has the desired effect (Windows treats writes to the extended region as overwrites and does not serialize them, the same as with Explorer copy).

Let me know if there's anything else you'd need before merging. :-)

axboe:
Pulled, thanks

PR #860, opened 2019-11-06
Automate fio tests

vincentkfu:
Jens, please consider the series of patches in this branch.

It consists of a script that (1) runs the jobs in t/jobs (checking the appropriate output afterwards) and (2) runs test scripts/programs (axmap, ieee754, lfsr-test, zbd, etc) in t/.

run-fio-tests.py can be run before or after patches are committed to determine if patches break any of the tested functionality. A full run completes in under 10 minutes.


axboe:
That is awesome! I'm really excited about this, I've been hoping to get closer to having some nice unit and regression tests that can be run. Pulled, with vigor.

PR #866, opened 2019-11-25
engines/rados: changed polling to completion callbacks methodology

aclamk:


axboe:
Haven't looked at the code yet, but the commit message is devoid of any sort of explanation on why this change is being made. A good commit message should always explain the reason for the change, maybe you can amend it and provide that?

PR #867, opened 2019-11-27
Doc: Fix documentation of exitall

hannesweisbach:
exitall does not kill all jobs, but only jobs of the same group.

Signed-off-by: Hannes Weisbach <hannes.weisbach@gmail.com>


HOWTO
```diff
@@ -2816,7 +2816,7 @@ Threads, processes and job synchronization
 
 	By default, fio will continue running all other jobs when one job finishes
 	but sometimes this is not the desired action.  Setting ``exitall`` will
-	instead make fio terminate all other jobs when one job finishes.
+	instead make fio terminate all other jobs in a group when one job finishes.
```
axboe: Probably should make that a bit more precise since we're changing it - rather than "in a group", it should be "[...] in the same group, then one job that group finishes."

hannesweisbach: Yea, you're right. I'll fix it tomorrow.

Would you be interested in an additional option ``exitgroup``, with the same semantics as exitall now has. And then having ``exitall`` actually exit all workers?

axboe: > Would you be interested in an additional option exitgroup, with the same semantics as exitall now has. And then having exitall actually exit all workers?

We can't break existing use cases by changing the meaning of it now, if you want that kind of behavior then we'd need a new option.

hannesweisbach: Hm. True …
I've actually already hacked a version up … (because I want to have grouping for log aggregation of only some jobs and still be able to quit after a specfic job has finished)

Do you have a preference for naming this new option? I can only come up with ``really_exitall`` ;)

axboe:
Naming is the hardest! How about just having an option that modifies what ```exitall``` does, defaulting to the same behavior as now? ```exitall_all``` (also lame name, but...) or something like that, with it defaulting to false. That avoids having conflicting options for the same thing. Or ```exitall_only_group``` with it defaulting to true.

HOWTO
sitsofe: `exit_every_group`?

hannesweisbach: ❤️

hannesweisbach:
This is what I came up with. I hope you don't mind me hijacking this PR for adding functionality 😬.

hannesweisbach:
Sorry, for the late reply … I completely missed the notification :(

My idea was to not touch exitall at all for backwards compatibility. exit_every_group was supposed to be a stand-alone/additional option. If exitall is specified it should take precedence for backwards compatibility.
If you think its better to have exit_every_group modify the behaviour of exitall, I'll do that.

I've thought a bit more about it, and one more issue is the interaction with stonewall. I think it might be useful to have exit_every_group not terminate everything but only all groups until the next stonewall. Or at least have it selectable.

Let's say, for example we have the following fio job file snippet:
```
group_reporting=1

[a]
rw=write

[b]
new_group=1
rw=rw
number_ios=123
exitall=1 # won't stop a
exit_every_group=1 # will stop everything:

[c]
stonewall=1
rw=rw
rwmixread=77
time_based=1
runtime=6
```

a and b should run at the same time until b has issued 123 ios. Both should be reported separately. After they finished, c should run for 6s. I believe this is currently not possible, because ``exit_all = 1`` would not stop a, and ``exit_every_group = 1`` will stop everything, including c and everything after c.

I'm a bit torn, but it feels a bit feature-creep-ish, but at the same time I might as well do it right …

I think in the latter case it might be best to make exitall a string option with the following values:
- "1", for backwards compatilibilty
- "group" as synonym for "1"
- "all" for TERMINATE_ALL
- "stonewall" terminate all currently running (i.e. non-stonewalled) threads (probably check for runstate >= TD_RUNNING in fio_terminate_threads())

What are your thoughts?

hannesweisbach:
I've prototyped what I explained verbally earlier - code is probably more concise then verbal explanation ;). This is WIP/RFC only, there is still cleanup/documentation/… missing.

I think exitall_on_error can also be folded into this code, with a bit of extra-code for staying backwards-compatible.

axboe:
> My idea was to not touch exitall at all for backwards compatibility. exit_every_group was supposed to be a stand-alone/additional option. If exitall is specified it should take precedence for backwards compatibility.

But that doesn't work at all, because then you have conflicting options. That makes it much worse, because if you just set exitall, then it doesn't matter what the other option is set at. If you introduce the option as a MODIFIER to exitall, then there isn't going to be a conflict ever and there's no backwards compatibility issue if the default is the same as before.

axboe:
Took a look at the current code and I still don't like it... Now you definitely have a risk of having backwards compatibility issues. Please re-read this comment again:

https://github.com/axboe/fio/pull/867#issuecomment-559186068

as that explains what I think is the best way to do it. In short, exitall retains it's current 0/1 values, that ensures it's 100% backwards compatible. Then you add the other option, exitall_only_group (or whatever name). This option defaults to true, but can be set to false.

Then just have the two callers that use groupid factor in exitall_only_group, and that should be it.

options.c
```diff
@@ -3947,13 +3941,28 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
 		.group	= FIO_OPT_G_PROCESS,
 	},
 	{
-		.name	= "exit_every_group",
-		.lname	= "Exit-every group on terminate",
-		.type	= FIO_OPT_STR_SET,
-		.cb	= str_exit_every_group_cb,
-		.help	= "Terminate all jobs of all groups when one jobs exits",
+		.name	= "exit_what",
+		.lname	= "Exit-all on terminate",
```
sitsofe: Should this be updated?

hannesweisbach: Yes, the documentation is also not ready yet. I think with the current implementaiton of exit_what I also need to edit cconv.c, right? I'm planning also to add an example.

Anyways I want to start updating documentation when the code is nailed down, otherwise it's too much of a moving goal post …

axboe: You need cconv additions if you make it a job option, instead of a global option. Which is recommended, global options are discouraged and honestly should not exist. They don't work over the network, for example.

hannesweisbach:
Adding exit_what as uint32_t to packed thread options resulted in an assert violation for the alignment of fp64 members. I changed exit_what and stonewall to uint16_t to keep alignment of following members and also the size of the struct.

I feel exit_what groups ok-ish with stonewall and new_group, hence I positioned it there.

I hope this is ok?

axboe:
Thanks for getting this to completion, pulled.

sitsofe:
Nice job @hannesweisbach !

PR #868, opened 2019-11-27
engines/rados: fix error with getting last instead of first element from list

aclamk:
A thinko has popped in implementation. Consumer function `getevent()` did pick last (newest) instead of first (oldest).


aclamk:
@axboe I am very sorry to make temporary worst possible latency in ops handling!

axboe:
Pulled, thanks.

PR #869, opened 2019-12-01
./configure: Fix --enable-libaio-uring

hasegaw:
Try compiling test code with libaio; otherwise, the test will fail.
CONFIG_LIBAIO_URING will be enabled if both of libaio and liburing
are available.

Signed-off-by: Takeshi HASEGAWA <hasegaw@gmail.com>

### DIFF
--- a/configure
+++ b/configure
@@ -607,9 +607,9 @@ int main(void)
   return 0;
 }
 EOF
-  if test "$libaio_uring" = "yes" && compile_prog "" "-luring" "libaio io_uring" ; then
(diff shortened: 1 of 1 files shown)


axboe:
This is wrong, it's a test case for using the liburing branch that has libaio emulation. If you link -laio in as well, then that defeats the purpose as you are then just using libaio. For reference, the liburing libaio emulation is here:

https://git.kernel.dk/cgit/liburing/log/?h=libaio-wrapper

hasegaw:
Ah I see - Thank you for review and your detailed comments.
I’ll try the branch.

PR #874, opened 2019-12-09
Use default version of the system python

anatol:
As the script is compatible both with python2 and python3 then it makes sense do not hardcode to python2.7 and use the system pytohn version.

### DIFF
--- a/tools/fiologparser.py
+++ b/tools/fiologparser.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python2.7
+#!/usr/bin/python
 # Note: this script is python2 and python 3 compatible.
 #
 # fiologparser.py


sitsofe:
So the current hashbang was a serious compromise when it was introduced because you can't please all of the people all of the time (see https://github.com/axboe/fio/commit/60023ade47e7817db1c18d9b7e511839de5c2c99 ). For the sake of distro packagers it's better for ALL hashbang lines to be the same across the whole project because they run a script that fixes them all up to be whatever they want.

sitsofe:
However I'm aware that upstream python 2 is being deprecated at the end of this year (https://pythonclock.org/ ) so maybe there's an argument for switching all the hashbang's to python 3?

sitsofe:
This one has been superceded by the commits leading up to https://github.com/axboe/fio/commits/c9beb194ef608dc49dd2f7537b90951f3c8b432d . Now all python scripts use
```
#!/usr/bin/env python3
```
and a few straggler scripts have been converted to be python3 compatible.

Issue #784, opened 2019-05-31
Failed to create dir  on windows


YongxueHong:
Hi, 
Failed to create a directory when using fio-3.13 on windows.
e.g:

> C:\Windows\System32>"C:\Program Files (x86)\fio\fio\fio.exe" --rw=randrw --bs=4k --iodepth=4 --direct=1 --filename=C\:\\fio_system_disk_test --name=fiotest --ioengine=windowsaio --thread --group_reporting --numjobs=4 --size=1G --runtime=300
> "C:\Program Files (x86)\fio\fio\fio.exe" --rw=randrw --bs=4k --iodepth=4 --direct=1 --filename=C\:\\fio_system_disk_test --name=fiotest --ioengine=windowsaio --thread --group_reporting --numjobs=4 --size=1G --runtime=300
> fiotest: (g=0): rw=randrw, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=windowsaio, iodepth=4
> ...
> fio-3.13
> Starting 4 threads
> 
> 
> Run status group 0 (all jobs):
> fio: failed to create dir (C:\): 13

It is ok for fio-2.2.10.
The format of option "filename" is changed ?

sitsofe:
@YongxueHong Well I can't say anything about fio 2.2.10 working because it's so ancient but wrt to fio 3.13 - have you read the documentation on `filename` (https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-filename )? If you haven't read it yet could you do so and let us know if anything it mentions helps?

YongxueHong:
@sitsofe Yeah, I have read the documents about the options "filename" then try it by any way which I could think, but still failed to work.

sitsofe:
@YongxueHong,

```
--filename=C:\fio_system_disk_test
            ^^
```

From https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-filename :
>**Each colon and backslash** in the wanted path must be escaped with a \ character.

But on top of that I can't tell what your shell is so you might have to put that in quotes too.

(Additionally while you're trying to solve this particular issue I recommend removing as many of the other options as possible to make it easier to narrow down the cause of the problem)

YongxueHong:
So sorry, My mistake.  
For the comment:

> Hi,
> Failed to create a directory when using fio-3.13 on windows.
> e.g:
> 
[...]

Issue #870, opened 2019-12-02
Windows fio: failed to create dir (d:): 22


Gt3pccb:
I am having this issue with FIO compiled from the latest branch because we need the fix for first allocations. 
This command used to work used fine and FIO is being run as Admin and permissions are everyone full
`--directory=d\:\`

PS D:\> FIO --thread  --direct=1 --ioengine=windowsaio --filesize=500GB --rw=write --iodepth=256 --bs=1MB --nrfiles=1 --name=XR2-4-Node --numjobs=1 `--directory=d\:\`

XR2-4-Node: (g=0): rw=write, bs=(R) 1024KiB-1024KiB, (W) 1024KiB-1024KiB, (T) 1024KiB-1024KiB, ioengine=windowsaio, iodepth=256
fio-3.16
Starting 1 thread
fio: failed to create dir (d:): 22

Run status group 0 (all jobs):

Any help would be great!

sitsofe:
This sounds very similar to #784. One trick is change the code to do a stat and then check the stat said it was a directory and if so there's nothing more to do and only try and do the mkdir if the stat says it failed (or wasn't a directory).

Gt3pccb:
Indeed,  “d” is the root of a drive, and there is no directory-folder to be created. The error remains the same for
=D\:\
=D\:\\
=D\:\FIO     (the directory is already there)
=D\:\\FIO   (the directory is already there)

The issue is that now I must choose between FIO with new truncate option that we need or use an older version of FIO without the truncate option. Not a good place to be.

Thank you
Astolfo

axboe:
> One trick is change the code to do a stat and then check the stat said it was a directory and if so there's nothing more to do and only try and do the mkdir if the stat says it failed (or wasn't a directory)

I really like that idea, and it improves the usability of the directory function a lot.

Gt3pccb:
Do you know what changed between let say from 3.12 to 3.16?

=D\:\\ or =D\:\\FIO  works just fine in 3.12 given that there is a mount point for “D:\” or “D:\FIO”

The mkdir would be very nice but it would be good to get the functionality back from 3.12

Thanks

axboe:
I'm guessing it's this commit:

```
commit 95af8dd57a7a9a564f99492bb1e04bb6d13b95b1
Author: Kyle Zhang <kyle@smartx.com>
Date:   Thu Apr 11 23:01:55 2019 +0800

[...]

Issue #864, opened 2019-11-14
fio doesn't create subdirectories as set at filename_format option 


StingerXIII:
I trying to launch a workload which writes files at the separate directories per each thread.
Here is my jobfile:

```
[global]
ioengine=psync
direct=1
rw=write
file_service_type=sequential:1
filename_format=$jobnum/$filenum
openfiles=1
create_on_open=1
end_fsync=1

[fstest]
numjobs=2
directory=/mnt/xfs/0
filesize=1M
nrfiles=10
```

The [manual](https://fio.readthedocs.io/en/latest/fio_man.html#cmdoption-arg-filename-format) says

> If you specify a path then the directories will be created up to the main directory for the file. So for example if you specify filename_format=a/b/c/$jobnum then the directories a/b/c will be created before the file setup part of the job. 

The error I get:
```
[root@hostname ~]# fio fio.ini 
fstest: (g=0): rw=write, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
...
fio-3.16
Starting 2 processes
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
disk util stat: No such file or directory
fio: pid=18011, err=2/file:filesetup.c:701, func=open(/mnt/xfs/0/0/0), error=No such file or directory
fio: pid=18012, err=2/file:filesetup.c:701, func=open(/mnt/xfs/0/1/0), error=No such file or directory


Run status group 0 (all jobs):
```

The directory /mnt/xfs/0 exists before I start a job.

sitsofe:
Hi @StingerXIII,

If you cd to `/mnt/xfs/0` and then remove `directory=/mnt/xfs/0` does it still fail?

sitsofe:
Hmm even
```
fio --filename_format=./1/1 --name=test --size=16k
```

seems to fail...

StingerXIII:
[...]

PR #879, opened 2019-12-16
filesetup: fix directory creation issues

sitsofe:
- Fix double directory prefix concatenation that was happening after
  95af8dd57a7a9a564f99492bb1e04bb6d13b95b1 ("filesetup: don't call
  create_work_dirs() for ioengine with FIO_DISKLESSIO") refactored some
  code
- Stop always using '/' and switch to using FIO_OS_PATH_SEPARATOR when
  concatenating the directory value to the filename in set_name_idx()
- Change the permissions directories are created with to have the
  execute bit so they can be entered(!)
- Fix bug where filenames that were absolute unix paths would prevent
  the creation of intermediate directories
- Remove the configure mkdir test and introduce fio_mkdir to hide mkdir
  platform differences
- Make Windows mkdir emulation more robust against multiple slashes
  after the drive letter and when dealing dealing with with paths
  containing drive letters in general

This commit makes the following work on Windows:
./fio --thread --name=doubleslashtest \
 --filename='C\:\\Windows\Temp\fio.tmp' --size=4k --bs=4k --rw=read

The next example now works when D:\1\ exists but D:\1\2\ doesn't:

./fio --thread --name=mkdirtest --directory='D\:\1' \
 --filename='2\fio.tmp' --size=4k --bs=4k --rw=read

The following now works on macOS/Linux when /tmp/1/ exists but /tmp/1/2/
doesn't:

./fio --name=mkdirtest --directory='/tmp/1' \
 --filename='2/fio.tmp' --size=4k --bs=4k --rw=read

Fixes https://github.com/axboe/fio/issues/784 ,
https://github.com/axboe/fio/issues/864 and
https://github.com/axboe/fio/issues/870

Tested-by: Astolfo Rueda <astolfor@microsoft.com>

(This one is still going through CI but I'm fairly sure it's good)

### DIFF
--- a/configure
+++ b/configure
@@ -2317,22 +2317,6 @@ fi
 print_config "cuda" "$cuda"
 
 ##########################################
-# mkdir() probe. mingw apparently has a one-argument mkdir :/
-mkdir_two="no"
-cat > $TMPC << EOF
-#include <sys/stat.h>
-#include <sys/types.h>
-int main(int argc, char **argv)
-{
-  return mkdir("/tmp/bla", 0600);
-}
-EOF
-if compile_prog "" "" "mkdir(a, b)"; then
-  mkdir_two="yes"
-fi
-print_config "mkdir(a, b)" "$mkdir_two"
-
-##########################################
 # check for cc -march=native
 build_native="no"
 cat > $TMPC << EOF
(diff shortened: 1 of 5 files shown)


sitsofe:
@vincentkfu I'm seeing a steady state test failure on macOS but I don't really know why - https://travis-ci.org/axboe/fio/jobs/625675665#L377 ...

vincentkfu:
The travis-ci check fails because of a change to the relevant xcode image. I have a fix for this issue and will submit the change soon: https://github.com/vincentkfu/fio/commit/684297570f9cbf49c7c94119ccae6313e5d619f8

sitsofe:
Rebased to pick up spurious test failure fix.

Issue #881, opened 2019-12-16
Windows Cygwin build error due to recent MinGW FORTIFY_SOURCE changes


wxlg1117:
```
Administrator@WIN12R2DC ~/fio
$ uname -a
CYGWIN_NT-6.3 WIN12R2DC 3.0.7(0.338/5/3) 2019-04-30 18:08 x86_64 Cygwin

Administrator@WIN12R2DC ~/fio
$ git log
commit 41ceb6c79aea52bd46ea13c5036722a294bb719e (HEAD -> master, origin/master, origin/HEAD)
Author: Vincent Fu <vincent.fu@wdc.com>
Date:   Tue Dec 10 12:54:42 2019 -0500

    t/run-fio-tests: relax acceptance criterion for t0011

    I saw a handful of benign failures for t0011 when run on
    travis/appveyor. This patch allows the test to still pass with a
    measured IOPS of 998 to 1002 and adds a debug print for the value
    actually observed.

    Signed-off-by: Jens Axboe <axboe@kernel.dk>




Administrator@WIN12R2DC ~/fio
$ ./configure
Forcing some known good options on Windows
Operating system              CYGWIN_NT-6.3
CPU                           x86_64
Big endian                    no
Target Windows version        7
Compiler                      x86_64-w64-mingw32-gcc
Cross compile                 no

Static build                  yes
Wordsize                      32
zlib                          yes
Linux AIO support             no
Linux AIO over io_uring       no
POSIX AIO support             no
POSIX AIO support needs -lrt  no
POSIX AIO fsync               no
POSIX pshared support         no
Solaris AIO support           no
__sync_fetch_and_add          yes
__sync_synchronize            yes
__sync_val_compare_and_swap   yes
libverbs                      no
rdmacm                        no
asprintf()                    yes
vasprintf()                   yes
Linux fallocate               no
POSIX fadvise                 no
POSIX fallocate               no
sched_setaffinity(3 arg)      no
sched_setaffinity(2 arg)      no
clock_gettime                 yes
CLOCK_MONOTONIC               yes
CLOCK_MONOTONIC_RAW           no
CLOCK_MONOTONIC_PRECISE       no
clockid_t                     yes
gettimeofday                  yes
fdatasync                     yes
sync_file_range               no
EXT4 move extent              no
Linux splice(2)               no
GUASI                         no
libnuma                       no
strsep                        no
strcasestr                    no
strlcat                       no
getopt_long_only()            yes
inet_aton                     no
[...]

PR #882, opened 2019-12-16
Makefile: add libssp for Windows

vincentkfu:
Recent Cygwin mingw releases require linking against libssp to avoid
undefined symbols. See
https://github.com/msys2/MINGW-packages/issues/5868 for discussion.

Fixes: https://github.com/axboe/fio/issues/881


vincentkfu:
As requested, updated commit message to mention FORTIFY_SOURCE and added signed-off-by.

PR #885, opened 2019-12-21
Travis updates

sitsofe:
- Stop doing two builds in the same xcode image (but this is slightly hidden due to whatever travis chooses as a default)
- Update Ubuntu build from a 16.04 container to an 18.04 one and add more libraries to it so we test whether we compile in more scenarios...


axboe:
Pulled, thanks Sitsofe.

PR #887, opened 2019-12-22
unittests: add unittests for oslib/{strcasestr,strsep}.c

kusumi:
Add some unittests to the framework I've written a year ago.
There were no bugs found this time, so nothing interesting here, just adding test cases.


sitsofe:
(FYI: The test 10 failure on Windows is spurious. See https://github.com/axboe/fio/commit/5e52534d963c2770be7f4d345e4db2a4ecac7730#commitcomment-36504957 and https://www.spinics.net/lists/fio/msg08170.html for details)

PR #903, opened 2020-01-10
Adding libaio/io_uring option cmdprio_percentage

trinof:
This adds the libaio option prio_percent which allows a certain percentage of commands to be sent at a higher priority. Since this only is relevant to reads on the drive side, only reads will be broken down into multiple priorities like this. The latencies of the two priority types will be reported separately in addition to the combined latencies. This will only affect normal and json reporting.

### DIFF
--- a/engines/io_uring.c
+++ b/engines/io_uring.c
@@ -70,6 +70,7 @@ struct ioring_data {
 struct ioring_options {
 	void *pad;
 	unsigned int hipri;
+	unsigned int cmdprio_percentage;
 	unsigned int fixedbufs;
 	unsigned int registerfiles;
 	unsigned int sqpoll_thread;
@@ -108,6 +109,26 @@ static struct fio_option options[] = {
 		.category = FIO_OPT_C_ENGINE,
 		.group	= FIO_OPT_G_IOURING,
 	},
+#ifdef FIO_HAVE_IOPRIO_CLASS
+	{
+		.name	= "cmdprio_percentage",
+		.lname	= "high priority percentage",
+		.type	= FIO_OPT_INT,
+		.off1	= offsetof(struct ioring_options, cmdprio_percentage),
+		.minval	= 1,
+		.maxval	= 100,
+		.help	= "Send high priority I/O this percentage of the time",
+		.category = FIO_OPT_C_ENGINE,
+		.group	= FIO_OPT_G_IOURING,
+	},
+#else
+	{
+		.name	= "cmdprio_percentage",
+		.lname	= "high priority percentage",
+		.type	= FIO_OPT_UNSUPPORTED,
+		.help	= "Your platform does not support I/O priority classes",
+	},
+#endif
 	{
 		.name	= "fixedbufs",
 		.lname	= "Fixed (pre-mapped) IO buffers",
@@ -313,11 +334,23 @@ static int fio_ioring_getevents(struct thread_data *td, unsigned int min,
 	return r < 0 ? r : events;
 }
 
+static void fio_ioring_prio_prep(struct thread_data *td, struct io_u *io_u)
+{
+	struct ioring_options *o = td->eo;
+	struct ioring_data *ld = td->io_ops_data;
+	if (rand_between(&td->prio_state, 0, 99) < o->cmdprio_percentage) {
+		ld->sqes[io_u->index].ioprio = IOPRIO_CLASS_RT << IOPRIO_CLASS_SHIFT;
+		io_u->flags |= IO_U_F_PRIORITY;
+	}
+	return;
+}
+
 static enum fio_q_status fio_ioring_queue(struct thread_data *td,
 					  struct io_u *io_u)
 {
 	struct ioring_data *ld = td->io_ops_data;
 	struct io_sq_ring *ring = &ld->sq_ring;
+	struct ioring_options *o = td->eo;
 	unsigned tail, next_tail;
 
 	fio_ro_check(td, io_u);
@@ -343,6 +376,8 @@ static enum fio_q_status fio_ioring_queue(struct thread_data *td,
 
 	/* ensure sqe stores are ordered with tail update */
 	write_barrier();
+	if (o->cmdprio_percentage)
+		fio_ioring_prio_prep(td, io_u);
 	ring->array[tail & ld->sq_ring_mask] = io_u->index;
 	*ring->tail = next_tail;
 	write_barrier();
@@ -618,6 +653,7 @@ static int fio_ioring_init(struct thread_data *td)
 {
 	struct ioring_options *o = td->eo;
 	struct ioring_data *ld;
+	struct thread_options *to = &td->o;
 
 	/* sqthread submission requires registered files */
 	if (o->sqpoll_thread)
@@ -640,6 +676,17 @@ static int fio_ioring_init(struct thread_data *td)
 	ld->iovecs = calloc(td->o.iodepth, sizeof(struct iovec));
 
 	td->io_ops_data = ld;
+
+	/*
+	 * Check for option conflicts
+	 */
+	if ((fio_option_is_set(to, ioprio) || fio_option_is_set(to, ioprio_class)) &&
+			o->cmdprio_percentage != 0) {
+		log_err("%s: cmdprio_percentage option and mutually exclusive "
+				"prio or prioclass option is set, exiting\n", to->name);
+		td_verror(td, EINVAL, "fio_io_uring_init");
+		return 1;
+	}
 	return 0;
 }
 
--- a/engines/libaio.c
+++ b/engines/libaio.c
@@ -16,7 +16,13 @@
 #include "../optgroup.h"
 #include "../lib/memalign.h"
 
+/* Should be defined in newest aio_abi.h */
+#ifndef IOCB_FLAG_IOPRIO
+#define IOCB_FLAG_IOPRIO    (1 << 1)
+#endif
+
 static int fio_libaio_commit(struct thread_data *td);
+static int fio_libaio_init(struct thread_data *td);
 
 struct libaio_data {
 	io_context_t aio_ctx;
@@ -44,6 +50,7 @@ struct libaio_data {
 struct libaio_options {
 	void *pad;
 	unsigned int userspace_reap;
+	unsigned int cmdprio_percentage;
 };
 
 static struct fio_option options[] = {
@@ -56,6 +63,26 @@ static struct fio_option options[] = {
 		.category = FIO_OPT_C_ENGINE,
 		.group	= FIO_OPT_G_LIBAIO,
 	},
+#ifdef FIO_HAVE_IOPRIO_CLASS
+	{
+		.name	= "cmdprio_percentage",
+		.lname	= "high priority percentage",
+		.type	= FIO_OPT_INT,
+		.off1	= offsetof(struct libaio_options, cmdprio_percentage),
+		.minval	= 1,
+		.maxval	= 100,
+		.help	= "Send high priority I/O this percentage of the time",
+		.category = FIO_OPT_C_ENGINE,
+		.group	= FIO_OPT_G_LIBAIO,
+	},
+#else
+	{
+		.name	= "cmdprio_percentage",
+		.lname	= "high priority percentage",
+		.type	= FIO_OPT_UNSUPPORTED,
+		.help	= "Your platform does not support I/O priority classes",
+	},
+#endif
 	{
 		.name	= NULL,
 	},
@@ -83,6 +110,17 @@ static int fio_libaio_prep(struct thread_data fio_unused *td, struct io_u *io_u)
 		io_prep_fsync(iocb, f->fd);
 
 	return 0;
+}
+
+static void fio_libaio_prio_prep(struct thread_data *td, struct io_u *io_u)
+{
+	struct libaio_options *o = td->eo;
+	if (rand_between(&td->prio_state, 0, 99) < o->cmdprio_percentage) {
+		io_u->iocb.aio_reqprio = IOPRIO_CLASS_RT << IOPRIO_CLASS_SHIFT;
+		io_u->iocb.u.c.flags |= IOCB_FLAG_IOPRIO;
+		io_u->flags |= IO_U_F_PRIORITY;
+	}
+	return;
 }
 
 static struct io_u *fio_libaio_event(struct thread_data *td, int event)
@@ -188,6 +226,7 @@ static enum fio_q_status fio_libaio_queue(struct thread_data *td,
 					  struct io_u *io_u)
 {
 	struct libaio_data *ld = td->io_ops_data;
+	struct libaio_options *o = td->eo;
 
 	fio_ro_check(td, io_u);
 
@@ -217,6 +256,9 @@ static enum fio_q_status fio_libaio_queue(struct thread_data *td,
 		io_u_mark_complete(td, 1);
 		return FIO_Q_COMPLETED;
 	}
+
+	if (o->cmdprio_percentage)
+		fio_libaio_prio_prep(td, io_u);
 
 	ld->iocbs[ld->head] = &io_u->iocb;
 	ld->io_us[ld->head] = io_u;
@@ -358,6 +400,8 @@ static int fio_libaio_post_init(struct thread_data *td)
 static int fio_libaio_init(struct thread_data *td)
 {
 	struct libaio_data *ld;
+	struct thread_options *to = &td->o;
+	struct libaio_options *o = td->eo;
 
 	ld = calloc(1, sizeof(*ld));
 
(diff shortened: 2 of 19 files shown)


axboe:
One minor comment - maybe add this as an engine flag, and make the code generic? Both libaio and io_uring support priority info, and I’m mostly viewing the former as dead and uninteresting at this point.

trinof:
Sure, I can look into making that adjustment. Looks like I could combine it with the existing hipri_percentage from the pvsync2 engine. hipri_percentage seems like a clearer option name anyway.

axboe:
Yeah, you could even just make them per engine options, just name them identically. Makes sense to differentiate this from the cfq/bfq priorities.

trinof:
I got my hands on a drive with priority support (enabled with the /sys/block/sd*/device/ncq_prio_enable file) and I'm not sure the consolidation of the various engine priorities makes too much sense. "priority" seems to be an overloaded word and the priority I'm interested in testing on a per-command basis is commands with a priority bit set in the eventual read command. That's how the libaio prio_percent implementation works, but doesn't appear to be what the pvsync2 hipri option does. I'm not sure what the RWF_PRI flag that the hipri option sets does, but it definitely does not affect the priority bit of the outgoing commands. Given that these two per-command priority options seem to be working on completely different types of priority, I think it makes sense to keep them with separate names. However, I could still set the priority_bit when sending pvsync2 commands which would let the two types of priorities be tracked separately as high and low priority commands.
I'll also try and see what type of priority the io_uring hipri command is setting and how to set it on a per-command basis, but it looks like I'll need to upgrade my kernel first.

sitsofe:
@trinof :

> That's how the libaio prio_percent implementation works, but doesn't appear to be what the pvsync2 hipri option does. I'm not sure what the RWF_PRI flag that the hipri option sets does, but it definitely does not affect the priority bit of the outgoing commands.

`RWF_HIPRI` basically tells the kernel it's worth polling the underlying block device for completion rather than waiting for the usual interrupt to arrive (maybe because you are trying to achieve lower latencies at the cost of more CPU). See https://lwn.net/Articles/670231/ and http://man7.org/linux/man-pages/man2/readv.2.html#DESCRIPTION (the later warns it's only usable if you are also using `O_DIRECT`).

axboe:
@trinof The pvsync2 is different, so don't mix it up with that. For what you care about, getting the NCQ prio bit set, that's all in the priority information carried with the submission. In that regard, libaio and io_uring does the same thing, they both carry the same priority information into the kernel. Those are the two I wanted you to unify, not the pvsync2 RWF_HIPRI.

trinof:
Thanks for the clarification. I've renamed the prio_percent option to hiprio_percent for clarity and added the hiprio_percent option to the io_uring engine. Let me know if you see any problems or have any other suggestions.

engines/io_uring.c
```diff
@@ -343,6 +365,14 @@ static enum fio_q_status fio_ioring_queue(struct thread_data *td,
 
 	/* ensure sqe stores are ordered with tail update */
 	write_barrier();
+	if (rand_between(&ld->prio_state, 0, 99) < o->hiprio_percent) {
```
axboe: Probably gate this on o->hiprio_percent being valid. If not set, we shouldn't do anything. This is a hot path, so generating random values if not enabled is a really bad idea. I'd make it:

```
if (o->hiprio_percent && rand_between(&ld->prio_state, 0, 99) < o->hiprio_percent) {
        ...
```

Interestingly, you seem to do this correctly for libaio, but not here.

engines/io_uring.c
```diff
@@ -108,6 +118,17 @@ static struct fio_option options[] = {
 		.category = FIO_OPT_C_ENGINE,
 		.group	= FIO_OPT_G_IOURING,
 	},
+	{
```
axboe: Does this need the same treatment as libaio in terms of availability?

engines/io_uring.c
```diff
@@ -640,6 +671,23 @@ static int fio_ioring_init(struct thread_data *td)
...
+	/*
+	 * Initialize hiprio_percent mechanisms if relevant
+	 */
+	if (!(fio_option_is_set(to, ioprio) ||
+			fio_option_is_set(to, ioprio_class) ||
+			(o->hiprio_percent == 0))) {
+		ld->prio_seed = FIO_RANDSEED * td->thread_number
+				+ FIO_RAND_NR_OFFS;
+		init_rand_seed(&ld->prio_state, ld->prio_seed, false);
+	}
```
axboe: Format is:

```
if (something) {
} else {
}
```

No new line before the else, please.

engines/libaio.c
```diff
@@ -122,7 +180,7 @@ struct aio_ring {
 #define AIO_RING_MAGIC	0xa10a10a1
 
 static int user_io_getevents(io_context_t aio_ctx, unsigned int max,
-			     struct io_event *events)
```
axboe: Unrelated change?

engines/libaio.c
```diff
@@ -162,8 +220,8 @@ static int fio_libaio_getevents(struct thread_data *td, unsigned int min,
 
 	do {
 		if (o->userspace_reap == 1
-		    && actual_min == 0
```
axboe: Also seem unrelated.

engines/libaio.c
```diff
@@ -226,7 +286,7 @@ static enum fio_q_status fio_libaio_queue(struct thread_data *td,
 }
 
 static void fio_libaio_queued(struct thread_data *td, struct io_u **io_us,
-			      unsigned int nr)
```
axboe: Same here

engines/libaio.c
```diff
@@ -368,6 +432,22 @@ static int fio_libaio_init(struct thread_data *td)
...
+	/*
+	 * Initialize hiprio_percent mechanisms if relevant
+	 */
+	if (!(fio_option_is_set(to, ioprio) ||
+			fio_option_is_set(to, ioprio_class) ||
+			(o->hiprio_percent == 0))) {
+		ld->prio_seed = FIO_RANDSEED * td->thread_number
+				+ FIO_RAND_NR_OFFS;
+		init_rand_seed(&ld->prio_state, ld->prio_seed, false);
+	}
```
axboe: Same style issue here.

fio.h
```diff
@@ -460,7 +460,7 @@ struct thread_data {
 	CUdevice  cu_dev;
 	CUcontext cu_ctx;
 	CUdeviceptr dev_mem_ptr;
-#endif
```
axboe: Also unrelated - it's clearly correct, but it should at least be a separate commit, not bundled with this.

ioengines.c
```diff
@@ -318,6 +318,7 @@ enum fio_q_status td_io_queue(struct thread_data *td, struct io_u *io_u)
 					sizeof(io_u->issue_time));
 	}
 
+
```
axboe: Unrelated

iolog.c
```diff
@@ -845,7 +845,7 @@ static void flush_hist_samples(FILE *f, int hist_coarseness, void *samples,
 	uint64_t *io_u_plat_before;
 
 	int stride = 1 << hist_coarseness;
-
```
axboe: Unrelated

iolog.c
```diff
@@ -1207,13 +1207,13 @@ void flush_log(struct io_log *log, bool do_append)
 
 		cur_log = flist_first_entry(&log->io_logs, struct io_logs, list);
 		flist_del_init(&cur_log->list);
-
```
axboe: Unrelated

iolog.c
```diff
@@ -1207,13 +1207,13 @@ void flush_log(struct io_log *log, bool do_append)
...
 		cur_log = flist_first_entry(&log->io_logs, struct io_logs, list);
 		flist_del_init(&cur_log->list);
-		
+
 		if (log->td && log == log->td->clat_hist_log)
 			flush_hist_samples(f, log->hist_coarseness, cur_log->log,
 			                   log_sample_sz(log, cur_log));
 		else
 			flush_samples(f, cur_log->log, log_sample_sz(log, cur_log));
-
```
axboe: Unrelated

server.c
```diff
@@ -1574,6 +1574,14 @@ void fio_server_send_ts(struct thread_stat *ts, struct group_run_stats *rs)
 	p.ts.cachehit		= cpu_to_le64(ts->cachehit);
 	p.ts.cachemiss		= cpu_to_le64(ts->cachemiss);
 
+	p.ts.priority_bit = ts->priority_bit;
```
axboe: Seems this needs a network protocol version bump. See FIO_SERVER_VER

stat.h
```diff
@@ -237,6 +237,12 @@ struct thread_stat {
 	fio_fp64_t ss_deviation;
 	fio_fp64_t ss_criterion;
 
+	bool priority_bit;
```
axboe: bool probably isn't a good choice for an on-wire variable type. Should be uint8_t or something like that. This structure is sent over the wire.

trinof:
Thanks for the quick review. I'll try to resolve all the issues--I think many of the apparently random white space changes were from when I was playing with part of the code to better understand it and though I removed the code afterwards I clearly didn't keep the white space perfectly the same. I'll plan to keep the pull request as a single commit unless you'd prefer the resolution of these issues as a separate commit?

axboe:
Please just amend the commit to fix it up. If the unrelated changes make sense on their own, then you should create at least one separate patch for those. Just don't bundle them with functional changes, that's never a good idea.

trinof:
I think the new commit has resolved all of the issues you found with your review. Let me know if there's anything else you'd like to see changed.

axboe:
Looks good in general, much improved. I'll add a few minor comments.

engines/io_uring.c
```diff
@@ -23,6 +23,10 @@
 #include "../lib/types.h"
 #include "../os/linux/io_uring.h"
 
+#ifndef FIO_RANDSEED
```
axboe: Both here and in libaio - and we have an existing one as well. You should add an extra element to td->rand_seeds[] instead, and have it be initialized in td_fill_rand_seeds_internal() like all the other ones. Then you also get the correct behavior in terms of 'rand_repeat', right now your sequence is always 100% repeatable which is probably not idea. It also means that you can remove some of the init/boiler plate code from libaio and io_uring and keep them in the spot where they belong, outside of the engine themselves.

engines/io_uring.c
```diff
@@ -108,6 +118,29 @@ static struct fio_option options[] = {
 		.category = FIO_OPT_C_ENGINE,
 		.group	= FIO_OPT_G_IOURING,
 	},
+#ifdef FIO_HAVE_IOPRIO_CLASS
```
axboe: This is also something that should not be in here (or libaio)

engines/io_uring.c
```diff
@@ -640,6 +683,22 @@ static int fio_ioring_init(struct thread_data *td)
 	ld->iovecs = calloc(td->o.iodepth, sizeof(struct iovec));
 
 	td->io_ops_data = ld;
+
+	/*
+	 * Initialize hiprio_percent mechanisms if relevant
+	 */
+	if (!(fio_option_is_set(to, ioprio) ||
```
axboe: With the above suggestion, this can go away and you're left with just checking for the conflicting options. Ditto in libaio.

io_u.h
```diff
@@ -30,6 +30,7 @@ enum {
  * The io unit
  */
 struct io_u {
+	uint8_t priority_bit;
```
axboe: This packs horribly, you're added a 7 byte hole here. But this field after 'numberio' instead, then you fill a hole instead of creating a new one and growing the size of the io_u.

iolog.h
```diff
@@ -41,7 +41,8 @@ union io_sample_data {
 struct io_sample {
 	uint64_t time;
 	union io_sample_data data;
-	uint32_t __ddir;
+	uint32_t __ddir :31;
```
axboe: Don't like this at all! Instead of playing games with this, just add a separate field. There's a 4 byte hole between __ddir and bs anyway.

axboe:
Added the comments, got caught up in other stuff yesterday. Almost there I think.

trinof:
I think I've addressed all of your latest concerns, but let me know if there's anything else that should be changed.

trinof:
I've renamed the option to `cmdprio_percentage` and removed the duplicate engine documentation. Let me know if there's anything else that you'd like to see changed.

stat.c
```diff
@@ -482,9 +482,12 @@ static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
 		display_lat("clat", min, max, mean, dev, out);
 	if (calc_lat(&ts->lat_stat[ddir], &min, &max, &mean, &dev))
 		display_lat(" lat", min, max, mean, dev, out);
+	if (ddir == DDIR_READ && calc_lat(&ts->clat_high_prio_stat, &min, &max, &mean, &dev))
```
axboe: Not sure I follow why you gate all of these on DDIR_READ?

stat.c
```diff
@@ -2782,7 +2969,7 @@ void add_clat_sample(struct thread_data *td, enum fio_ddir ddir,
 		if (!hw->hist_last)
 			hw->hist_last = elapsed;
 		this_window = elapsed - hw->hist_last;
-
```
axboe: Should go in the other patch.

engines/io_uring.c
```diff
@@ -343,6 +365,14 @@ static enum fio_q_status fio_ioring_queue(struct thread_data *td,
 
 	/* ensure sqe stores are ordered with tail update */
 	write_barrier();
+	if (o->cmdprio_percentage && rand_between(&td->prio_state, 0, 99) < o->cmdprio_percentage) {
```
axboe: I'd put this out of line - which you did in libaio, but make it ala:

```
if (o->cmdprio_percentage)
    fio_ioring_prio_prep(bla);
```

for both of them.

engines/io_uring.c
```diff
@@ -343,6 +365,14 @@ static enum fio_q_status fio_ioring_queue(struct thread_data *td,
...
+		dprint(FD_IO, "Enable PRIO\n");
```
axboe: And probably ditch these prio messages for both on/off in both engines.

engines/libaio.c
```diff
@@ -85,6 +112,24 @@ static int fio_libaio_prep(struct thread_data fio_unused *td, struct io_u *io_u)
...
+static void fio_libaio_prio_prep(struct thread_data *td, struct io_u *io_u)
+{
+	struct libaio_options *o = td->eo;
+	if (o->cmdprio_percentage == 0)
+		return;
+	if (rand_between(&td->prio_state, 0, 99) < o->cmdprio_percentage) {
+		dprint(FD_IO, "Enable PRIO\n");
+		io_u->iocb.aio_reqprio = IOPRIO_CLASS_RT << IOPRIO_CLASS_SHIFT;
+		io_u->iocb.u.c.flags |= IOCB_FLAG_IOPRIO;
+		io_u->priority_bit = 1;
```
axboe: Is this correct? From my quick look, you want (1U << 13) set, which means RT priority class. This gets translated into NCQ prio in libata at least, whereas any other class does not.

stat.c
trinof: Current Seagate implementation of handling the priority bit will only affect reads, and my understanding is WD's implementation is the same. There's no guarantee this will be true in the future, but I figured it is better to only track reads for now.

axboe: That limitation on the device front should not be carried into fio. You're setting the priority bit regardless of data direction, you should print the stats similarly as well.

engines/libaio.c
trinof: I can see on the bus trace this sets the priority bit as desired, and it looks like `IOPRIO_CLASS_RT` is 1 and `IOPRIO_CLASS_SHIFT` is 13.

axboe: Sorry was my being blind, thought the priority_bit was the field in the iocb. Yes, the other bit looks correct, you just want priority class RT set with no data value, which is 1U << 13, or CLASS_RT << CLASS_SHIFT as you did.

axboe: But then that begs the question why you add priority_bit, instead of just using an io_u flag?

axboe: Then you also don't have to clear it if prio isn't set.

stat.c
trinof: I'll update it to also collect high/normal priority completion latencies separately for reads, writes and trims. I wonder if it would make sense to be able to set cmdprio_percentage separately for each data direction?

engines/libaio.c
trinof: That's a good idea, I didn't think about adding a new io_u flag.

trinof: I tried leaving the io_u->flags alone when priority is not set, but that seemed to leave the priority flag set after it was set once.

trinof:
I think I addressed all your suggestions in the latest review, although I am clearing the io_u priority flag. I also noticed that the priority_bit field added to the thread_stat structure wasn't really necessary and could be replaced by passing the priority_bit through a couple more calls. Let me know if there are any other changes you'd like to see.

axboe:
For the flag clear, I'd just add that flag to __get_io_u() where it clears the other flags. Don't do it unconditionally in the prep path. The common case is nobody using priorities, ever, so I don't want any extra added memory accesses for that if that can be avoided (and this one easily can). I'll check the patch, but the flag definitely needs to go in __get_io_u().

engines/io_uring.c
```diff
@@ -313,11 +334,23 @@ static int fio_ioring_getevents(struct thread_data *td, unsigned int min,
 	return r < 0 ? r : events;
 }
 
+static void fio_ioring_prio_prep(struct thread_data *td, struct io_u *io_u) {
```
axboe: No brace on the same line, functions must be:

```
type fn(args)
{
}
```

axboe:
Just one minor comment - address that and the flag clear mentioned above, and I think it's good to go.

trinof:
Those two issues have been addressed, thank you for your time and consideration! Hopefully this one can be pulled, but as always let me know if there's anything that should be changed.

axboe:
Pulled, just did one small fixup as your alignments were off in stat.h - but done! Thanks for pulling this to completion.

Issue #732, opened 2019-01-09
verify_only option fails with verify: bad header numberio 1, wanted 0


tomerg-lb:
**Created the following configuration file: write_to_file.fio**
[write_with_verify]
filename=/tmp/fiofile
direct=1
rw=write
bs=4096
verify=md5
size=128k
do_verify=1

**I then run:**
fio write_to_file.fio

**Result:**
write_with_verify: (g=0): rw=write, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
fio-3.11
Starting 1 process

write_with_verify: (groupid=0, jobs=1): err= 0: pid=4205: Wed Jan  9 17:43:43 2019
   read: IOPS=6400, BW=25.0MiB/s (26.2MB/s)(128KiB/5msec)
    clat (usec): min=106, max=164, avg=113.44, stdev=11.66
     lat (usec): min=107, max=165, avg=113.96, stdev=11.68
    clat percentiles (usec):
     |  1.00th=[  108],  5.00th=[  108], 10.00th=[  108], 20.00th=[  109],
     | 30.00th=[  109], 40.00th=[  109], 50.00th=[  109], 60.00th=[  111],
     | 70.00th=[  113], 80.00th=[  114], 90.00th=[  127], 95.00th=[  135],
     | 99.00th=[  165], 99.50th=[  165], 99.90th=[  165], 99.95th=[  165],
     | 99.99th=[  165]
  write: IOPS=6400, BW=25.0MiB/s (26.2MB/s)(128KiB/5msec)
    clat (usec): min=108, max=415, avg=134.05, stdev=56.11
     lat (usec): min=119, max=442, avg=145.99, stdev=58.65
    clat percentiles (usec):
     |  1.00th=[  110],  5.00th=[  110], 10.00th=[  110], 20.00th=[  111],
     | 30.00th=[  111], 40.00th=[  115], 50.00th=[  118], 60.00th=[  122],
     | 70.00th=[  131], 80.00th=[  137], 90.00th=[  161], 95.00th=[  215],
     | 99.00th=[  416], 99.50th=[  416], 99.90th=[  416], 99.95th=[  416],
     | 99.99th=[  416]
  lat (usec)   : 250=98.44%, 500=1.56%
  cpu          : usr=25.00%, sys=0.00%, ctx=65, majf=0, minf=14
[...]

PR #905, opened 2020-01-14
verify: Fix test to not check for numberio when verify_only is true

gwendalcr:
io->numberio can not be populated when verify_only is true, because
do_dry_run() build and complete IOs immediately, so it can not replicate
the numberio that was produced when the data was layered on the media.

Without this fix, using write_random
[write_stress]
filename=${FILENAME}
size=${FILESIZE}
verify_only=${VERIFY_ONLY}
readwrite=randwrite
bs=4k
ioengine=libaio
iodepth=32
direct=1
do_verify=1
verify=crc32c

'VERIFY_ONLY=1 FILENAME=/dev/sda1 FILESIZE=1M fio write_random' passes,
but
'VERIFY_ONLY=0 FILENAME=/dev/sda1 FILESIZE=1M fio write_random' fails:
"""verify_only option fails with verify: bad header numberio 1, wanted
0""".
The fix addresses the problem by not checking numberio.

Signed-off-by: Gwendal Grignou <gwendal@chromium.org> #732

### DIFF
--- a/verify.c
+++ b/verify.c
@@ -845,13 +845,11 @@ static int verify_header(struct io_u *io_u, struct thread_data *td,
 	 * For read-only workloads, the program cannot be certain of the
 	 * last numberio written to a block. Checking of numberio will be
 	 * done only for workloads that write data.  For verify_only,
-	 * numberio will be checked in the last iteration when the correct
-	 * state of numberio, that would have been written to each block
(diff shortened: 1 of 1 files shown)


gwendalcr:
Ignore, I have a better patch, to populate numberio in do_dry_run() properly.

PR #910, opened 2020-01-18
Fix several additional Coverity complaints

bvanassche:


sitsofe:
Thanks for submitting these Coverity fixes Bart!

PR #912, opened 2020-01-23
engines/filestat: add "lstat" option to use lstat(2)

kusumi:
Use `stat(2)` by default, but `lstat(2)` if "lstat" bool option specified.


kusumi:
But there looks to be no way for this ioengine to stat symlink...

axboe:
I pulled this, but in hindsight, we might want to make this a posval option, and have "stat" be the default, "lstat" available. This would allow for addition of eg statx, without having potentially conflicting options.

PR #926, opened 2020-02-07
client: Improve handle_job_opt()

bvanassche:


bvanassche:
Does anyone want to comment on this pull request?

client.c
```diff
@@ -1140,37 +1140,30 @@ static void handle_gs(struct fio_client *client, struct fio_net_cmd *cmd)
...
-		const char *pos = "";
+		if (!job_opt_object)
+			return;
 
-		if (p->value)
-			pos = p->value;
-
-		json_object_add_value_string(job_opt_object, p->name, pos);
+		json_object_add_value_string(job_opt_object,
+					     strdup((char *)pdu->name) ? : "",
```
sitsofe, line 1153: If we want to match was was there before don't we want to pass NULL if duplication of `pdu->name` couldn't allocate memory?

sitsofe: I'm left wondering how the duplicated name could ever be freed because it might have been dynamically allocated or it might be constant (if it is the empty string we pass here). Having said that, this is how the code was before so I guess there's no harm?

sitsofe:
@bvanassche Sorry for the delay - this one slipped away...

client.c
bvanassche: > If we want to match was was there before don't we want to pass NULL if duplication of `pdu->name` couldn't allocate memory?

I think not passing NULL to the `name` argument of `json_object_add_value_string()` is a bug fix. The `name` argument is passed indirectly to `strdup()` and the `strdup()` argument must not be NULL.

bvanassche: > I'm left wondering how the duplicated name could ever be freed because it might have been dynamically allocated or it might be constant (if it is the empty string we pass here). Having said that, this is how the code was before so I guess there's no harm?

Is this perhaps the code that frees the memory allocated by the code touched by this patch?
```
	json_free_object(root);
```

sitsofe: @bvanassche My guess is that it it would be done by `json_free_value()` over in https://github.com/axboe/fio/blob/master/json.c#L155 as that is (eventually) called by functions called from `json_free_object()`. But again how is that meant to handle the constant string `""`? Do we simply never try and free these objects?

sitsofe: > I think not passing NULL to the `name` argument of `json_object_add_value_string()` is a bug fix. The `name` argument is passed indirectly to `strdup()` and the `strdup()` argument must not be NULL.

Fair enough. Tracing through `strdup_escape()` shows it has a call to `strlen()` early on and that call expects its argument to not be NULL.

bvanassche: > @bvanassche My guess is that it it would be done by `json_free_value()` over in https://github.com/axboe/fio/blob/master/json.c#L155 as that is (eventually) called by functions called from `json_free_object()`. But again how is that meant to handle the constant string `""`? Do we simply never try and free these objects?

Please take a look at the new patch series that has been pushed. An additional memory leak has been fixed.

sitsofe:
@axboe this one LGTM!

PR #929, opened 2020-02-13
Makefile: don't fail to remove conditionally compiled binaries on clean

kusumi:
`rm -f $(T_PROGS) $(T_TEST_PROGS)` in `clean` target is intended to
remove all binaries under t directory, but it actually fails to remove
`t/{fio-btrace2fio,io_uring,read-to-pipe-async}` files for not
evaluating platform specific variables on `make clean`.

```
 # uname -r
 5.4.17-200.fc31.x86_64
 # make clean
 # file t/* | grep ELF | awk '{print $1,$2,$3,$4,$5}'
 t/fio-btrace2fio: ELF 64-bit LSB executable,
 t/io_uring: ELF 64-bit LSB executable,
 t/read-to-pipe-async: ELF 64-bit LSB executable,
```

The easiest way to fix this is to hard-code these conditionally
compiled binaries.


sitsofe:
@axboe This one seems reasonable to me...

PR #943, opened 2020-03-15
io_uring: only use rand48_r functions for glibc

neheb:
These break compilation under musl. They are also glibc specific.

### DIFF
--- a/t/io_uring.c
+++ b/t/io_uring.c
@@ -63,7 +63,9 @@ struct file {
 struct submitter {
 	pthread_t thread;
 	int ring_fd;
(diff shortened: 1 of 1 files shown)


axboe:
This silently breaks seeding, for repeated runs or for verification purposes.

neheb:
That's unfortunate.

axboe:
Ah I'm blind, it's just t/io_uring. I'll fix this up a bit differently, we should just use the local helpers.

axboe:
Killed it, closing this one.

PR #956, opened 2020-04-09
rdmaio: add multithread support

dmonakhov:
Use dedicated port for each thread similar to netio engine
Let's make RDMA engine first class citizen


dmonakhov:
Please ignore this PR, I've prepared a bunch of other cleanups and bugfixes. So i'll create new one.

PR #958, opened 2020-04-13
configure/Makefile: don't override user CFLAGS

Hi-Angel:
It is a usual practice to build sw by passing `CFLAGS="-foo"` on
configure stage. It didn't work with FIO though. This commit fixes two
problems:

* configure: this script was overriding user CFLAGS
* Makefile: this script was appending its own CFLAGS instead of
  prepending them. The problem with this one is that it sets a -O3
option, but a user may have wanted to disable optimization, so they set
-O0 option. And by appending our CFLAGS we make user CFLAGS to not work.


axboe:
configure does have --extra-cflags for adding CFLAGS, but I agree it's less surprising to setting CFLAGS.

PR #959, opened 2020-04-14
RDMAIO fixes

dmonakhov:
Hi, please take a look at this PR,
It's primary idea is to make rdmaio engine alive. 
# Cleanup and fixes
0c17458 rdmaio: cleanup rdma_poll_wait
353d133 rdmaio: add helper function to detect control messages
0788eae rdmaio: allow random io semantics
0ba8446 rdmaio: fix verb=recv case
# New features
693d60d rdmaio: add multithread support
# New feature which break compatibility with old configs
36b6f50 rdmaio: enable fio-native io-direction semantics

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1900,8 +1900,15 @@ I/O engine
 		**rdma**
 			The RDMA I/O engine supports both RDMA memory semantics
 			(RDMA_WRITE/RDMA_READ) and channel semantics (Send/Recv) for the
-			InfiniBand, RoCE and iWARP protocols. This engine defines engine
-			specific options.
+			InfiniBand, RoCE and iWARP protocols. This engine defines specific
+			io derection transformation matrix:
+			With :option: `direct` enbled:
+				`DDIR_READ`  :  `IBV_WC_RDMA_READ`
+				`DDIR_WRITE` :  `IBV_WC_RDMA_WRITE`
+			With options.direct disabled:
+				`DDIR_READ`  :  `IBV_WC_RECV`
+				`DDIR_WRITE` :  `IBV_WC_SEND`
+			This engine defines engine specific options.
 
 		**falloc**
 			I/O engine that does regular fallocate to simulate data transfer as
@@ -2169,11 +2176,13 @@ with the caveat that when used on the command line, they must come after the
 	hostname if the job is a TCP listener or UDP reader. For unix sockets, the
 	normal :option:`filename` option should be used and the port is invalid.
 
-.. option:: listen : [netsplice] [net]
+.. option:: listen : [netsplice] [net] [rdma]
 
 	For TCP network connections, tell fio to listen for incoming connections
 	rather than initiating an outgoing connection. The :option:`hostname` must
 	be omitted if this option is used.
+	For RDMA engine, tell fio to listen for incoming RDMA-CM connection,
+	rather than initiating an outgoing connection.
 
 .. option:: pingpong : [netsplice] [net]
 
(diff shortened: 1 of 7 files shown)


dmonakhov:
Original PR introduce of-by-one array access https://travis-ci.org/github/dmonakhov/fio/jobs/674809391
PR updated

axboe:
You should at least update the existing example rdma job files as well, since they will no longer work.

dmonakhov:
Ок, will do it

sitsofe:
@dmonakhov any luck with Jens' suggestions?

dmonakhov:
Sorry, was busy by other tasks. I'll update pr this week

sitsofe:
@dmonakhov ping?

dmonakhov:
Not sure I've understand last comment. Should I fix  the commit message?

vincentkfu:
The comment was to point out a misspelling in the patch where direction was spelled as "derection"

dmonakhov:
Ops. It is appeared that I've forget to convert port to network order. Please take a look at updated version

dmonakhov:
Ping, do you have a change to take a look at new version?

engines/rdma.c
```diff
@@ -1029,19 +1053,37 @@ static int fio_rdmaio_close_file(struct thread_data *td, struct fio_file *f)
 	return 0;
 }
 
-static int aton(struct thread_data *td, const char *host,
-		     struct sockaddr_in *addr)
+static int fill_sockaddr(struct thread_data *td, const char *host,
+			 struct sockaddr_storage *sin, int af, unsigned short port)
```
sitsofe, line 1057: Looks like the style in this file for multi-lines is indenting with the bracket - https://github.com/axboe/fio/blob/ba8fe8672c82e2ebaa7f5ccb74d079933cc20eed/engines/rdma.c#L237 so would you be able to rewrap this?

HOWTO
```diff
@@ -2245,13 +2254,6 @@ with the caveat that when used on the command line, they must come after the
...
-.. option:: verb=str : [rdma]
-
-	The RDMA verb to use on this side of the RDMA ioengine connection. Valid
-	values are write, read, send and recv. These correspond to the equivalent
-	RDMA verbs (e.g. write = rdma_write etc.). Note that this only needs to be
-	specified on the client side of the connection. See the examples folder.
-
 .. option:: bindname=str : [rdma]
 
 	The name to use to bind the local RDMA-CM connection to a local RDMA device.
```
sitsofe, line 2259: You got the HOWTO but you skipped the man page (`fio.1`, yeah I know)...

examples/rdmaio-client.fio
```diff
@@ -1,13 +1,16 @@
 # Example rdma client job
 [global]
 ioengine=rdma
-hostname=[hostname]
-port=[port]
-verb=[read/write/send/recv]
+hostname=${RDMA_SERVER}
+port=8888
+# Data exchange method
+# verb=[read/write/send/recv]
```
sitsofe: Did you say verb has gone away as an option in the previous commit?

engines/rdma.c
dmonakhov: I'm not sure I understand this peace, can you please elaborate.

sitsofe: Oops, I missed this. For long lines that get split we normally line up at the bracket of the previous line so I think it would be better as:
```
static int fill_sockaddr(struct thread_data *td, const char *host, struct
                         sockaddr_storage *sin, int af, unsigned short port)
```

Notice where the s of sockaddr_storage is compared to the bracket of the previous line.

Issue #971, opened 2020-04-24
Repeated timestamps in latency log files


pzoxiuv:
This seems to be the same as https://github.com/axboe/fio/issues/947, but I'm using the latest from master and still seeing the issue.

Config file:
```
[global]
directory=/var/data/
size=64m

[test]
rw=write
time_based
runtime=1m
write_lat_log=/host/fio_out_lat
log_avg_msec=50
```

And here's a segment of the log file:
```
3250, 2407, 1, 0, 0
3300, 2467, 1, 0, 0
3350, 2459, 1, 0, 0
3400, 2502, 1, 0, 0
3450, 2493, 1, 0, 0
3500, 2497, 1, 0, 0
3550, 2471, 1, 0, 0
5471, 194581, 1, 0, 0
5475, 4039507, 1, 0, 0
5475, 13797, 1, 0, 0
5475, 6633, 1, 0, 0
5475, 6187, 1, 0, 0
5475, 5457, 1, 0, 0
...
5475, 5063, 1, 0, 0
5475, 5510, 1, 0, 0
5475, 5537, 1, 0, 0
5500, 5902, 1, 0, 0
5553, 5831, 1, 0, 0
5600, 6004, 1, 0, 0
```
The first several lines are as expected, but then there's 38 lines with the same timestamp.  It seems like maybe the timestamp just isn't being updated?  I think there's the correct number of lines: (5500-3550)/50 = 39

pzoxiuv:
I think the issue is that when [add_log_sample](https://github.com/axboe/fio/blob/master/stat.c#L2711) saves a log entry it always sets the last log time (`iolog->avg_last`) to `iolog->avg_last` plus the sampling window duration (`iolog->avg_msec`), even if more than one sampling window has elapsed.  E.g. if `iolog->avg_last` is 600ms, `iolog->avg_msec` is 50ms, and `add_log_sample` is called at t=800ms, `iolog->avg_last` will be set to 650ms.  So next time `add_log_sample` gets called (probably with t=800ms again) a new log entry will be saved with the same t=800ms timestamp.

One fix would be to change line https://github.com/axboe/fio/blob/master/stat.c#L2754 from
`iolog->avg_last[ddir] = elapsed - (this_window - iolog->avg_msec);`
to
`iolog->avg_last[ddir] = elapsed - (elapsed % iolog->avg_msec);`

sitsofe:
[...]

PR #982, opened 2020-05-12
stat: fix duplicate timestamps in iops/latency/bandwidth logs

pzoxiuv:
add_log_sample() checks if the current time is greater than the last log time
plus the sampling window duration ("log_avg_msec").  If so it records the log
entry and sets the "last log time" to the current "last log time" plus
log_avg_msec.  This is incorrect in cases where more than log_avg_msec has
elapsed between calls to add_log_sample().  The effect is that the very next
time add_log_sample() is called a new log entry will be recorded, possibly with
the same timestamp as the previous log entry.

This fix sets the last log time to the current time, rounded down to the nearest
multiple of log_avg_msec (e.g. if current time is 203 and log_avg_msec is 50 the
last log time will be set to 200, so the next log entry should be recorded at
time=250).  This prevents duplicate log entries.

Fixes https://github.com/axboe/fio/issues/971

### DIFF
--- a/stat.c
+++ b/stat.c
@@ -2739,7 +2739,7 @@ static unsigned long add_log_sample(struct thread_data *td,
 
 	__add_stat_to_log(iolog, ddir, elapsed, td->o.log_max != 0, priority_bit);
 
-	iolog->avg_last[ddir] = elapsed - (this_window - iolog->avg_msec);
+	iolog->avg_last[ddir] = elapsed - (elapsed % iolog->avg_msec);
 	return iolog->avg_msec;
 }
 


sitsofe:
I don't know if this is sufficient... Don't we have to clamp all sample period starting points downwards including the first one? I'm also wondering when fio handles its own stats whether it accounts for periods where no sampling actually took place (although this isn't an issue for the log generation)...

pzoxiuv:
> Don't we have to clamp all sample period starting points downwards including the first one?

I'm not sure what you mean by this?

sitsofe:
I mean, shouldn't we align the very first sample downwards relative the the starting epoch? If the first second has nothing in it then that is somehow significant...

jeffreyalien:
I believe this change along with the changes I have proposed in [https://github.com/axboe/fio/pull/1114] have resolved the issues we've been seeing with the status interval creep.  Could you also review PR# 1114 and comment on the changes?

sitsofe:
@pzoxiuv Do you think this one is going to be solved by https://github.com/axboe/fio/pull/1119 which essentially absorbs your fix?

sitsofe:
Closing because this has been superseded by #1119 .

PR #984, opened 2020-05-13
fix man page and example for latency_window

liu-song-6:
The code uses latency_window as millisecond, but the man page says it is
microsecond. Fix the man page and the example.


liu-song-6:
Or maybe we change the code to microsecond?

PR #987, opened 2020-05-18
fix unit of latency_window

liu-song-6:
latency_window has unit of microseconds, and is compared against
usec_window. Therefore, there is no need to fix it up to nanoseconds.

### DIFF
--- a/init.c
+++ b/init.c
@@ -956,7 +956,6 @@ static int fixup_options(struct thread_data *td)
 	 */
 	o->max_latency *= 1000ULL;
 	o->latency_target *= 1000ULL;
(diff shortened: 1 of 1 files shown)


sitsofe:
This patch LGTM (the only place I see latency_window being compared is in https://github.com/axboe/fio/blob/e1ce5b385c65803b170e512a1b5256bbe99f98f0/io_u.c#L1488 where it is compared against a variable `usec_window`).

@axboe what do you think?

sitsofe:
About the only thing I could say is maybe the commit title should be with `init: fix unit of latency_window`

sitsofe:
@axboe this one LGTM...

PR #988, opened 2020-05-18
Add option latency_run to continue enable latency_target

liu-song-6:
Currently, latency_target run will exist once fio find the highest queue
depth that meets latency_target. Add option latency_run. If set, fio will
continue running and try to meet latency_target by adusting queue depth.

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -2551,6 +2551,13 @@ I/O latency
 	defaults to 100.0, meaning that all I/Os must be equal or below to the value
 	set by :option:`latency_target`.
 
+.. option:: latency_run=bool
+
+	Used with :option:`latency_target`. If false (default), fio will find
+	the highest queue depth that meets :option:`latency_target` and exit. If
+	true, fio will continue running and try to meet :option:`latency_target`
+	by adjusting queue depth.
+
 .. option:: max_latency=time
 
 	If set, fio will exit the job with an ETIMEDOUT error if it exceeds this
--- a/io_u.c
+++ b/io_u.c
@@ -1391,6 +1391,7 @@ static bool __lat_target_failed(struct thread_data *td)
 		td->latency_qd_low--;
 
 	td->latency_qd = (td->latency_qd + td->latency_qd_low) / 2;
+	td->latency_stable_count = 0;
 
 	dprint(FD_RATE, "Ramped down: %d %d %d\n", td->latency_qd_low, td->latency_qd, td->latency_qd_high);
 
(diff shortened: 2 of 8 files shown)


HOWTO
```diff
@@ -2551,6 +2551,12 @@ I/O latency
 	defaults to 100.0, meaning that all I/Os must be equal or below to the value
 	set by :option:`latency_target`.
 
+.. option:: latency_run=bool
+
+	If not set, fio will find the highest queue depth that meets
+	:option:`latency_target` and exit. If set, fio will continue running and
+	try to meet :option:`latency_target` by addjusting queue depth.
```
sitsofe, line 2558: Typo - (you have two ds in `addjusting` when you only need one). Can you also put what does this option defaults to?

io_u.c
```diff
@@ -1439,6 +1440,18 @@ static void lat_target_success(struct thread_data *td)
 
 	td->latency_qd_low = td->latency_qd;
 
+	if (td->latency_qd + 1 == td->latency_qd_high) {
+		td->latency_stable_count++;
+		/*
+		 * Succeed on the same qd for 3 times in a row create more
+		 * head room to increase queue depth.
```
sitsofe, line 1447: I don't quite understand this comment. Is it saying:

> If the same queue depth is seen three times in a row then create more headroom so the depth can continue increasing?

io_u.c
```diff
@@ -1439,6 +1440,18 @@ static void lat_target_success(struct thread_data *td)
 
 	td->latency_qd_low = td->latency_qd;
 
+	if (td->latency_qd + 1 == td->latency_qd_high) {
```
sitsofe, line 1443: This means we would never be OK in the case where the user set a latency maximum and we started out at it but it turned out to be OK?

axboe:
If we're changing the wording, I'd also like to see it changed so that it's clear it only matters if latency_target is set. That's not clear from the current description.

HOWTO
liu-song-6: Will fix. Thanks!

io_u.c
liu-song-6: I should have explained this better. 
For latency_run == 0 case, the behavior is the same as existing code. 
For latency_run == 1 case, we hit this condition when latency_qd_low == latency_qd == (latency_qd_high - 1). With existing code, latency_qd won't increase even if we hit lat_target_success() multiple times. For example, latency_qd == 3 and latency_qd_high == 4. It is very likely that we will hit latency_target with some margin with qd of 3, but we will miss latency_target with qd of 4. With existing code, we will stick with qd of 3 for the rest of the run, and result latency is lower than latency_target. 
This change tries to fix this by allowing latency_qd to increase in such cases. The heuristic here is a little too naive though...

liu-song-6:
@sitsofe @axboe Does this version look good to you? Thanks!

axboe:
Pulled, thanks.

Issue #716, opened 2018-11-16
rate_min is limited to less than 4G


rustyscottweber:
Fio version is 3.6.  Not terribly out of date, but not up to date.

```
fio jobby.fio 
max value out of range: 4294967296 (4294967295 max)
fio: failed parsing rate_min=4G
fio: job global dropped
```
Job file
```
[global]

# read-only bandwith test
# runs for 100 minutes
# fails when 1GB/s is not met

rw=read
runtime=100m
rate_min=12G

log_avg_msec=1000
write_bw_log=ess-ion
write_iops_log=ess-ion

bs=4m
direct=1
disable_bw_measurement=0
disable_clat=1
disable_lat=1
disable_slat=1
exitall=1
group_reporting
iodepth=32
ioengine=libaio
numjobs=1
overwrite=1
size=100%
thread=0
time_based=1
verify_backlog=0

# run for each FUSIONIO LUN

[jobbyjob-sdb]
filename=/dev/sdb
[jobbyjob-sdc]
filename=/dev/sdc
[jobbyjob-sdd]
filename=/dev/sde
[jobbyjob-sde]
filename=/dev/sde
```
We would like to be able to fail our test should we be reading less than 8 Gibibytes from any disk. 32 Gibibytes for the entire system.

sbates130272:
@sitsofe would you prefer a patch to the mailing list or a PR for this?

sitsofe:
PR will be fine.

sbates130272:
@sitsofe please see the referenced PR. This appears to be working for us on our systems.

sitsofe:
(@sbates130272 Just in case it doesn't get seen: do you think `rate_bytes` will be OK (see https://github.com/axboe/fio/pull/990#issuecomment-631003550 ))

axboe:
@sitsofe On 32-bit it'd need a patch too.

sitsofe:
@axboe ah so that's going to mean 64 bit Windows (LLP64?) needs something too right?

axboe:
Yeah, or weird setups like Windows where unsigned long is 32-bit even on 64-bit environments.

sitsofe:
OK, I guess we'd better re-open this one until a few more patches arrive.

axboe:
@sbates130272 Care to fix that one up, please?

sbates130272:
@axboe @sitsofe Looks like this was addressed in #991. Thanks! LMK if you need anything from me in this regard.

[...]

PR #991, opened 2020-05-19
Fix 32-bit/LLP64 platform truncation issues

sitsofe:
- After 140a6888 ("rate: Convert the rate and rate_min options to
  FIO_OPTS_ULL") landed 32-bit/LLP64 platforms need additional changes
  to cope with 64 bit I/O rate values
- The seed is 64 bit but was being being truncated to 32 bits in
  td_fill_rand_seeds_internal() on bit/LLP64 platforms

Prior to this commit when running an fio compiled with

CC=clang-9 ./configure \
  --extra-cflags="-fsanitize=undefined,implicit-integer-truncation "
  "-fno-builtin"

using this job

./fio --ioengine=null --bs=1M --rate=6G --rate_min=5G --name=test --size=100G

warnings like the following were produced

init.c:996:27: runtime error: implicit conversion from type 'uint64_t' (aka 'unsigned long long') of value 5942511153023025289 (64-bit, unsigned) to type 'unsigned int' changed the value to 2914779273 (32-bit, unsigned)
[..]
backend.c:212:25: runtime error: implicit conversion from type 'unsigned long long' of value 12886999040 (64-bit, unsigned) to type 'unsigned long' changed the value to 2097152 (32-bit, unsigned)

inside a 32-bit Ubuntu 18.04 docker container.

Fixes: https://github.com/axboe/fio/issues/716

### DIFF
--- a/fio.h
+++ b/fio.h
@@ -319,7 +319,7 @@ struct thread_data {
 	 */
 	uint64_t rate_bps[DDIR_RWDIR_CNT];
 	uint64_t rate_next_io_time[DDIR_RWDIR_CNT];
-	unsigned long rate_bytes[DDIR_RWDIR_CNT];
+	unsigned long long rate_bytes[DDIR_RWDIR_CNT];
 	unsigned long rate_blocks[DDIR_RWDIR_CNT];
 	unsigned long long rate_io_issue_bytes[DDIR_RWDIR_CNT];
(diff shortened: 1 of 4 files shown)


sitsofe:
This might require some eyeballing as a took a fairly brute force approach to silencing warnings.

axboe:
LGTM, thanks!

PR #996, opened 2020-05-24
Fix parse_format()

bvanassche:
This series of three patches is the result of the analysis of a Coverity complaint about the parse_format() function.


axboe:
Thanks Bart

PR #998, opened 2020-05-27
Unbreak zbd.c build on NetBSD and OpenBSD

kusumi:
Neither platform has `pthread_mutexattr_setpshared(3)`,
so NetBSD and OpenBSD can't support (and have never supported)
`zonemode=zbd` unless pthread part in zbd.c gets implemented without
using it.

Since it's their pthread issue that they haven't implemented
`pthread_mutexattr_setpshared(3)`, just define an NOP macro and fail
on runtime, rather than bringing back compile time detection.

Fixes: b76949618d55 ("fio: Generalize zonemode=zbd")

### DIFF
--- a/filesetup.c
+++ b/filesetup.c
@@ -1268,7 +1268,12 @@ done:
 	td_restore_runstate(td, old_state);
 
 	if (td->o.zone_mode == ZONE_MODE_ZBD) {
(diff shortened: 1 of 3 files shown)


kusumi:
There are *official* patches for this from zbd folks following my below post, so I'll just close this one.
https://www.spinics.net/lists/fio/msg08600.html

Issue #890, opened 2020-01-02
test failures with t0010-b7aae4ba.fio


vincentkfu:
Appveyor and Travis tests having been failing with t0010-b7aae4ba.fio:
```
# Expected result: fio runs and completes the job
# Buggy result: fio segfaults
#
[test]
ioengine=null
size=10g
io_submit_mode=offload
iodepth=16
```

Previously this only happened on Windows but there have been recent failures on other platforms. Failures occurring on multiple platforms suggests that there is a genuine problem.

- Windows failures (both x86 and x64)
  - https://ci.appveyor.com/project/axboe/fio/builds/29724441/job/5yc42vujc3y46r5b
  - https://ci.appveyor.com/project/axboe/fio/builds/29704848/job/4n6fq1rct22d70y3
  - https://ci.appveyor.com/project/axboe/fio/builds/29643383/job/lswsoca5laxj7vw6
  - https://ci.appveyor.com/project/vincentkfu/fio/builds/29688203/job/nc776ub9dt43w3qw
  - https://ci.appveyor.com/project/vincentkfu/fio/builds/29578210/job/vj083pvj5nma9q9k
  - https://ci.appveyor.com/project/vincentkfu/fio/builds/29342164/job/43q67r342g6wyd1f
- macOS failures
  -  https://travis-ci.org/axboe/fio/jobs/631746909
  - https://travis-ci.org/axboe/fio/jobs/635404562
- Linux failure
  - https://travis-ci.org/axboe/fio/jobs/633377483
  - https://travis-ci.org/axboe/fio/jobs/647005719

vincentkfu:
The AppVeyor builds run configure with the --disable-native option, preventing the addition of march=native to CFLAGS. My own automated Windows builds typically do not run configure with --disable-native.

When I built fio *with* --disable-native, I saw one t0010 failure in 100 trials . Without --disable-native, I saw no failures in 100 trials.

disable-native  | march=native | Failures | Trials
---------------- | ------------ | -------- | ------
 yes | no | 1 | 100
 no | yes | 0 | 100

bcran:
I'm running a set of tests now. So far 250 runs with no failures (built _with_ `--disable-native`).

bcran:
I got a failure after a few hundred runs, which confirmed that I _can_ replicate the problem on my ThreadRipper system. I'll run a script overnight which runs fio via gdb.

bcran:
It looks like it's getting to the end and hanging, waiting for threads to exit.

[...]

PR #999, opened 2020-05-27
Testing improvements

vincentkfu:
Jens, please consider this pull request which does the following:

- upload AppVeyor test artifacts whether tests fail or pass instead of only when they all pass 
- enable TravisCI arm64 builds

arm64 builds are working for my fio fork: https://travis-ci.com/github/vincentkfu/fio/builds/168409097

TravisCI says that arm64 support is still at an [alpha](https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support) stage and there have been some issues only [recently resolved](https://travis-ci.community/t/disk-quota-exceeded-on-arm64/7619/11). So adding arm64 builds may cause some test noise, but it is useful to support an additional platform.


vincentkfu:
The gcc arm64 build for this PR failed because of https://github.com/axboe/fio/issues/890

https://travis-ci.org/github/axboe/fio/builds/691840772

axboe:
LGTM!

PR #1003, opened 2020-05-30
Fix two memory leaks in the client code

bvanassche:


client.c
```diff
@@ -486,6 +486,10 @@ int fio_client_add(struct client_ops *ops, const char *hostname, void **cookie)
 	dprint(FD_NET, "client: added <%s>\n", client->hostname);
 	*cookie = client;
 	return 0;
+
+err:
+	fio_put_client(client);
+	return -1;
```
sitsofe, line 491: I'm curious - why put it here and not inline? The regular code doesn't fall through to it...

bvanassche: In the Linux kernel there is a habit of grouping all error handling code at the end of a function. Anyway, I have modified the code as suggested.

PR #1015, opened 2020-06-12
fio_generate_plots: fix scale on gnuplot graphs

avlasov-mos-de:
Since log files for latencies have values in nanoseconds in order to get microseconds we divide by 1000
and for milliseconds by 1000000


sitsofe:
@axboe this one looks fine to me. The only nit-pick is there's no no [SoB](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) and the wrapping on the commit message looks a bit weird but those are somewhat minor...

PR #1016, opened 2020-06-13
Fix a potentially infinite loop in check_overlap()

bvanassche:


bvanassche:
@vincentkfu , do you want to have a look at the patches in this pull request?

vincentkfu:
This does not mean that there isn't one, but I don't see the bug. If there is no overlap, the `for_each_td()` loop finishes with `overlap=false`, and the `do/while` loops ends as well. If overlap is found, we break out of the `for_each_td()` loop with the mutex unlocked and re-acquire the lock in the next iteration of the `do/while` loop.

Actually one improvement I can see is to set `overlap=false` at the top of the `do/while` loop in case this is not the first time through and we continue in each iteration of `for_each_td()`.

I added that line, restored the original locking, and ran the following job with the assertions added in your second patch and saw no problems. I also saw no problems using  your patches unmodified.

```
[global]
bs=4k
ioengine=null
size=100g
runtime=10
io_submit_mode=offload
serialize_overlap=1
numjobs=64
group_reporting=1

[flow]
```
Incidentally, for `t/jobs/t0011` to trigger this code `io_submit_mode=overlap` needs to be added. My changes are at https://github.com/vincentkfu/fio/tree/overlap.

bvanassche:
Thanks for the feedback. I have updated the commit message of patch 1/3 to make it clear that that patch is not a bug fix. Please take another look.

vincentkfu:
1. I would feel better about a new test job rather than modifying an existing one. The current jobs should produce an error in any commit before the hash in the filename and should run without issue for any hash at or after the one in the filename. In other words t0011 works with 5d2788d5 or later but produces an error with commits before 5d2788d5. Adding extra options to existing jobs probably doesn't do any harm but may make things messy if jobs are used to detect multiple bugs.
2. To exercise the code in question, both io_submit_mode=overlap and serialize_overlap=1 need to be included.
3. I like the error checking in patch 2/3, although it would be nice to know how much it hurts performance.
4. Consider adding overlap=false at the top of the do/while loop. I can envision a scenario where its absence will cause an endless loop.
5. As for patch 1/3, I have mixed feelings. As far as I can tell the current code seems to work correctly (apart from the issue in item 4). Patch 1/3 seems to be equivalent. The benefit of the patch is to silence a coverity complaint. The advantage to the current code is that the colleagues I developed the feature for captured a bus trace and confirmed that there was no overlap. I don't know that there is the wherewithal to do this testing again, although it might be possible to enable debug=io and confirm the absence of overlap by analyzing the log.

bvanassche:
Thanks again for the feedback. Based on that feedback the entire patch series has been reworked. Please take another look.

Regarding the overhead of the error checking mutex: I expect that overhead to be much smaller than the time spent in the loop inside check_overlap().

vincentkfu:
The patches look good to me. However, the Windows tests are failing with the new t/jobs/t0012.fio.

bvanassche:
Reopening this pull request.

sitsofe:
LGTM but I do wonder why test 13 is now failing... Is the MinGW's POSIX mutex implementation deficient?

bvanassche:
Thanks @sitsofe for having drawn my attention to the test failure on Windows. I had not yet noticed that failure. Since I couldn't reproduce the failure locally in a Windows VM, I ran a sent several tests to the Appveyor CI server. Reverting the error checking mutex change was not sufficient to make the overlap test pass on Appveyor. Upgrading the Appveyor image however makes that test pass :-)

bvanassche:
@sitsofe I think I have found the root cause of the t0013 hangs. See also https://github.com/axboe/fio/pull/1034.

PR #1021, opened 2020-06-19
Zone Append command support for NVMe Zoned Namespaces (ZNS)

krish-reddy:
Add Zone Append command support for NVMe Zoned Namespaces (ZNS) defined in NVM Express TP4053.

1. Added a new FIO option zone_append.
    When zone_append option is enabled, the existing write path will send Zone Append command with LBA offset as start of the Zone.
2. libaio: support for Zone Append command in libaio IO engine
	(New Opcode added : IO_CMD_ZONE_APPEND [9])
3. iouring: support for Zone Append command in io_uring IO engine
	(New Opcodes added : IORING_OP_ZONE_APPEND [34], IORING_OP_ZONE_APPENDV [35] and IORING_OP_ZONE_APPEND_FIXED [36])
4. t/zbd: Add support to verify Zone Append (ZNS)) command with libaio, io_uring IO engine tests

Corresponding code changes in Linux Kernel are submitted for review in LKML, as part of the below patchset
https://lkml.org/lkml/2020/6/17/801
https://lkml.org/lkml/2020/6/17/802
https://lkml.org/lkml/2020/6/17/803
https://lkml.org/lkml/2020/6/17/804

This is a RFC and the Linux Kernel interface is under submission and review.
Feedback / Comments will help in improving this further.

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1010,6 +1010,14 @@ Target file/device
 	:option:`zonesize` bytes of data have been transferred. This parameter
 	must be zero for :option:`zonemode` =zbd.
 
+.. option:: zone_append=bool
+
+	This parameter applies to :option:`zonemode` =zbd only.
+
+	For :option:`rw` =write, or :option: `rw` =randwrite the request
+	points to start LBA of the zone. On successful completion the device
+	returns the offset where data has been placed.
+
 .. option:: read_beyond_wp=bool
 
 	This parameter applies to :option:`zonemode` =zbd only.
--- a/engines/libaio.c
+++ b/engines/libaio.c
@@ -16,6 +16,10 @@
 #include "../optgroup.h"
 #include "../lib/memalign.h"
 
+#ifndef IO_CMD_ZONE_APPEND
+#define IO_CMD_ZONE_APPEND	9
+#endif
+
 /* Should be defined in newest aio_abi.h */
 #ifndef IOCB_FLAG_IOPRIO
 #define IOCB_FLAG_IOPRIO    (1 << 1)
@@ -123,7 +127,13 @@ static int fio_libaio_prep(struct thread_data *td, struct io_u *io_u)
 		if (o->nowait)
 			iocb->aio_rw_flags |= RWF_NOWAIT;
 	} else if (io_u->ddir == DDIR_WRITE) {
-		io_prep_pwrite(iocb, f->fd, io_u->xfer_buf, io_u->xfer_buflen, io_u->offset);
+		if ((td->o.zone_mode == ZONE_MODE_ZBD || td->o.zone_mode == ZONE_MODE_STRIDED) &&
+		     td->o.zone_append) {
+			io_prep_pwrite(iocb, f->fd, io_u->xfer_buf, io_u->xfer_buflen, io_u->zone_slba);
+			iocb->aio_lio_opcode = IO_CMD_ZONE_APPEND;
+		}
+		else
+			io_prep_pwrite(iocb, f->fd, io_u->xfer_buf, io_u->xfer_buflen, io_u->offset);
 		if (o->nowait)
 			iocb->aio_rw_flags |= RWF_NOWAIT;
 	} else if (ddir_sync(io_u->ddir))
@@ -228,6 +238,20 @@ static int fio_libaio_getevents(struct thread_data *td, unsigned int min,
 		} else {
 			r = io_getevents(ld->aio_ctx, actual_min,
 				max, ld->aio_events + events, lt);
+			if (td->o.zone_mode == ZONE_MODE_ZBD
+			    || td->o.zone_mode == ZONE_MODE_STRIDED) {
+				struct io_event *ev;
+				struct io_u *io_u;
+				for (unsigned event = 0; event < r; event++) {
+					ev = ld->aio_events + event;
+					io_u = container_of(ev->obj, struct io_u, iocb);
+					if (td->o.zone_append
+					    && td->o.do_verify && td->o.verify
+					    && (io_u->ddir== DDIR_WRITE)) {
+						io_u->ipo->offset = io_u->zone_slba + ev->res2;
+					}
+				}
+			}
 		}
 		if (r > 0)
 			events += r;
--- a/fio.1
+++ b/fio.1
@@ -782,6 +782,12 @@ sequential workloads and ignored for random workloads. For read workloads,
 see also \fBread_beyond_wp\fR.
 
 .TP
+.BI zone_append
+For \fBrw\fR=write and \fBrw\fR=randwrite requests if zone_append is enabled,
+the request points at the start LBA of the zone. On successful completion
+the device will return the offset where the data has been placed.
+
+.TP
 .BI read_beyond_wp \fR=\fPbool
 This parameter applies to \fBzonemode=zbd\fR only.
 
--- a/io_u.h
+++ b/io_u.h
@@ -94,19 +94,25 @@ struct io_u {
 	};
 
 	/*
+	 * ZBD Zone Append Support
+	 */
+	unsigned long long zone_slba;
+
+	/*
 	 * ZBD mode zbd_queue_io callback: called after engine->queue operation
 	 * to advance a zone write pointer and eventually unlock the I/O zone.
 	 * @q indicates the I/O queue status (busy, queued or completed).
 	 * @success == true means that the I/O operation has been queued or
 	 * completed successfully.
 	 */
-	void (*zbd_queue_io)(struct io_u *, int q, bool success);
+	void (*zbd_queue_io)(struct thread_data *, struct io_u *, int q,
+			     bool success);
 
 	/*
 	 * ZBD mode zbd_put_io callback: called in after completion of an I/O
 	 * or commit of an async I/O to unlock the I/O target zone.
 	 */
-	void (*zbd_put_io)(const struct io_u *);
+	void (*zbd_put_io)(struct thread_data *, const struct io_u *);
 
 	/*
 	 * Callback for io completion
--- a/os/linux/io_uring.h
+++ b/os/linux/io_uring.h
@@ -1,4 +1,4 @@
-/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
+/* SPDX-License-Identifier: (GPL-2.0 WITH Linux-syscall-note) OR MIT */
 /*
  * Header file for the io_uring interface.
  *
@@ -23,7 +23,10 @@ struct io_uring_sqe {
 		__u64	off;	/* offset into file */
 		__u64	addr2;
 	};
-	__u64	addr;		/* pointer to buffer or iovecs */
+	union {
+		__u64	addr;	/* pointer to buffer or iovecs */
+		__u64	splice_off_in;
+	};
 	__u32	len;		/* buffer size or number of iovecs */
 	union {
 		__kernel_rwf_t	rw_flags;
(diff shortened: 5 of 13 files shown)


sitsofe:
@damien-lemoal Could you cast your eye over this one? I'm a little worried about just how many paths this touches...

@krish-reddy Re your profile - I admire your enthusiasm :-)

HOWTO
```diff
@@ -1010,6 +1010,14 @@ Target file/device
 	:option:`zonesize` bytes of data have been transferred. This parameter
 	must be zero for :option:`zonemode` =zbd.
 
+.. option:: zone_append=bool
+
+	This parameter applies to :option:`zonemode` =zbd only.
+
+	For :option:`rw` =write, or :option: `rw` =randwrite the request
+	points to start LBA of the zone. On successful completion the device
```
damien-lemoal, line 1018: Please use fio terms: request -> io_u, LBA -> offset etc.

fio.1
```diff
@@ -781,6 +781,12 @@ zone have been read (read workloads). This parameter is valid only for
 sequential workloads and ignored for random workloads. For read workloads,
 see also \fBread_beyond_wp\fR.
 
+.TP
+.BI zone_append
+For \fBrw\fR=write and \fBrw\fR=randwrite requests if zone_append is enabled,
+the request points at the start LBA of the zone. On successful completion
+the device will return the offset where the data has been placed.
```
damien-lemoal, line 788: See previous comment.

io_u.h
```diff
@@ -93,20 +93,26 @@ struct io_u {
 		struct workqueue_work work;
 	};
 
+	/*
+	 * ZBD Zone Append Support
+	 */
+	unsigned long long zone_slba;
```
damien-lemoal, line 99: offset, not lba

zbd.c
```diff
@@ -1283,8 +1295,24 @@ static void zbd_put_io(const struct io_u *io_u)
 	       "%s: terminate I/O (%lld, %llu) for zone %u\n",
 	       f->file_name, io_u->offset, io_u->buflen, zone_idx);
 
-	ret = pthread_mutex_unlock(&z->mutex);
-	assert(ret == 0);
+	if (td->o.zone_append && z->pending_ios > 0) {
+		pthread_mutex_lock(&z->mutex);
+		z->pending_ios--;
```
damien-lemoal, line 1300: pending_ios is updated under a lock but tested outside of the lock... This smells like atomicity problems for some arch... shouldn't pending_ios be an atomic (and then beware of compilation issues for other OSes/compilers).

zbd.c
```diff
@@ -1283,8 +1295,24 @@ static void zbd_put_io(const struct io_u *io_u)
...
+		if (!z->pending_ios)
+			pthread_cond_broadcast(&z->reset_cond);
+		ret = pthread_mutex_unlock(&z->mutex);
+		assert(ret == 0);
+	}
+
+	if (!td->o.zone_append) {
+		ret = pthread_mutex_unlock(&z->mutex);
+		assert(ret == 0);
+	}
```
damien-lemoal, line 1314: The previous hunk could be moved here as "} else if (z->pending_ios > 0) {". That will simplify reading and reduce condition tests.

zbd.c
```diff
@@ -1524,16 +1552,65 @@ enum io_u_action zbd_adjust_block(struct thread_data *td, struct io_u *io_u)
...
+proceed:
+		/*
+		 * Check for zone full condition again. For zone append request
+		 * the zone may already be reset, written and full while we
+		 * were waiting for our turn
+		 */
+		if (zbd_zone_full(f, zb, min_bs)) {
+			goto reset;
+		}
+
```
damien-lemoal, line 1606: The additional wait label and goto are not nice... There is reset and proceed already, adding one more makes the code like spaghetti and hard to read. This needs refactoring first.

Also, for the regular write case, it looks like nobody is signaling the condition, so the pthread_cond_wait may block forever, no ?

zbd.c
```diff
@@ -1524,16 +1552,65 @@ enum io_u_action zbd_adjust_block(struct thread_data *td, struct io_u *io_u)
...
 		/* Make writes occur at the write pointer */
 		assert(!zbd_zone_full(f, zb, min_bs));
 		io_u->offset = zb->wp;
+
+		if (td->o.zone_append)
+			io_u->zone_slba = zb->start;
```
damien-lemoal, line 1612: Why not use io_u->offset ? Is that because of bad interactions with random map etc ?

os/linux/io_uring.h
```diff
@@ -1,4 +1,4 @@
-/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
+/* SPDX-License-Identifier: (GPL-2.0 WITH Linux-syscall-note) OR MIT */
```
damien-lemoal, line 1: This change seems unrelated to zone append and probably should be a different patch with a commit message justifying it.

os/linux/io_uring.h
```diff
@@ -185,10 +265,32 @@ struct io_uring_params {
...
+	__u16 flags;	/* IO_URING_OP_* flags */
+	__u32 resv2;
+};
+
+struct io_uring_probe {
+	__u8 last_op;	/* last opcode supported */
+	__u8 ops_len;	/* length of ops[] array below */
+	__u16 resv;
+	__u32 resv2[3];
+	struct io_uring_probe_op ops[0];
```
damien-lemoal, line 293: All of the above changes do not look like they are specific to ZNS but rather simply complete fio support for the full iouring interface. So the commit title is not correct (for some reasons, I fail to see the entire commit message using github web interface)

engines/libaio.c
```diff
@@ -123,7 +127,13 @@ static int fio_libaio_prep(struct thread_data *td, struct io_u *io_u)
 		if (o->nowait)
 			iocb->aio_rw_flags |= RWF_NOWAIT;
 	} else if (io_u->ddir == DDIR_WRITE) {
-		io_prep_pwrite(iocb, f->fd, io_u->xfer_buf, io_u->xfer_buflen, io_u->offset);
+		if ((td->o.zone_mode == ZONE_MODE_ZBD || td->o.zone_mode == ZONE_MODE_STRIDED) &&
```
damien-lemoal, line 130: You initially mentioned that zoneappend option is valid only for zonemode=zbd.  So why adding strided here ? You will not have any zone information with strided mode. Or I am missing something here ?

engines/libaio.c
```diff
@@ -123,7 +127,13 @@ static int fio_libaio_prep(struct thread_data *td, struct io_u *io_u)
...
+		     td->o.zone_append) {
+			io_prep_pwrite(iocb, f->fd, io_u->xfer_buf, io_u->xfer_buflen, io_u->zone_slba);
+			iocb->aio_lio_opcode = IO_CMD_ZONE_APPEND;
+		}
+		else
+			io_prep_pwrite(iocb, f->fd, io_u->xfer_buf, io_u->xfer_buflen, io_u->offset);
```
damien-lemoal, line 136: If you can set io_u->offset to zone->start in zbd_adjust_block() for zone append, you would only need to set the opcode here. That is, if we define a new opcode... Back to first defining the kernel interface as it is not clear yet.

engines/libaio.c
```diff
@@ -228,6 +238,20 @@ static int fio_libaio_getevents(struct thread_data *td, unsigned int min,
 		} else {
 			r = io_getevents(ld->aio_ctx, actual_min,
 				max, ld->aio_events + events, lt);
+			if (td->o.zone_mode == ZONE_MODE_ZBD
+			    || td->o.zone_mode == ZONE_MODE_STRIDED) {
```
damien-lemoal, line 242: Same comment as above. Why consider strided mode here ?

damien-lemoal:
(@sitsofe @axboe)

All if this work needs a V2 but that can be reviewed, and more importantly tested, only once we get a clear definition and patches for the kernel side.  Right now, the interface/semantic of zone append issuing from user space is still not clear.

Other notes:
1) Support for zone append option when a regular block device is being used needs to be added. zonemode=zbd accepts regular block devices and fio will emulate zones in this case. zoneappend can also be emulated in this case using regular writes issued with an offset set to zone->wp, exactly the same way as for regular writes.
2) Please also send your patches as emails to fio mailing list (fio@vger.kernel.org). That makes reviewing far easier (I am not a big fan of github web interface for reviews)

damien-lemoal:
Another thing: It also be nice to add zone append test cases under t/zbd/.

javigon:
@damien-lemoal 

Thanks for the review!

We will send a V2 for the kernel append patches today. They apply on top of the current ZNS patches so we can start testing to address other comments while the kernel interface stabilizes.

os/linux/io_uring.h
krish-reddy: Yes, this file is not related to the ZNS patch. This is to keep the io_uring.h file updated with the latest 5.8 kernel's io_uring.h
This file will be submitted separately as a separate patch.

engines/libaio.c
krish-reddy: Yeah, strided mode is not needed for now. This will be removed in the v2 patches.
We can check for the strided mode support later.

engines/libaio.c
krish-reddy: New opcodes will be removed in the v2 patches and we will be using RWF_ZONE_APPEND flag as specified in the comments for the Kernel patches.
io_u->offset cannot be used due to the random map generator issue.

engines/libaio.c
krish-reddy: Strided mode support will be removed in the v2 patches.

os/linux/io_uring.h
krish-reddy: This file will be sent as a separate patch, not part of ZNS.

zbd.c
krish-reddy: Sure, this code will be refactored in the v2 patches.
The pthread_cond_wait will be called only in case of zone append, not in case of zone write.

zbd.c
krish-reddy: Sure, this code will be refactored in the v2 patches.

zbd.c
krish-reddy: Sure, this code will be refactored in the v2 patches.

io_u.h
krish-reddy: Will be renamed to zone_start_offset in the v2 patches.

HOWTO
krish-reddy: Sure, will be modified as per the comment.

fio.1
krish-reddy: Sure, will be modified as per the comment.

zbd.c
ankit-sam: Yes so we decided to add a new variable.

engines/libaio.c
damien-lemoal: You are missing my point: zonemode=zbd supports zones, zonemode=strided does not. There will be no zone information with it, so no possibility of doing zone append.

engines/libaio.c
damien-lemoal: And verify will not work either. Lots of things need to be documented about zone append use, in the commit messages and man page.

engines/libaio.c
krish-reddy: ok. this check will be removed for strided zone mode.

engines/libaio.c
krish-reddy: verify will work fine, just as normal write (we are updating io_u->ipo->offset)

damien-lemoal: The offset being written is used as data for write buffers snd checked on read, no ? So verify cannot work for zone append without modifications, I think.

krish-reddy: io_u->ipo->offset is updated as the offset returned from Linux kernel block layer in the fio_libaio_getevents()

io_u->offset is updated from ipo->offset at the time of verification in the get_next_verify() function of verify.c
This offset is used to perform read and verify the data written earlier.

damien-lemoal: I do not get it... Will need to check the code again. My understanding is that since io_u->offset is part of the data written, one cannot make verify work since that offset may not be equal to the one where the data was actually written. So on read, you can get mismatches of offset in data with the offset being read.

PR #1022, opened 2020-06-22
Make most code that uses barriers faster by using load acquire and store release instead

bvanassche:


axboe:
Thanks Bart, looks good.

sitsofe:
@bvanassche Do you think we should up the version of GCC we check for (https://github.com/axboe/fio/blob/653241de1eb5b9abe21cb6feb036df202d388c68/compiler/compiler-gcc4.h#L12 ) to 4.7?

bvanassche:
In compiler.h I found the following:
~~~
#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 1)
#include "compiler-gcc4.h"
#else
#error Compiler too old, need at least gcc 4.1.0
#endif
~~~
According to https://gcc.gnu.org/wiki/C11Status, support for atomics was added in version 4.9.

sitsofe:
https://gcc.gnu.org/gcc-4.7/changes.html and https://gcc.gnu.org/wiki/Atomic/GCCMM seem to hint that atomics went into 4.7 (but I'm not sure in what form).

It's kind of a mess that we're checking the minimum compiler in multiple places. Further, I'm not actually sure I've tried to compile a modern fio with an old compiler so it could very well be it needs gcc 4.9 (or something later). I guess that will be the real test if I can dig out a docker container from somewhere...

sitsofe:
OK looks like I was wrong and @bvanassche was right - you need at least GCC 4.9 according to #1023 ...

(What were the odds - I ask for it to be tested and then someone does one unprovoked?!)

axboe:
> (What were the odds - I ask for it to be tested and then someone does one unprovoked?!)

Indeed! Well at least now it should be obvious that 4.9+ is needed.

PR #1024, opened 2020-06-23
Changed signedness of seqlock.sequence fixing comparison-warning

safl:
I might be off here, but it seems like ``seqlock.sequence`` should be ``unsigned`` to avoid comparing signed with unsigned in the function ``read_seqlock_retry()``.


axboe:
Doesn't really matter, but it's cleaner. Thanks!

PR #1034, opened 2020-07-04
Fix race conditions in the workqueue mechanism

bvanassche:
Since test t0013 often hangs on Travis and Appveyor, I analyzed these hangs further. My conclusion is that these hangs are caused by race conditions in the workqueue mechanism. Hence this pull request.


axboe:
Thanks Bart, pulled.

PR #1040, opened 2020-07-14
Fix: dev-dax engine building with make

harish-24:
With recent changes to Makefile dev-dax skips building. This patch fixes that.

Signed-off-by: Harish <harish@linux.ibm.com>


axboe:
Thanks, indeed it currently doesn't. Pulled.

Issue #1039, opened 2020-07-14
Job doesn't stop at runtime when rate is less than the minimum block size


shradha-khapra:
## Details of the Bug
 Job file was executed for specified runtime but instead of stopping execution after that interval it keeps running indefinitely.

## Expected output
Execution halt for FIO after the specified runtime.

### During specified runtime
![output_during_runtime](https://user-images.githubusercontent.com/34954810/87420219-860d5680-c5f2-11ea-84a7-fd79c2806ce5.png)

### After specified runtime
![output_after_runtime](https://user-images.githubusercontent.com/34954810/87420267-9ae9ea00-c5f2-11ea-92de-b4ceaa4984e0.png)
   
## OS Details
```
NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.4 LTS"
VERSION_ID="18.04"
```
## FIO version
```
fio-3.1
```

## FIO installation
```
sudo apt install fio
```

## Job File used
```
[reader-writer]
ioengine=libaio
iodepth=4
rw=rw
runtime=30
filename=/dev/sdc
bs=4096
rate=50.497800,50.354600
```

sitsofe:
@shradha-khapra can you check if the information in https://github.com/axboe/fio/issues/1017#issuecomment-644544867 clarifies things? If so can you close this ticket. Thanks!

> [...] are you aware that by default without extra options `runtime` means "exit IF you run longer than this time" (so it's a cap) rather than "run for exactly this time"? See the [fio documentation for runtime](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-runtime)  and [timebased](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-time-based).

sitsofe:
Actually I think the above is an incorrect guess. My next question is has your disk "gone away"? Can you check what ps says about the state of all your fio processes? This might be another case of https://github.com/axboe/fio/issues/1004#issuecomment-637227364 but we will need more information to know for sure...

vincentkfu:
With the rate option you are limiting bandwidth to 50 bytes/sec but your block size is 4096 bytes. fio will work better with a more reasonable rate value, but probably it should better handle unusual cases like this.

shradha-khapra:
[...]

PR #1044, opened 2020-07-19
Adding Fix for Indefinite run for fio

sribs:
PR to the Linked Issue:

io_u: fix asymptotically high time run in accordance with blocksize of fio when using rate with timeout
Fixes: https://github.com/axboe/fio/issues/1039
signed-off-by : Sriharsha B S sribs@microsoft.com

@sitsofe , It is almost the same change that you have proposed. However, I have added another check to check if it takes in negative values.

Thanks,
Sri

### DIFF
--- a/io_u.c
+++ b/io_u.c
@@ -680,7 +680,23 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
 	if (td->o.io_submit_mode == IO_MODE_INLINE)
 		io_u_quiesce(td);
 
+	if (td->o.timeout && ((usec + now) > td->o.timeout)) {
+		/*
+		 * check if the usec is capable of taking negative values
+		 * however it is unlikely, but still a possibility.
+		 */
+		if (now > td->o.timeout) {
+			ddir = DDIR_INVAL;
+			return ddir;
+		}
+		usec = td->o.timeout - now;
+	}
 	usec_sleep(td, usec);
+
+	now = utime_since_now(&td->epoch);
+	if ((td->o.timeout && (now > td->o.timeout)) || td->terminate)
+		ddir = DDIR_INVAL;
+
 	return ddir;
 }
 
@@ -896,6 +912,10 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
 
 	set_rw_ddir(td, io_u);
 
+	if (io_u->ddir == DDIR_INVAL) {
+		dprint(FD_IO, "Invalid Direction received ddir = %d", io_u->ddir);
+		return 1;
+	}
 	/*
 	 * fsync() or fdatasync() or trim etc, we are done
 	 */
--- a/stat.c
+++ b/stat.c
@@ -414,6 +414,18 @@ static void display_lat(const char *name, unsigned long long min,
 	free(maxp);
 }
 
+static double convert_agg_kbytes_percent(struct group_run_stats *rs, int ddir, int mean)
+{
+	double p_of_agg = 100.0;
+	if (rs->agg[ddir] < 1024) {
+		p_of_agg = mean * 100 / (double) (rs->agg[ddir] / 1024.0);
+
+		if (p_of_agg > 100.0)
+			p_of_agg = 100.0;
+	}
+	return p_of_agg;
+}
+
 static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
 			     int ddir, struct buf_output *out)
 {
@@ -551,11 +563,7 @@ static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
 		else
 			bw_str = "kB";
 
-		if (rs->agg[ddir]) {
-			p_of_agg = mean * 100 / (double) (rs->agg[ddir] / 1024);
-			if (p_of_agg > 100.0)
-				p_of_agg = 100.0;
-		}
+		p_of_agg = convert_agg_kbytes_percent(rs, ddir, mean);
 
 		if (rs->unit_base == 1) {
 			min *= 8.0;
@@ -1376,11 +1384,7 @@ static void add_ddir_status_json(struct thread_stat *ts,
 	}
 
 	if (calc_lat(&ts->bw_stat[ddir], &min, &max, &mean, &dev)) {
-		if (rs->agg[ddir]) {
-			p_of_agg = mean * 100 / (double) (rs->agg[ddir] / 1024);
-			if (p_of_agg > 100.0)
-				p_of_agg = 100.0;
-		}
+		p_of_agg = convert_agg_kbytes_percent(rs, ddir, mean);
 	} else {
 		min = max = 0;
 		p_of_agg = mean = dev = 0.0;


io_u.c
```diff
@@ -680,7 +680,27 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
...
+	 * Check if the usec is taking negative values.
+	 * However it is unlikely, But still a possibility.
+	 * If this is not taken care, usec_sleep() can take a 
+	 * huge nap which is capable of not exiting the process indefinitely
+	 * although the process/thread may be sleeping.
+	 */
+	if (usec == (1 << (sizeof(usec) - usec))) {
+		ddir = DDIR_INVAL;
+		return ddir;
+	}
```
sitsofe, line 696: You only need to check for negativity in the case where we need to do adjustment (i.e. around 684). At that point you can simply check if now is bigger than td->o.timeout. If it is then you can return immediately with an invalid ddir as it's clear we don't want to sleep. If it's not then you go on and do the subtraction.

io_u.c
```diff
@@ -895,7 +915,13 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
 		goto out;
 
 	set_rw_ddir(td, io_u);
-
+	/*
+	 * check if we have received Invalid Direction
+	 */
```
sitsofe, line 920: If we're keeping the comment keep it all lower case to match the style of the other comments. I'd argue you don't need this particular comment as it is basically saying what the condition of the if is saying (contrast with the comment below talking about fsync where's not immediately obvious the if is dealing with fsync...

stat.c
```diff
@@ -552,7 +552,7 @@ static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
 			bw_str = "kB";
 
 		if (rs->agg[ddir]) {
-			p_of_agg = mean * 100 / (double) (rs->agg[ddir] / 1024);
+			p_of_agg = mean * 100 / (double) (rs->agg[ddir] / 1024.0);
```
sitsofe, line 555: This one is tricky. Before we did integer arithmetic to convert rs->agg[ddir] into kilobytes which mean it could just come out to 0. Arguably the if above was trying to catch that but it dates from a time before this conversion was happening. Maybe it would be better to check if rs->agg[ddir] was less than 1024 before going into the body. There's also similar code around line 1380 so maybe we should turn this into a function?

io_u.c
```diff
@@ -895,7 +915,13 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
...
+	if (io_u->ddir == DDIR_INVAL) {
+		dprint(FD_IO, "Invalid Direction received ddir = %d", io_u->ddir);
+		return 1;
```
sitsofe, line 923: @axboe by doing this we bail out of the do_io() while loop (https://github.com/axboe/fio/blob/5090d1d0f2a109c276384c93308566b7a3bfa5ad/backend.c#L997 ) and won't have done `fio_mark_td_terminate()` (see https://github.com/axboe/fio/blob/5090d1d0f2a109c276384c93308566b7a3bfa5ad/backend.c#L965 ). Basically the intent is to stop us processing the io_u because this diff changes things so we return earlier than we would have done according to the rate we were trying to achieve which means the io_u is sent earlier and skews the numbers. However it's become a bit messy...

io_u.c
sribs: I'm not sure if I'm doing anything different here although I agree that yours is more efficient.

I'm checking for negative value here. However, the fact that ```usec``` is of type ```unsigned int```, I'm performing 2's compliment to check if the value is negative.

However, if you feel that earlier comment suits to this context, I can fix that

io_u.c
sribs: I will remove this comment

io_u.c
sitsofe: @sribs I agree the outcome is hopefully the same but in the case that timeout adjustment is not needed we will still evaluate the diff with your existing code. Also my case theoretically will handle very big values of usec where multiplying by two would cause overflow (but I think we've got other problems if usec is that big :-) ).

It's minor but I think
```c
if (now > td->o.timeout)
```
is easier to read than
```c
if (usec == (1 << (sizeof(usec) - usec))) {
```

and will need less comment.

stat.c
sribs: Done. Create a function

io_u.c
```diff
@@ -680,21 +680,21 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
...
-	 * If this is not taken care, usec_sleep() can take a 
-	 * huge nap which is capable of not exiting the process indefinitely
-	 * although the process/thread may be sleeping.
-	 */
-	if (usec == (1 << (sizeof(usec) - usec))) {
-		ddir = DDIR_INVAL;
-		return ddir;
 	}
-
+
```
sitsofe, line 697: I'd strip trailing whitespace

io_u.c
```diff
@@ -680,21 +680,21 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
...
 		io_u_quiesce(td);
 
-	if (td->o.timeout && ((usec + now) > td->o.timeout))
+	if (td->o.timeout && ((usec + now) > td->o.timeout)) {
+		/*
+		 * check if the usec is capable of taking negative values.
+		 * However it is unlikely, but still a possibility.
+		 * If this is not taken care, usec_sleep() can take a 
+		 * huge nap which is capable of not exiting the process indefinitely
+		 * although the process/thread may be sleeping.
```
sitsofe, line 689: I'd move most of this into the commit message

sitsofe:
I don't think the squashing went as you were expecting - https://github.com/axboe/fio/pull/1044 (that lists 7 commits). I fiddled around with an interactive rebase and end up with this:
```diff
From 1f315844ee0092b5d16a33533e9ec7370f0ec859 Mon Sep 17 00:00:00 2001
From: sribs <sriharsha.bs@outlook.com>
Date: Fri, 24 Jul 2020 12:11:07 +0530
Subject: [PATCH] Squashing all changes to a single commit.

Fixes: https://github.com/axboe/fio/issues/1039
io_u: fix asymptotically high time run in accordance with blocksize of fio when using rate with timeout
stat: Adding changes to check and convert rs->agg[ddir] in KBytes

io_u:
Check if the usec is taking negative values.
However it is unlikely, But still a possibility.
If this is not taken care, usec_sleep() can take a
huge nap which is capable of not exiting the process indefinitely
although the process/thread may be sleeping

Fixes: https://github.com/axboe/fio/issues/1039
io_u: fix asymptotically high time run in accordance with blocksize of fio when using rate with timeout
stat: Adding changes to check and convert rs->agg[ddir] in KBytes

io_u:
Check if the usec is taking negative values.
However it is unlikely, But still a possibility.
If this is not taken care, usec_sleep() can take a
huge nap which is capable of not exiting the process indefinitely
although the process/thread may be sleeping

Removing extra comments
---
 io_u.c | 20 ++++++++++++++++++++
 stat.c | 24 ++++++++++++++----------
 2 files changed, 34 insertions(+), 10 deletions(-)

diff --git a/io_u.c b/io_u.c
index 7f50906b..798c28f3 100644
--- a/io_u.c
+++ b/io_u.c
@@ -680,7 +680,23 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
 	if (td->o.io_submit_mode == IO_MODE_INLINE)
 		io_u_quiesce(td);
 
+	if (td->o.timeout && ((usec + now) > td->o.timeout)) {
+		/*
+		 * check if the usec is capable of taking negative values
+		 * however it is unlikely, but still a possibility.
+		 */
+		if (now > td->o.timeout) {
+			ddir = DDIR_INVAL;
+			return ddir;
+		}
+		usec = td->o.timeout - now;
+	}
 	usec_sleep(td, usec);
+
+	now = utime_since_now(&td->epoch);
+	if ((td->o.timeout && (now > td->o.timeout)) || td->terminate)
+		ddir = DDIR_INVAL;
+
 	return ddir;
 }
 
@@ -896,6 +912,10 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
 
 	set_rw_ddir(td, io_u);
 
+	if (io_u->ddir == DDIR_INVAL) {
+		dprint(FD_IO, "Invalid Direction received ddir = %d", io_u->ddir);
+		return 1;
+	}
 	/*
 	 * fsync() or fdatasync() or trim etc, we are done
 	 */
diff --git a/stat.c b/stat.c
index b3951199..2e46f1b7 100644
--- a/stat.c
+++ b/stat.c
@@ -414,6 +414,18 @@ static void display_lat(const char *name, unsigned long long min,
 	free(maxp);
 }
 
+static double convert_agg_kbytes_percent(struct group_run_stats *rs, int ddir, int mean)
+{
+	double p_of_agg = 100.0;
+	if (rs->agg[ddir] < 1024) {
+		p_of_agg = mean * 100 / (double) (rs->agg[ddir] / 1024.0);
+
+		if (p_of_agg > 100.0)
+			p_of_agg = 100.0;
+	}
+	return p_of_agg;
+}
+
 static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
 			     int ddir, struct buf_output *out)
 {
@@ -551,11 +563,7 @@ static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
 		else
 			bw_str = "kB";
 
-		if (rs->agg[ddir]) {
-			p_of_agg = mean * 100 / (double) (rs->agg[ddir] / 1024);
-			if (p_of_agg > 100.0)
-				p_of_agg = 100.0;
-		}
+		p_of_agg = convert_agg_kbytes_percent(rs, ddir, mean);
 
 		if (rs->unit_base == 1) {
 			min *= 8.0;
@@ -1376,11 +1384,7 @@ static void add_ddir_status_json(struct thread_stat *ts,
 	}
 
 	if (calc_lat(&ts->bw_stat[ddir], &min, &max, &mean, &dev)) {
-		if (rs->agg[ddir]) {
-			p_of_agg = mean * 100 / (double) (rs->agg[ddir] / 1024);
-			if (p_of_agg > 100.0)
-				p_of_agg = 100.0;
-		}
+		p_of_agg = convert_agg_kbytes_percent(rs, ddir, mean);
 	} else {
 		min = max = 0;
 		p_of_agg = mean = dev = 0.0;
-- 
```
there were a few conflicts along the way though so it may be easier/quicker to "start over" having downloaded the above:
```
% git reset --hard fa443634fbfa38fd5d6418a96a45022c78b90df4

HEAD is now at fa443634 Merge branch 'patch-1' of https://github.com/avlasov-mos-de/fio
% git am 0001-Squashing-all-changes-to-a-single-commit.patch
Applying: Squashing all changes to a single commit.
% git log -2 --oneline
fa443634 (HEAD -> master) Merge branch 'patch-1' of https://github.com/avlasov-mos-de/fio
e0424151 t/zbd: Enable regular block devices for test case #47
```

If you can arrange for something like that as the only "new" commit on your branch (but maybe with a more brief commit message). Once you're down to a single commit you can use `git commit --amend` (https://docs.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message ) to repeatedly change the last commit.

sribs:
@sitsofe, @axboe : I'm rejecting this PR and raising another PR from a different branch.
New PR: https://github.com/axboe/fio/pull/1050

PR #1050, opened 2020-07-24
io_u: fix exit failure case when using rates and timeout

sribs:
Fixes: https://github.com/axboe/fio/issues/1039
io_u: fix asymptotically high time run in accordance with blocksize of fio when using rate with timeout
stat: Adding changes to check and convert rs->agg[ddir] in KBytes

io_u:
Check if the usec is taking negative values.
However it is unlikely, But still a possibility.
If this is not taken care, usec_sleep() can take a
huge nap which is capable of not exiting the process indefinitely
although the process/thread may be sleeping

Signed-off-by: Sriharsha B S <sribs@microsoft.com>

### DIFF
--- a/io_u.c
+++ b/io_u.c
@@ -680,7 +680,22 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
 	if (td->o.io_submit_mode == IO_MODE_INLINE)
 		io_u_quiesce(td);
 
+	if (td->o.timeout && ((usec + now) > td->o.timeout)) {
+		/*
+		 * check if the usec is capable of taking negative values
+		 */
+		if (now > td->o.timeout) {
+			ddir = DDIR_INVAL;
+			return ddir;
+		}
+		usec = td->o.timeout - now;
+	}
 	usec_sleep(td, usec);
+
+	now = utime_since_now(&td->epoch);
+	if ((td->o.timeout && (now > td->o.timeout)) || td->terminate)
+		ddir = DDIR_INVAL;
+
 	return ddir;
 }
 
@@ -896,6 +911,10 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
 
 	set_rw_ddir(td, io_u);
 
+	if (io_u->ddir == DDIR_INVAL) {
+		dprint(FD_IO, "invalid direction received ddir = %d", io_u->ddir);
+		return 1;
+	}
 	/*
 	 * fsync() or fdatasync() or trim etc, we are done
 	 */
--- a/stat.c
+++ b/stat.c
@@ -414,6 +414,18 @@ static void display_lat(const char *name, unsigned long long min,
 	free(maxp);
 }
 
+static double convert_agg_kbytes_percent(struct group_run_stats *rs, int ddir, int mean)
+{
+	double p_of_agg = 100.0;
+	if (rs && rs->agg[ddir] > 1024) {
+		p_of_agg = mean * 100 / (double) (rs->agg[ddir] / 1024.0);
+
+		if (p_of_agg > 100.0)
+			p_of_agg = 100.0;
+	}
+	return p_of_agg;
+}
+
 static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
 			     int ddir, struct buf_output *out)
 {
@@ -551,11 +563,7 @@ static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
 		else
 			bw_str = "kB";
 
-		if (rs->agg[ddir]) {
-			p_of_agg = mean * 100 / (double) (rs->agg[ddir] / 1024);
-			if (p_of_agg > 100.0)
-				p_of_agg = 100.0;
-		}
+		p_of_agg = convert_agg_kbytes_percent(rs, ddir, mean);
 
 		if (rs->unit_base == 1) {
 			min *= 8.0;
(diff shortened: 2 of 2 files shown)


sribs:
@sitsofe : can you review this ?

sitsofe:
OK nice job on squashing it down to one commit! Next up:

The commit title is now a bit strange. I'd change the message to something like this:

```
io_u: fix infinite run when using rate with timeout

Previously when using a rate smaller than the minimum block size, fio could end up failing to exit at a designated timeout. Fix this and also adjust stat.c to no longer trigger division by zero on bandwidths greater than 0 but less than 1 kilobyte a second.
```

sitsofe:
@sribs I'll get round to reviewing it eventually but it can be variable amount of time. Don't worry I'm keeping an eye on this :-)

io_u.c
```diff
@@ -680,7 +680,23 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
 	if (td->o.io_submit_mode == IO_MODE_INLINE)
 		io_u_quiesce(td);
 
+	if (td->o.timeout && ((usec + now) > td->o.timeout)) {
+		/*
+		 * check if the usec is capable of taking negative values
+		 * however it is unlikely, but still a possibility.
```
sitsofe, line 686: You can skip the last line of the comment

io_u.c
```diff
@@ -896,6 +912,10 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
 
 	set_rw_ddir(td, io_u);
 
+	if (io_u->ddir == DDIR_INVAL) {
+		dprint(FD_IO, "Invalid Direction received ddir = %d", io_u->ddir);
```
sitsofe, line 916: Direction -> direction (lowercase d)

stat.c
```diff
@@ -414,6 +414,18 @@ static void display_lat(const char *name, unsigned long long min,
 	free(maxp);
 }
 
+static double convert_agg_kbytes_percent(struct group_run_stats *rs, int ddir, int mean)
```
sitsofe, line 417: Nice! One question is should you just take agg and expect the caller to do pass you the value rather than a pointer that you have to deference?

sribs:
Sure. I can add that commit message

stat.c
```diff
@@ -414,6 +414,18 @@ static void display_lat(const char *name, unsigned long long min,
...
+{
+	double p_of_agg = 100.0;
+	if (rs->agg[ddir] < 1024) {
```
sitsofe, line 420: Is this comparison the wrong way round?

Also if I've contributed nothing because my rate is so small should that count as 100% or 0%?

stat.c
sribs: I prefer the pointer here to be dereferenced since I prefer providing meaningful insight to the function since this aggregation is specific to this run_stats or am I missing something here

stat.c
sribs: I agree....missed it. Will change this in this commit

sitsofe:
@sribs your commit message still looks odd - https://github.com/axboe/fio/pull/1050/commits .

sribs:
> @sribs your commit message still looks odd - https://github.com/axboe/fio/pull/1050/commits .

It is because, I have sqashed all the commits into a single commit but retained every squashed commit message. I Think it is evident before my latest push wherein I had pushed 4 extra commits by mistake

sribs:
> > @sribs your commit message still looks odd - https://github.com/axboe/fio/pull/1050/commits .
> 
> It is because, I have sqashed all the commits into a single commit but retained every squashed commit message. I Think it is evident before my latest push wherein I had pushed 4 extra commits by mistake

Do you want me to only have the recent commit message ?

axboe:
> Do you want me to only have the recent commit message ?

Yes, having essentially the same stuff duplicated tons of times makes no sense. You also need to fix the title of it, and ensure that the length of each line is 72/74 chars at most.

As an example of a good commit message format, see:

https://git.kernel.dk/cgit/fio/commit/?id=236d23a86d6e3c14b700496e0f196f88513a5f3e

You want the first line to be a title, and preferable prefixed with whatever it's touching. For yours, something ala:

```
io_u: fix exit failure case when using rates and timeout
```

or something like that. Then an empty line, then an explanation for WHY the change is made. Things like Signed-off-by: or a Fixes line go at the end.

sitsofe:
> Do you want me to only have the recent commit message ?

That's right (I see Jens has already responded).

sribs:
> > Do you want me to only have the recent commit message ?
> 
> Yes, having essentially the same stuff duplicated tons of times makes no sense. You also need to fix the title of it, and ensure that the length of each line is 72/74 chars at most.
> 
> As an example of a good commit message format, see:
> 
> https://git.kernel.dk/cgit/fio/commit/?id=236d23a86d6e3c14b700496e0f196f88513a5f3e
> 
> You want the first line to be a title, and preferable prefixed with whatever it's touching. For yours, something ala:
> 
> ```
> io_u: fix exit failure case when using rates and timeout
> ```
> 
> or something like that. Then an empty line, then an explanation for WHY the change is made. Things like Signed-off-by: or a Fixes line go at the end.

Sure. That sounds great. I wasn't aware about the format. Will push it right away

sribs:
I have pushed the recent changes to the commit message. Please review the same

axboe:
The commit format looks good to me now. Only one comment, which isn't a huge deal for this one, but something that you should be wary of in general. When your commit message contains phrases like:

"Also [...]"

then that's a clear indication that you are doing two different things in a single commit, and that this really should have been two separate commits. These two changes have nothing to do with each other, hence they should be two different commits.

sribs:
> The commit format looks good to me now. Only one comment, which isn't a huge deal for this one, but something that you should be wary of in general. When your commit message contains phrases like:
> 
> "Also [...]"
> 
> then that's a clear indication that you are doing two different things in a single commit, and that this really should have been two separate commits. These two changes have nothing to do with each other, hence they should be two different commits.

@axboe , I have made that change as well :+1:

sribs:
Thanks folks for reviewing this. Do note that I'm super new to contributing to community projects and I really appreciate the fact that your patient enough to provide constructive feedback

axboe:
Well, thanks for not giving up! At least these lessons will apply in a lot of other places, too. Looks fine to me now, I've pulled it.

sitsofe:
@sribs I would also like to thank you for sticking with this to the end. I know I kept coming back with requests for change but you had kept your part of the bargain (by submitting a pull request) so I had to help you get it in :-). It was definitely a collaborative effort!

PR #1053, opened 2020-07-25
configure: check for C11 atomics support

sitsofe:
So I know you've merged something already to address this but I was fiddling with about with this in the morning to check if it worked. I figure I'll post it for posterity even if it doesn't get merged (I'd have to rebase it anyway).

### DIFF
--- a/configure
+++ b/configure
@@ -496,20 +496,6 @@ else
 fi
 
 ##########################################
-# check compiler version
-
-cat > $TMPC <<EOF
-#include "$(pwd)/compiler/compiler.h"
-int main(void)
-{
-  return 0;
-}
-EOF
-if ! compile_prog "" "" "compiler check"; then
-  fatal "Your compiler is too old, needs GCC 4.9 or higher"
-fi
-
-##########################################
 # check endianness
 if test "$bigendian" != "yes" ; then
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -550,6 +550,26 @@ else
...
+int main(void)
+{
+  _Atomic unsigned v;
+  atomic_load(&v);
+  return 0;
+}
+EOF
+if ! compile_prog "" "" "C11 atomics"; then
+  echo
+  echo "From fio 3.21 compiler support for C11 atomics is required."
```
axboe, line 566: I'd remove the version info here, it isn't really relevant. It'll just end up being a historic thing that's then removed later, and it doesn't even really add anything right now. They obviously have this version (or newer) if they run into this :-)

axboe:
Honestly, I just pushed out what I had, since I didn't hear back. I'm fine with this one too, and it's probably better since it doesn't rely on version checking.

configure
sitsofe: Fair point - done.

sitsofe:
OK updated - hopefully this looks better...

axboe:
LGTM!

axboe:
LGTM!

PR #1055, opened 2020-07-26
Travis cleanup

sitsofe:
- Enable the cuda and libiscsi ioengines in travis (this way we get at least compile time testing)
- Fix up bugs found by the above
- Add a search for a differently named tcmalloc
- Cosmetic changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
     - os: osx
       compiler: gcc
 
-before_install:
+install:
   - ci/travis-install.sh
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,34 +9,26 @@ arch:
   - amd64
   - arm64
 env:
-  matrix:
-    - BUILD_ARCH="x86"
-    - BUILD_ARCH="x86_64"
   global:
     - MAKEFLAGS="-j 2"
 matrix:
   include:
+    - os: linux
+      compiler: gcc
+      arch: amd64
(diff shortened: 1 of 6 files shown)


sitsofe:
I guess it's an open question as to whether we want CI to build ioengines that aren't actually run (insofar as they had errors so I guess no one is building them on a recent gcc / clang) so I can always remove that commit if you think it doesn't win us anything.

@axboe you may want to:
- Sign out of travis
- Remove travis authorization from https://github.com/settings/apps/authorizations / https://github.com/settings/applications
- Sign back in to Travis (https://travis-ci.org/signin?redirectUrl=https%3A%2F%2Ftravis-ci.org%2F )
- Click Sync Account on https://travis-ci.org/account/repositories

because I've noticed Travis results never show up on on commits (e.g. click the tick to the left of the title on https://github.com/axboe/fio/commit/e235e74dfd0627b17690194f957da509b8ace808 and you'll only see Appveyor. I've no idea why this happened but I had the same issue on my github fork.

axboe:
Thanks pulled, and went through the sign-out/reauth sequence. Hopefully that does the trick...

sitsofe:
(Curiously when the issue was happening even though travis couldn't report back it would still run the tests)

I think you've nailed it - the latest commit (https://github.com/axboe/fio/commit/6feb475a9e775db911ee766a9671f244b8c1d35e ) has a pending link to travis.

PR #1061, opened 2020-08-01
Prevent that fio hangs when using io_submit_mode=offload

bvanassche:
This patch has been tested by running the following shell command:

for ((i=0;i<1000;i++)); do echo $i; python3 t/run-fio-tests.py -o 10; done

Fixes: d28174f0189c ("workqueue: ensure we see deferred error for IOs")

### DIFF
--- a/rate-submit.c
+++ b/rate-submit.c
@@ -97,8 +97,11 @@ static int io_workqueue_fn(struct submit_worker *sw,
 			td->cur_depth -= ret;
 	}
 
-	if (error || td->error)
+	if (error || td->error) {
+		pthread_mutex_lock(&td->io_u_lock);
 		pthread_cond_signal(&td->parent->free_cond);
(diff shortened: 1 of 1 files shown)


sitsofe:
(Windows failure is in IOPs so nothing to do with this patch)

LGTM. This makes me wonder if the `td_io_u_free_notify()` call in `put_io_u()` should actually be guarded by an `if (needs_lock)` but I guess one spurious wake isn't going to be an issue.

bvanassche:
@sitsofe Isn't using `if (needs_lock)` an anti-pattern? My own preference for code that needs locking is to obtain a lock unconditionally at the start of the block that needs protection and to release the same lock at the end of the code block that needs protection.

sitsofe:
@bvanassche Hmm but doesn't pushing the locking into `put_io_u()` mean it the critical region can be smaller and is arguably easier to request (`needs_lock` strongly hints at the requirement)?

PR #1062, opened 2020-08-06
filelock: fix wrong file trylock assertion.

bardavid:
Earlier solution to increase MAX_FILELOCKS for high parallelism might not be necessary given the wrong assertion

`> fio: filelock.c:182: __fio_lock_file: Assertion !trylock' failed. `

https://github.com/axboe/fio/commit/43661a66a8c3b9fb5ff28b463c9ab0aac91a0355

### DIFF
--- a/filelock.c
+++ b/filelock.c
@@ -179,7 +179,7 @@ static bool __fio_lock_file(const char *fname, int trylock)
 	fio_sem_up(&fld->lock);
 
 	if (!ff) {
-		assert(!trylock);
+		assert(trylock);
 		return true;
 	}
 


sitsofe:
To me this fix looks correct:
- If trylock is non-zero
   - The variable's name suggests we're willing to accept `fio_hash_get()` returning `NULL`
      - Therefore we should assert on trylock directly

@bardavid My suggestions are:
- Can you prefix the area affected to the commit title and drop the trailing `.` (e.g. `filelock: fix wrong file trylock assertion`)
- Could you add your Signed-off-by ?

Tracing through the code:
- Because `__fio_lock_file()` is static all calls must be in the `filelock.c`. There are only two calls to it (https://github.com/axboe/fio/blob/87622bf5295880682bfad5ba14116cf5facbaf2c/filelock.c#L215 ) and only `fio_trylock_file()` sets `trylock` to 1.
- The only call to `fio_trylock_file()` is in iolog.c (https://github.com/axboe/fio/blob/87622bf5295880682bfad5ba14116cf5facbaf2c/iolog.c#L1237 )
- `write_*_log()` calls `__write_log()` which calls `finish_log()` and try is passed all the way down and all those functions are also static.
- Function pointers for `write_*_log()` are set in `log_types[]` over in https://github.com/axboe/fio/blob/87622bf5295880682bfad5ba14116cf5facbaf2c/iolog.c#L1683 
- These pointers seem to be called from https://github.com/axboe/fio/blob/87622bf5295880682bfad5ba14116cf5facbaf2c/iolog.c#L1728 .
- So try/trylock will be 1 for all but the last log file. https://github.com/axboe/fio/commit/243bfe190245a10e9d0981bf2a7c722edc4c43d4 says this was done to allow logs to be written in parallel where possible.

@axboe was this done to try and guarantee there were enough locks to simultaneously write at least one log for every job?

sitsofe:
@axboe This one LGTM!

sitsofe:
(@axboe But ICYMI have you seen my question over in https://github.com/axboe/fio/pull/1062#issuecomment-670315412 ?)

axboe:
Sorry, dropped the ball. Now merged, thanks!

bardavid:
@axboe  thank you, it's my first open source delivery ever ..

axboe:
Hopefully it's the start of a glorious career in open source :-)

sitsofe:
@axboe :

> @axboe was this [the trylocks] done to try and guarantee there were enough locks to simultaneously write at least one log for every job?

sitsofe:
> it's my first open source delivery ever ..

@bardavid Nifty! It was a tidy submission and fixed an issue. I wish you luck on your future submissions - if they're going to be like this I know I'll be happy to seem them :-)

Issue #741, opened 2019-02-20
fio how to run mix workload all 4 patterns in one job file?


mallilinux1978:
Example: 60% randread, 20% randwrite, 10% seqread, 10% seqwrite

i tried a bit by reading fio guide but not getting it quite right so appreciated if anyone has done this already and share the job file please?

[sybase-like]	
bs=2k
rw=randrw
percentage_random=80
rwmixread=70
rwmixwrite=30

sitsofe:
@mallilinux1978 Sadly this is going to be hard as you want to balance your workload more than two ways ([flow](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-flow) almost gets you there but not quite).

szaydel:
@mallilinux1978, it is possible to just setup two or more independent configurations and run them all concurrently. Or are you looking for consolidated reporting?

mallilinux1978:
yes all as one test or a job from FIO perspective. Basically we're trying to simulate OLTP workload where every transaction will do 4 patterns for example table space are random and logs are sequential. from a reporting perspective, these metrics are they key (IOPs, Latency & Throughput) in total as well as read/write ratio. 

Other synthetic tools like EMC IORATE and VDBENCH support 4 patterns in one test.

sitsofe:
If VDBench is working for you you may want to stick with it in this particular case  ;-)

FWIW this a feature I've wanted to code up (I was going to abuse flow by adding another config item that would basically make it work with "shares" calculated from a total) but there's never the time...

szaydel:
This may be an open and shut case for the moment as @sitsofe suggested, but I am just going to ask one last thing. Why not create separate job blocks without a barrier between them? @sitsofe, correct me if I am wrong, but that should not be impossible given these requirements, however, ordering may be something that you cannot realistically control.

sitsofe:
@szaydel running four different workloads (randread, randwrite, seqread, seqwrite) is easy (you could just create seperate jobs for these in the jobfile and ensure you don't use stonewall as you said). Maintaining the ratios between all the jobs versus all the other jobs is what makes this hard. Balancing one against one or 2vs2 (two on one side, two on the other) could be done via `flow` etc. but all vs all when you have three or more jobs is where I think you'll struggle...

mallilinux1980:
Any idea when this will be introduced?

sitsofe:
[...]

PR #1063, opened 2020-08-10
flow: add ability for weight-based flow control on multiple jobs

bardavid:
should solve https://github.com/axboe/fio/issues/741

removed `flow_watermark` as it's not needed with new flow ..

changed `flow` to accept only natural numbers and changed some fio jobs accordingly.. i ran the 1:2:3 scenario and 1:8 scenario and it yielded expected results.


bardavid:
i'm not sure why github views wrong spacing in the code, any idea what am i doing wrong ?

PR #1064, opened 2020-08-10
flow: add ability for weight-based flow control on multiple jobs

bardavid:
flow: add ability for weight-based flow control on multiple jobs …

Fixes: #741 - fio how to run mix workload all 4 patterns in one job file

Previously, 'flow' was used to regulate activity between 2 or more jobs with
certain weight.  However, the implementation was ineffective when regulating
more than 2 jobs with different weights, since it relied on a single,
shared counter, and was missing logic to track individual job's activity
and regulate it according to it's requested weight in proportion to other jobs'
requested weight and activity.

This commit modifies the 'flow' functionality to track each job's
activity and the total activity belonging to a 'flow_id' group.
Using this new information, the 'flow' logic is then modified to
regulate a job's activity in proportion to other jobs in the
same group.  Similar to previous behavior, the regulation is done by
stalling a job's activity.

New jobs joining an existing 'flow_id' group or jobs repead from said
group will cause re-calibration of the weights and alter the activity as a
result.  For example, 3 jobs (j1,j2,j3) with weights (1:2:3)
respectively will maintain activity ratio of 1:2:3.  Thus, j1 will have
activity of 1/6, j2 - 2/6 and j3 - 1/2.

If j3 exits prematurely, the proportion is re-calibrated
such that j1 will strive to have activity of 1/3 and j2 - 2/3.

Re-calibration phase may cause a hiccup in the activity of one or more
jobs since the new desired proportion might not be achieved quickly
enough.


fio.1
```diff
@@ -2549,21 +2549,18 @@ The ID of the flow. If not specified, it defaults to being a global
...
-\fBflow\fR parameter stands for how much should be added or subtracted to the
-flow counter on each iteration of the main I/O loop. That is, if one job has
-`flow=8' and another job has `flow=\-1', then there will be a roughly 1:8
-ratio in how much one runs vs the other.
-.TP
-.BI flow_watermark \fR=\fPint
-The maximum value that the absolute value of the flow counter is allowed to
-reach before the job must wait for a lower value of the counter.
+Weight in token-based flow control. If this value is used, 
+then Fio regulates the activity between two or more jobs
```
sitsofe, line 2553: I'd go with lower fio except at the start of a sentence.

flow.c
```diff
@@ -7,7 +7,8 @@ struct fio_flow {
 	unsigned int refs;
 	struct flist_head list;
 	unsigned int id;
-	long long int flow_counter;
+	volatile unsigned long long flow_counter;
```
sitsofe, line 10: Why is this volatile? Are you using it from a signal handler?

flow.c
```diff
@@ -35,9 +38,10 @@ int flow_threshold_exceeded(struct thread_data *td)
 		return 1;
 	}
 
-	/* No synchronization needed because it doesn't
-	 * matter if the flow count is slightly inaccurate */
-	flow->flow_counter += td->o.flow;
+	// increment flow(shared counter, therefore atomically) and job-specific counter
```
sitsofe, line 41: Please use old-style C comments (even though it's single line)

HOWTO
```diff
@@ -2861,11 +2861,6 @@ Threads, processes and job synchronization
 	``flow=8`` and another job has ``flow=-1``, then there will be a roughly 1:8
 	ratio in how much one runs vs the other.
 
-.. option:: flow_watermark=int
-
-	The maximum value that the absolute value of the flow counter is allowed to
-	reach before the job must wait for a lower value of the counter.
-
```
sitsofe, line 2868: I'm not sure about throwing away the `flow_watermark` behaviour as that will just break old jobs (but I'm all for the examples being converted to the new way as you did below). It would be better to deprecate the old option and remove it entirely in the future.

flow.c
bardavid: the variable is incremented from different jobs, i wanted to get the value from memory instead of cache to increase the accuracy of the flow.   do you think i should fetch it using atomic api instead ?

bardavid:
> * How does this cope with the online conditions changing (new jobs arriving e.g. because of stonewall/wait_for, existing ones exiting while others are still running)?

it doesn't, i didn't consider it. thanks.
do you think each job that exits should decrement his counter from the shared flow counter along its weight from total weight ? that way the jobs that remain will re-calibrate with the other remaining jobs

bardavid:
> > * How does this cope with the online conditions changing (new jobs arriving e.g. because of stonewall/wait_for, existing ones exiting while others are still running)?
> 
> it doesn't, i didn't consider it. thanks.
> do you think each job that exits should decrement his counter from the shared flow counter along its weight from total weight ? that way the jobs that remain will re-calibrate with the other remaining jobs.
> 
> so for example if we have a 1:2:3 ratio and job3 finishes earlier, the remaining jobs will continue with 1:2 ratio; in that case since both remaining jobs have enlarged their ratio from total (job1 from 1/6 to 1/3, job2 from 1/3 to 2/3) we will get a surge of operations until ratio has been achieved

@sitsofe  I fixed according to your comments,

I ended up modifying the code as I proposed above. I had to add an explicit call to `flow_exit_job(td);` when the job is reaped since it wasn't called on the new test case I added (t0014) which tests re-calibration after one of the jobs ends earlier. It seems to work though I'm not sure whether the explicit call I added is the best possible solution.

bardavid:
the checks seem to run on older versions of the commit, how can i rerun them ?

sitsofe:
Do another empty commit - you can always squash it later...

sitsofe:
@bardavid Hmm looks like its iops ratio mismatching. You might want to look further into that test...

sitsofe:
I'm not great with atomics but I've posted some comments.

flow.c
```diff
@@ -35,9 +38,10 @@ int flow_threshold_exceeded(struct thread_data *td)
 		return 1;
 	}
 
-	/* No synchronization needed because it doesn't
-	 * matter if the flow count is slightly inaccurate */
-	flow->flow_counter += td->o.flow;
+	/* increment flow(shared counter, therefore atomically) and job-specific counter */
+	__sync_fetch_and_add(&flow->flow_counter, 1);
```
sitsofe, line 42: Probably better to use C11 atomics everywhere and avoid the old-style GCC ones.

flow.c
```diff
@@ -78,13 +83,20 @@ static struct fio_flow *flow_get(unsigned int id)
...
+static void flow_put(struct fio_flow *flow, unsigned long long flow_counter, unsigned int weight)
 {
+	unsigned long long old_val;
+
 	if (!flow_lock)
 		return;
 
 	fio_sem_down(flow_lock);
 
+	old_val = __sync_fetch_and_sub(&flow->flow_counter, flow_counter);
```
sitsofe, line 95: As above re C11 atomics

t/run-fio-tests.py
```diff
@@ -420,6 +420,91 @@ def check_result(self):
...
+        # there are 9 samples for job1 and job2, 4 samples for job3
+        iops1 = 0.0
+        iops2 = 0.0
+        iops3 = 0.0
+        for i in range(9):
+            if i < 4:
+                iops3 = iops3 + float(iops_files[2][i].split(',')[1])
+            elif i == 4:
+                ratio1 = iops1/iops2
+                ratio2 = iops1/iops3
```
sitsofe, line 482: Minor: spaces around the operator?

thread_options.h
```diff
@@ -618,6 +615,13 @@ struct thread_options_pack {
 	fio_fp64_t latency_percentile;
 	uint32_t latency_run;
 
+	/*
+	 * flow support
+	 */
+	int32_t flow_id;
+	uint32_t flow;
+	uint32_t flow_sleep;
```
sitsofe, line 623: (FYI: FIO_SERVER_VER in https://github.com/axboe/fio/blob/0ce7e3e584be6f4f834fb2fec8e52deb23e1dd12/server.h#L51 would normally need upping on network format changes but it's only just been upped and we haven't cut another release so I wouldn't change it this time round).

flow.c
```diff
@@ -16,17 +17,19 @@ static struct fio_sem *flow_lock;
...
 
-	if (td->o.flow > 0)
-		flow_counter = flow->flow_counter;
-	else
-		flow_counter = -flow->flow_counter;
+	flow_counter_ratio = (double)td->flow_counter / __atomic_load_n(&flow->flow_counter, __ATOMIC_RELAXED);
+	flow_weight_ratio = (double)td->o.flow / __atomic_load_n(&flow->total_weight, __ATOMIC_RELAXED);
 
-	if (flow_counter > td->o.flow_watermark) {
+	/* each thread/process executing a fio job will stall based on the expected user ratio
```
sitsofe, line 28: Maybe wrap this at 80?

sitsofe:
@bardavid do you know if everything is still happy in client/server mode when using flow?

bardavid:
> @bardavid do you know if everything is still happy in client/server mode when using flow? Additionally when the tests fail could you output what the desired ratios are too?

t0014 and t0012 seem okay in client/server mode..
 I changed the failure output to your request

thread_options.h
bardavid: thanks, I wasn't aware of that

bardavid:
the tests are having a trouble remaining accurate. since i made the flow counters dynamic in that they can handle new jobs incoming and others prematurely ending (they re-calibrate the weight between the remaining jobs), i experience a surge in IOs due to jobs taken a LONG time to shutdown, giving the ones that are still alive a huge leverage.

i'll limit the max rate to make the tests more consistent

bardavid:
I'm having  some problems with the tests not getting correct IO ratios. although locally it works perfect. I added a debug commit.. i suspect it relates to constraints held by the Travis VM such as process niceness or whatever.

i'll address comments after

sitsofe:
@bardavid Don't forget you can create a Travis account and you can have CI on commits that are entirely within in your repo too.

flow.c
```diff
@@ -16,17 +17,19 @@ static struct fio_sem *flow_lock;
...
+	double flow_counter_ratio, flow_weight_ratio;
 
 	if (!flow)
 		return 0;
 
-	if (td->o.flow > 0)
-		flow_counter = flow->flow_counter;
-	else
-		flow_counter = -flow->flow_counter;
+	flow_counter_ratio = (double)td->flow_counter / atomic_load_relaxed(&flow->flow_counter);
```
bardavid, line 25: not sure if should just replace atomic loads with memory barrier

bardavid:
@sitsofe  what should i do next to escalate this PR ?

HOWTO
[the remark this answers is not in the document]
sitsofe: My apologises I think I mixed this one up.

thread_options.h
sitsofe: Looks like we've had a release so we'll need this.

sitsofe (changes requested):
We'll need at least the server version bump.

flow.c
bvanassche: I prefer atomic_load_relaxed() since using that primitive makes it clear which variable is shared with other threads. That is not the case when using memory barriers.

arch/arch.h
```diff
@@ -36,6 +36,13 @@ extern unsigned long arch_flags;
 
 #define ARCH_CPU_CLOCK_WRAPS
 
+#define atomic_inc(p, v)					\
```
bvanassche, line 39: Please rename this macro into atomic_add() since the implementation adds a value to an atomic counter instead of incrementing that counter by one.

sitsofe (approved):

sitsofe:
@axboe assuming you're happy with the non-backwards compatible change to flow this one LGTM!

axboe:
I actually pulled this, but then inspected the final result, and then un-pulled it. There are multiple issues in here, just from a quick look:

- Lots of places have weird line lengths and breaks. Please keep it at 80 in general.
- There are lots of places that have empty spaces at the end of lines. git diff shows these quite nicely
- Please always follow the existing code style. Comments are either one line /* like this */ or nicely broken over multiple lines. Look at the other comments in the code, none of this please:

```
/* foo foo foo
 * bar bar bar */
```

it should be:

```
/*
 * This is a comment over
 * multiple lines.
 */
```

- The main option change, "flow", is changed to FIO_OPT_ULL from FIO_OPT_INT, but the thread_options type is still an integer.
- For a change like this, the commit messages should be a *lot* more descriptive. Why is the change made? What does it enable that we could not do before?

sitsofe (changes requested):
@bardavid my apologies - it looks like I missed quite a bit during my review (frustratingly GitHub doesn't show end of line whitespace clearly but it's very visible in git itself). Would you be able to fix up Jens' feedback?

bardavid:
> @bardavid my apologies - it looks like I missed quite a bit during my review (frustratingly GitHub doesn't show end of line whitespace clearly but it's very visible in git itself). Would you be able to fix up Jens' feedback?

sure, i'm on it

bardavid:
> I actually pulled this, but then inspected the final result, and then un-pulled it. There are multiple issues in here, just from a quick look:
> 
> * Lots of places have weird line lengths and breaks. Please keep it at 80 in general.
> * There are lots of places that have empty spaces at the end of lines. git diff shows these quite nicely
> * Please always follow the existing code style. Comments are either one line /* like this */ or nicely broken over multiple lines. Look at the other comments in the code, none of this please:
> 
> ```
> /* foo foo foo
>  * bar bar bar */
> ```
> 
> it should be:
> 
> ```
> /*
>  * This is a comment over
>  * multiple lines.
>  */
> ```
> 
> * The main option change, "flow", is changed to FIO_OPT_ULL from FIO_OPT_INT, but the thread_options type is still an integer.
> * For a change like this, the commit messages should be a _lot_ more descriptive. Why is the change made? What does it enable that we could not do before?



thanks for the review.

i fixed accordingly, although i wasn't sure if the same comments standard was being followed on the python files as well.  Please tell me if those need fixup as well.

regarding:
> The main option change, "flow", is changed to FIO_OPT_ULL from FIO_OPT_INT, but the thread_options type is still an integer.
"flow" is set to uint32_t / unsigned int on the thread_options.. perhaps you're referring to "flow_id"  ? anyway that wasn't changed

axboe:
BTW, this is readily apparent in that the examples/flow.fio doesn't work properly. You set flow_sleep=1000, but then flow is set after that. The latter set will clobber flow_sleep, and you end up with a flow_sleep value of 0 instead of 1000. I'm going to just change it to FIO_OPT_INT again for now, since the current state is most definitely very broken.

Issue #1065, opened 2020-08-13
version 3.18 sets stonewall incorrectly on s390x


tdi2004:
Hi @hannesweisbach,
assume the following jobfile:
```
[global]
blocksize_range=4k-32k
rw=randrw
percentage_random=50
rwmixread=50
iodepth=4
ioengine=libaio
verify=crc32c
verify_fatal=1
verify_dump=1
verify_backlog=100000
time_based
runtime=14d
exitall_on_error=1
numjobs=1

[LV1]
size=133722k
directory=/mnt1/LV1 

[LV2]
size=124315k
directory=/mnt1/LV2

[LV3]
size=124873k
directory=/mnt1/LV3

[LV4]
size=133380k
directory=/mnt1/LV4

[LV5]
size=168411k
directory=/mnt1/LV5

```
with commit 64402a8a7b158c1b7a778b5f2508e625a6fc6317 a change in behaviour was introduced.
While older fio versions started the fio parent PID and one fio child PID per each directory to run fio on several mountpoints in parallel, this is now broken. fio runs only on the first directory specified:

```
root@t83lp32:~/fio-fio-3.18# ./fio jobfile.fio 
LV1: (g=0): rw=randrw, bs=(R) 4096B-32.0KiB, (W) 4096B-32.0KiB, (T) 4096B-32.0KiB, ioengine=libaio, iodepth=4
LV2: (g=1): rw=randrw, bs=(R) 4096B-32.0KiB, (W) 4096B-32.0KiB, (T) 4096B-32.0KiB, ioengine=libaio, iodepth=4
LV3: (g=2): rw=randrw, bs=(R) 4096B-32.0KiB, (W) 4096B-32.0KiB, (T) 4096B-32.0KiB, ioengine=libaio, iodepth=4
LV4: (g=3): rw=randrw, bs=(R) 4096B-32.0KiB, (W) 4096B-32.0KiB, (T) 4096B-32.0KiB, ioengine=libaio, iodepth=4
LV5: (g=4): rw=randrw, bs=(R) 4096B-32.0KiB, (W) 4096B-32.0KiB, (T) 4096B-32.0KiB, ioengine=libaio, iodepth=4
fio-3.18
Starting 5 processes
Jobs: 1 (f=1): [m(1),P(4)][0.0%][r=95.0MiB/s,w=48.1MiB/s][r=5726,w=2854 IOPS][eta 69d:23h:59m:46s]    
```
It shows only one job! Previous version showed five in my case.
[...]

PR #1066, opened 2020-08-14
Use unsigned short type for exit_what and stonewall

XeS0r:
Fixes: 64402a8 ("Expand choices for exitall")
Fixes: https://github.com/axboe/fio/issues/1065
Signed-off-by: André Wild <wild.andre.ae@gmail.com>

### DIFF
--- a/cconv.c
+++ b/cconv.c
@@ -237,8 +237,8 @@ void convert_thread_options_to_cpu(struct thread_options *o,
 	o->loops = le32_to_cpu(top->loops);
 	o->mem_type = le32_to_cpu(top->mem_type);
 	o->mem_align = le32_to_cpu(top->mem_align);
-	o->exit_what = le16_to_cpu(top->exit_what);
-	o->stonewall = le16_to_cpu(top->stonewall);
+	o->exit_what = le32_to_cpu(top->exit_what);
+	o->stonewall = le32_to_cpu(top->stonewall);
 	o->new_group = le32_to_cpu(top->new_group);
 	o->numjobs = le32_to_cpu(top->numjobs);
 	o->cpus_allowed_policy = le32_to_cpu(top->cpus_allowed_policy);
@@ -437,8 +437,8 @@ void convert_thread_options_to_net(struct thread_options_pack *top,
 	top->loops = cpu_to_le32(o->loops);
 	top->mem_type = cpu_to_le32(o->mem_type);
 	top->mem_align = cpu_to_le32(o->mem_align);
-	top->exit_what = cpu_to_le16(o->exit_what);
-	top->stonewall = cpu_to_le16(o->stonewall);
+	top->exit_what = cpu_to_le32(o->exit_what);
+	top->stonewall = cpu_to_le32(o->stonewall);
 	top->new_group = cpu_to_le32(o->new_group);
 	top->numjobs = cpu_to_le32(o->numjobs);
 	top->cpus_allowed_policy = cpu_to_le32(o->cpus_allowed_policy);
--- a/examples/exitwhat.fio
+++ b/examples/exitwhat.fio
@@ -1,7 +1,7 @@
 # We want to run fast1 as long as slow1 is running, but also have a cumulative
 # report of fast1 (group_reporting=1/new_group=1).  exitall=1 would not cause
 # fast1 to stop after slow1 is done. Setting exit_what=stonewall will cause
-# alls jobs up until the next stonewall=1 setting to be stopped, when job slow1
+# alls jobs up until the next stonewall setting to be stopped, when job slow1
 # finishes.
 # In this example skipping forward to slow2/fast2. slow2 has exit_what=all set,
 # which means all jobs will be cancelled when slow2 finishes. In particular,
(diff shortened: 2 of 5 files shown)


sitsofe:
@XeS0r Nice! Could you:

- Change the title so it prefixes the area (init: use unsigned short type for exit_what and stonewall)

And that's about all...

sitsofe:
Actually your fix has uncovered a bug in https://github.com/axboe/fio/blob/5c79c32f6afb39fae910912475e8fea786c3353e/examples/exitwhat.fio#L35 - stonewall doesn't take any value (see https://fio.readthedocs.io/en/latest/fio_man.html#cmdoption-arg-stonewall ). Could you fix that up too?

sitsofe:
The more I think about it the more I wonder - wouldn't it be better just to convert both these values (exit_what, stonewall) to be ints rather than shorts even if they could be smaller (even FIO_BOOL seems to go into an int)? numa_mem_mode gets away with it because it's not a fully fledged option passed in via the command line...

XeS0r:
> Actually your fix has uncovered a bug in
> 
> https://github.com/axboe/fio/blob/5c79c32f6afb39fae910912475e8fea786c3353e/examples/exitwhat.fio#L35
> - stonewall doesn't take any value (see https://fio.readthedocs.io/en/latest/fio_man.html#cmdoption-arg-stonewall ). Could you fix that up too?

I already fixed this issue. But I want to make sure that I don't find anything else. Now that I'm aware of the unit tests.

XeS0r:
> The more I think about it the more I wonder - wouldn't it be better just to convert both these values (exit_what, stonewall) to be ints rather than shorts even if they could be smaller (even FIO_BOOL seems to go into an int)? numa_mem_mode gets away with it because it's not a fully fledged option passed in via the command line...

I can also change them both to int if you prefer that solution.

sitsofe:
If it looks like it would be smaller that would be the way to go (we would also likely need to bump FIO_VER in https://github.com/axboe/fio/blob/master/server.h#L51 )...

axboe:
Yeah just make them ints, and increment the ```FIO_SERVER_VER```.

sitsofe:
@XeS0r Looks good:
- Can you add a prefix to the commit message (see https://github.com/axboe/fio/pull/1050#issuecomment-663606909 for tips on the "house style")
- Could you fix up the incorrect test file https://github.com/axboe/fio/blob/5c79c32f6afb39fae910912475e8fea786c3353e/examples/exitwhat.fio#L35 ?

sitsofe:
Ah one more:
- I think you need to fix up the packed structure too - https://github.com/axboe/fio/blob/5c79c32f6afb39fae910912475e8fea786c3353e/thread_options.h#L497

XeS0r:
> @XeS0r Looks good:
> 
>     * Can you add a prefix to the commit message (see [#1050 (comment)](https://github.com/axboe/fio/pull/1050#issuecomment-663606909) for tips on the "house style")
> 

Yes sure, hopefully this prefix is fulfilling your requirements.

>     * Could you fix up the incorrect test file https://github.com/axboe/fio/blob/5c79c32f6afb39fae910912475e8fea786c3353e/examples/exitwhat.fio#L35
>        ?

I've also fixed the comment.

examples/exitwhat.fio
```diff
@@ -32,7 +32,7 @@ iodepth=32
 rate=300,300,300
 
 [slow2]
-stonewall=1
```
axboe, line 35: If this change in job files is required, then we'll have to fix it so that's not the case. If it's not required, don't make the change in the job file.

cconv.c
```diff
@@ -237,8 +237,8 @@ void convert_thread_options_to_cpu(struct thread_options *o,
 	o->loops = le32_to_cpu(top->loops);
 	o->mem_type = le32_to_cpu(top->mem_type);
 	o->mem_align = le32_to_cpu(top->mem_align);
-	o->exit_what = le16_to_cpu(top->exit_what);
```
axboe, line 240: Pretty sure this breaks alignment and compilation will fail...

axboe:
Looks good to me now, except you still have the exitwhat.fio change in there. Please get rid of that.

axboe:
Not saying it's incorrect, but it doesn't matter and it's not related. You can do that as a separate commit, if you'd like. Ditto the fio.1 change, that should be a separate commit as well.

axboe:
Ah screw it, I'll just pull it as it is, not worth changing it for imho.

axboe:
Thanks for fixing this issue!

examples/exitwhat.fio
XeS0r: Stonewalls usage is not documented like this. I've changed the man page and also fixed some other stuff in the example. Now it's up to you if we should use `stonewall=1` in the example or not. Both will work and it's not required to set a value. But all other examples just use stonewall without any value assigned.

cconv.c
XeS0r: Yes, I've now removed one of the manual paddings to get it aligned again.

XeS0r:
I would have changed that too. Was not sure If you want separate commits. Next time I'm aware of that. Thanks for your patience.

Issue #1033, opened 2020-07-04
Add option to allow use of O_DSYNC (rather than O_SYNC).


anarazel:
Hi,

Using O_DSYNC has obvious advantages over O_SYNC, but right now it can't be benchmarked with fio. Would you accept a patch that either allows sync/dsync as parameters for --sync or adds a new --dsync?

Thanks

axboe:
Yeah I think that'd be fine. If it turns out clean enough to turn "sync" into a posval kind of setup where "dsync" becomes a valid option, than that'd be best. If not, then I guess adding "dsync" would be the next best thing.

fxober:
Is this feature merged, there is a use case at Intel we would like to test?

axboe:
Andres, did you ever write that patch?

anarazel:
I did write and use it locally. Probably needs a bit more polish.... See PR...

PR #1080, opened 2020-09-04
Add O_DSYNC support

anarazel:
This is my first stab at this, there might be more places that need to be touched.

What do you think about my approach of extending --sync rather than adding a --dsync? And about just storing the flag bits in `td->o.sync_io`?

See  #1033

### DIFF
--- a/engines/libpmem.c
+++ b/engines/libpmem.c
@@ -193,6 +193,7 @@ static enum fio_q_status fio_libpmem_queue(struct thread_data *td,
 
 	dprint(FD_IO, "DEBUG fio_libpmem_queue\n");
 	dprint(FD_IO,"td->o.odirect %d td->o.sync_io %d \n",td->o.odirect, td->o.sync_io);
+	/* map both O_SYNC / DSYNC to not using NODRAIN */
 	flags = td->o.sync_io ? 0 : PMEM_F_MEM_NODRAIN;
 	flags |= td->o.odirect ? PMEM_F_MEM_NONTEMPORAL : PMEM_F_MEM_TEMPORAL;
 
--- a/options.c
+++ b/options.c
@@ -3733,14 +3733,32 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
 	{
 		.name	= "sync",
 		.lname	= "Synchronous I/O",
-		.type	= FIO_OPT_BOOL,
+		.type	= FIO_OPT_STR,
 		.off1	= offsetof(struct thread_options, sync_io),
-		.help	= "Use O_SYNC for buffered writes",
-		.def	= "0",
-		.parent = "buffered",
+		.help	= "Use synchronous write IO",
+		.def	= "none",
 		.hide	= 1,
 		.category = FIO_OPT_C_IO,
 		.group	= FIO_OPT_G_IO_TYPE,
+		.posval = {
+			  { .ival = "none",
+			    .oval = 0,
+			  },
+			  { .ival = "0",
+			    .oval = 0,
+			  },
+			  { .ival = "sync",
+			    .oval = O_SYNC,
+			  },
+			  { .ival = "1",
+			    .oval = O_SYNC,
+			  },
+#ifdef O_DSYNC
+			  { .ival = "dsync",
+			    .oval = O_DSYNC,
+			  },
+#endif
+		},
 	},
 #ifdef FIO_HAVE_WRITE_HINT
 	{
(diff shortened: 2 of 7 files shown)


sitsofe:
- Could you add `Fixes: https://github.com/axboe/fio/issues/1033`
- WIll this be backwards compatible for people who are using sync=1 ?
- fio assumes we always have `O_SYNC` (and defines it on Windows where it doesn't exist) so you can skip the ifdef on that section
- Does the HOWTO and fio.1 need updating?
- @lukaszstolarczuk - is the pmem change OK?

options.c
```diff
@@ -3717,14 +3717,35 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
 	{
 		.name	= "sync",
 		.lname	= "Synchronous I/O",
-		.type	= FIO_OPT_BOOL,
+		.type	= FIO_OPT_STR,
 		.off1	= offsetof(struct thread_options, sync_io),
 		.help	= "Use O_SYNC for buffered writes",
```
axboe, line 3722: This should probably change to "Use O_SYNC/O_DSYNC for buffered writes"

options.c
```diff
@@ -3717,14 +3717,35 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
 	{
 		.name	= "sync",
 		.lname	= "Synchronous I/O",
-		.type	= FIO_OPT_BOOL,
+		.type	= FIO_OPT_STR,
```
axboe, line 3720: I think this is a good way to do it, as it's backwards compatible and avoids adding a new option.

axboe:
pmem and option change looks fine to me, main one missing is the documentation update. Once that's done, I'd say it's good to go.

sitsofe:
@anarazel ping. Could you:
- Fix up the option description
- Add docs to the HOWTO and fio.1?
- Add Fixes: https://github.com/axboe/fio/issues/1033

And then this will be mergeable!

sitsofe (changes requested):
(Minor changes requested)

engines/libpmem.c
```diff
@@ -193,6 +193,7 @@ static enum fio_q_status fio_libpmem_queue(struct thread_data *td,
 
 	dprint(FD_IO, "DEBUG fio_libpmem_queue\n");
 	dprint(FD_IO,"td->o.odirect %d td->o.sync_io %d \n",td->o.odirect, td->o.sync_io);
+	/* probably fine to treat both O_SYNC / DSYNC as NODRAIN */
 	flags = td->o.sync_io ? 0 : PMEM_F_MEM_NODRAIN;
```
lukaszstolarczuk, line 197: behavior seems fine, but the comment is the other way around ;) if `td->o.sync_io` is set (to either one) than we don't want to set up the flag "NODRAIN"

options.c
anarazel: Why is O_SYNC documented to be about buffered writes? As far as I understand both O_SYNC and O_DSYNC have effects for O_DIRECT style IO as well?

axboe:
You are correct, that should remove “buffered” from the wording.

engines/libpmem.c
anarazel: Oops, you're right of course. Should be fixed now.

anarazel:
@sitsofe 

>    Could you add Fixes: https://github.com/axboe/fio/issues/1033

Done

>  WIll this be backwards compatible for people who are using sync=1 ?

Yep, should continue to work - it looks like bool opts only allow 0/1, which I added as alternatives to `sync={sync,dsync,none}`.

> fio assumes we always have O_SYNC (and defines it on Windows where it doesn't exist) so you can skip the ifdef on that section

Done for `O_SYNC`. Or would you prefer to deal with `O_DSYNC` the same way?

> Does the HOWTO and fio.1 need updating?

Updated. I am however somewhat confused about how exactly fio.1 is maintained - HOWTO can be converted to manpage with `make -C doc man`, but it looks like the fio.1 in the root dir is manually maintained?

@axboe 
> You are correct, that should remove “buffered” from the wording.

Tried doing that. Not quite sure what the deal with 'parent' is in options though? Hope this makes sense.

axboe:
> Updated. I am however somewhat confused about how exactly fio.1 is maintained - HOWTO can be converted to manpage with make -C doc man, but it looks like the fio.1 in the root dir is manually maintained?

I think last time this was checked, there was still bits missing or not in sync. But yeah, that would be the goal.

> Not quite sure what the deal with 'parent' is in options though?

It's for options that only makes sense with a higher level option. Like ```offset_increment``` which doesn't make sense without ```offset```, which is why it's listed as its parent.

PR #1081, opened 2020-09-05
Rework the interval timer code

bvanassche:
Make the interval timer more accurate.

### DIFF
--- a/gettime.c
+++ b/gettime.c
@@ -524,23 +524,33 @@ uint64_t mtime_since_now(const struct timespec *s)
 	return mtime_since(s, &t);
 }
 
-uint64_t mtime_since(const struct timespec *s, const struct timespec *e)
+/*
+ * Returns *e - *s in milliseconds as a signed integer. Note: rounding is
+ * asymmetric. If the difference yields +1 ns then 0 is returned. If the
+ * difference yields -1 ns then -1 is returned.
+ */
+int64_t rel_time_since(const struct timespec *s, const struct timespec *e)
 {
-	int64_t sec, usec;
+	int64_t sec, nsec;
 
 	sec = e->tv_sec - s->tv_sec;
-	usec = (e->tv_nsec - s->tv_nsec) / 1000;
-	if (sec > 0 && usec < 0) {
+	nsec = e->tv_nsec - s->tv_nsec;
+	if (nsec < 0) {
 		sec--;
-		usec += 1000000;
+		nsec += 1000ULL * 1000 * 1000;
 	}
+	assert(0 <= nsec && nsec < 1000ULL * 1000 * 1000);
 
-	if (sec < 0 || (sec == 0 && usec < 0))
-		return 0;
+	return sec * 1000 + nsec / (1000 * 1000);
+}
 
-	sec *= 1000;
-	usec /= 1000;
-	return sec + usec;
+/*
+ * Returns *e - *s in milliseconds as an unsigned integer. Returns 0 if
+ * *e < *s.
+ */
+uint64_t mtime_since(const struct timespec *s, const struct timespec *e)
+{
+	return max(rel_time_since(s, e), (int64_t)0);
 }
 
 uint64_t time_since_now(const struct timespec *s)
--- a/gettime.c
+++ b/gettime.c
@@ -239,19 +239,13 @@ static unsigned long get_cycles_per_msec(void)
 {
 	struct timespec s, e;
 	uint64_t c_s, c_e;
-	enum fio_cs old_cs = fio_clock_source;
 	uint64_t elapsed;
 
-#ifdef CONFIG_CLOCK_GETTIME
-	fio_clock_source = CS_CGETTIME;
-#else
-	fio_clock_source = CS_GTOD;
-#endif
-	__fio_gettime(&s);
+	fio_get_mono_time(&s);
 
 	c_s = get_cpu_clock();
 	do {
-		__fio_gettime(&e);
+		fio_get_mono_time(&e);
 		c_e = get_cpu_clock();
 
 		elapsed = ntime_since(&s, &e);
@@ -259,7 +253,6 @@ static unsigned long get_cycles_per_msec(void)
 			break;
 	} while (1);
 
-	fio_clock_source = old_cs;
 	return (c_e - c_s) * 1000000 / elapsed;
 }
 
--- a/gettime.c
+++ b/gettime.c
@@ -127,18 +127,33 @@ static void fio_init gtod_init(void)
 
 #endif /* FIO_DEBUG_TIME */
 
-#ifdef CONFIG_CLOCK_GETTIME
-static int fill_clock_gettime(struct timespec *ts)
+/*
+ * Queries the value of the monotonic clock if a monotonic clock is available
+ * or the wall clock time if no monotonic clock is available. Returns 0 if
+ * querying the clock succeeded or -1 if querying the clock failed.
+ */
+int fio_get_mono_time(struct timespec *ts)
 {
-#if defined(CONFIG_CLOCK_MONOTONIC_RAW)
-	return clock_gettime(CLOCK_MONOTONIC_RAW, ts);
-#elif defined(CONFIG_CLOCK_MONOTONIC)
-	return clock_gettime(CLOCK_MONOTONIC, ts);
+	int ret;
+
+#ifdef CONFIG_CLOCK_GETTIME
+#if defined(CONFIG_CLOCK_MONOTONIC)
+	ret = clock_gettime(CLOCK_MONOTONIC, ts);
 #else
-	return clock_gettime(CLOCK_REALTIME, ts);
+	ret = clock_gettime(CLOCK_REALTIME, ts);
 #endif
+#else
+	struct timeval tv;
+
+	ret = gettimeofday(&tv, NULL);
+	if (ret == 0) {
+		ts->tv_sec = tv.tv_sec;
+		ts->tv_nsec = tv.tv_usec * 1000;
+	}
+#endif
+	assert(ret <= 0);
+	return ret;
 }
-#endif
 
 static void __fio_gettime(struct timespec *tp)
 {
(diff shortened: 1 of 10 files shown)


bvanassche:
@sitsofe, can you take a look?

backend.c
```diff
@@ -1458,16 +1458,17 @@ static bool keep_running(struct thread_data *td)
 	return false;
 }
 
-static int exec_string(struct thread_options *o, const char *string, const char *mode)
+static int exec_string(struct thread_options *o, const char *string,
+		       const char *mode)
```
sitsofe, line 1462: Can't tell if this indenting is correct (might just be github)

sitsofe (commented):
Minor comments but the overall approach looks good.

gettime.c
```diff
@@ -516,23 +516,29 @@ uint64_t mtime_since_now(const struct timespec *s)
...
 	sec = e->tv_sec - s->tv_sec;
-	usec = (e->tv_nsec - s->tv_nsec) / 1000;
-	if (sec > 0 && usec < 0) {
+	nsec = e->tv_nsec - s->tv_nsec;
+	if (nsec < 0) {
 		sec--;
-		usec += 1000000;
+		nsec += 1000ULL * 1000 * 1000;
 	}
+	assert(0 <= nsec && nsec < 1000ULL * 1000 * 1000);
```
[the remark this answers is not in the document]
bvanassche: I think nsec will be negative for e.g. the following example:
e = { 0, 0 }
s = { 0, 1000000001 }

gettime.c
```diff
@@ -516,23 +516,29 @@ uint64_t mtime_since_now(const struct timespec *s)
 	return mtime_since(s, &t);
 }
 
-uint64_t mtime_since(const struct timespec *s, const struct timespec *e)
+/* Returns *e - *s in milliseconds as a signed integer. */
+int64_t rel_time_since(const struct timespec *s, const struct timespec *e)
```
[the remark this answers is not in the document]
bvanassche: Please take a look at the has_expired() function added by a later patch. That function checks whether or not a timer has expired by examining the sign of the rel_time_since() return value.

backend.c
bvanassche: In my editor indentation looks correct. Maybe github has trouble with displaying tabs correctly?

vincentkfu:
Perhaps the delays are related to `timeout` for the `select()` system call. The man page says:

> Note that the timeout interval will be rounded up to the system clock granularity, and kernel scheduling delays mean that the blocking interval may overrun by a small amount.

Consider making `timeout` slightly less than `msec_to_next_event` and skipping `read_from_pipe()` when `msec_to_next_event` is very small.

gettime.c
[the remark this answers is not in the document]
sitsofe: Ah go it.

gettime.c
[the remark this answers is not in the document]
sitsofe: OK I follow now.

sitsofe:
The latest change gets rid of `CLOCK_MONOTONIC_RAW` entirely (and you give the rationale in one of the commits). The only issue I have is that `CLOCK_MONOTONIC` can and is slewed by things like NTP. Technically this means a second is more likely to actually BE a second but it also means that the frequency the clock is ticking at is variable between NTP adjustments. I guess it depends what's more important to fio: consistent intervals or accurate seconds? I think when timing small periods the first is preferred but I'm happy to be convinced otherwise.

(Coincidentally the current approach  "solves" what https://github.com/axboe/fio/pull/857 was trying to address on old kernels)

bvanassche:
> Perhaps the delays are related to `timeout` for the `select()` system call. The man page says:
> 
> > Note that the timeout interval will be rounded up to the system clock granularity, and kernel scheduling delays mean that the blocking interval may overrun by a small amount.
> 
> Consider making `timeout` slightly less than `msec_to_next_event` and skipping `read_from_pipe()` when `msec_to_next_event` is very small.

Hmm ... why to skip read_from_pipe()? read_from_pipe() reads from a non-blocking pipe and should not take more than a few microseconds.

I have been experimenting with decreasing msec_to_next_event but that did not yield the desired effect.

A high-res timer has been introduced instead.

bvanassche:
> The latest change gets rid of `CLOCK_MONOTONIC_RAW` entirely (and you give the rationale in one of the commits). The only issue I have is that `CLOCK_MONOTONIC` can and is slewed by things like NTP. Technically this means a second is more likely to actually BE a second but it also means that the frequency the clock is ticking at is variable between NTP adjustments. I guess it depends what's more important to fio: consistent intervals or accurate seconds? I think when timing small periods the first is preferred but I'm happy to be convinced otherwise.
> 
> (Coincidentally the current approach "solves" what #857 was trying to address on old kernels)

I do not understand the concern about variability of time intervals when using CLOCK_MONOTONIC? ntpd only needs one hour to measure the clock drift. Once the clock drift has been measured it is saved in a file and that file is read the next time ntpd starts (/var/lib/ntp/drift or /var/lib/chrony/drift). See also http://doc.ntp.org/4.1.0/ntpd.htm.

Additionally, ntpd supports frequency deviations of at most 500 ppm (0.05%). Any clock crystal that deviates more is considered as defective by ntpd.

Stepwise adjustments of the clock by ntpd affect the real-time clock but not the monotonic clock.

bvanassche:
@sitsofe @vincentkfu This patch series is ready for re-review. The controversial patches have been dropped from this series.

helper_thread.c
```diff
@@ -172,90 +199,119 @@ static uint8_t wait_for_action(int fd, unsigned int timeout_ms)
...
+
+#ifdef _SC_CLK_TCK
+	clk_tck = sysconf(_SC_CLK_TCK);
+#else
+	/*
+	 * The timer frequence is variable on Windows. Instead of trying to
+	 * query it, use 100 Hz. See also
+	 * https://carpediemsystems.co.uk/2019/07/18/windows-system-timer-granularity/.
+	 */
+	clk_tck = 100;
```
sitsofe, line 296: I remember reading Bruce Dawson's article (https://randomascii.wordpress.com/2013/07/08/windows-timer-resolution-megawatts-wasted/ ) that mentioned this was introduced with Windows 8. This is tricky because the timer may be less accurate than 100Hz (1000ms / 15.6ms ~ 64 Hz)? I guess likewise if something has forced the whole system to have an increased clock rate (https://stackoverflow.com/a/48011619 ) then the accuracy will be much finer. Why did you pick 100? Should fio just request more accurate timers?

helper_thread.c
```diff
@@ -188,14 +193,36 @@ static uint8_t wait_for_action(int fd, unsigned int timeout_ms)
...
+		 */
+		struct itimerspec delta = {};
+
+		delta.it_value.tv_sec = timeout.tv_sec;
+		delta.it_value.tv_nsec = timeout.tv_usec * 1000;
+		res = timerfd_settime(timerfd, 0, &delta, NULL);
+		assert(res == 0);
+		FD_SET(timerfd, &rfds);
+	}
+#endif
```
sitsofe, line 212: I feel a little sad seeing this in this particular file but there's absolutely nothing portable that provides improved granularity AND works on all platforms short of reaching for some giant library like libevent (and arguably non-Linux platforms are no worse off than today) without open coding. My only thought is: should "a more granular select" be pushed into platform specific code?

helper_thread.c
```diff
@@ -188,14 +193,36 @@ static uint8_t wait_for_action(int fd, unsigned int timeout_ms)
...
 	if (res < 0) {
 		log_err("fio: select() call in helper thread failed: %s",
 			strerror(errno));
 		return A_EXIT;
 	}
 	if (FD_ISSET(fd, &rfds))
 		read_from_pipe(fd, &action, sizeof(action));
+	if (timerfd >= 0 && FD_ISSET(timerfd, &rfds)) {
+		res = read(timerfd, &exp, sizeof(exp));
+		assert(res == sizeof(exp));
```
sitsofe, line 224: How frequently will the if body be called?

sitsofe:
(Aside: https://github.com/axboe/fio/pull/982 appears to clamp log invervals down to the nearest interval. If we had that approach and people were choosing to use `log_avg_msec` would impact what we should do here?)

helper_thread.c
bvanassche: I can change this into 64 Hz if you want. This value should be the lowest timer frequency of any system fio runs on.

helper_thread.c
bvanassche: Making select() more granular would be suboptimal because a generic implementation either would have to select a timerfd from a pool or dynamically create the timerfd. The code in this pull request only creates one timerfd per helper thread. I think a similar comment applies to poll() and epoll().

helper_thread.c
bvanassche: Every time the timeout for `timerfd` has been reached. However, the time spent in the read() call should be small since all that call does is to read 8 bytes from kernel memory.

bvanassche:
> (Aside: #982 appears to clamp log invervals down to the nearest interval. If we had that approach and people were choosing to use `log_avg_msec` would impact what we should do here?)

Recently complaints have been posted on the fio mailing list about log_avg_msec not being applied accurately. I'm concerned clamping log intervals down would make some users unhappy ...

helper_thread.c
sitsofe: Oof that would be very complicated... OK understood.

helper_thread.c
sitsofe: I think that would be better.

sitsofe:
> Recently complaints have been posted on the fio mailing list about log_avg_msec not being applied accurately. I'm concerned clamping log intervals down would make some users unhappy ...

:-) OK let's go press on with what you're proposing as it seems a more detailed approach.

bvanassche:
> (Aside: #982 appears to clamp log invervals down to the nearest interval. If we had that approach and people were choosing to use `log_avg_msec` would impact what we should do here?)

Hmm ... I would need to spend some time to understand that code properly before I can answer this question.

helper_thread.c
bvanassche: The constant 100 has been changed into 64.

bvanassche:
@axboe , I think this branch is ready for merging.

axboe:
Done, thanks Bart

PR #1082, opened 2020-09-06
Makefile/configure: fix guasi build

sitsofe:
The (fairly old!) GUASI engine didn't build but these fixes seemed to get it working again. I doubt it's an engine that sees much use these days but amazingly it still works if you get all the pieces...


axboe:
We should probably just drop it... I'll pull these to make it working again, but probably no point in carrying this one forward.

axboe:
BTW, please add your Signed-off-by to commits in the future!

Issue #1096, opened 2020-10-06
Avoid arithmetic on a pointer to void error in parse.h


chengli-rutgers:
Hi guys --

We found there is an arithmetic on a pointer to void error in parse.h:128 when compiled with c++14 compiler. 
```
parse.h:128:13 error: arithmetic on a pointer to void
              return ret + offset;
                       ~~~ ^
```

To broaden fio's impact, we would like to see it can be compiled with recent c++ compilers.

More specifically, we would like to make the following change from
```
return ret + offset;
```
to 
```
return (void*) ((uint_8*) ret + offset);
```
We would like to get an opinion on whether this is something the fio community can help fix. Or, we could work on a patch to fix this issue as well.

axboe:
Sure, if it's done cleanly enough. But have you checked the scope of this? It'll be quite a bit more than just that one spot in parse.h

chengli-rutgers:
Thanks for the comments. The compiler only complains this spot so far. Please let us know if anything else must be handled if we change this line.

Our current plan is to only change this line and cherry pick it back to our branch.

axboe:
You should be able to add -Wpointer-arith to the Makefile, that'll find a bunch more. I'm surprised your compiler only complains about the parse.h instance.

chengli-rutgers:
You are right. The mainstream fio contains a bunch more.

Looks like we walked around most of the other `arithmetic on a pointer` instances by not linking them. But, the one in the parse.h are shared by many core source code so it is hard to walk around. So, the compiler only complains this particular instance. 

Is it still a valid patch (though not comprehensive) if we only change the instance in parse.h?

axboe:
Yeah that's fine, feel free to send that patch in. Do note spacing though, follow the fio general codestyle. Which means it needs to look like this:

```
return (void *) ((uint_8 *) ret + offset);
```

and I think it'd be cleaner as:

```
return (void *) ((uintptr_t) ret + offset);
```

sitsofe:
The original instance of this has been closed by https://github.com/axboe/fio/pull/1098 . As no one is working on the comprehensive case mentioned in https://github.com/axboe/fio/issues/1096#issuecomment-704614087 do we win much from keeping this open?

PR #1098, opened 2020-10-07
Update parse.h to avoid arithmetic on a pointer to void cast (#1096)

chengli-rutgers:
axboe#1096 shows that a compiler complains an arithmetic on a pointer to void error when -Wpointer-arith is turned on in Makefile. Although there are many violating instances, we only fix the one in parse.h because it is shared across many core source code files. This patch fixes the arithmetic on a pointer to void issue by casting it to a uintptr_t type first and perform manipulation, and then cast it back to void* type.

### DIFF
--- a/parse.h
+++ b/parse.h
@@ -125,7 +125,7 @@ static inline void *td_var(void *to, const struct fio_option *o,
 	else
 		ret = to;
 
-	return ret + offset;
+	return (void *) ((uintptr_t) ret + offset);
 }
 
 static inline int parse_is_percent(unsigned long long val)


sitsofe:
Hmm, 
- Could you squash these commits into one?
- Could you add your signed-off-by line?
- Could you reformat the commit message?

See https://github.com/axboe/fio/pull/1050#issuecomment-663606909 for the house style :-)

sitsofe (changes requested):
(Please see other comment for the requested changes)

sitsofe (approved):
LGTM!

Issue #1100, opened 2020-10-13
Avoid global variable initialization in getopt_long.c


chengli-rutgers:
Hi guys --

Our Address Sanitizer (ASAN) analysis complains about a few global variables that are initialized in the getopt_long.c file. More specifically, the following variables can be cleaned up without a global initialization:

```
char *optarg = NULL;
int optind = 0, opterr = 0, optopt = 0;
```

1) optarg is initialized in oslib/getopt_long.c:49 so no need to initialize it in line 19.
2) optind is initialized in init.c:2477 so no need to initialize it in line 20.
3) opterr is initialized in oslib/getopt_long.c:188 so no need to initialize it in line 20.
4) opterr appears not to be used.

Also, we assume global variables will be initialized to zero by default.

So, we propose to change the code to the following:

```
char *optarg;
int optind, opterr, optopt;
```
Is it something that the fio community can help fix? Or, we can work on a patch to fix it as well.

axboe:
Would be great if you can send a patch.

PR #1102, opened 2020-10-13
getopt_long: avoid variable global initialization

chengli-rutgers:
Issue #1100 shows that an address sanitizer(ASAN) complains about a few
global variables that are initialized globally in the getopt_long.c file.
We look into these variables and found they do not need to be initialized
globally. This patch fixes the issue by cleaning up the global
initialization for these variables.

Signed-off-by: Cheng Li <chenglii@cs.rutgers.edu>

### DIFF
--- a/oslib/getopt_long.c
+++ b/oslib/getopt_long.c
@@ -16,8 +16,8 @@
 
 #include "getopt.h"
 
-char *optarg = NULL;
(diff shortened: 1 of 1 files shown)


sitsofe:
Could you add
```
Fixes https://github.com/axboe/fio/issues/1100
```

just above your signed off by? But otherwise LGTM!

PR #1103, opened 2020-10-17
stat: Increase depth levels to log latency

ikegami-t:
Previously latency log depth levels are 10 for nsec and usec and 12 for msec.
But to measure latency variation on a finer granularity required to increase.
So increase the depth leves as 20 for nsec and usec and 22 for msec.

Signed-off-by: Tokunori Ikegami <ikegami.t@gmail.com>


sitsofe:
I don't know we can accept this as-is it's bound to break people depending on the previous behaviour who aren't checking the headings... Also wouldn't the documentation have to be updated (https://github.com/axboe/fio/blob/8bfe330eb42739d503d35d0b7d96f98c5c544204/HOWTO#L3884-L3890 )?

ikegami-t:
Hi,

> I don't know we can accept this as-is - it's bound to break people depending on the previous behaviour who aren't checking the headings... Also wouldn't the documentation have to be updated (
> 
> https://github.com/axboe/fio/blob/8bfe330eb42739d503d35d0b7d96f98c5c544204/HOWTO#L3884-L3890
> 
> )?

Thanks for the comment.
Just update the document as mentioned.

Regards,
Ikegami

sitsofe:
@axboe what are your thoughts on this one?

HOWTO
```diff
@@ -3881,13 +3881,17 @@ field was introduced or whether it's specific to some terse version):
 
         <=1, 2, 4, 8, 16, 32, >=64
 
+    I/O latencies nanoseconds::
+
+        <=2, 3, 4, 6, 8, 10, 15, 20, 35, 50, 75, 100, 175, 250, 375, 500, 625, 750, 875, 1000
+
```
sitsofe, line 3887: You would have to fix up the man page (fio.1) too...

ikegami-t: Just fixed the man page also.

Regards,
Ikegami

sitsofe:
(Re-pinging @axboe for his thoughts on this)

sitsofe (changes requested):
I think this is dramatic enough that `FIO_SERVER_VER` will need ramping (see https://github.com/axboe/fio/blob/master/server.h#L51 ). I'm also unsure if all the tools that scoop up latency ranges will need re-testing:

- https://github.com/axboe/fio/blob/master/tools/hist/fio-histo-log-pctiles.py
- https://github.com/axboe/fio/blob/master/tools/hist/half-bins.py
- https://github.com/axboe/fio/blob/master/t/latency_percentiles.py
- https://github.com/axboe/fio/blob/master/tools/hist/fiologparser_hist.py

Just check if they work initially and report back.

ikegami-t:
> I think this is dramatic enough that `FIO_SERVER_VER` will need ramping (see https://github.com/axboe/fio/blob/master/server.h#L51 ). I'm also unsure if all the tools that scoop up latency ranges will need re-testing:

Just ramped up the version.

> * https://github.com/axboe/fio/blob/master/tools/hist/fio-histo-log-pctiles.py
> * https://github.com/axboe/fio/blob/master/tools/hist/half-bins.py
> * https://github.com/axboe/fio/blob/master/t/latency_percentiles.py
> * https://github.com/axboe/fio/blob/master/tools/hist/fiologparser_hist.py
> 
> Just check if they work initially and report back.

Also tested the tools with the changes and seems to work correctly as the same behavior with the original fio version.

sitsofe:
@ikegami-t Will this impact the terse output https://fio.readthedocs.io/en/latest/fio_man.html#terse-output ?

ikegami-t:
> @ikegami-t Will this impact the terse output https://fio.readthedocs.io/en/latest/fio_man.html#terse-output ?

Yes the terse output will be increased the number of microsecond and millisecond latency outputs since those depth levels increased.

sitsofe:
Hmm that's going to be a problem because the column headings are no longer going to match and CSV is not self describing...

ikegami-t:
> Hmm that's going to be a problem because the column headings are no longer going to match and CSV is not self describing...

Okay now I am thinking to add an option to change the log latency depth levels as not changed the terse output by default. Thank you.

ikegami-t:
> > Hmm that's going to be a problem because the column headings are no longer going to match and CSV is not self describing...
> 
> Okay now I am thinking to add an option to change the log latency depth levels as not changed the terse output by default. Thank you.

To fix the issue changed to add an option to change the lat stats depth levels so please review the changes again.

sitsofe:
@ikegami-t can you squash all these into one commit?

ikegami-t:
> @ikegami-t can you squash all these into one commit?

Yes just squashed the changes as mentioned. Thank you.

HOWTO
```diff
@@ -299,6 +299,10 @@ Command line options
 	Use the directory specified by `path` for generated state files instead
 	of the current working directory.
 
+.. option:: --latency-stats=nr
+
+	Change default latency stats number.
+
```
sitsofe, line 305: In the documentation could you say what the default number of buckets is? Could you also add a note to the CSV documentation warning that the number of buckets may vary?

sitsofe (changes requested):

stat.c
```diff
@@ -798,11 +805,25 @@ static void show_ddir_status(struct group_run_stats *rs, struct thread_stat *ts,
...
 }
 
-static bool show_lat(double *io_u_lat, int nr, const char **ranges,
-		     const char *msg, struct buf_output *out)
+static int stat_get_range(int i)
+{
+	int range = STAT_LAT_RANGE_MAX;
+
+	if (i + 1 >= stat_lat_nr)
+		range *= i + 2 - stat_lat_nr;
```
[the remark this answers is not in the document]
ikegami-t: This is for the millisecond latency ranges since it includes ranges "2000" and ">=2000".

HOWTO
ikegami-t: Updated the description for your comments. Thank you.

sitsofe: Ah I meant:
- https://github.com/axboe/fio/blob/master/HOWTO#L4048-L4054
- https://github.com/axboe/fio/blob/master/fio.1#L3751-L3768
- https://github.com/axboe/fio/blob/master/HOWTO#L4082-L4087
- https://github.com/axboe/fio/blob/master/fio.1#L3810-L3816

I think just rewrite the description to something like the following:
```
stat: make latency buckets configurable

Allow the user to set how many latency buckets there are at the finest 
granularity (when unset the default is 10). The number of larger granularity 
buckets is calculated as the finest + 2.
```

Out of interest could this cause errors because the system timer doesn't offer a granularity that low? I'm not sure what the smallest piece of time your typical server can measure is (see https://stackoverflow.com/questions/6498972/faster-equivalent-of-gettimeofday/ for some overheads reading various clocks)...

sitsofe (changes requested):

HOWTO
ikegami-t: > Ah I meant:
> 
> * https://github.com/axboe/fio/blob/master/HOWTO#L4048-L4054
> * https://github.com/axboe/fio/blob/master/fio.1#L3751-L3768
> * https://github.com/axboe/fio/blob/master/HOWTO#L4082-L4087
> * https://github.com/axboe/fio/blob/master/fio.1#L3810-L3816

Noted so updated the documentation again.

> I think just rewrite the description to something like the following:
> 
> ```
> stat: make latency buckets configurable
> 
> Allow the user to set how many latency buckets there are at the finest 
> granularity (when unset the default is 10). The number of larger granularity 
> buckets is calculated as the finest + 2.
> ```

Yes updated the description also.

> Out of interest could this cause errors because the system timer doesn't offer a granularity that low? I'm not sure what the smallest piece of time your typical server can measure is (see https://stackoverflow.com/questions/6498972/faster-equivalent-of-gettimeofday/ for some overheads reading various clocks)...

Currently the minimum latency bucket is 2 nanosecond or lower so it should be worked the changes I think.

Regards,
Ikegami

sitsofe (approved):
LGTM @axboe this one looks ready for review/pulling.

PR #1109, opened 2020-10-29
stat: fix bandwidth log maximum log entry count

pjakuszew:
When using `--bandwidth-log`, resulting aggregated logs will
contain only 1024 lines, which do not contain enough information
when fio is running for a longer period of time. This commit
changes size cap to the value of MAX_LOG_ENTRIES (instead of
DEF_LOG_ENTRIES)

### DIFF
--- a/stat.c
+++ b/stat.c
@@ -2435,7 +2435,7 @@ static struct io_logs *get_new_log(struct io_log *iolog)
 	 * forever
 	 */
 	if (!iolog->cur_log_max)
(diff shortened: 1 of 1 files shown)


pjakuszew:
This was confirmed to be working with 25m-long run time, not marking this as complete pull request until we finish a 24 hour test

pjakuszew:
This is actually not a proper fix for this problem, closing this PR

PR #1119, opened 2020-11-05
stat: Prevent the BW and IOPs logging interval from creeping up

jeffreyalien:
With the current FIO version, we're seeing the interval for writing 
the BW and IOPS logs increase by 1 msec over time.  The --log_avg_msec
option is used to set the time interval for logging BW and IOPs.  See
the following file for an example:
[int-creep-example.txt](https://github.com/axboe/fio/files/5503565/int-creep-example.txt)

This issue will cause dip ( low ) in IOPS and BW plot vs time because 
parsing/plot tool can't get correct sum total IOPS and BW from 
multiple jobs when there is 1-2us increment in timestamp.

The issue was introduced in version 3.18 with the following 
commit - 31eca64.  And these changes made it slightly better: 0f77d30,
and PR #973 (0e59dd6), but the problem still exists.

This commit also contains the change from PR #982.  I believe we need
both changes to totally resolve the issue of the BW and IOPs logging
interval creep.

Signed-off-by: Jeff Lien <jeff.lien@wdc.com>

### DIFF
--- a/stat.c
+++ b/stat.c
@@ -2747,7 +2747,8 @@ static unsigned long add_log_sample(struct thread_data *td,
 
 	__add_stat_to_log(iolog, ddir, elapsed, td->o.log_max != 0, priority_bit);
 
-	iolog->avg_last[ddir] = elapsed - (this_window - iolog->avg_msec);
+	iolog->avg_last[ddir] = elapsed - (elapsed % iolog->avg_msec);
+
 	return iolog->avg_msec;
 }
 
@@ -2985,7 +2986,7 @@ static int __add_samples(struct thread_data *td, struct timespec *parent_tv,
 	next_log = avg_time;
 
 	spent = mtime_since(parent_tv, t);
-	if (spent < avg_time && avg_time - spent >= LOG_MSEC_SLACK)
+	if (spent < avg_time && avg_time - spent > LOG_MSEC_SLACK)
 		return avg_time - spent;
 
 	if (needs_lock)
@@ -3078,13 +3079,16 @@ static int add_iops_samples(struct thread_data *td, struct timespec *t)
 int calc_log_samples(void)
 {
 	struct thread_data *td;
-	unsigned int next = ~0U, tmp;
+	unsigned int next = ~0U, tmp = 0, next_mod = 0, log_avg_msec_min = -1U;
 	struct timespec now;
 	int i;
+	long elapsed_time = 0;
 
 	fio_gettime(&now, NULL);
 
 	for_each_td(td, i) {
+		elapsed_time = mtime_since_now(&td->epoch);
+
 		if (!td->o.stats)
 			continue;
 		if (in_ramp_time(td) ||
@@ -3095,16 +3099,33 @@ int calc_log_samples(void)
 		if (!td->bw_log ||
 			(td->bw_log && !per_unit_log(td->bw_log))) {
 			tmp = add_bw_samples(td, &now);
-			if (tmp < next)
-				next = tmp;
+
+			if (td->bw_log)
+				log_avg_msec_min = min(log_avg_msec_min, (unsigned int)td->bw_log->avg_msec);
 		}
 		if (!td->iops_log ||
 			(td->iops_log && !per_unit_log(td->iops_log))) {
 			tmp = add_iops_samples(td, &now);
-			if (tmp < next)
-				next = tmp;
+
+			if (td->iops_log)
+				log_avg_msec_min = min(log_avg_msec_min, (unsigned int)td->iops_log->avg_msec);
 		}
+
+		if (tmp < next)
+			next = tmp;
 	}
+
+	/* if log_avg_msec_min has not been changed, set it to 0 */
+	if (log_avg_msec_min == -1U)
+		log_avg_msec_min = 0;
+
+	if (log_avg_msec_min == 0)
+		next_mod = elapsed_time;
+	else
+		next_mod = elapsed_time % log_avg_msec_min;
+
+	/* correction to keep the time on the log avg msec boundary */
+	next = min(next, (log_avg_msec_min - next_mod));
 
 	return next == ~0U ? 0 : next;
 }


stat.c
```diff
@@ -3087,17 +3091,25 @@ int calc_log_samples(void)
...
+
+			if (td->iops_log)
+				log_avg_msec = td->iops_log->avg_msec;
 		}
+
+		if (tmp < next)
+			next = tmp;
 	}
 
+	next_mod = log_avg_msec==0 ? elapsed_time: elapsed_time%log_avg_msec;
```
axboe, line 3110: Please rewrite this using ```if``` instead, that's much easier to read. Also remember to have spaces between operators and variables. Goes for the next line, too.

stat.c
```diff
@@ -3087,17 +3091,25 @@ int calc_log_samples(void)
...
+	next = min(next, (log_avg_msec-next_mod));  /* correction to keep the time on the log avg msec boundary */
```
sitsofe, line 3111: This line looks a bit long - maybe reposition the comment?

sitsofe:
Could you also wrap the body of the commit message at 78 characters?

stat.c
```diff
@@ -3070,13 +3071,16 @@ static int add_iops_samples(struct thread_data *td, struct timespec *t)
...
-	unsigned int next = ~0U, tmp;
+	unsigned int next = ~0U, tmp = 0, next_mod = 0, log_avg_msec = 0;
 	struct timespec now;
 	int i;
+	long elapsed_time = 0;
 
 	fio_gettime(&now, NULL);
 
 	for_each_td(td, i) {
+		elapsed_time = mtime_since_now(&td->epoch);
```
sitsofe, line 3082: This is the time since the job started...

stat.c
```diff
@@ -3087,17 +3091,29 @@ int calc_log_samples(void)
...
+			if (td->iops_log)
+				log_avg_msec = td->iops_log->avg_msec;
 		}
+
+		if (tmp < next)
+			next = tmp;
 	}
 
+	if (log_avg_msec == 0)
+		next_mod = elapsed_time;
```
sitsofe, line 3111: So this say the time that started since the job started... Won't get bigger steadily?

axboe:
> Could you also wrap the body of the commit message at 78 characters?

I like 74 chars :-)

Also, please squash the two commits into one, no point having a change + fixup patch when it could be one correct change instead.

stat.c
```diff
@@ -3087,17 +3091,29 @@ int calc_log_samples(void)
...
+	else
+		next_mod = elapsed_time % log_avg_msec;
+
+	next = min(next, (log_avg_msec - next_mod));  /* correction to keep the time on the log avg msec boundary */
```
sitsofe, line 3115: So `log_avg_msec - next_mod` is depending on wraparound to make it large in the `log_avg_msec` case? This feels a bit strange...

vincentkfu:
Hello Jeff:

I think two different features are conflated here. The issue addressed by my April patch was related to the `--status-interval` option but your patch appears to be modifying code for `--write_bw_log` and related options. Is your `stat-int-creep.txt` output from `--write-bw-log`? I would make sure to distinguish the two (related but distinct) features. If the issue the patch is addressing is related to the output of `--write-bw-log` and relatives, then the commit message and the title of this pull request should not repeatedly refer to `--status-interval`. As far as I can tell your patch does not affect the `--status-interval` option.

stat.c
jeffreyalien: ok, I'll move the comment.

jeffreyalien:
Vincent, How does this look?
   stat: Prevent the interval for BW and IOPs logging from creeping up
    
    With the current FIO version, we're seeing the interval for writing 
	the BW and IOPS increase by 1 msec over time.
    
    This issue will cause dip ( low ) in IOPS and BW plot vs time because 
	parsing/plot tool can't get correct sum total IOPS and BW from 
	multiple jobs when there is 1-2us increment in timestamp.
    
    The issue was introduced in version 3.18 with the following 
	commit - 31eca64.  And this commit made it slightly better - 0f77d30,
	but the problem still exists.
    
    This commit also contains the change from PR #982.  I believe we need
	both changes to totally resolve the issue of the BW and IOPs logging
	interval creep.

vincentkfu:
Jeff, the commit message looks much better to me. It might be helpful to mention that `--log_avg_msec` is the option that sets the logging interval.

Also, I am a little skeptical that 0f77d30 helped the problem, but I don't have the bandwidth to actually test this. I think that another patch in the pull request https://github.com/axboe/fio/pull/973 (0e59dd68fcce7a5c77ba5bf1239d79a28c4006a0) is more likely to have helped the issue.

stat.c
jeffreyalien: Yes, next_mod will get steadily bigger here, but the next calculation on line 3115 is still correct.

stat.c
jeffreyalien: next_mod is the actual time expired in the interval and (log_avg_msec - next_mod)  is the actual time left in interval.  So when next_mod is 0 (elapsed_time is an even multiple of log_avg_msec), it indicates it's time to log the IOPs and BW and start a new interval.  Since it's based on elapsed time now, it will generate more accurate next value.  Make sense?

axboe:
@jeffreyalien  Can you squash the commits, please?

jeffreyalien:
@axboe Finally got it squashed down to one commit.  Let me know if you have any other comments.  Thanks.

axboe:
Thanks! Would be great if @sitsofe and/or @vincentkfu could sign off on this.

stat.c
sitsofe: Ah I see!

stat.c
```diff
@@ -3087,17 +3091,30 @@ int calc_log_samples(void)
...
+				log_avg_msec = td->bw_log->avg_msec;
 		}
 		if (!td->iops_log ||
 			(td->iops_log && !per_unit_log(td->iops_log))) {
 			tmp = add_iops_samples(td, &now);
-			if (tmp < next)
-				next = tmp;
+
+			if (td->iops_log)
+				log_avg_msec = td->iops_log->avg_msec;
```
sitsofe, line 3103: If both are on it doesn't matter who (iops/bw_log) sets `log_avg_msec` (because they both derive their value from the same original option). However the value MAY only be in one of  `bw_log` or `iops_log` structs so we are actually forced to check both to be safe...

stat.c
```diff
@@ -3087,17 +3091,30 @@ int calc_log_samples(void)
...
 		}
+
+		if (tmp < next)
+			next = tmp;
 	}
 
+	if (log_avg_msec == 0)
```
sitsofe, line 3110: I'm have some questions here:
- Can different threads have different log_avg_msec values? If so does it mean we need the lowest one that is greater than 0?
- Could we just do 
 `max(td->o.iops_avg_time, td->o.bw_avg_time)` and check if the result were 0 to see if either one had been set within a loop iteration?

sitsofe (commented):
I have some questions. I think the idea is good though because we get clamping and we only log a sample when we're definitely within the slack period (and maybe it's better not to log anything in the final incomplete period because you never saw all of it...)

sitsofe:
Also mentioning @bvanassche because of previously mentioned views on clamping.

stat.c
```diff
@@ -2977,7 +2978,7 @@ static int __add_samples(struct thread_data *td, struct timespec *parent_tv,
 	next_log = avg_time;
 
 	spent = mtime_since(parent_tv, t);
-	if (spent < avg_time && avg_time - spent >= LOG_MSEC_SLACK)
+	if (spent < avg_time && avg_time - spent > LOG_MSEC_SLACK)
```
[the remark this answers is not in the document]
jeffreyalien: yes, that's correct.

stat.c
jeffreyalien: yes, that was my thought on this.

stat.c
jeffreyalien: I'm not sure if different threads can have different log_avg_msec values.  @vincentkfu or @bvanassche  do you know the answer to this?

stat.c
```diff
@@ -2739,7 +2739,8 @@ static unsigned long add_log_sample(struct thread_data *td,
 
 	__add_stat_to_log(iolog, ddir, elapsed, td->o.log_max != 0, priority_bit);
 
-	iolog->avg_last[ddir] = elapsed - (this_window - iolog->avg_msec);
+	iolog->avg_last[ddir] = elapsed - (elapsed % iolog->avg_msec);
```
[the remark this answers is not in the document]
jeffreyalien: yes

stat.c
axboe: It's a per-job option, so safe to assume that different threads can have different values.

jeffreyalien:
@sitsofe, @bvanassche @vincentkfu do you have any additional concerns or comments?  If not, could you signoff on this PR so it can be merged?  Thanks

stat.c
sitsofe: @jeffreyalien Doesn't `log_avg_msec` only account for the "last" thread with it set?

sitsofe:
@jeffreyalien I think my previous comment about threads having different log_avg_sec values still stands which is why I left it open...

stat.c
jeffreyalien: @sitsofe Yes, you're correct about the log_avg_msec being from the last thread.  I could add code track and use the minimum log_avg_msec  of all the threads.   I believe that should solve the potential issue of having threads with different values.

sitsofe: @jeffreyalien sounds good!

jeffreyalien:
@sitsofe Just committed a change to use the minimum log_avg_msec value now.  Please review and let me know any feedback.  Thanks

stat.c
```diff
@@ -3070,13 +3071,16 @@ static int add_iops_samples(struct thread_data *td, struct timespec *t)
 int calc_log_samples(void)
 {
 	struct thread_data *td;
-	unsigned int next = ~0U, tmp;
+	unsigned int next = ~0U, tmp = 0, next_mod = 0, log_avg_msec_min = 0xFFFFFFFF;
```
sitsofe, line 3074: Doing a quick search of the code base for the following:

```
% git grep '== ~0' | wc -l 
       1
% git grep '== -1U' | wc -l
      27
% git grep -i '== 0xFF' | wc -l 
       0
% git grep '== .*INT_MAX' | wc -l   
       5
```

It looks like there's a preference for using `-1U` as a magic maximum sentinel. Could you swap this to use that?

sitsofe (commented):
Trivial style comment otherwise LGTM!

stat.c
sitsofe: Yup!

stat.c
jeffreyalien: Changed 0xffffffff to -1U.

sitsofe (approved):
LGTM - thanks for hanging in there! @axboe could we pull this?

sitsofe:
(@jeffreyalien let's give it a few more days to hear from @vincentkfu @bvanassche and if everything remains quiet we can ping **aboxe** again)

sitsofe:
@jeffreyalien did you want to credit @pzoxiuv (`Inspired-by: Alex Merenstein <mmerenstein@>` I notice Alex has a web page on https://www.fsl.cs.stonybrook.edu/~alex/) in your commit message?

jeffreyalien:
@sitsofe I can add Inpired-by Alex, but not sure how to do that.  I didn't see an option in the commit command.  Do I just add it as part of the commit message.

sitsofe:
> Do I just add it as part of the commit message.

Yup!

jeffreyalien:
Added "Inspired-by: Alex Merenstein" to commit message.

jeffreyalien:
@axboe There have been no additional comments on this.  Can it be merged now?

axboe:
I think so, thanks for working through this!

PR #1120, opened 2020-11-06
Add cufile I/O engine

bsmith94:
Please review. I hope that the code adheres to the patterns I observed in the other I/O engines, and am sure there are things that need to be corrected.

All feedback is welcome.

As of GPUDirect Storage 0.9.0, the configure invocation is:

```
$ CFLAGS="-I/usr/local/cuda/include -I/usr/local/cuda/lib64" LDFLAGS="-L/usr/local/cuda/lib64" ./configure --enable-cufile
```

GPUDirect Storage (GDS) is obtained from here: https://developer.nvidia.com/gpudirect-storage

GDS requires a Tesla or better GPU, x86_64 CPU, Ubuntu OS 18 or 20, MOFED 5.1 and Mellanox HCA >= ConnectX-5.

----

The cufile I/O engine uses NVIDIA GPUDirect Storage (GDS) cuFile API to perform
I/O directly against GPU buffers via nvidia-fs and a GDS-supported filesystem.

'configure --enable-cufile' enables the cufile engine.

CFLAGS must specify the location of cuFile headers, e.g.
CFLAGS="-I /usr/local/gds/lib"

LDFLAGS must specify the location of cuFile libraries .e.g
LDFLAGS="-L /usr/local/gds/lib"

cuFile will eventually be part of CUDA, and will then not require
the special CFLAGS and LDFLAGS settings.

cufile engine options:

    cufile_gpu_ids : colon-separated list of GPU-indexes to use, per subjob.
       If the quantity of IDs in the list is less than the quantity of subjobs,
       the last ID in the list will be used for remaining subjobs.

    cufile_posixio : if 1, the cufile engine will instead use POSIX I/O with
       CPU buffers and perform cudaMemcpy() as necessary. This provides
       relative performance for a CUDA application the uses POSIX I/O instead
       of the cuFile API.

Note that enabling validation necessitates cudaMemcpy() to populate
the GPU buffer on a write and populate the CPU buffer on a read.

### DIFF
--- a/engines/libcufile.c
+++ b/engines/libcufile.c
@@ -1,0 +1,627 @@
+/*
+ * Copyright (c)2020 System Fabric Works, Inc. All Rights Reserved.
+ * mailto:info@systemfabricworks.com
+ *
+ * License: GPLv2, see COPYING.
+ *
+ * libcufile engine
+ *
+ * fio I/O engine using the NVIDIA cuFile API.
+ *
+ */
+
+#include <stdlib.h>
+#include <unistd.h>
+#include <errno.h>
+#include <string.h>
+#include <sys/time.h>
+#include <sys/resource.h>
+#include <cufile.h>
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <pthread.h>
+
+#include "../fio.h"
+#include "../lib/pow2.h"
+#include "../optgroup.h"
+#include "../lib/memalign.h"
+
+#define ALIGNED_4KB(v) (((v) & 0x0fff) == 0)
+
+#define LOGGED_BUFLEN_NOT_ALIGNED     0x01
+#define LOGGED_GPU_OFFSET_NOT_ALIGNED 0x02
+#define GPU_ID_SEP ":"
+
+enum {
+	IO_CUFILE    = 1,
+	IO_POSIX     = 2
+};
+
+struct libcufile_options {
+	struct thread_data *td;
+	char               *gpu_ids;       /* colon-separated list of GPU ids,
+					      one per job */
+	void               *cu_mem_ptr;    /* GPU memory */
+	void               *junk_buf;      /* buffer to simulate cudaMemcpy with
+					      posix I/O write */
+	int                 my_gpu_id;     /* GPU id to use for this job */
+	unsigned int        cuda_io;       /* Type of I/O to use with CUDA */
+	size_t              total_mem;     /* size for cu_mem_ptr and junk_buf */
+	int                 logged;        /* bitmask of log messages that have
+					      been output, prevent flood */
+};
+
+struct fio_libcufile_data {
+	CUfileDescr_t  cf_descr;
+	CUfileHandle_t cf_handle;
+};
+
+static struct fio_option options[] = {
+	{
+		.name	  = "gpu_dev_ids",
+		.lname	  = "libcufile engine gpu dev ids",
+		.type	  = FIO_OPT_STR_STORE,
+		.off1	  = offsetof(struct libcufile_options, gpu_ids),
+		.help	  = "GPU IDs, one per subjob, separated by " GPU_ID_SEP,
+		.category = FIO_OPT_C_ENGINE,
+		.group	  = FIO_OPT_G_LIBCUFILE,
+	},
+	{
+		.name	  = "cuda_io",
+		.lname	  = "libcufile cuda io",
+		.type	  = FIO_OPT_STR,
+		.off1	  = offsetof(struct libcufile_options, cuda_io),
+		.help	  = "Type of I/O to use with CUDA",
+		.def      = "cufile",
+		.posval   = {
+			    { .ival = "cufile",
+			      .oval = IO_CUFILE,
+			      .help = "libcufile nvidia-fs"
+			    },
+			    { .ival = "posix",
+			      .oval = IO_POSIX,
+			      .help = "POSIX I/O"
+			    }
+		},
+		.category = FIO_OPT_C_ENGINE,
+		.group	  = FIO_OPT_G_LIBCUFILE,
+	},
+	{
+		.name	 = NULL,
+	},
+};
+
+static int running = 0;
+static int cufile_initialized = 0;
+static pthread_mutex_t running_lock = PTHREAD_MUTEX_INITIALIZER;
+
+#define check_cudaruntimecall(fn, rc)                                               \
+	do {                                                                        \
+		cudaError_t res = fn;                                               \
+		if (res != cudaSuccess) {                                           \
+			const char *str = cudaGetErrorName(res);                    \
+			log_err("cuda runtime api call failed %s:%d : err=%d:%s\n", \
+				#fn, __LINE__, res, str);                           \
+			rc = -1;                                                    \
+		} else                                                              \
+			rc = 0;                                                     \
+	} while(0)
+
+static const char *fio_libcufile_get_cuda_error(CUfileError_t st)
+{
+	if (IS_CUFILE_ERR(st.err))
+		return cufileop_status_error(st.err);
+	return "unknown";
+}
+
+/*
+ * Assign GPU to subjob roundrobin, similar to how multiple
+ * entries in 'directory' are handled by fio.
+ */
+static int fio_libcufile_find_gpu_id(struct thread_data *td)
+{
+	struct libcufile_options *o = td->eo;
+	int gpu_id = 0;
+
+	if (o->gpu_ids != NULL) {
+		char *gpu_ids, *pos, *cur;
+		int i, id_count, gpu_idx;
+
+		for (id_count = 0, cur = o->gpu_ids; cur != NULL; id_count++) {
+			cur = strchr(cur, GPU_ID_SEP[0]);
+			if (cur != NULL)
+				cur++;
+		}
+
+		gpu_idx = td->subjob_number % id_count;
+
+		pos = gpu_ids = strdup(o->gpu_ids);
+		if (gpu_ids == NULL) {
+			log_err("strdup(gpu_ids): err=%d\n", errno);
+			return -1;
+		}
+
+		i = 0;
+		while (pos != NULL && i <= gpu_idx) {
+			i++;
+			cur = strsep(&pos, GPU_ID_SEP);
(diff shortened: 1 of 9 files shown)


engines/cufile.c
```diff
@@ -0,0 +1,586 @@
...
+		cudaError_t res = fn;										\
+		if (res != cudaSuccess) {									\
+			const char *str = cudaGetErrorName(res);				\
+			log_err("cuda runtime api call failed %s:%d : %s\n",	\
+					#fn, __LINE__, str);							\
+			rc = -1;												\
+		} else {													\
+			rc = 0;													\
+		}															\
+	} while(0)
```
sitsofe, line 95: I'll assume GitHub has done a poor job of showing how these line up :-)

engines/cufile.c
```diff
@@ -0,0 +1,586 @@
...
+	char* cur = NULL;
+	char* pos = NULL;
+	CUfileError_t status;
+	int gpu_id = 0;
+	int i;
+	int initialized;
+	int rc;
+
+	pthread_mutex_lock(&running_lock);
+	if (running == 0) {
```
sitsofe, line 120: If this is running in a separate process to other versions of this engine will that be an issue?

engines/cufile.c
```diff
@@ -0,0 +1,586 @@
...
+	running++;
+	initialized = cufile_initialized;
+	pthread_mutex_unlock(&running_lock);
+
+	if (!o->use_posixio && !initialized) {
+		return 1;
+	}
+
+	if (o->gpu_ids) {
+		for (i = 0, cur = strtok_r(o->gpu_ids, GPU_ID_SEP, &pos);
```
sitsofe, line 142: Ah string splitting in C is always nasty. However I would feel better if we were to `strsep`...

engines/cufile.c
```diff
@@ -0,0 +1,586 @@
...
+	}
+
+	return 0;
+}
+
+static inline int fio_cufile_pre_write(struct thread_data *td,
+									   struct cufile_options *o,
+									   struct io_u *io_u,
+									   size_t gpu_offset)
+{
```
sitsofe, line 163: Again I can tell if this will really line up in real life...

engines/cufile.c
```diff
@@ -0,0 +1,586 @@
...
+		}
+	} else {
+		o->junk_buf = NULL;
+	}
+
+	dprint(FD_MEM, "Alloc %zu for GPU %d\n", total_mem, o->my_gpu_id);
+	check_cudaruntimecall(cudaMalloc(&o->cu_mem_ptr, total_mem), rc);
+	if (rc != 0) {
+		goto exit_error;
+	}
```
sitsofe, line 489: House style is that single line conditional bodies without an else don't get braces.

engines/cufile.c
```diff
@@ -0,0 +1,586 @@
...
+	check_cudaruntimecall(cudaMemset(o->cu_mem_ptr, 0xab, total_mem), rc);
+	if (rc != 0) {
+		goto exit_error;
+	}
```
sitsofe, line 493: As above.

engines/cufile.c
```diff
@@ -0,0 +1,586 @@
...
+
+	if (!o->use_posixio) {
+		status = cuFileBufRegister(o->cu_mem_ptr, total_mem, 0);
+		if (status.err != CU_FILE_SUCCESS) {
+			log_err("cuFileBufRegister: %d:%s\n", status.err,
+					cuFileGetErrorString(status));
```
sitsofe, line 499: I'd line up the following line with the open bracket.

engines/cufile.c
```diff
@@ -0,0 +1,586 @@
...
+	.iomem_free          = fio_cufile_iomem_free,
+	.cleanup             = fio_cufile_cleanup,
+	.flags               = FIO_SYNCIO,
+	.options             = options,
+	.option_struct_size  = sizeof(struct cufile_options)
+};
+
+void fio_init fio_cufile_register(void)
+{
+	const char* diskless = getenv("FIO_CUFILE_DISKLESSIO");
```
sitsofe, line 576: Hmm this is different. We don't typically pass parameters via environment variable... What forced you to do so here?

sitsofe (changes requested):
- Please prefix your commit with something like `ioengine: [...]`
- Please add your Signed-off-by
- We're missing documentation for this ioengine in the HOWTO and fio.1
- We're missing an example jobfile
- Won't compilation flags depend entirely on where CUDA was installed? If it is in the usual place it would hopefully be auto-detected... 

Out of curiosity how come we need another CUDA-based engine? What's this one testing differently?

engines/cufile.c
bsmith94: github is using a different tab width than my editor. I'll change the whitespace to the line continuation backslash to use spaces. That should tame it.

engines/cufile.c
bsmith94: Sorry, I'm not sure that I understand the question. This is designed to operate in both multi-process and multi-threaded mode. If you are referring to client/server operation, there is no explicit data transfer implemented by this engine. Everything happens through the filesystem. So, I believe the answer is: "No, there should not be any issue."

engines/cufile.c
bsmith94: Will do.

engines/cufile.c
bsmith94: I will adjust my editor settings and reformat the code.

engines/cufile.c
bsmith94: OK, will do.

engines/cufile.c
bsmith94: OK, will do.

engines/cufile.c
bsmith94: OK, will do.

engines/cufile.c
bsmith94: Agreed, that is out of the ordinary. The reason for using an environment variable is because the underlying filesystem may or may not be diskless. Using an fio_option to set the flag didn't appear to work. I was also unclear on the total ramifications of setting FIO_DISKLESSIO. I see it used in engines/cpu.c and engines/rdma.c, which makes perfect sense, but also in engines/glusterfs_sync.c, and that made things murkier. My current take is that FIO_DISKLESSIO should be set if the engine *should never* access local disk.

Long story short, I don't believe that FIO_DISKLESSIO needs to be set and will remove that code.

bsmith94:
> Won't compilation flags depend entirely on where CUDA was installed? 

Yes. The existing work to support `configure --enable-cuda` also requires CFLAGS and LDFLAGS to add the CUDA directories and that doesn't appear to be documented anywhere. I added that note as a point of reference.

> Out of curiosity how come we need another CUDA-based engine? What's this one testing differently?

The existing CUDA code tests GPUDirect through use of `--iomem=cudamalloc` and `--ioengine=rdma`. That tests I/O between two GPUs via an RDMA interface.

The cufile I/O engine tests performance of nvidia-fs I/O with GPU buffers, RDMA and a GPUDirectStorage-supported filesystem (e.g. ext4 on nvme, nfsordma, lustre). This is an entirely different use case than what is implemented by  engines/rdma.c. 

`--iomem=cudamalloc` is not used for cufile because we still need the malloc'd io buffers in order to implement verify. cufile does its own allocation of GPU buffers, with the same capacity of what is malloc'd for io buffers.

The option `--cufile_posixio` allows comparison of I/O between "CUDA using nvidia-fs" and "CUDA using POSIX I/O". The default behavior of `--cufile_posixio=0` uses nvidia-fs to do I/O directly with the GPU buffers--the data is never copied between the GPU and RAM (unless verify is enabled). In the POSIX I/O case, I/O is performed in a RAM buffer and that has to be synced up with the GPU buffer via cudaMemcpy.

sitsofe:
@bsmith94 Could you squash the new changes into the original commit and change the original's title/body?

bsmith94:
@sitsofe I did not realize that mentioning the pull request number in my commit to a different branch would auto-update the pull request. Sorry about that. I still have a few things to address, and the plan is to squash everything into a single commit.

engines/libcufile.c
```diff
@@ -0,0 +1,624 @@
...
+	pthread_mutex_unlock(&running_lock);
+
+	if (o->cuda_io == IO_CUFILE && !initialized)
+		return 1;
+
+	o->my_gpu_id = fio_libcufile_find_gpu_id(td);
+	if (o->my_gpu_id < 0)
+		return 1;
+
+	dprint(FD_MEM, "Subjob %d use GPU %d\n", td->subjob_number, o->my_gpu_id);
```
sitsofe, line 190: use -> uses

examples/libcufile-cufile.fio
```diff
@@ -0,0 +1,43 @@
...
+rw=randwrite
+
+[verify]
+rw=write
+verify=md5
+
+[randverify]
+rw=randwrite
+verify=md5
+
```
sitsofe, line 43: Maybe trim the trailing newline

examples/libcufile-posix.fio
```diff
@@ -0,0 +1,42 @@
...
+rw=randwrite
+
+[verify]
+rw=write
+verify=md5
+
+[randverify]
+rw=randwrite
+verify=md5
+
```
sitsofe, line 42: Maybe trim the trailing newline

sitsofe (approved):
Some trivial comments but LGTM! @axboe this look pullable...

engines/libcufile.c
bsmith94: ok

examples/libcufile-cufile.fio
bsmith94: ok. I think the sections should be stonewalled. Thoughts?

examples/libcufile-posix.fio
bsmith94: ok

bsmith94: Should I squash the changes into one commit or push a new commit?

bsmith94: I squashed.

bsmith94:
Thanks for helping me get this pulled.

Issue #1113, opened 2020-10-30
Aggregated logs generated by --bandwidth-log option are capped at 1024 lines


pjakuszew:
Hello,

I have discovered that `--bandwidth-log` command line option produces incomplete logs on longer test run times. This was discovered on fio-3.23 built from source on Ubuntu 18.04.4.

Used command:
```
./fio-3.23/fio --filename=/dev/nvme1n1p1 --direct=1 --rw=randrw --refill_buffers --norandommap --randrepeat=0 --ioengine=libaio --bs=4k --rwmixread=50 --iodepth=16 --numjobs=16 --runtime=24h --time_based --group_reporting --name=4kxxxtest --output=output.log --bandwidth-log
```

I have tried to determine the cause of bug but was unable to find it, so far.

I suspect that `io_logs` struct containing the aggregated logs does not have its entry count incremented properly, which makes the following statement to fail and the resulting `aggr-*_bw.log` files are capped at 1024 lines (current value of `DEF_LOG_ENTRIES` defined in iolog.h):
https://github.com/axboe/fio/blob/fio-3.23/stat.c#L2431-L2441
```
	/*
	 * Cap the size at MAX_LOG_ENTRIES, so we don't keep doubling
	 * forever
	 */
	if (!iolog->cur_log_max)
		new_samples = DEF_LOG_ENTRIES;
	else {
		new_samples = iolog->cur_log_max * 2;
		if (new_samples > MAX_LOG_ENTRIES)
			new_samples = MAX_LOG_ENTRIES;
	}
```

This issue does not happen when `write_bw_log` option is used; per-job bandwidth logs contain full metrics. I tried to compare the sample append differences between `write_bw_log` and `bandwidth-log` options, but frankly I'm not that familiar with C to understand it.

I have temporarily mitigated this in my test case by changing `DEF_LOG_ENTRIES` to `MAX_LOG_ENTRIES` at https://github.com/axboe/fio/blob/fio-3.23/stat.c#L2436 but I would appreciate any tips about the proper way of fixing this issue.

pjakuszew:
I have added fio output and bandwidth logs in attachments, they were produced from following command on my computer:
```
[...]

PR #1136, opened 2020-11-27
Grow bandwidth log stats to MAX_LOG_ENTRIES

pmoust:
This commit enables agg logs to grow to MAX_LOG_ENTRIES, like any thread logs.

Fixes: https://github.com/axboe/fio/issues/1113
Co-authored-by: Sitsofe Wheeler <sitsofe@yahoo.com>
Tested-by: Panagiotis Moustafellos <pmoust@elastic.co>
Signed-off-by: Panagiotis Moustafellos <pmoust@elastic.co>

### DIFF
--- a/eta.c
+++ b/eta.c
@@ -507,6 +507,7 @@ bool calc_thread_status(struct jobs_eta *je, int force)
 		calc_rate(unified_rw_rep, rate_time, io_bytes, rate_io_bytes,
 				je->rate);
 		memcpy(&rate_prev_time, &now, sizeof(now));
+		regrow_agg_logs();
 		for_each_rw_ddir(ddir) {
 			add_agg_sample(sample_val(je->rate[ddir]), ddir, 0, 0);
 		}
--- a/stat.c
+++ b/stat.c
@@ -2536,6 +2536,14 @@ void regrow_logs(struct thread_data *td)
 	td->flags &= ~TD_F_REGROW_LOGS;
 }
 
+void regrow_agg_logs(void)
+{
+	enum fio_ddir ddir;
+
+	for (ddir = 0; ddir < DDIR_RWDIR_CNT; ddir++)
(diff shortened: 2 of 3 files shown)


stat.c
```diff
@@ -2536,6 +2536,14 @@ void regrow_logs(struct thread_data *td)
 	td->flags &= ~TD_F_REGROW_LOGS;
 }
 
+void regrow_agg_logs(void)
+{
+	enum fio_ddir ddir;
+
+	for (ddir = 0; ddir < DDIR_RWDIR_CNT; ddir++)
+	regrow_log(agg_io_log[ddir]);
```
sitsofe, line 2544: I think this would need to be indented?

eta.c
```diff
@@ -508,6 +508,7 @@ bool calc_thread_status(struct jobs_eta *je, int force)
 				je->rate);
 		memcpy(&rate_prev_time, &now, sizeof(now));
 		for_each_rw_ddir(ddir) {
+			regrow_agg_logs();
```
sitsofe, line 511: I'm ashamed to say I clearly wasn't thinking here when I wrote the original patch :-(

I'd move `regrow_agg_logs()` above the for because that function loops over all the directions anyway...

sitsofe (changes requested):

sitsofe:
@pmoust Could you also prefix the commit title with an area like
```
stat: allow bandwidth log stats to [...]
```

(and thanks for the shout out :-) )

eta.c
pmoust: I actually though about popping it above, agreed - fixed in 76204de

pmoust:
Addressed comments in 76204de

sitsofe (approved):
LGTM! @axboe can this one be pulled?

sitsofe:
Thanks @pmoust !

PR #1140, opened 2020-12-02
engine/rbd: tweaks for properly handling direct mode and writearound/back caching

dillaman:


sitsofe:
Don't you want to flush the cache before you even do I/O when you're going direct so that you minimise cache operations (i.e. discarding it because you're writing through it but there was something unwritten there first)?

dillaman:
> Don't you want to flush the cache before you even do I/O when you're going direct so that you minimise cache operations (i.e. discarding it because you're writing through it but there was something unwritten there first)?

The image was just opened in the previous lines so there is no pending IO. We use a similar technique in the "rbd bench" CLI tool. I just happened to notice that `fio` was 5x slower than `rbd bench` so I wanted to address it since most folks use `fio` to test Ceph clusters.

engines/rbd.c
```diff
@@ -227,12 +227,30 @@ static int _fio_rbd_connect(struct thread_data *td)
...
 		log_err("rbd_open failed.\n");
 		goto failed_open;
 	}
 
+	if (!td->o.odirect) {
+		/*
+		 * ensure cache enables writeback/around mode unless explicitly
+		 * configured for writethrough mode
+		 */
+		r = rbd_flush(rbd->image);
```
sitsofe, line 248: Should this just call fio_rbd_invalidate to be safe?

sitsofe:
> I just happened to notice that fio was 5x slower than rbd bench so I wanted to address it since most folks use fio to test Ceph clusters.

Ah got it! One minor comment because I was going through matching it up to the existing cache invalidate logic.

engines/rbd.c
dillaman: Hmm, I am probably not understanding the concern but it might be just due to my lack of understanding the `fio` internals. At this point, since `librbd` has just opened the image, there wouldn't be anything to invalidate since we don't have the equivalent of a global kernel pagecache-like buffer shared between images.

sitsofe: Ah sorry I mixed up flush with invalidate.

sitsofe (approved):
LGTM! @axboe would you pull these ones?

PR #1150, opened 2020-12-20
num2str: fix precision loss bug when the fractional part is close to 1

### LINKED ISSUE #1117: num2str precision loss bug
https://github.com/axboe/fio/blob/08a3f6fef1fe6c9fcd18d5ed40ca81097922bb14/lib/num2str.c#L112-L113

when `(double) modulo  / (double) thousand` is evaluated to near 1, say `0.965`, and the precision specifier is 1, it will be formatted as `1.0`, which results in precision loss  here

https://github.com/axboe/fio/blob/08a3f6fef1fe6c9fcd18d5ed40ca81097922bb14/lib/num2str.c#L115-L116

gloit042:
Fix #1117

example:
	The result of num2str(11999999999999, 4, 1, 0, N2S_NONE) should be "12.0G", but current result is "11.0G". 

Signed-off-by: Jiahao Li <gloit042@gmail.com>

### DIFF
--- a/lib/num2str.c
+++ b/lib/num2str.c
@@ -112,6 +112,9 @@ done:
 	sprintf(tmp, "%.*f", (int)(maxlen - strlen(tmp) - 1),
 		(double)modulo / (double)thousand);
 
+	if (tmp[0] == '1')
+		num++;
+
 	if (asprintf(&buf, "%llu.%s%s%s", (unsigned long long) num, &tmp[2],
 		     unitprefix[post_index], unitstr[units]) < 0)
 		buf = NULL;


sitsofe:
@gloit042 Could you include an example call that your patches fixes in the commit message?

sitsofe:
@kusumi do you have any thoughts on this patch?

sitsofe:
(@gloit042 thanks for the motivating example)

lib/num2str.c
```diff
@@ -112,6 +112,9 @@ char *num2str(uint64_t num, int maxlen, int base, int pow2, enum n2s_unit units)
 	sprintf(tmp, "%.*f", (int)(maxlen - strlen(tmp) - 1),
 		(double)modulo / (double)thousand);
 
+	if (tmp[0] == '1')
+		num++;
```
sitsofe, line 116: OK so if the modulo / thousand has been rounded up to 1 when printing it as a float then we assume we have to do a carry. Could you add a test to `unittests/lib/num2str.c` to ensure this behaviour is preserved?

sitsofe (approved):
Minor comment about a unittest otherwise LGTM!

PR #1154, opened 2020-12-23
Fix compilation on macOS

sitsofe:
- Use a different method to avoid weak linking
- Fix the inability to compile on old OSX 10.11 by adding a shim for clock_gettime().

It's a bit hacky to shim clock_gettime() but it looks like fio has grown a dependence on it being available (all the CI platforms fio builds for have it so it's understandable).

@axboe, @bvanassche any thoughts?

### DIFF
--- a/configure
+++ b/configure
@@ -45,6 +45,7 @@ print_config() {
 
 # Default CFLAGS
 CFLAGS="-D_GNU_SOURCE -include config-host.h $CFLAGS"
+CONFIGURE_CFLAGS="-Werror-implicit-function-declaration"
 BUILD_CFLAGS=""
 
 # Print a helpful header at the top of config.log
@@ -88,14 +89,14 @@ do_cc() {
 }
 
 compile_object() {
-  do_cc $CFLAGS -Werror-implicit-function-declaration -c -o $TMPO $TMPC
+  do_cc $CFLAGS $CONFIGURE_CFLAGS -c -o $TMPO $TMPC
 }
 
 compile_prog() {
   local_cflags="$1"
   local_ldflags="$2 $LIBS"
   echo "Compiling test case $3" >> config.log
-  do_cc $CFLAGS -Werror-implicit-function-declaration $local_cflags -o $TMPE $TMPC $LDFLAGS $local_ldflags
+  do_cc $CFLAGS $CONFIGURE_CFLAGS $local_cflags -o $TMPE $TMPC $LDFLAGS $local_ldflags
 }
 
 feature_not_found() {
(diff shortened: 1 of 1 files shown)


bvanassche:
Disabling weak symbols seems weird to me since weak symbols are important in Unix environments. So I think it's an improvement that weak symbols are no longer disabled although I'm not familiar with `-Werror=partial-availability`.

Regarding the second patch: how about dropping support for platforms that do not provide clock_getttime()? clock_gettime() was introduced more than 20 years ago so there is no good excuse for POSIX platforms for not providing clock_gettime() today. See also https://pubs.opengroup.org/onlinepubs/7908799/xsh/clock_settime.html.

If it would be desirable to provide a clock_gettime() backport for older OS/X versions, please take a look at https://stackoverflow.com/questions/5167269/clock-gettime-alternative-in-mac-os-x.

sitsofe:
The weak symbol thing is better explained over in https://github.com/axboe/fio/commit/ccf2d89d39b21bc8c7b497b40be5b82eadb80863 . Arguably it could be refined as they only need to be avoided at configure time.

With respect to dropping support for platforms without clock_getttime() - you have a good point (but it goes a tiny bit further as it looks like CLOCK_MONOTONIC and CLOCK_REALTIME must also be available (I think the last big platform without CLOCK_MONOTONIC was RHEL3 and fio doesn't even easily build on RHEL7). A lot of the older platforms (HPUX, AIX) are already unsupportable due to a lack of GCC 4.9/clang 3.6 and so long as recent Linux/BSDs/macOS/Windows (and maybe Minix ;-) are alive I'm happy enough.

Re a proper `clock_gettime()` on old macOS - I saw that stackoverflow but it's unclear how threadsafe those shims are and whether making them thread safe also makes them slow (https://github.com/joyent/libuv/commit/211bf4ec37a73dea06d15233ffa747e9387b8947 shows what libuv did).

@axboe would you be happy with change that mandate `clock_getttime()` and drop some of the conditionals?

sitsofe:
@bvanassche I've removed the contentious patch from the branch for now and I've changed the original patch to be limited to configure time so we can at least get fio building with Xcode 11.4+ on modern macOS in the meantime.

bvanassche:
@sitsofe So a detection mechanism for clock_gettime() is provided but no backport? Can you take a look at https://github.com/bvanassche/fio/tree/osx-clock-gettime? There is a patch on that branch that provides a backport of clock_gettime() to OS/X versions that do not provide that function.

sitsofe:
@axboe could you pull this one?

Issue #1005, opened 2020-06-02
Floating point precision issue with percentile calculations


vincentkfu:
There have been two ~~32-bit~~ Windows latency test failures where there is an inconsistency between latency percentiles reported by fio and corresponding percentiles calculated from latency logs:

1. https://ci.appveyor.com/project/vincentkfu/fio/builds/33008676/job/ywsyeaymrb3htht2
2. https://ci.appveyor.com/project/axboe/fio/builds/33178585/job/qveg19ae3gmacb1u

Artifacts are available for failure 2.  The error was:

```
slat percentiles match
clat percentiles match
lat percentiles match
slat percentiles match
Error with clat 99.950000th percentile: fio: 3063808, expected: 3006600, proportional delta: 0.019027
Rank: 43978, index: 43977
Error with lat 99.950000th percentile: fio: 3325952, expected: 3125300, proportional delta: 0.064202
Rank: 43978, index: 43977
**********Test 10 FAILED**********
```

The problem occurred with the write workload. fio reported a 99.95th percentile clat of 3063808 but the test script determined that the value should have been 3006600 based on the clat log. The values do not match exactly because fio uses latency bins to produce the percentiles, but the delta was larger than the expected 0.78125% maximum. Investigation reveals that this is just a floating point precision issue and may be just something we have to live with.

There are 44000 write IOs and stat.c evaluates the following condition:

```(unsigned long long) 43978 >= (long double) 99.95 / 100 * 44000```

This is false because the expression on the right-hand-side evaluates to 43978.00000000000125055521. So fio uses the next higher latency for the 99.95th percentile instead of the one identified by the python script.

The right-hand side really should evaluate to 43978:
```
$ echo "44000*99.95/100" | bc
43978
```
The total number of write IOs was counted as 44000 by:
```
$ awk '$3=="1,"{print $2}' < latency010_clat.1.log | cut -d, -f 1 | wc
  44000   44000  262771
```
The 43978th and 43979th values are:
```
$ awk '$3=="1,"{print $2}' < latency010_clat.1.log | cut -d, -f 1 | sort -n | head -n 43979 | tail -n 2
3006600
3071900
```

We can reconstruct the calculations for the first AppVeyor failure and also confirm that this was a floating point precision issue.

Here is some code showing the results of the floating point calculations: https://onlinegdb.com/SJ4U8B43I

I'm documenting the issue here so that we can consider this when future test failures occur. I don't know of any way to resolve this problem. So I will close this issue as well.

[...]

PR #1159, opened 2020-12-29
Extend flexibility of non-uniform random distributions

aclamk:
This change affects options random_distribution and file_service_type.
For pareto, zipf and gauss distribution a contept of `center` is implemented.
It allows to fix in place a value that is most probable to access.
Example:
fio --randseed=1 --ioengine=libaio --rw=randwrite --nrfiles=16 --bs=4k \
    --size=256m --allow_file_create=1 --write_iolog=log.txt \
     --file_service_type=gauss:10:0.1 --filename_format=object.\$filenum --name=x
    
cat log.txt |grep write |cut -f 1 -d " " |sort |uniq -c | sort -n | \
sed "s/[.]/ /" | while read a b c; do echo $c $b $a; done |sort -n
0 object 13429
1 object 17928
2 object 14724
3 object 7845
4 object 2476
5 object 468
6 object 44
7 object 3
12 object 24
13 object 318
14 object 1795
15 object 6482

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1361,7 +1361,7 @@ I/O type
 	limit reads or writes to a certain rate.  If that is the case, then the
 	distribution may be skewed. Default: 50.
 
-.. option:: random_distribution=str:float[,str:float][,str:float]
+.. option:: random_distribution=str:float[:float][,str:float][,str:float]
 
 	By default, fio will use a completely uniform random distribution when asked
 	to perform random I/O. Sometimes it is useful to skew the distribution in
@@ -1395,6 +1395,14 @@ I/O type
 	option. If a non-uniform model is used, fio will disable use of the random
 	map. For the **normal** distribution, a normal (Gaussian) deviation is
 	supplied as a value between 0 and 100.
+
+	The second, optional float is allowed for **pareto**, **zipf** and **normal** distributions.
+	It allows to set base of distribution in non-default place, giving more control
+	over most probable outcome. This value is in range [0-1] which maps linearly to
+	range of possible random values.
+	Defaults are: random for **pareto** and **zipf**, and 0.5 for **normal**.
+	If you wanted to use **zipf** with a `theta` of 1.2 centered on 1/4 of allowed value range,
+	you would use ``random_distibution=zipf:1.2:0.25``.
 
 	For a **zoned** distribution, fio supports specifying percentages of I/O
 	access that should fall within what range of the file or device. For
--- a/fio.1
+++ b/fio.1
@@ -1132,7 +1132,7 @@ first. This may interfere with a given rate setting, if fio is asked to
 limit reads or writes to a certain rate. If that is the case, then the
 distribution may be skewed. Default: 50.
 .TP
-.BI random_distribution \fR=\fPstr:float[,str:float][,str:float]
+.BI random_distribution \fR=\fPstr:float[:float][,str:float][,str:float]
 By default, fio will use a completely uniform random distribution when asked
 to perform random I/O. Sometimes it is useful to skew the distribution in
 specific ways, ensuring that some parts of the data is more hot than others.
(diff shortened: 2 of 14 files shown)


aclamk:
This supersedes #1155.

sitsofe (changes requested):

HOWTO
```diff
@@ -1396,6 +1396,14 @@ I/O type
...
 	supplied as a value between 0 and 100.
 
+	The second, optional float is allowed for **pareto**, **zipf** and **normal** distributions.
+	It allows to set base of distribution in non-default place, giving more control
+	over most probable outcome. This value is in range [0-1] which maps linearly to
+	range of possible random values.
+	Defaults are: random for **pareto** and **zipf**, and 0.5 for **normal**"
+	If you wanted to use **zipf** with a `theta` of 1.2 centered on 1/4 of allowed value range,
+	you would use ``random_distibution=zipf:1.2:0.25``.
+
```
sitsofe, line 1406: fio.1 also needs (manual :-( ) updating too.

aclamk:
@sitsofe Added fixes. Meld them into previous commit? Maybe meld all together?

aclamk:
@sitsofe Do you have idea, what is intended use of multiple distributions that can be provided with --random_distribution ?

Hint in manual is "random_distribution=str:float[,str:float][,str:float]", but code does not care for it. Does it?
Was it intended to behave like "--directory" option, shuffling different value for different job?

sitsofe:
@aclamk Each [extra comma separated value influences the remaining "directions"](https://fio.readthedocs.io/en/latest/fio_man.html#cmdoption-arg-blocksize):
- The first alone: read, writes and trims are all use the same distribution
- The first, second: reads use the first, writes/trims use the second
- Reads get one distributions, writes another and trims a third

From the [documentation for random_distribution](https://fio.readthedocs.io/en/latest/fio_man.html#cmdoption-arg-random-distribution):

> [...] Like bssplit, it’s possible to specify separate zones for reads, writes, and trims. If just one set is given, it’ll apply to all of them.

Does it really not behave that way?

sitsofe:
(Travis CI failure looks spurious. See #1005 for why)

aclamk:
@sitsofe 
Thanks! I could not see it.
I guess that extra [,str:float] only apply to zones. I see code for it in zones, I do not see code for it in other distributions.

I will update documentation.

And as a side note, I would love separate distributions for read and writes :D

options.c
```diff
@@ -44,6 +44,21 @@ static char *get_opt_postfix(const char *str)
 	return strdup(p);
 }
 
+static bool split_parse_distr(const char *str, double *val, double *center)
+{
+	char *p = strdup(str);
```
sitsofe, line 49: What if p is NULL?

sitsofe (approved):
LGTM! The only minor point is whether people will understand what moving the base means but time will tell.

sitsofe:
@aclamk could you squash all these commits together?

aclamk (commented):
> @aclamk could you squash all these commits together?

Done.

sitsofe (changes requested):
Strangely this seems to fail to build with a compilation error - https://travis-ci.org/github/axboe/fio/jobs/753505992 . Can you move variable definitions to the top of the function?

aclamk:
@sitsofe 
This is my bad. I forgot that legacy C does not like mixing declaration and code.
Fixed.

fio.1
```diff
@@ -1168,6 +1168,14 @@ option. If a non\-uniform model is used, fio will disable use of the random
 map. For the \fBnormal\fR distribution, a normal (Gaussian) deviation is
 supplied as a value between 0 and 100.
 .P
+The second, optional float is allowed for \fBpareto\fR, \fBzipf\fR and \fBnormal\fR
+distributions. It allows to set base of distribution in non-default place, giving
+more control over most probable outcome. This value is in range [0-1] which maps linearly to
+range of possible random values.
+Defaults are: random for \fBpareto\fR and \fBzipf\fR, and 0.5 for \fBnormal\fR.
+If you wanted to use \fBzipf\fB with a `theta` of 1.2 centered on 1/4 of allowed value range,
```
sitsofe, line 1176: Second `\fB` should be `\fR` ?

io_u.c
```diff
@@ -1326,8 +1326,10 @@ static struct fio_file *__get_next_file(struct thread_data *td)
 	if (f && fio_file_open(f) && !fio_file_closing(f)) {
 		if (td->o.file_service_type == FIO_FSERVICE_SEQ)
 			goto out;
-		if (td->file_service_left--)
-			goto out;
+		if (td->file_service_left) {
+		  td->file_service_left--;
+		  goto out;
+		}
```
sitsofe, line 1332: Is this fixing a different bug? If it's a patch already submitted maybe you can rebase to check if it's already in place?

sitsofe (changes requested):

io_u.c
aclamk: It was bug I found during work on this. Has already been merged.

fio.1
aclamk: Fixed!

sitsofe (approved):
LGTM! @axboe could you consider this one for pulling?

sitsofe:
Thanks @aclamk!

PR #1160, opened 2021-01-05
Fuzzing

catenacyber:
This Pull Request adds a fuzz target, and fixes one bug uncovered by it.
This fuzz target looks for bugs in `./fio --parse-only file`

If this gets merged, it will allow oss-fuzz integration with https://github.com/google/oss-fuzz/pull/4916

What do you think of it ?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -346,6 +346,23 @@ T_MEMLOCK_PROGS = t/memlock
 T_TT_OBJS = t/time-test.o
 T_TT_PROGS = t/time-test
 
+T_FUZZ_OBJS = t/fuzz/fuzz_parseini.o
+T_FUZZ_OBJS += $(OBJS)
+ifdef CONFIG_ARITHMETIC
+T_FUZZ_OBJS += lex.yy.o y.tab.o
+endif
+# in case there is no fuzz driver defined by environment variable LIB_FUZZING_ENGINE, use a simple one
+# For instance, with compiler clang, address sanitizer and libFuzzer as a fuzzing engine, you should define
+# export CFLAGS="-fsanitize=address,fuzzer-no-link"
+# export LIB_FUZZING_ENGINE="-fsanitize=address"
+# export CC=clang
+# before running configure && make
+# You can adapt this with different compilers, sanitizers, and fuzzing engines
+ifndef LIB_FUZZING_ENGINE
+T_FUZZ_OBJS += t/fuzz/onefile.o
+endif
+T_FUZZ_PROGS = t/fuzz/fuzz_parseini
+
 T_OBJS = $(T_SMALLOC_OBJS)
 T_OBJS += $(T_IEEE_OBJS)
 T_OBJS += $(T_ZIPF_OBJS)
@@ -359,6 +376,7 @@ T_OBJS += $(T_PIPE_ASYNC_OBJS)
 T_OBJS += $(T_MEMLOCK_OBJS)
 T_OBJS += $(T_TT_OBJS)
 T_OBJS += $(T_IOU_RING_OBJS)
+T_OBJS += $(T_FUZZ_OBJS)
 
 ifneq (,$(findstring CYGWIN,$(CONFIG_TARGET_OS)))
     T_DEDUPE_OBJS += $(WINDOWS_OBJS)
@@ -382,6 +400,7 @@ endif
 ifneq (,$(findstring Linux,$(CONFIG_TARGET_OS)))
 T_TEST_PROGS += $(T_IOU_RING_PROGS)
 endif
+T_TEST_PROGS += $(T_FUZZ_PROGS)
 
 PROGS += $(T_PROGS)
 
@@ -532,6 +551,13 @@ t/ieee754: $(T_IEEE_OBJS)
 
 fio: $(FIO_OBJS)
 	$(QUIET_LINK)$(CC) $(LDFLAGS) -o $@ $(FIO_OBJS) $(LIBS) $(HDFSLIB)
+
+t/fuzz/fuzz_parseini: $(T_FUZZ_OBJS)
+ifndef LIB_FUZZING_ENGINE
+	$(QUIET_LINK)$(CC) $(LDFLAGS) -o $@ $(T_FUZZ_OBJS) $(LIBS) $(HDFSLIB)
+else
+	$(QUIET_LINK)$(CXX) $(LDFLAGS) -o $@ $(T_FUZZ_OBJS) $(LIB_FUZZING_ENGINE) $(LIBS) $(HDFSLIB)
+endif
 
 gfio: $(GFIO_OBJS)
 	$(QUIET_LINK)$(CC) $(filter-out -static, $(LDFLAGS)) -o gfio $(GFIO_OBJS) $(LIBS) $(GFIO_LIBS) $(GTK_LDFLAGS) $(HDFSLIB)
--- a/init.c
+++ b/init.c
@@ -327,6 +327,7 @@ void free_threads_shm(void)
 
 static void free_shm(void)
 {
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
 	if (nr_segments) {
 		flow_exit();
 		fio_debug_jobp = NULL;
@@ -343,6 +344,7 @@ static void free_shm(void)
 	fio_filelock_exit();
 	file_hash_exit();
 	scleanup();
+#endif
 }
 
 static int add_thread_segment(void)
--- a/options.c
+++ b/options.c
@@ -5064,7 +5064,7 @@ static char *fio_keyword_replace(char *opt)
 		struct fio_keyword *kw = &fio_keywords[i];
 
 		while ((s = strstr(opt, kw->word)) != NULL) {
-			char *new = malloc(strlen(opt) + 1);
+			char *new = calloc(strlen(opt) + 1, 1);
 			char *o_org = opt;
 			int olen = s - opt;
 			int len;
@@ -5081,6 +5081,7 @@ static char *fio_keyword_replace(char *opt)
 			 * in too
 			 */
 			opt += strlen(kw->word) + olen;
+			/* keeps final zero thanks to calloc */
 			if (strlen(opt))
 				memcpy(new + olen + len, opt, opt - o_org - 1);
 
--- a/t/fuzz/fuzz_parseini.c
+++ b/t/fuzz/fuzz_parseini.c
@@ -1,0 +1,41 @@
+#include "fio.h"
+
+static int initialized = 0;
+
+const char *const fakeargv[] = {(char *) "fuzz",
+	(char *) "--output", (char *) "/dev/null",
+	(char *) "--parse-only",
+	0};
+
+int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
+{
+	char *fuzzedini;
+
+	if (size < 2)
+		return 0;
+
+	if (initialized == 0) {
+		if (fio_init_options()) {
+			printf("Failed fio_init_options\n");
+			return 1;
+		}
+
+		parse_cmd_line(4, (char **) fakeargv, 0);
+		sinit();
+
+		initialized = 1;
+	}
+	fuzzedini = malloc(size);
+	if (!fuzzedini) {
+		printf("Failed malloc\n");
+		return 1;
+	}
+	/* final character is type for parse_jobs_ini */
+	memcpy(fuzzedini, data, size - 1);
+	/* ensures final 0 */
+	fuzzedini[size - 1] = 0;
+
+	parse_jobs_ini(fuzzedini, 1, 0, data[size - 1]);
+	free(fuzzedini);
+	return 0;
+}
--- a/t/fuzz/onefile.c
+++ b/t/fuzz/onefile.c
@@ -1,0 +1,51 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <stdio.h>
+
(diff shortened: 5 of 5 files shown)


t/fuzz/fuzz_parseini.c
```diff
@@ -0,0 +1,37 @@
...
+        initialized = 1;
+    }
+    char * fuzzedini = malloc(size);
+    if (!fuzzedini) {
+        printf("Failed malloc\n");
+        return 1;
+    }
+    // final character is type for parse_jobs_ini
+    memcpy(fuzzedini, data, size - 1);
+    // ensures final 0
```
sitsofe, line 31: Could we have C style comments?

t/fuzz/fuzz_parseini.c
```diff
@@ -0,0 +1,37 @@
+#include "fio.h"
+
+static int initialized = 0;
+
+int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+    if (size < 2) {
```
sitsofe, line 6: We prefer single line condition bodies not to have braces

t/fuzz/fuzz_parseini.c
```diff
@@ -0,0 +1,37 @@
+#include "fio.h"
+
+static int initialized = 0;
+
+int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
```
sitsofe, line 5: Can you move function braces on to the line below function (see fio.c as an example).

init.c
```diff
@@ -325,6 +325,7 @@ void free_threads_shm(void)
 	cur_segment = 0;
 }
 
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
sitsofe, line 328: Maybe move the ifndef into the body so you don't need it in two places?

sitsofe (changes requested):
Looks useful!

sitsofe:
Can you also:
- Add your Signed-off-by to the commit messages
- Add an area prefix to the your commit's title (see other commits as an example)

Thanks!

sitsofe:
(You'll have to fix the compilation errors)

t/fuzz/fuzz_parseini.c
catenacyber: ok

t/fuzz/fuzz_parseini.c
catenacyber: ok

t/fuzz/fuzz_parseini.c
catenacyber: ok

init.c
catenacyber: Good idea

catenacyber:
> Can you also:
> Add your Signed-off-by to the commit messages
> Add an area prefix to the your commit's title (see other commits as an example)

Done.

I will check CI now

t/fuzz/fuzz_parseini.c
```diff
@@ -0,0 +1,38 @@
...
+    if (size < 2)
+        return 0;
+
+    if (initialized == 0) {
+        if (fio_init_options()) {
+            printf("Failed fio_init_options\n");
+            return 1;
+        }
+
+        char *fakeargv[] = {(char *) "fuzz",
```
axboe, line 16: Variable declarations go at the top. And this should follow the general coding style, tabs are tabs, pointer '*' goes with the variable name, etc.

catenacyber: > Variable declarations go at the top.

Top of the block or the function ?
Changed it to top of the block.

> tabs are tabs, pointer '*' goes with the variable name

This should be good

axboe: Top of the block is fine.

catenacyber:
I do not understand appveyor failure... Do you ?

sitsofe:
> I do not understand appveyor failure... Do you ?

Maybe the C++ compiler is incompatible with the installed zlib? I'm not sure either I'm afraid.

catenacyber:
> Maybe the C++ compiler is incompatible with the installed zlib

Thanks, I was looking for a difference and forgot about CXX

We only need CXX instead of regular C compiler when the fuzz driver is C++ itself (for instance libFuzzer)

options.c
```diff
@@ -5064,7 +5064,7 @@ static char *fio_keyword_replace(char *opt)
 		struct fio_keyword *kw = &fio_keywords[i];
```
sitsofe, line 5064: Maybe the area in the git commit title should be `options: ` ?

sitsofe (commented):
Could you include some hints as to what's needed to build the fuzzer? Imagine someone only had GCC installed - what would they need to have and then what would they do to get things going?

t/fuzz/fuzz_parseini.c
```diff
@@ -0,0 +1,41 @@
+#include "fio.h"
+
+static int initialized = 0;
+
+const char *const fakeargv[] = {(char *) "fuzz",
+	(char *) "--output", (char*) "/dev/null",
```
sitsofe, line 6: (char *)?

t/fuzz/onefile.c
```diff
@@ -0,0 +1,51 @@
...
+		fclose(fp);
+		return 2;
+	}
+	if (fread(data, size, 1, fp) != 1) {
+		fclose(fp);
+		free(data);
+		return 2;
+	}
+
+	//lauch fuzzer
```
sitsofe, line 45: C-style comments please (and there's a minor typo)

sitsofe (commented):
Could you include some hints as to what's needed to build the fuzzer? Imagine someone only had GCC installed - what would they need to have and then what would they do to get things going?

catenacyber:
> Could you include some hints as to what's needed to build the fuzzer? Imagine someone only had GCC installed - what would they need to have and then what would they do to get things going?

To build a right fuzzer, one has to define the fuzzing engine (or fuzzing driver) through the environment variable `LIB_FUZZING_ENGINE`. There are multiple fuzzing engines like libFuzzer and AFL.
To get more interesting results, you should also set `CFLAGS` to include  `-fsanitize=address` (or another sanitizer)

For libFuzzer with gcc, this page seems to show how to do https://github.com/google/clusterfuzz/issues/920#issuecomment-526761713

t/fuzz/onefile.c
catenacyber: ok thanks

t/fuzz/fuzz_parseini.c
catenacyber: indeed

options.c
catenacyber: ok

sitsofe:
> To build a right fuzzer [...]

Ah I meant as a comment that sits alongside the code/documentation so anyone who looks at it will learn what you posted here.

catenacyber:
> I meant as a comment that sits alongside the code/documentation

I added the `LIB_FUZZING_ENGINE` comment in Makefile.
Do you want some other place in the code ? Which ones ?

And it is hard to cover all use cases as there are many : multiple compilers, multiple fuzzing engines, multiple sanitizers...
See https://github.com/OISF/suricata/blob/master/src/tests/fuzz/README for instance

sitsofe:
> And it is hard to cover all use cases as there are many 

Ah I see what you mean. A link or two (just like those you've posted in above comments) would be more than enough! :-)

Makefile
```diff
@@ -346,6 +346,23 @@ T_MEMLOCK_PROGS = t/memlock
...
+T_FUZZ_OBJS += $(OBJS)
+ifdef CONFIG_ARITHMETIC
+T_FUZZ_OBJS += lex.yy.o y.tab.o
+endif
+# in case there is no fuzz driver defined by environment variable LIB_FUZZING_ENGINE, use a simple one
+# For instance, with compiler clang, address sanitizer and libFuzzer as a fuzzing engine, you should define
+# export CFLAGS="-fsanitize=address,fuzzer-no-link"
+# export LIB_FUZZING_ENGINE="-fsanitize=address"
+# export CC=clang
+# before running make
```
sitsofe, line 359: Should this be
`# before running configure`
? If you've never run configure then make will invoke it as a convenience but that's more of a special case...

catenacyber: Indeed, thanks

sitsofe (approved):
LGTM. @axboe could you consider pulling this one?

sitsofe:
Thanks @catenacyber!

catenacyber:
Thanks, now waiting for https://github.com/google/oss-fuzz/pull/4916 to get oss-fuzz running

catenacyber:
oss-fuzz started finding more bugs.
Do you have access to them ? 
https://oss-fuzz.com/testcase?key=5913552609214464 for instance ?
@sitsofe should you be added to https://github.com/google/oss-fuzz/blob/master/projects/fio/project.yaml#L5 ?

axboe:
I did get a ton of email on that! That was actually my worry with the fuzzing, it really requires that someone at the other end has time to look into these. It'd be great to give Sitsofe access as well as a start.

catenacyber:
https://github.com/google/oss-fuzz/pull/4978 is to give Sitsofe access to oss-fuzz bug reports

> I did get a ton of email on that! That was actually my worry with the fuzzing, it really requires that someone at the other end has time to look into these

There are indeed quite some to look at in the beginning. But this workload should decrease over time as bugs get fixed

sitsofe:
@catenacyber So I've been taking a quick poke at the ever growing pile of reports but there's a problem - I'm really struggling to reproduce the problems reported locally. I'm currently on macOS but even with the sanitizers turned on I can't reproduce most of the issue locally which really impairs my ability to investigate an issue (since I don't have that much time to start with). Is there some sort of environment which I can interact with interactively and make the issue happen?

sitsofe:
@catenacyber Another query - when I compile with the undefined sanitizer and without setting the `LIB_FUZZING_ENGINE` environment variable I get the following error when trying to use `fuzz_parseini`:

```
LDFLAGS="-fsanitize=undefined" eatmydata ./configure --extra-cflags="-fsanitize=undefined"
make -j $(nproc)
[...]
# printf '[t]\nioengine=null\nsize=8k\n' > /tmp/test.fio 
# ./t/fuzz/fuzz_parseini /tmp/test.fio 
fio_sem.c:151:2: runtime error: member access within null pointer of type 'struct fio_sem'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior fio_sem.c:151:2 in 
UndefinedBehaviorSanitizer:DEADLYSIGNAL
==16784==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x000000000060 (pc 0x000000536f7f bp 0x000000000000 sp 0x7fffa1bb3d70 T16784)
==16784==The signal is caused by a READ memory access.
==16784==Hint: address points to the zero page.
    #0 0x536f7f in fio_sem_down /fio/fio_sem.c:151:2
    #1 0x54be02 in file_hash_exit /fio/filehash.c:112:2
    #2 0x481317 in free_shm /fio/init.c:345:2
    #3 0x7fd43d7a2a26  (/lib/x86_64-linux-gnu/libc.so.6+0x49a26)
    #4 0x7fd43d7a2bdf in exit (/lib/x86_64-linux-gnu/libc.so.6+0x49bdf)
    #5 0x7fd43d7800b9 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b9)
    #6 0x40f0cd in _start (/fio/t/fuzz/fuzz_parseini+0x40f0cd)

DEDUP_TOKEN: fio_sem_down--file_hash_exit--free_shm
UndefinedBehaviorSanitizer can not provide additional info.
SUMMARY: UndefinedBehaviorSanitizer: SEGV /fio/fio_sem.c:151:2 in fio_sem_down
==16784==ABORTING
```

But this doesn't happen when with the following
```
# ./fio --parse-only /tmp/test.fio 
# echo $?
0
```

catenacyber:
> Is there some sort of environment which I can interact with interactively and make the issue happen?

The most reliable way to reproduce oss-fuzz bugs is to use their Docker cf https://google.github.io/oss-fuzz/advanced-topics/reproducing/
You can also use your local code in their Docker cf https://google.github.io/oss-fuzz/advanced-topics/reproducing/#reproduce-using-local-source-checkout

On MacOS, I recompiled libFuzzer myself instead of using the system one (cf fuzzing https://github.com/OISF/suricata/blob/master/src/tests/fuzz/README).
And I use `export ASAN_OPTIONS="detect_leaks=1"` because clang sets it off by default on MacOS

>  I get the following error when trying to use fuzz_parseini:

This may be a bug in the fuzz target (ie fuzz_parseini.c) and not in fio itself.
It looks like some initialization is missing (hence the NULL dereference)

Note that oss-fuzz also uses `-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` in CFLAGS

Does that help ?

catenacyber:
And thanks for fixing already many of the reported bugs 👍

sitsofe:
@catenacyber thanks, it's all useful information! What I have noticed is that some of the issues reported seem to be flakey even though the code for fio itself isn't changing so maybe there's some hidden state being built up...

catenacyber:
> What I have noticed is that some of the issues reported seem to be flakey even though the code for fio itself isn't changing so maybe there's some hidden state being built up...

Indeed, there are some bugs which seem unreproducible.
That likely means the fuzz target is not stateless (maybe some global variable), as you say :  some hidden state is being built up...
I recommend to focus on the reproducible test cases first. An unreproducible bug may have the same root cause ;-)

PR #1161, opened 2021-01-07
engines/cpu: Adding qsort capabilities

ErwanAliasr1:
This commit adds cpumode option into the cpuio engine.
By default, cpumode=noop to keep the current behavior.

If cpumode is set to qsort, fio will use a qsort algorithm
instead of the noop instructions to load the processor.
This mode will consume more cpu power and will be useful to
increase the pressure on the thermal and electrical components.

The expected cpu load is selected as per noop via the cpuload option.
qsort() consumes a lot of energy so the duration
of every loop will vary over time as the power management & cpu clock
changes. To ensure a proper calibration, the thinktime is adjusted after every
qsort() computation to be as precise as possible.

To give an order of magnitude, on an AMD 7502P (TDP=180W) :
cpuload=30, numjobs=64: packagewatt = 134W
cpuload=50, numjobs=64: packagewatt = 167W
cpuload=70, numjobs=64: packagewatt = 180W

The example file is updated to reflect this new capabilities.

The qsort code is coming from stress-qsort.c from stress-ng tool.
This software is also GPLv2 but author was informed and agreed with this usage.

### DIFF
--- a/engines/cpu.c
+++ b/engines/cpu.c
@@ -8,11 +8,26 @@
 #include "../fio.h"
 #include "../optgroup.h"
 
+// number of 32 bit integers to sort
+size_t qsort_size = (256 * (1ULL << 10)); // 256KB
+
+struct mwc_t {
+	uint32_t w;
+  uint32_t z;
+};
+
+enum stress_mode {
+	FIO_CPU_NOOP = 0,
+	FIO_CPU_QSORT = 1,
+};
+
 struct cpu_options {
 	void *pad;
 	unsigned int cpuload;
 	unsigned int cpucycle;
+	enum stress_mode cpumode;
 	unsigned int exit_io_done;
+	int32_t *qsort_data;
 };
 
 static struct fio_option options[] = {
@@ -24,6 +39,26 @@ static struct fio_option options[] = {
 		.help	= "Use this percentage of CPU",
 		.category = FIO_OPT_C_ENGINE,
 		.group	= FIO_OPT_G_INVALID,
+	},
+	{
+		.name     = "cpumode",
+		.lname    = "cpumode",
+		.type     = FIO_OPT_STR,
+		.help     = "Stress mode",
+		.off1     = offsetof(struct cpu_options, cpumode),
+		.def      = "noop",
+		.posval = {
+			  { .ival = "noop",
+			    .oval = FIO_CPU_NOOP,
+			    .help = "NOOP instructions",
+			  },
+			  { .ival = "qsort",
+			    .oval = FIO_CPU_QSORT,
+			    .help = "QSORT computation",
+			  },
+		},
+		.category = FIO_OPT_C_ENGINE,
+		.group    = FIO_OPT_G_INVALID,
 	},
 	{
 		.name	= "cpuchunks",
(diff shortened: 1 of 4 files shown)


engines/cpu.c
```diff
@@ -52,6 +90,85 @@ static struct fio_option options[] = {
 	},
 };
 
+/*
+ *  mwc32()
+ *      Multiply-with-carry random numbers
+ *      fast pseudo random number generator, see
+ *      http://www.cse.yorku.ca/~oz/marsaglia-rng.html
+ */
+uint32_t mwc32(mwc_t *mwc) {
```
axboe, line 99: brace on next line

engines/cpu.c
```diff
@@ -52,6 +90,85 @@ static struct fio_option options[] = {
...
+static int stress_qsort_cmp_3(const void *p1, const void *p2)
+{
+	const int8_t *i1 = (const int8_t *)p1;
+	const int8_t *i2 = (const int8_t *)p2;
+
+	/* Force re-ordering on 8 bit value */
+	return *i1 - *i2;
+}
+
+static int do_qsort(struct thread_data *td) {
```
axboe, line 144: Same here (brace)

engines/cpu.c
```diff
@@ -63,14 +180,57 @@ static enum fio_q_status fio_cpuio_queue(struct thread_data *td,
...
-	usec_spin(co->cpucycle);
+	switch (co->cpumode) {
+		case FIO_CPU_NOOP: 	usec_spin(co->cpucycle); break;
+		case FIO_CPU_QSORT: do_qsort(td); break;
+	}
+
 	return FIO_Q_COMPLETED;
 }
 
+static int noop_init(struct thread_data *td) {
```
axboe, line 191: Ditto

engines/cpu.c
```diff
@@ -63,14 +180,57 @@ static enum fio_q_status fio_cpuio_queue(struct thread_data *td,
 		return FIO_Q_BUSY;
 	}
 
-	usec_spin(co->cpucycle);
+	switch (co->cpumode) {
```
axboe, line 183: I do prefer having separate lines, makes it easier to read for a switch.

axboe:
Just some minor style change requests, looks good overall. One question on calibration - 5 seconds is a looong time, why so long? What's the accuracy drop going to eg 500 ms instead?

engines/cpu.c
ErwanAliasr1: fixed

engines/cpu.c
ErwanAliasr1: fixed

engines/cpu.c
ErwanAliasr1: fixed

engines/cpu.c
ErwanAliasr1: done

ErwanAliasr1:
> Just some minor style change requests, looks good overall. One question on calibration - 5 seconds is a looong time, why so long? What's the accuracy drop going to eg 500 ms instead?

Sorry about the 5sec leftover, it was useless and ineffective in the code.

 I changed my mind on this. At each do_qsort() loop, the thinktime is adjusted regarding the current performance. This avoids having an initial calibration value that become wrong over time as per the cpu clock change and thread concurrency.

This corrective loop ensures we always have the right precision regarding the current context.

ErwanAliasr1:
Pushed a new version with the fixes you requested.

axboe:
That sounds much better than a calibration!

ErwanAliasr1:
> That sounds much better than a calibration!

Yup. That's what I figured out when running on 32C/64T systems ;)

engines/cpu.c
```diff
@@ -8,11 +8,26 @@
 #include "../fio.h"
 #include "../optgroup.h"
 
+// number of 32 bit integers to sort
+size_t qsort_size = (256 * (1ULL << 10)); // 256KB
+
+typedef struct {
```
axboe, line 14: Lets get rid of this typedef, generally not a fan unless it's a specific single unit type.

engines/cpu.c
```diff
@@ -52,6 +87,87 @@ static struct fio_option options[] = {
...
+	qsort(co->qsort_data, qsort_size, sizeof(*(co->qsort_data)), stress_qsort_cmp_2);
+
+	/* And re-order by byte compare */
+	qsort((uint8_t *)co->qsort_data, qsort_size * 4, sizeof(uint8_t), stress_qsort_cmp_3);
+
+	/* Reverse sort this again */
+	qsort(co->qsort_data, qsort_size, sizeof(*(co->qsort_data)), stress_qsort_cmp_2);
+	fio_get_mono_time(&now);
+
+	// Adjusting cpucycle automatically to be as close as possible to the expected cpuload
```
axboe, line 161: Let's use /**/ style comments throughout

engines/cpu.c
```diff
@@ -63,14 +179,65 @@ static enum fio_q_status fio_cpuio_queue(struct thread_data *td,
...
 
-	usec_spin(co->cpucycle);
+	switch (co->cpumode) {
+		case FIO_CPU_NOOP:
+			usec_spin(co->cpucycle);
+			break;
+		case FIO_CPU_QSORT:
+			do_qsort(td);
+			break;
+	}
```
axboe, line 189: Probably wants to include a default: statement with a log_err() of some sort (and error return). Just a bit of defensive programming.

engines/cpu.c
```diff
@@ -63,14 +179,65 @@ static enum fio_q_status fio_cpuio_queue(struct thread_data *td,
...
+{
+	struct cpu_options *co = td->eo;
+
+	// Setting up a default entropy
+	mwc_t mwc = {
+	        (521288629UL),
+	        (362436069UL)
+	};
+
+	if ((co->qsort_data = calloc(qsort_size, sizeof(*co->qsort_data))) == NULL) {
```
axboe, line 221: Separate lines for assignment and check of return

engines/cpu.c
```diff
@@ -89,12 +263,27 @@ static int fio_cpuio_init(struct thread_data *td)
 
 	o->nr_files = o->open_files = 1;
 
-	log_info("%s: ioengine=%s, cpuload=%u, cpucycle=%u\n",
-		td->o.name, td->io_ops->name, co->cpuload, co->cpucycle);
+	switch (co->cpumode) {
```
axboe, line 266: This one also needs breaking up like the previous one.

axboe: And probably the default case belongs here, so we catch it at init time as we should. Avoids the check in the runtime/fast path.

engines/cpu.c
```diff
@@ -89,12 +263,27 @@ static int fio_cpuio_init(struct thread_data *td)
...
+		case FIO_CPU_NOOP: noop_init(td);  break;
+		case FIO_CPU_QSORT: qsort_init(td); break;
+	}
 
+	// Let's restore the previous state and clear stats.
+	reset_all_stats(td);
```
axboe, line 272: Why is it doing a stats reset?

engines/cpu.c
ErwanAliasr1: sure, done.

engines/cpu.c
```diff
@@ -89,12 +263,27 @@ static int fio_cpuio_init(struct thread_data *td)
...
+	td_set_runstate(td, td_previous_state);
 	return 0;
 }
 
+static void fio_cpuio_cleanup(struct thread_data *td)
+{
+	struct cpu_options *co = td->eo;
+
+	switch (co->cpumode) {
```
axboe, line 281: Ditto on breaking it into multiple lines.

examples/cpuio.fio
```diff
@@ -1,8 +1,15 @@
...
+[burn_2x50_noop]
 cpuload=50
+numjobs=2
+cpumode=noop
 
+[burn_2x50%_qsort]
+stonewall
+cpuload=50
+numjobs=2
+cpumode=qsort
```
axboe, line 15: Should probably include a comment here, it's nice if examples are well documented.

engines/cpu.c
ErwanAliasr1: sure, sorry.

engines/cpu.c
ErwanAliasr1: I didn't on purpose. As I do a switch() on an enum, GCC would complain if any value of the num isn't present here. If any add a new value in the enum, GCC will request the developer to add the proper case statement.
A default statement, would kill that behavior
Is it fine for you ?

engines/cpu.c
ErwanAliasr1: ok, done.

engines/cpu.c
axboe: Instead of having to worry about if this is true for any gcc that we use (and with the flags set), I'd prefer to have it as part of the init check where you have one. The rest don't need it. And it's still possible that someone sets this to a non-enum value, compiler won't warn about that.

engines/cpu.c
ErwanAliasr1: yup and fix the last one too. Sorry.

engines/cpu.c
ErwanAliasr1: yup fixed.

engines/cpu.c
ErwanAliasr1: Thought it was needed in this case. If not, I remove it.

examples/cpuio.fio
ErwanAliasr1: sure.

ErwanAliasr1:
Pushed the last version with all the fixes included

ErwanAliasr1:
thx for the merge

PR #1162, opened 2021-01-08
engines/cpu: Fix td_vmsg() call

bvanassche:
Make sure that the third td_vmsg() argument is a fully expanded string and
also that the fourth argument is a string argument. This was detected by
Coverity.

Fixes: b213922390fe ("engines/cpu: style cleanups")


axboe:
Thanks Bart

PR #1157, opened 2020-12-27
Add GitHub Actions / retire Travis

sitsofe:
- Add GitHub actions CI
- Retire Travis CI :-(. See https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing for how abuse forced Travis to change its billing model

PR #1163, opened 2021-01-08
Provide a clock_gettime() backport for OS/X

bvanassche:


bvanassche:
@sitsofe, do you perhaps want to review this pull request?

sitsofe:
@bvanassche will do (I should be done by Sunday).

.travis.yml
```diff
@@ -24,7 +24,12 @@ matrix:
     # Latest xcode image (needs periodic updating)
     - os: osx
       compiler: clang
-      osx_image: xcode11.2
+      osx_image: xcode12.3
+      arch: amd64
+    # Oldest xcode version supported by Homebrew
+    - os: osx
+      compiler: clang
+      osx_image: xcode10.3
```
sitsofe, line 32: Commit title does not match this (Given the comment I suspect the title is the one in error)

os/mac/clock_gettime.c
```diff
@@ -0,0 +1,56 @@
...
+		tp->tv_sec = tv.tv_sec;
+		tp->tv_nsec = tv.tv_usec * 1000;
+		return 0;
+	}
+	uint64_t delta_ns = (mach_absolute_time() - mach_timestart) *
+		(__uint128_t)mach_timebase.numer / mach_timebase.denom;
+	tp->tv_sec = delta_ns / (uint64_t)1e9;
+	tp->tv_nsec = delta_ns - tp->tv_sec * (uint64_t)1e9;
+	return 0;
+}
```
sitsofe, line 56: My only thought is that modern macOS doesn't need this and really can do clock_gettime so we should only use this fallback when we don't have it. Would it be possible to guard this in a `#ifndef CONFIG_CLOCK_GETTIME`?

.travis.yml
sitsofe: I guess one additional snag is that usage of travis sounds like it is going to become expensive (see https://github.com/axboe/fio/pull/1157 for details).

bvanassche: > Commit title does not match this (Given the comment I suspect the title is the one in error)

Right, I have fixed the commit title.

bvanassche: > I guess one additional snag is that usage of travis sounds like it is going to become expensive (see #1157 for details).

So we will need to run OS/X builds on Appveyor?

os/mac/clock_gettime.c
bvanassche: > My only thought is that modern macOS doesn't need this and really can do clock_gettime so we should only use this fallback when we don't have it. Would it be possible to guard this in a `#ifndef CONFIG_CLOCK_GETTIME`?

This patch allows to remove the clock_gettime() configure check. Since the configure-time clock_gettime() check is fragile, I propose to remove that check instead of adding an #ifndef / #endif pair in this patch.

.travis.yml
sitsofe: > So we will need to run OS/X builds on Appveyor?

https://github.com/axboe/fio/pull/1157 adds macOS builds via GitHub Actions. The bad thing is that older macOS builds will no longer be available but that was always one of Travis' strengths  (e.g. Appveyor only supports two macOS releases - https://www.appveyor.com/docs/macos-images-software/ ).

os/mac/clock_gettime.c
sitsofe: With modern macOS supporting the regular code path we should support that as the first class citizen and deprioritise platform specific code. While it's not pretty I'd prefer one of the following:
- Use ifdefs around the above so modern macOS gets a regular path and older macOS gets a best effort path which can eventually be dropped
- Drop support for platforms without CLOCK_GETTIME. This would be done turning the check into a configure time failure, removing the CONFIG_CLOCK_GETTIME symbol and dropping code that is trying to work around the lack of it. [OS X 10.11 is five years old](https://en.wikipedia.org/wiki/OS_X_El_Capitan) and fio has already dropped support for Linux platforms that old. Further, macOS tools like homebrew have a policy of only supporting the last two macOS releases which means that [10.14 is the oldest macOS homebrew currently supports](https://github.com/Homebrew/brew/blob/1a9c3fa06e7fcc85239ec390ce1a0a4d268f3751/Library/Homebrew/os/mac.rb#L53) so we'd be in good company.

PR #1165, opened 2021-01-14
filesetup: add option to prepopulate file with data

lukaszstolarczuk:
When read job is executed and file is zero'ed, results are
higher than expected (because reading zero page).

Signed-off-by: Łukasz Stolarczuk <lukasz.stolarczuk@intel.com>

**note:**
I've done this for the libpmem engine purpose (where it makes a difference), but I believe it may be useful for other engines as well. I've tested it also with mmap, but performance difference there was just minimal (if any).

### DIFF
--- a/cconv.c
+++ b/cconv.c
@@ -255,6 +255,7 @@ void convert_thread_options_to_cpu(struct thread_options *o,
 	o->stats = le32_to_cpu(top->stats);
 	o->fadvise_hint = le32_to_cpu(top->fadvise_hint);
 	o->fallocate_mode = le32_to_cpu(top->fallocate_mode);
+	o->prepopulate = le32_to_cpu(top->prepopulate);
 	o->zero_buffers = le32_to_cpu(top->zero_buffers);
 	o->refill_buffers = le32_to_cpu(top->refill_buffers);
 	o->scramble_buffers = le32_to_cpu(top->scramble_buffers);
@@ -456,6 +457,7 @@ void convert_thread_options_to_net(struct thread_options_pack *top,
 	top->stats = cpu_to_le32(o->stats);
 	top->fadvise_hint = cpu_to_le32(o->fadvise_hint);
 	top->fallocate_mode = cpu_to_le32(o->fallocate_mode);
+	top->prepopulate = cpu_to_le32(o->prepopulate);
 	top->zero_buffers = cpu_to_le32(o->zero_buffers);
 	top->refill_buffers = cpu_to_le32(o->refill_buffers);
 	top->scramble_buffers = cpu_to_le32(o->scramble_buffers);
--- a/filesetup.c
+++ b/filesetup.c
@@ -338,6 +338,95 @@ error:
 	return ret;
 }
 
+/*
+ * Prepopulate regular file with data.
+ * Leaves f->fd open on success, caller must close.
+ */
+static int prepopulate_file(struct thread_data *td, struct fio_file *f)
+{
+	int flags;
+	unsigned long long left, bs;
+	char *b = NULL;
+
+	if (read_only) {
+		log_err("fio: refusing extend of file due to read-only\n");
+		return 0;
+	}
+	if (td->o.fill_device) {
+		log_err("fio: prepopulate option with fill_device makes no sense\n");
+		return 0;
+	}
+
+	flags = O_WRONLY;
+	if (td->o.allow_create)
+		flags |= O_CREAT;
+
+#ifdef WIN32
+	flags |= _O_BINARY;
+#endif
+
+	dprint(FD_FILE, "open file %s, flags %x\n", f->file_name, flags);
+	f->fd = open(f->file_name, flags, 0644);
+	if (f->fd < 0) {
+		int err = errno;
+
+		if (err == ENOENT && !td->o.allow_create)
+			log_err("fio: file creation disallowed by "
+					"allow_file_create=0\n");
+		else
+			td_verror(td, err, "open");
+		return 1;
+	}
+
+	left = f->real_file_size; // XXX ?? f->file_offset + f->io_size
+	bs = td->o.max_bs[DDIR_WRITE];
+	if (bs > left)
+		bs = left;
+
+	b = malloc(bs);
+	if (!b) {
+		td_verror(td, errno, "malloc");
+		goto err;
+	}
+
+	while (left && !td->terminate) {
+		ssize_t r;
+
+		if (bs > left)
+			bs = left;
+
+		fill_io_buffer(td, b, bs, bs);
+
+		r = write(f->fd, b, bs);
+
+		if (r > 0) {
+			left -= r;
+		} else {
+			td_verror(td, errno, "write");
+			goto err;
+		}
+	}
+
+	if (td->terminate) {
+		dprint(FD_FILE, "terminate unlink %s\n", f->file_name);
+		td_io_unlink_file(td, f);
+	} else if (td->o.create_fsync) {
+		if (fsync(f->fd) < 0) {
+			td_verror(td, errno, "fsync");
+			goto err;
+		}
+	}
+
+	free(b);
+	return 0;
+err:
+	close(f->fd);
+	f->fd = -1;
+	if (b)
+		free(b);
+	return 1;
+}
+
 unsigned long long get_rand_file_size(struct thread_data *td)
 {
 	unsigned long long ret, sized;
@@ -1240,6 +1329,44 @@ int setup_files(struct thread_data *td)
 
 			err = __file_invalidate_cache(td, f, old_len,
 								extend_len);
+
+			/*
+			 * Shut up static checker
+			 */
+			if (f->fd != -1)
+				close(f->fd);
+
+			f->fd = -1;
+			if (err)
+				break;
+		}
+		temp_stall_ts = 0;
+	}
+
+	if (err)
+		goto err_out;
+
+	/*
+	 * Prepopulate regular files with data. It might be expected
+	 * to read some "real" data instead of zero'ed files (if no writes
+	 * to file occurred prior to a read job).
+	 */
+	if (td->o.prepopulate && td_read(td)) {
+		temp_stall_ts = 1;
+
+		for_each_file(td, f, i) {
+			assert(f->filetype == FIO_TYPE_FILE);
+			if (output_format & FIO_OUTPUT_NORMAL) {
+				log_info("%s: Prepopulating IO file (%s)\n",
+							o->name, f->file_name);
+			}
+
+			err = prepopulate_file(td, f);
+			if (err)
+				break;
+
+			err = __file_invalidate_cache(td, f, f->file_offset,
+								f->io_size);
 
 			/*
 			 * Shut up static checker
--- a/options.c
+++ b/options.c
@@ -2525,6 +2525,16 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
 		},
 	},
 	{
+		.name	= "prepopulate",
+		.lname	= "Prepopulate",
+		.type	= FIO_OPT_BOOL,
+		.off1	= offsetof(struct thread_options, prepopulate),
+		.help	= "Prepopulate files with some data",
+		.def	= "0",
+		.category = FIO_OPT_C_FILE,
+		.group	= FIO_OPT_G_INVALID,
+	},
+	{
 		.name	= "fadvise_hint",
 		.lname	= "Fadvise hint",
 		.type	= FIO_OPT_STR,
--- a/thread_options.h
+++ b/thread_options.h
@@ -230,6 +230,7 @@ struct thread_options {
 	unsigned int stats;
 	unsigned int fadvise_hint;
 	enum fio_fallocate_mode fallocate_mode;
+	unsigned int prepopulate;
 	unsigned int zero_buffers;
 	unsigned int refill_buffers;
 	unsigned int scramble_buffers;
@@ -619,6 +620,9 @@ struct thread_options_pack {
 	uint64_t max_latency;
 	fio_fp64_t latency_percentile;
 	uint32_t latency_run;
+
+
+	uint32_t prepopulate;
 
 	/*
 	 * flow support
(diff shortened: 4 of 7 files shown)


options.c
```diff
@@ -2524,6 +2524,16 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
...
 	},
+	{
+		.name	= "prepopulate",
+		.lname	= "Prepopulate",
+		.type	= FIO_OPT_BOOL,
+		.off1	= offsetof(struct thread_options, prepopulate),
+		.help	= "Prepopulate files with some data",
+		.def	= "0",
+		.category = FIO_OPT_C_FILE,
+		.group	= FIO_OPT_G_INVALID,
```
lukaszstolarczuk, line 2535: I'm really not sure if the group was picked properly, pls verify

thread_options.h
```diff
@@ -620,6 +621,9 @@ struct thread_options_pack {
 	fio_fp64_t latency_percentile;
 	uint32_t latency_run;
 
+
+	uint32_t prepopulate;
```
lukaszstolarczuk, line 625: I wanted to put this just next to fallocate option (above in the file), but I've got static assert in libfio.c:367.
And I'm not even sure if we need this - pls verify if I can do better here...

filesetup.c
```diff
@@ -338,6 +338,95 @@ static bool pre_read_file(struct thread_data *td, struct fio_file *f)
...
+{
+	int flags;
+	unsigned long long left, bs;
+	char *b = NULL;
+
+	if (read_only) {
+		log_err("fio: refusing extend of file due to read-only\n");
+		return 0;
+	}
+	if (td->o.fill_device) {
```
lukaszstolarczuk, line 355: I assume this makes no sense, but pls verify

filesetup.c
```diff
@@ -338,6 +338,95 @@ static bool pre_read_file(struct thread_data *td, struct fio_file *f)
...
+
+		if (err == ENOENT && !td->o.allow_create)
+			log_err("fio: file creation disallowed by "
+					"allow_file_create=0\n");
+		else
+			td_verror(td, err, "open");
+		return 1;
+	}
+
+	left = f->real_file_size; // XXX ?? f->file_offset + f->io_size
```
lukaszstolarczuk, line 381: and need a little help here - which one is correct? or maybe should cover both (if io_size is set)

sitsofe (changes requested):
> add option to prepopulate file with data

I thought this was already done on pure read jobs when the file doesn't already exist?

```|
% rm fio.tmp
% ls -l fio.tmp
ls: fio.tmp: No such file or directory
% fio --name=test --filename=fio.tmp --size=64 --bs=32 --rw=read
test: (g=0): rw=read, bs=(R) 32B-32B, (W) 32B-32B, (T) 32B-32B, ioengine=psync, iodepth=1
fio-3.25-37-g4008
Starting 1 process
test: Laying out IO file (1 file / 0MiB)

test: (groupid=0, jobs=1): err= 0: pid=5447: Fri Jan 15 12:49:51 2021
  read: IOPS=2000, BW=62.5KiB/s (64.0kB/s)(64B/1msec)
    clat (nsec): min=6000, max=73000, avg=39500.00, stdev=47376.15
     lat (nsec): min=6000, max=74000, avg=40000.00, stdev=48083.26
    clat percentiles (nsec):
     |  1.00th=[ 5984],  5.00th=[ 5984], 10.00th=[ 5984], 20.00th=[ 5984],
     | 30.00th=[ 5984], 40.00th=[ 5984], 50.00th=[ 5984], 60.00th=[73216],
     | 70.00th=[73216], 80.00th=[73216], 90.00th=[73216], 95.00th=[73216],
     | 99.00th=[73216], 99.50th=[73216], 99.90th=[73216], 99.95th=[73216],
     | 99.99th=[73216]
  lat (usec)   : 10=50.00%, 100=50.00%
  cpu          : usr=0.00%, sys=0.00%, ctx=0, majf=0, minf=34
     issued rwts: total=2,0,0,0 short=0,0,0,0 dropped=0,0,0,0

   READ: bw=62.5KiB/s (64.0kB/s), 62.5KiB/s-62.5KiB/s (64.0kB/s-64.0kB/s), io=64B (64B), run=1-1msec
% hexdump -C fio.tmp
00000000  35 e0 28 cc 38 a0 99 16  06 9c 6a a9 f2 cd e9 0a  |5.(.8.....j.....|
00000010  80 53 2a 07 09 e5 0d 15  70 4a 25 f7 0b 39 9d 18  |.S*.....pJ%..9..|
00000020  a8 71 09 48 d3 ad 5f c5  35 2e 2d b0 b5 51 5a 13  |.q.H.._.5.-..QZ.|
00000030  c6 25 eb 1e 20 72 c3 13  b8 64 fa 70 ce 5e 52 18  |.%.. r...d.p.^R.|
00000040
```

Could you illustrate the problem with a small test file? Is this patch explicitly for cases where the file already exists?

lukaszstolarczuk:
I think FIO_DISKLESSIO is to blame. You used the psync engine, we've noticed the problem with the libpmem engine.
```
$ ls /mnt/pmem/ -l
total 0

$ ./fio --name=read --size=64 --bs=32 --rw=read --ioengine=libpmem --directory=/mnt/pmem
read: (g=0): rw=read, bs=(R) 32B-32B, (W) 32B-32B, (T) 32B-32B, ioengine=libpmem, iodepth=1
fio-3.25-37-g4008-dirty
Starting 1 process

read: (groupid=0, jobs=1): err= 0: pid=420106: Fri Jan 15 17:13:48 2021
  ...

   READ: bw=62.5KiB/s (64.0kB/s), 62.5KiB/s-62.5KiB/s (64.0kB/s-64.0kB/s), io=64B (64B), run=1-1msec

$ ls /mnt/pmem -l
total 4
-r-------- 1 luke luke 64 sty 15 17:13 read.0.0

$ hexdump -C /mnt/pmem/read.0.0
00000000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000040
```
and with the prepopulate option:
```
$ rm /mnt/pmem/read.0.0
rm: remove write-protected regular file '/mnt/pmem/read.0.0'? y

$ ./fio --name=read --size=64 --bs=32 --rw=read --ioengine=libpmem --directory=/mnt/pmem --prepopulate=1
read: (g=0): rw=read, bs=(R) 32B-32B, (W) 32B-32B, (T) 32B-32B, ioengine=libpmem, iodepth=1
fio-3.25-38-g3c93e0
Starting 1 process
read: Prepopulating IO file (/mnt/pmem/read.0.0)

read: (groupid=0, jobs=1): err= 0: pid=421702: Fri Jan 15 17:21:30 2021
  ...

   READ: bw=62.5KiB/s (64.0kB/s), 62.5KiB/s-62.5KiB/s (64.0kB/s-64.0kB/s), io=64B (64B), run=1-1msec

$ ls /mnt/pmem -l
total 4
-rw-r--r-- 1 luke luke 64 sty 15 17:21 read.0.0

$ hexdump -C /mnt/pmem/read.0.0
00000000  35 e0 28 cc 38 a0 99 16  06 9c 6a a9 f2 cd e9 0a  |5.(.8.....j.....|
00000010  80 53 2a 07 09 e5 0d 15  70 4a 25 f7 0b 39 9d 18  |.S*.....pJ%..9..|
00000020  a8 71 09 48 d3 ad 5f c5  35 2e 2d b0 b5 51 5a 13  |.q.H.._.5.-..QZ.|
00000030  c6 25 eb 1e 20 72 c3 13  b8 64 fa 70 ce 5e 52 18  |.%.. r...d.p.^R.|
00000040
```

Maybe the fix is much simpler and I should just update some file laying rules. If you have any advice, please share.

sitsofe:
> we've noticed the problem with the libpmem engine.

Hmm that probably is the case where fio doesn't think layout has to be done. Maybe you get the same with the psync ioengine when the file already exists at a big enough size:

```
% rm fio.tmp                              
% dd if=/dev/zero of=fio.tmp bs=64 count=1
1+0 records in
1+0 records out
64 bytes transferred in 0.000141 secs (453438 bytes/sec)
% fio --name=test --overwrite=0 --filename=fio.tmp --size=64 --bs=32 --rw=read
test: (g=0): rw=read, bs=(R) 32B-32B, (W) 32B-32B, (T) 32B-32B, ioengine=psync, iodepth=1
fio-3.25-37-g4008
Starting 1 process

test: (groupid=0, jobs=1): err= 0: pid=6264: Fri Jan 15 17:04:43 2021
  read: IOPS=2000, BW=62.5KiB/s (64.0kB/s)(64B/1msec)
    clat (nsec): min=2000, max=44000, avg=23000.00, stdev=29698.48
     lat (nsec): min=3000, max=45000, avg=24000.00, stdev=29698.48
    clat percentiles (nsec):
     |  1.00th=[ 2008],  5.00th=[ 2008], 10.00th=[ 2008], 20.00th=[ 2008],
     | 30.00th=[ 2008], 40.00th=[ 2008], 50.00th=[ 2008], 60.00th=[43776],
     | 70.00th=[43776], 80.00th=[43776], 90.00th=[43776], 95.00th=[43776],
     | 99.00th=[43776], 99.50th=[43776], 99.90th=[43776], 99.95th=[43776],
     | 99.99th=[43776]
  lat (usec)   : 4=50.00%, 50=50.00%
  cpu          : usr=0.00%, sys=0.00%, ctx=0, majf=6, minf=27
     issued rwts: total=2,0,0,0 short=0,0,0,0 dropped=0,0,0,0

   READ: bw=62.5KiB/s (64.0kB/s), 62.5KiB/s-62.5KiB/s (64.0kB/s-64.0kB/s), io=64B (64B), run=1-1msec
% hexdump -C fio.tmp
00000000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000040
```

Looking back, you're not the first to encounter this issue (see https://www.spinics.net/lists/fio/msg08619.html ) although the problem you see is different.

> Maybe the fix is much simpler and I should just update some file laying rules

It doesn't look totally straightforward...

If the ioengine is "normal" and disk based:
- We can just force going through the existing path.

If the ioengine is "diskless":
- We don't know how to do writes without the ioengine's help. Do we ask it to provide a "do synchronous write" function which could be excruciatingly slow?

Maybe what is really needed is a fake write job "conditioning" pass that is only done when an appropriate option is on (`precondition=1` although maybe this gives people the wrong impression given things like PTS). What are your thoughts?

lukaszstolarczuk:
@sitsofe, I hope I got your idea right. I think the per engine function to do the pre-write/prepopulate file(s) with data is a good option. I seems like something we want to enable per engine and to be turned on all the time, not just for some selected measurements.

I prepared a solution on separate branch, please take a look and tell me if this is something you agree with:
https://github.com/axboe/fio/compare/master...lukaszstolarczuk:per-engine-pre-write-function?expand=1

If so, I'll open a new PR.

Also, should I advertise this new io_ops (in a documentation)?

lukaszstolarczuk:
@sitsofe, ping.
What do you think on the new proposed solution?

sitsofe:
@lukaszstolarczuk thanks for the ping - I'll take a look on the weekend.

filesetup.c
```diff
@@ -1254,6 +1343,44 @@ int setup_files(struct thread_data *td)
...
 
+	if (err)
+		goto err_out;
+
+	/*
+	 * Prepopulate regular files with data. It might be expected
+	 * to read some "real" data instead of zero'ed files (if no writes
+	 * to file occurred prior to a read job).
+	 */
+	if (td->o.prepopulate && td_read(td)) {
```
sitsofe, line 1354: I wouldn't optimize only for reads unless your going to introduce an additional state (which would be kinda awesome because it opens the opportunity to use this to entirely replace the existing code). There can be reasons why you want to force pre-population when you're doing writes (e.g. job stops part way through and then reads are done and the reads perform faster than expected).

cconv.c
```diff
@@ -456,6 +457,7 @@ void convert_thread_options_to_net(struct thread_options_pack *top,
 	top->stats = cpu_to_le32(o->stats);
 	top->fadvise_hint = cpu_to_le32(o->fadvise_hint);
 	top->fallocate_mode = cpu_to_le32(o->fallocate_mode);
+	top->prepopulate = cpu_to_le32(o->prepopulate);
```
sitsofe, line 460: We may need to bump `FIO_SERVER_VER` in server.h depending on if a new fio release is done before this goes in. Hard to say at this point! :-P

options.c
sitsofe: I'd agree with this because things like `fallocate()` are similar.

thread_options.h
sitsofe: Theoretically you can do it if you but you'll need to add some the padding because there are some variables that HAVE to be aligned on 64 bit boundaries and that's what the static assert is grumbling about. It's not great to have a lot of holes but make sure you rebase your patch as a change to this data structure happened recently and maybe that makes it easier to put this where you want (but it may not).

sitsofe (changes requested):
In general this is looking good and I'm wondering if we can rip out the existing pre-population code and make it call into this. The only other thing might be to introduce a new op for ioengines to support pre-population if they aren't file based.

cconv.c
lukaszstolarczuk: I believe it's not needed if we move to use io_ops instead of the parameter.

filesetup.c
lukaszstolarczuk: Now I believe I should use file_off + io_size, but I'm not sure how to apply offset into the `while` loop with writes.

(I moved this issue into a new PR)

filesetup.c
lukaszstolarczuk: I see and agree. Done.

thread_options.h
lukaszstolarczuk: I believe I don't have to worry about it anymore.

lukaszstolarczuk (commented):
Thanks for the review. I'm reading your comments and I believe the largest part for moving this into a separate io_ops is done, see new PR: #1182

Issue #1130, opened 2020-11-23
How can run fio on SMB share


mirim-123:
I need run fio on smb share in windows
File path is \\10.10.x.x\Test1

I try run fio with following command:
"C:\Program Files\fio\fio.exe" --name=warmup --readwrite=write --bs=4k --ba=64k --filename=\\\\10.10.x.x\Test1 --size=1GB --ioengine=windowsaio --runtime=10000 --time_based --iodepth=128 --numjobs=1 --output=c:\output2.txt

And get result:
fio: failed to create dir (\\\\): No such file or directory

Is there any way to do it right?

Thanks

sitsofe:
I think the "quick" workaround for you for now would be to map the share as a drive and then access it through a drive letter.

I can well believe that UNC paths need work (I think https://github.com/axboe/fio/blob/9107a641e67f27003fa6cbe7b55b1ec6a0239197/os/os-windows.h#L205 would need some love). I'm not able to work on this but I'd be happy to review patches so I'll label this up.

mirim-123:
Thanks! I will try it . hope it is the answer

mirim-123:
It is work well!
Thanks you

sitsofe:
I'll leave this one open as it's a genuine issue if you can't set up a mapping.

sitsofe:
(@mirim-123 but I'm glad you were able to workaround your particular issue!)

PR #1172, opened 2021-01-20
Fix Windows UNC path parsing

fkouteib:
This is a fix for #1130 by adding support for Windows UNC path. The issue is the logic for parsing the path was previously early exiting when encountering leading '\\' and stopping before it parsed the full server name, resulting in error message "fio: failed to create dir (\\): No such file or directory".

Unfortunately, I don't have access to a Windows SMB share to fully test the changes. So I isolated the path parsing changes and verified them independently for correctness.

### DIFF
--- a/filesetup.c
+++ b/filesetup.c
@@ -928,7 +928,17 @@ int longest_existing_path(char *path) {
 	sprintf(buf, "%s", path);
 	done = false;
 	while (!done) {
-		buf_pos = strrchr(buf, FIO_OS_PATH_SEPARATOR);
+#if WIN32
+		// Support for Windows UNC paths (i.e. '\\server1\...').
+		if ((buf[0] == FIO_OS_PATH_SEPARATOR) && (buf[1] == FIO_OS_PATH_SEPARATOR)) {
+			buf_pos = strrchr(buf + 2, FIO_OS_PATH_SEPARATOR);
+		}
+		else
+#endif
+		{
+			buf_pos = strrchr(buf, FIO_OS_PATH_SEPARATOR);
+		}
+
 		if (!buf_pos) {
 			done = true;
 			offset = 0;


sitsofe:
@fkouteib Thanks for submitting this pull request!

I feel we should try to isolate platform specific changes to platform specific files. Can you prototype an `fio_next_path_seperator()` with a generic fallback for POSIX platforms? You can initially do that by adding defines and functions to os/os-windows.h and os/os.h and adding a file that is always linked in https://github.com/axboe/fio/blob/master/Makefile#L245-L250 . I can add more suggestions if that's helpful...

With respect to your commit message:
- Could you add your signed-off-by line?
- Can you prefix your commit message title with an appropriate `section: ` ?
- Can you reformat the body of your commit to a width of 72 characters?
(see https://github.com/axboe/fio/pull/1050#issuecomment-663606909 for some discussion/examples)

You don't need an SMB share to test a UNC path - even local paths can be UNC ones (e.g. `\\?\C:\Users`) - see https://docs.microsoft.com/en-gb/windows/win32/fileio/naming-a-file for details. Can you test your patch with that and check it still works with old fashioned absolute paths (e.g. `C:\Users` and assuming you're currently "on" the C drive `\Users` see https://docs.microsoft.com/en-gb/windows/win32/fileio/naming-a-file#fully-qualified-vs-relative-paths for documentation)?

It's worth noting that your change will be the first step in improved Windows path support and will be reviewed standalone. So long as this patch fixes your immediate issue and doesn't regress anything we should look into moulding it and getting it accepted!

Longer term, fio on Windows currently uses old fashioned non-unicode aware functions which may go on to cause trouble with more complicated/longer paths. See https://docs.microsoft.com/en-us/windows/win32/intl/conventions-for-function-prototypes for Windows and unicode...

filesetup.c
```diff
@@ -928,7 +928,17 @@ int longest_existing_path(char *path) {
 	sprintf(buf, "%s", path);
 	done = false;
 	while (!done) {
-		buf_pos = strrchr(buf, FIO_OS_PATH_SEPARATOR);
+#if WIN32
+		// Support for Windows UNC paths (i.e. '\\server1\...').
```
sitsofe, line 932: Can you use regular C comment style (rather than C++)?

sitsofe:
A few more questions:
- What happens if the path is just a single character (e.g. `a`)?
- What happens if later parts of the path have double backslashes (e.g. `foo\bar\\baz`)?

sitsofe (changes requested):
(marking as changes requested)

sitsofe (changes requested):
(Setting to changes requested so I skip over this one)

PR #1173, opened 2021-01-22
bug: ESX does not have timerfd support

brianredbeard:
Running fio on an ESX host fails due to the lack of support for file
descriptor based notifications of timer expiration.  This change
bypasses patch 696378af added to increase timer accuracy.

Signed-off-by: Brian 'Redbeard' Harrington <redbeard@redhat.com>

### DIFF
--- a/configure
+++ b/configure
@@ -2695,6 +2695,7 @@ print_config "Windows PDB generation" "$pdb"
 ##########################################
 # check for timerfd support
 timerfd_create="no"
+if test "$esx" != "yes" ; then
 cat > $TMPC << EOF
 #include <sys/time.h>
 #include <sys/timerfd.h>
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -2695,17 +2695,19 @@ print_config "Windows PDB generation" "$pdb"
...
-cat > $TMPC << EOF
+if test "$esx" != "yes" ; then
+  cat > $TMPC << EOF
 #include <sys/time.h>
 #include <sys/timerfd.h>
 
 int main(int argc, char **argv)
 {
-	return timerfd_create(CLOCK_MONOTONIC, TFD_NONBLOCK);
+  return timerfd_create(CLOCK_MONOTONIC, TFD_NONBLOCK);
```
sitsofe, line 2705: I admire your diligence fixing the indentation :-)

configure
```diff
@@ -2695,17 +2695,19 @@ print_config "Windows PDB generation" "$pdb"
 ##########################################
 # check for timerfd support
 timerfd_create="no"
-cat > $TMPC << EOF
```
sitsofe, line 2698: Looking good! One minor suggestion - can you change the prefix from `bug:` to `configure:` as fio commit messages are prefixed by area rather than change type.

sitsofe (changes requested):

brianredbeard:
You betcha. Thanks for doing the yeoman's work.

sitsofe:
@axboe This LGTM - could you pull it?

sitsofe (approved):

PR #1177, opened 2021-02-05
num2str(): Fix invalid unsigned arithmetic.

ihsinme:
I believe your checks are not correct.
in my opinion they are equivalent to checks !=.
I suggest a simple fix.

### DIFF
--- a/lib/num2str.c
+++ b/lib/num2str.c
@@ -105,7 +105,7 @@ done:
 	/*
 	 * Fill in everything and return the result.
 	 */
-	assert(maxlen - strlen(tmp) - 1 > 0);
+	assert(maxlen > ( strlen(tmp) + 1 ));
 	assert(modulo < thousand);
 	sprintf(tmp, "%.*f", (int)(maxlen - strlen(tmp) - 1),
 		(double)modulo / (double)thousand);


sitsofe (changes requested):
@ihsinme: Interesting! So a few points:

- Could you fill out the description with a scenario this fixes? From what I understand `maxlen` is only the number of digits and strlen returns the size of a string without the terminating null so why do you need `strlen(tmp) + 1`?
- Can you prefix the area on your commit message (see https://github.com/axboe/fio/commits/master ) for the style.
- Can you add your Signed-off-by ?

ihsinme:
good day.
I'm sorry to be late with the reply.
but I probably need your help on some of your questions.

> * Could you fill out the description with a scenario this fixes? From what I understand `maxlen` is only the number of digits and strlen returns the size of a string without the terminating null so why do you need `strlen(tmp) + 1`?

correcting the error, I try not to change the code as much as possible. if you see rationality in removing `-1` I will do it.

> * Can you prefix the area on your commit message (see https://github.com/axboe/fio/commits/master ) for the style.

if I understand correctly I have to change the name. if so, tell me the starting prefix.

> * Can you add your Signed-off-by ?

this point is not clear to me, probably due to lack of experience. could you please more extensively indicate what you mean.

thanks

sitsofe:
@ihsinme 

> correcting the error, I try not to change the code as much as possible. if you see rationality in removing -1 I will do it.

It's more a case of the reverse. Can you tell me the scenario/inputs that generate this error and can you tell me why the existing code needs this change to avoid the problem? I'm NOT saying "you're wrong there's not issue"! Rather, I'm seeking to understand the problem scenario so I can see WHY your fix is correct and accept it :-).

> if I understand correctly I have to change the name. if so, tell me the starting prefix.

I'd rather you picked one but may the commits titles you see on https://github.com/axboe/fio/commits/master/lib/num2str.c will give you some more focussed ideas.

ihsinme:
I am considering this problem.
```cpp
unsigned long A;

// A= 70; (A - 11) = 59; true
// A= 11; (A - 11) = 0; false
// A= 1; (A - 11) = 18446744073709551607; true
// A= 0; (A - 11) = 18446744073709551606; true
if (A - 11 > 0)
```

sitsofe:
@ihsinme Are you saying type promotion and widening will ruin check due to wraparound ("integer underflow") in the final two cases you outlined? If so do you happen to know what are the inputs to `num2str` generate this problem? The reason being if they exist it would be good to add the such inputs to [unittests/lib/num2str.c](https://github.com/axboe/fio/blob/master/unittests/lib/num2str.c#L17).

Regarding the commit message itself:
- It's your commit message (e.g. https://github.com/axboe/fio/pull/1177/commits/44271d956937b6f3b0d1c7ae49400fd7436d4f97 ) that needs the amending. Please ensure any amendments are squashed together so they arrive as one commit in this particular scenario
  - Please add your Signed-off-by (https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for ) to it
  - Please update the title of the commit. The title you have on the pull request would be appropriate for the commit and the only adjustments I could suggest are perhaps skipping the brackets and the trailing full stop
  - Please update the description of the commit message outlining the problem case
  - If possible add a unit test case too

Thanks!

ihsinme:
good day.
I'm sorry to be late with the reply.
I have revised your code again.
and noticed that there is a sign comparison above, which will exclude the overflow that I am talking about.

PR #1180, opened 2021-02-08
options: allow separate values for max_latency

venkatrag1:
Add ability to specify latency limit separately for read, write, trim similar to blocksize, rate parameters etc


thread_options.h
```diff
@@ -203,7 +203,7 @@ struct thread_options {
 	enum fio_memtype mem_type;
 	unsigned int mem_align;
 
-	unsigned long long max_latency;
+	unsigned long long max_latency[DDIR_RWDIR_CNT];
```
sitsofe, line 206: We're OK with changing this data structure and not ramping FIO_SERVER_VER in server.h because the last ramp in https://github.com/axboe/fio/commit/a87c90fd72823d5438d724e5a57ced8d1f7bed3f has yet to make it into an official release.

fio.1
```diff
@@ -2397,10 +2397,11 @@ Used with \fBlatency_target\fR. If false (default), fio will find the highest
 queue depth that meets \fBlatency_target\fR and exit. If true, fio will continue
 running and try to meet \fBlatency_target\fR by adjusting queue depth.
 .TP
-.BI max_latency \fR=\fPtime
+.BI max_latency \fR=\fPtime[,time][,time]
 If set, fio will exit the job with an ETIMEDOUT error if it exceeds this
 maximum latency. When the unit is omitted, the value is interpreted in
-microseconds.
+microseconds. Comma-separated values may be specified for reads, writes,
+and trims as described in \fBblocksize\fR.
```
sitsofe, line 2404: You got the man page too - nice!

sitsofe (commented):
@venkatrag1 Looking good! Two minor comments:

- Could you shorten the title of the commit message and swap the area? Maybe something like `options: allow separate values for max_latency` and you can put any extra detail in the commit message body.
- Could you add your Signed-off-by ?

thread_options.h
venkatrag1: Thanks a lot for the quick review! I will shorten the title and add the details in commit message as suggested. I will also add my name and email address in Signed-off-by.

venkatrag1:
Converting to draft to handle Test009 of latency_profiles.py

io_u.c
```diff
@@ -1888,8 +1888,8 @@ static void account_io_completion(struct thread_data *td, struct io_u *io_u,
 				icd->error = ops->io_u_lat(td, tnsec);
 		}
 
-		if (td->o.max_latency && tnsec > td->o.max_latency)
-			lat_fatal(td, icd, tnsec, td->o.max_latency);
+		if (ddir_rw(idx) && td->o.max_latency[idx] && tnsec > td->o.max_latency[idx])
```
venkatrag1, line 1891: This represents a change in behavior for --max_latency. Unlike the rate params which only check read, write and trim; currently the max_latency and latency_target checks currently also apply to syncs. 
This line will cause that behavior to change in the case where only one option is specified without commas - previously that would have applied the latency limit to sync also, but with this line will cause it to apply only to read, write and trim. Wanted to check what the best options are over here - 
1) Keep this change for max_latency, also introduce the ddir_rw check for latency_target so that latency checks are also consistent with BW checks, and what is logged in lat_log (syncs are not logged)
2) Keep the original functionality of max_latency for the default case - this will involve adding some way to distinguish between the case where user passes the same limits for read,write,trim vs when user passes only one value and this gets applied to other two operations in parse.c

options.c
```diff
@@ -3728,8 +3728,10 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
 	{
 		.name	= "max_latency",
 		.lname	= "Max Latency (usec)",
-		.type	= FIO_OPT_STR_VAL_TIME,
-		.off1	= offsetof(struct thread_options, max_latency),
+		.type	= FIO_OPT_ULL,
```
venkatrag1, line 3731: This change was needed to apply limits to trim and it works because is_time is set so the time strings get decoded correctly.

sitsofe:
(I'll take a look at this over the weekend)

io_u.c
sitsofe: I'd suggest 2. Which value will sync's take for max_latency when two or more are present?

venkatrag1: Sure, I'll add the ddir_rw checks for latency_target too. 

This change will take take sync's out of the equation for max_latency and latency target, which is the current behavior with rate, rate_min, rate_iops etc. 

If only one option is specificifed it will apply the same value to read, write and trim.
 Two options (with either them possibly left blank using comma - eg max_latency=,10ms or max_latency=10ms,) will apply it to read and write only. Three options will apply the values to read, write and trim in that order. 

This allows us to specify trim as the fourth option (plus include in one option) in the same manner for latency and rate params if we choose to do so in the future.

sitsofe: "This allows us to specify trim" do you mean "This allows us to specify sync"?

venkatrag1: That's correct, sorry about the typo. Corrected the above comment to say sync instead of trim.

sitsofe (approved):
Smooth work @venkatrag1, this LGTM!

@axboe I'd say this one is ready for pulling :-)

venkatrag1:
Hi @sitsofe @axboe , wanted to see if I can roll in one more change as part of this PR. I wanted to log the io_unit information as well from lat_fatal, when latency_exceeded condition occurs so that it become easy to correlate the latency event with blktrace/write_iolog traces.
Screen Shot 2021-02-26 at 7 47 31 PM
Please let me know your thoughts on printing this information in the next line like this.
Screen Shot 2021-02-26 at 7 47 53 PM
vs extending the existing print string.
Thanks in advance for the guidance!

PR #1181, opened 2021-02-08
engine/rados: Remove mandatory creation

aclamk:
Changed connection procedure so it no longer attempts to create objects.
This change will affect tests that read objects without creating them first.

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -2294,6 +2294,12 @@ with the caveat that when used on the command line, they must come after the
         Poll store instead of waiting for completion. Usually this provides better
         throughput at cost of higher(up to 100%) CPU utilization.
 
+.. option:: touch_objects=bool : [rados]
+
+        During initialization, touch (create if do not exist) all objects (files).
+        Touching all objects affects ceph caches and likely impacts test results.
+        Enabled by default.
+
 .. option:: skip_bad=bool : [mtd]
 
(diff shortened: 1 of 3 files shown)


aclamk:
The problem with current implementation is that touching (writing 0 bytes) to each object takes a lot of time, and also messes with ceph caches, inducing huge errors to tests related to cache performance.

ifed01 (approved):
LGTM

sitsofe:
I'd have to think about this but at a minimum could this be turned into an option?

aclamk:
@sitsofe I think that I synchronized rados engine with others.
I do not see codes in other engines that create files on init/setup.

sitsofe:
@aclamk The problem is going to be the change in behaviour. Technically some other ioengines may have creation done "for free" when they aren't diskless (see https://github.com/axboe/fio/blob/master/filesetup.c#L1237 ) so hence the preference for an option...

sitsofe (changes requested):
I'd like to see this as an option...

aclamk:
@sitsofe:
So I will turn it into an option.

sitsofe (approved):
@aclamk Apologies for the long delay and thank you for your patience.

LGTM!

sitsofe:
@axboe can this one be pulled?

PR #1182, opened 2021-02-08
filesetup: add engine's io_ops to prepopulate file with data

lukaszstolarczuk:
In some cases (e.g. engine marked as diskless) files are not laid out.
If the first job is a read job, results are higher than expected
(because reading zero page). Each engine should deliver func to
prepopulate file with data to avoid this situation.

Signed-off-by: Łukasz Stolarczuk <lukasz.stolarczuk@intel.com>

### DIFF
--- a/filesetup.c
+++ b/filesetup.c
@@ -338,6 +338,95 @@ error:
 	return ret;
 }
 
+/*
+ * Generic function to prepopulate regular file with data.
+ * Useful if you want to make sure I/O engine has data to read.
+ * Leaves f->fd open on success, caller must close.
+ */
+int generic_prepopulate_file(struct thread_data *td, struct fio_file *f)
+{
+	int flags;
+	unsigned long long left, bs;
+	char *b = NULL;
+
+	/* generic function for regular files only */
+	assert(f->filetype == FIO_TYPE_FILE);
+
+	if (read_only) {
+		log_err("fio: refusing to write a file due to read-only\n");
+		return 0;
+	}
+
+	flags = O_WRONLY;
+	if (td->o.allow_create)
+		flags |= O_CREAT;
+
+#ifdef WIN32
+	flags |= _O_BINARY;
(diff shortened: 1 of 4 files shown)


axboe:
Looks pretty straight forward, but needs a bump of ```FIO_IOOPS_VERSION```.

filesetup.c
```diff
@@ -338,6 +338,99 @@ static bool pre_read_file(struct thread_data *td, struct fio_file *f)
...
+
+		if (err == ENOENT && !td->o.allow_create)
+			log_err("fio: file creation disallowed by "
+					"allow_file_create=0\n");
+		else
+			td_verror(td, err, "open");
+		return 1;
+	}
+
+	left = f->real_file_size; // XXX: ?? f->file_offset + f->io_size
```
lukaszstolarczuk, line 385: I'm not sure - if I use the offset + io_size, how to apply offset into the write...?

lukaszstolarczuk (commented):
FYI, this is a continuation of #1165.

filesetup.c
axboe: f->real_file_size should be the actual size of the file, f->file_offset is just where IO would be done. So I think you should lay out as f->real_file_size as you do.

lukaszstolarczuk: Ok, I've removed the comment.

lukaszstolarczuk:
@sitsofe, I've removed my check for `fill_device` - I believe it was redundant. I've done some testing and I believe you're right, it may be useful in some cases to run both together - fill_device is done before prepopulation so it should be fine.

sitsofe:
(I'll look at this one over the weekend)

sitsofe (approved):
Got to say I really like this one - LGTM!

sitsofe:
@axboe could you pull this?

axboe:
Agree! Done

PR #1183, opened 2021-02-10
$clientuid keyword to differentiate clients in client/server mode

tarasglek:
We been using fio heavily, but getting it to put an IP into filename was a struggle.

Previously we had to specify `directory=/` in config, which would prepend /<ip> to any filenames. Having explicit format string for this seems better :)

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -809,6 +809,8 @@ Target file/device
 
 		**$jobname**
 				The name of the worker thread or process.
+		**$clientuid**
+				IP of the fio process when using client/server mode.
 		**$jobnum**
 				The incremental number of the worker thread or process.
 		**$filenum**
--- a/fio.1
+++ b/fio.1
@@ -584,6 +584,9 @@ string:
 .B $jobname
 The name of the worker thread or process.
 .TP
+.B $clientuid
+IP of the fio process when using client/server mode.
+.TP
 .B $jobnum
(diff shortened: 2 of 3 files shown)


axboe:
A few comments:

- Does 'client_sockaddr_str work on all platforms? Might need a configure check
- Please update the HOWTO as well
- Commit message should have explanation of the change, and include a Signed-off-by line at the bottom. Commit message is basically just what you put in here. But I want the history in the git tree, not on github

Hope that makes sense!

tarasglek:
Thanks for quick reply. How does this look?

axboe:
Better! But:

- Don't add my signed-off-by, I do that myself when I sign off on it.
- You didn't do the configure test for client_sockaddr_str, and make the keyword depend on that

tarasglek:
@axboe client_sockaddr_str is an fio global variable. It even has a windows codepath to populate it.

axboe:
> @axboe client_sockaddr_str is an fio global variable. It even has a windows codepath to populate it.

Ah duh, no worries then! Just kill my SOB from the commit, and then it looks fine to me.

tarasglek:
killed the SOB

axboe:
LGTM, pulled - thanks.

sitsofe:
(@tarasglek Wow a contribution from another person who I've been an internet fan of! ;-)

PR #1185, opened 2021-02-11
engines/libpmem: minor fixes and code cleanup

lukaszstolarczuk:
- set file open/create mode always to RW,
- do not call drain on close, it's redundant,
- code and example cleanup.

### DIFF
--- a/engines/libpmem.c
+++ b/engines/libpmem.c
@@ -2,7 +2,7 @@
  * libpmem: IO engine that uses PMDK libpmem to read and write data
  *
  * Copyright (C) 2017 Nippon Telegraph and Telephone Corporation.
- * Copyright 2018-2020, Intel Corporation
+ * Copyright 2018-2021, Intel Corporation
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License,
@@ -97,16 +97,9 @@ static int fio_libpmem_file(struct thread_data *td, struct fio_file *f,
 			    size_t length, off_t off)
 {
 	struct fio_libpmem_data *fdd = FILE_ENG_DATA(f);
-	mode_t mode = 0;
+	mode_t mode = S_IWUSR | S_IRUSR;
 	size_t mapped_len;
 	int is_pmem;
-
-	if(td_rw(td))
-		mode = S_IWUSR | S_IRUSR;
-	else if (td_write(td))
-		mode = S_IWUSR;
-	else
-		mode = S_IRUSR;
 
 	dprint(FD_IO, "DEBUG fio_libpmem_file\n");
 	dprint(FD_IO, "f->file_name = %s td->o.verify = %d \n", f->file_name,
--- a/engines/libpmem.c
+++ b/engines/libpmem.c
@@ -212,12 +212,6 @@ static int fio_libpmem_close_file(struct thread_data *td, struct fio_file *f)
 	int ret = 0;
 
 	dprint(FD_IO, "DEBUG fio_libpmem_close_file\n");
-	dprint(FD_IO, "td->o.odirect %d\n", td->o.odirect);
-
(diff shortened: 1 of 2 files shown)


engines/libpmem.c
```diff
@@ -185,8 +175,9 @@ static enum fio_q_status fio_libpmem_queue(struct thread_data *td,
 	io_u->error = 0;
 
 	dprint(FD_IO, "DEBUG fio_libpmem_queue\n");
-	dprint(FD_IO,"td->o.odirect %d td->o.sync_io %d \n",td->o.odirect, td->o.sync_io);
-	/* map both O_SYNC / DSYNC to not using NODRAIN */
+	dprint(FD_IO, "td->o.odirect %d td->o.sync_io %d\n",
+					td->o.odirect, td->o.sync_io);
```
sitsofe, line 179: I think the recent style in fio has been tending towards lining up with the brackets:
```
printf("foo = %d\n",
          i);
```

examples/libpmem.fio
```diff
@@ -17,16 +17,6 @@ thread
...
-#
-# In case of 'scramble_buffers=1', the source buffer
-# is rewritten with a random value every write operations.
-#
-# But when 'scramble_buffers=0' is set, the source buffer isn't
-# rewritten. So it will be likely that the source buffer is in CPU
-# cache and it seems to be high performance.
-#
-scramble_buffers=0
-
```
sitsofe, line 29: What made you ditch this? Maybe add the reason to the description...

engines/libpmem.c
```diff
@@ -2,7 +2,7 @@
  * libpmem: IO engine that uses PMDK libpmem to read and write data
```
sitsofe, line 2: I think the style is that there's an empty line above the S-o-b in the commit message.

engines/libpmem.c
```diff
@@ -212,12 +212,6 @@ static int fio_libpmem_close_file(struct thread_data *td, struct fio_file *f)
 	int ret = 0;
 
 	dprint(FD_IO, "DEBUG fio_libpmem_close_file\n");
-	dprint(FD_IO, "td->o.odirect %d\n", td->o.odirect);
-
-	if (!td->o.odirect) {
-		dprint(FD_IO,"pmem_drain\n");
-		pmem_drain();
-	}
```
sitsofe, line 236: Out of curiosity do know why the original author might have believed it was?

sitsofe (commented):
Only minor comments.

sitsofe (changes requested):
@lukaszstolarczuk any comment on the comments (even if they're minor)?

engines/libpmem.c
lukaszstolarczuk: I'm not sure. In general you want your data to be persistent on the device (on NVDIMM), so maybe it was added just like that. It's not a mistake it is here, it's just seems redundant.
Anyway, I extended the commit msg a little bit, just FYI.

lukaszstolarczuk (commented):
Thanks for the review and ping. Sorry for the delay, I just switched my mind to other tasks and it slipped my mind.

sitsofe (approved):
LGTM!

@axboe can you pull this one?

engines/libpmem.c
ErwanAliasr1: Shouldn't we have pmem_drain() before the unmap() to ensure everything is persistent ?

lukaszstolarczuk: As I stated in the commit message:
- if we set the direct flag to 1, we use "PMEM_F_MEM_NONTEMPORAL", so non-temporal stores, so we don't need to worry about it (we avoid the need to flush processor caches)
- if we set the direct flag to 0, we simply don't care about this (at least in here, in the benchmarking. We rather don't want that in real-life scenarios)

PR #1187, opened 2021-02-12
hack buffer_pattern option to provide large sample file

aclamk:
*This is not actual PR*

Background:
This thing allows to provide large file as source of data for IO operations.
It allowed me to test different compression algorithms and different options.

When testing balance between compression gained and CPU power invested it is necessary to provide actual data
that can be compressed differently by different algorithms.

Data provided in --buffer-pattern is here considered to be a sample of customer data.

Request:
Can I please get some hints how to implement above feature properly, so it could be passed via client-server path?
Or can it be a feature that cannot be used in client - server mode?

### DIFF
--- a/cconv.c
+++ b/cconv.c
@@ -170,7 +170,7 @@ void convert_thread_options_to_cpu(struct thread_options *o,
 	o->verify_offset = le32_to_cpu(top->verify_offset);
 
 	memcpy(o->verify_pattern, top->verify_pattern, MAX_PATTERN_SIZE);
-	memcpy(o->buffer_pattern, top->buffer_pattern, MAX_PATTERN_SIZE);
+	//	memcpy(o->buffer_pattern, top->buffer_pattern, MAX_PATTERN_SIZE);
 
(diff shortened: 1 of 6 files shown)


sitsofe:
I'm going to have to come back to this one as I'm almost out of time for reviews today but I agree client/server mode might complicate things. Is the expectation that there is one buffer used in all writes and said buffer can come from a file?

aclamk:
@sitsofe 
Yes, ONE buffer that provides as data source for all writes is ok.
I mean, for as long as it is not always read from beginning, but works as circular buffer.

sitsofe:
Interesting... I'm going to have to come back to this one to think more on it so I'll warn you it may be a while :-)

aclamk:
@sitsofe 
Would it be acceptable to pass pattern from server to client and let it evaluate it?
But naturally, reading files on different machines will increase probability (by error) of supplying different content.

Other alternative is to directly transfer the data. Maybe it will have significant impact on performance, but if feature is unlikely to be used in server-client mode it will just need to be present for correctness and completeness?

PR #1194, opened 2021-02-24
fix loop with unreachable exit condition

ihsinme:
this is my second try.
this time I'm sure there is an error.

the essence of the error lies in a possible hit in the eternal cycle while, when exiting the higher cycle by means of break.

keeping in mind your requirements, I will immediately ask for help in describing the functionality of the place I am correcting, so that I can create tests and poc. with less effort.
thank

Signed-off-by: Sergey Valentey <ihsinme@gmail.com>

### DIFF
--- a/server.c
+++ b/server.c
@@ -1909,7 +1909,7 @@ static int fio_append_iolog_gz(struct sk_entry *first, struct io_log *log)
 			break;
 		}
 		flist_add_tail(&entry->list, &first->next);
-	} while (ret != Z_STREAM_END);
+	}
 
 	ret = deflateEnd(&stream);
 	if (ret == Z_OK)


ihsinme:
Do you need more information for this PR?

sitsofe:
@ihsinme Thanks for dropping by. I haven't had a chance to look at this as I currently have my hands full with `${DAYJOB}` but it will get looked at in the end ;-)

axboe:
The change itself looks fine, but the commit message is unreadable. I'd rework that:

- Improve the wording of it so it makes sense without having to look at the code change :-)
- Break up overly long lines. A commit message should be no wider than 72-74 chars.

ihsinme:
if I fix the commit message like this?

```
The presence of a second while cycle may result in an error.
I suggest removing it.
```
at the same time, please clarify whether it is necessary to additionally correct the name of the commit and the PR message?

ihsinme:
@axboe please respond to my suggestion for a fix.

PR #1198, opened 2021-03-04
Unified RW Reporting - Separate and Mixed

bjpaupor:
Add functionality to the unified_rw_reporting parameter to output separate and mixed stats when set to 2.

To achieve this, the stats are gathered on a per/data direction (read/write/etc) basis and then summed into the mixed stats when generating output.


jeffreyalien (approved):

axboe:
Please squash this down to a clean series, looks like a bit of a work-in-progress so far. Don't want to merge (or review :-)) that.

bjpaupor:
Will do, thanks for the update!

PR #1200, opened 2021-03-09
engines/io_uring: fix compilation conflict with Android NDK

DevriesL:
This rename variable __errno to init_err, because Android NDK already uses __errno in errno.h, then it will throw a compilation error `called object type 'int' is not a function or function pointer` if we declare a variable as same name.


axboe:
Thanks!

PR #1201, opened 2021-03-10
options: allow separate values for max_latency

venkatrag1:
Adding the ability to specify max_latency limits separately for read, write, trim.
Uses comma-separated values similar to blocksize, rate parameters etc.

Also modifies the behavior for the one option only case for max_latency and latency_target
by excluding syncs. This makes the latency limits similar to the rate checks, in that they
only apply to reads, writes and trims.

Extends the output printed on latency exceeded event, to display io_unit information using format
similar to that iolog write.

Signed-off-by: Venkat Ramesh <venkatraghavan@fb.com>

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -2663,11 +2663,12 @@ I/O latency
 	true, fio will continue running and try to meet :option:`latency_target`
 	by adjusting queue depth.
 
-.. option:: max_latency=time
+.. option:: max_latency=time[,time][,time]
 
 	If set, fio will exit the job with an ETIMEDOUT error if it exceeds this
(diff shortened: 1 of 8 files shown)


venkatrag1:
Hi @sitsofe @axboe ,

This pull request is the same as the previously approved PR (https://github.com/axboe/fio/pull/1180).

The only addition is that I ramped up the FIO_SERVER_VER to 89 since I noticed that a87c90f has made it into fio-3.26 release and this would be the first modification to thread_options since that release.

I ended up polluting my PR while trying to rebase to upstream and squash my commits, so ended up closing that PR and generating this one afresh. Thanks a lot!

PR #1210, opened 2021-03-18
Option unified-rw-reporting=2, Mixed and Split Statistics Output

bjpaupor:
Add functionality to the unified_rw_reporting parameter to output separate and mixed stats when set to 2.

To achieve this, the stats are gathered on a per/data direction (read/write/etc) basis and then summed into the mixed stats when generating output.


stat.c
```diff
@@ -1080,9 +1285,12 @@ static void show_thread_status_normal(struct thread_stat *ts,
 
 	for_each_rw_ddir(ddir) {
 		if (ts->io_bytes[ddir])
-			show_ddir_status(rs, ts, ddir, out);
+		show_ddir_status(rs, ts, ddir, out);
```
axboe, line 1288: indentation?

stat.c
```diff
@@ -1205,7 +1413,7 @@ static void show_ddir_status_terse(struct thread_stat *ts,
 					&minv);
 	else
 		len = 0;
-
+
```
axboe, line 1416: Whitespace

stat.c
```diff
@@ -282,6 +282,45 @@ bool calc_lat(struct io_stat *is, unsigned long long *min,
 	return true;
 }
 
+void show_mixed_group_stats(struct group_run_stats *rs, struct buf_output *out) {
```
axboe, line 285: Brace placement

stat.c
```diff
@@ -306,7 +345,7 @@ void show_group_stats(struct group_run_stats *rs, struct buf_output *out)
 		max = num2str(rs->max_bw[i], rs->sig_figs, 1, i2p, rs->unit_base);
 		maxalt = num2str(rs->max_bw[i], rs->sig_figs, 1, !i2p, rs->unit_base);
 		log_buf(out, "%s: bw=%s (%s), %s-%s (%s-%s), io=%s (%s), run=%llu-%llumsec\n",
-				rs->unified_rw_rep ? "  MIXED" : str[i],
+				(rs->unified_rw_rep == 1) ? "  MIXED" : str[i],
```
axboe, line 348: Not specific to this location, but in general - do we want to make these (now) three states specific by using some kind of enum/deifne for them? == 1 or == 2 means nothing when you are reading the code, but symbolics would help with that.

Might even then make sense to turn the option into a posval instead, and use the same values there.

bjpaupor:
Howdy @axboe, I've squashed those commits per your request if you'd like to review this.

bjpaupor:
@axboe I agreed with and made the changes you suggested, going with creating the defines for each option here.

axboe:
Looks better. The commit message needs a rewrite though, it's a bit of a mess right now.

And it's not using posval yet, seems like it should. See eg the fallocate option. It's quite a common way to do it, if an option goes from 0/1 to all of a sudden having more than 2 options. It retains compatibility and has a nicer interface for the user. For this one, the options would be 0 (and none), 1, and then 'both' for the new mode.

bjpaupor:
Ah, the man page is a bit messed up after that last push. I'll fix that shortly.

bjpaupor:
@axboe Changes should be fixed now, it seems like the posval option is working as intended.

axboe:
Thanks!

PR #1219, opened 2021-04-20
Fix `filesize` description

Nikratio:
Setting `size` does not actually limit the total size of files created.

Fixes: axboe#1218

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -1836,11 +1836,12 @@ I/O size
 
 .. option:: filesize=irange(int)
 
-	Individual file sizes. May be a range, in which case fio will select sizes
-	for files at random within the given range and limited to :option:`size` in
-	total (if that is given). If not given, each created file is the same size.
-	This option overrides :option:`size` in terms of file size, which means
-	this value is used as a fixed size or possible range of each file.
+	Individual file sizes. May be a range, in which case fio will select sizes for
+	files at random within the given range. If not given, each created file is the
+	same size. This option overrides :option:`size` in terms of file size, i.e. if
+	:option:`filesize` is specified then :option:`size` becomes merely the default
+	for :option:`io_size` and has no effect at all if :option:`io_size` is set
+	explicitly.
 
 .. option:: file_append=bool
 
--- a/fio.1
+++ b/fio.1
@@ -1625,10 +1625,10 @@ In this case \fBio_size\fR multiplies \fBsize\fR= value.
 .TP
 .BI filesize \fR=\fPirange(int)
 Individual file sizes. May be a range, in which case fio will select sizes
-for files at random within the given range and limited to \fBsize\fR in
-total (if that is given). If not given, each created file is the same size.
-This option overrides \fBsize\fR in terms of file size, which means
-this value is used as a fixed size or possible range of each file.
+for files at random within the given range. If not given, each created file
+is the same size. This option overrides \fBsize\fR in terms of file size, 
+i.e. \fBsize\fR becomes merely the default for \fBio_size\fR (and
+has no effect it all if \fBio_size\fR is set explicitly).
 .TP
 .BI file_append \fR=\fPbool
 Perform I/O after the end of the file. Normally fio will operate within the


sitsofe (changes requested):
So I think this behaviour stems from https://github.com/axboe/fio/blob/6202c70d8d5cbdd3fb4bc23b96f691cbd25a327e/filesetup.c#L1216-L1229 . We would have to confirm `f->real_file_size` doesn't shrink but if it doesn't then I think your change is correct. You would also have to update the HOWTO (which has the text duplicated in a different format).

sitsofe:
@Nikratio overall looks it's the right direction. In addition to comments elsewhere:

- Can you check the code is doing what we now suspect?
- I would argue these commits could be squashed into one and some general header given
- Can you add a prefix to the commit titles (see https://github.com/axboe/fio/commits/e7e536b665bd6a9d3e936e0847dbbb6957101da4/fio.1 for ideas)?
- Can you add your Signed-off-by ?

Nikratio:
Thanks for the review! I can't justify spending time on understanding the implementation. I just noticed that the command was not behaving as documented and wanted to report this.

I have made the same change to `HOWTO` and added a Signed-Off-By header, prefix, and general description in the top commit. This should indeed be squashed into one - if you like the patch, please merge with "Squash & Rebase" and drop all commit descriptions other than the last one. Feel free to adjust the description as well if I didn't get it quite right.

sitsofe:
I did a search and it looks like the "filesize limited to size" behaviour was introduced in https://github.com/axboe/fio/commit/9c60ce649d3f976d196709d4399bb7c540ce97b5 but was later removed as part of https://github.com/axboe/fio/commit/7bb48f84ac78cac1f90e3e04d0220d90d6a64a6b#diff-c7128ba06f81c567584dfc2b3d851d0b7124c8c73f52d70be6bc3bce4723b43eL133 (but you'll have to click "Load diff" to see line 133).

@axboe do you know if this was intentional?

HOWTO
```diff
@@ -1837,11 +1837,11 @@ I/O size
...
 	Individual file sizes. May be a range, in which case fio will select sizes
-	for files at random within the given range and limited to :option:`size` in
-	total (if that is given). If not given, each created file is the same size.
-	This option overrides :option:`size` in terms of file size, which means
-	this value is used as a fixed size or possible range of each file.
-
+        for files at random within the given range. If not given, each created file
+        is the same size. This option overrides :option:`size` in terms of file size, 
+        i.e. :option:`size` becomes merely the default for :option:`io_size` (and
+        has no effect it all if :option:`io_size` is set explicitly).
```
bardavid, line 1843: Typo: has no effect if :option..

Nikratio: Fixed, thanks.

sitsofe (approved):
LGTM - @axboe can you pull this one?

vincentkfu:
This looks good to me. I ran some simple tests and the outcomes were consistent with the relationship among `size`, `filesize`, and `io_size` described here.

`size` by itself sets `io_size` as well as `filesize`:

```
root@ubuntu:~/fio-dev/fio-canonical# ./fio --name=test --size=512K
...
test: Laying out IO file (1 file / 0MiB)
...
  read: IOPS=25.6k, BW=100MiB/s (105MB/s)(512KiB/5msec)
...
   READ: bw=100MiB/s (105MB/s), 100MiB/s-100MiB/s (105MB/s-105MB/s), io=512KiB (524kB), run=5-5msec
...
root@ubuntu:~/fio-dev/fio-canonical# ls -l test.0.0
-rw-r--r-- 1 root root 524288 Jan 28 20:39 test.0.0
```


`size` only determines `io_size` when `filesize` is also specified:
```
root@ubuntu:~/fio-dev/fio-canonical# ./fio --name=test --filesize=1M --size=512K
...
test: Laying out IO file (1 file / 1MiB)
...
  read: IOPS=21.3k, BW=83.3MiB/s (87.4MB/s)(512KiB/6msec)

...
   READ: bw=83.3MiB/s (87.4MB/s), 83.3MiB/s-83.3MiB/s (87.4MB/s-87.4MB/s), io=512KiB (524kB), run=6-6msec
root@ubuntu:~/fio-dev/fio-canonical# ls -l test.0.0
-rw-r--r-- 1 root root 1048576 Jan 28 20:39 test.0.0
```


`size` is ignored when both `filesize` and `io_size` are specified:
```
root@ubuntu:~/fio-dev/fio-canonical# ./fio --name=test --filesize=1M --size=512K --io_size=1M
...
  read: IOPS=32.0k, BW=125MiB/s (131MB/s)(1024KiB/8msec)
...
   READ: bw=125MiB/s (131MB/s), 125MiB/s-125MiB/s (131MB/s-131MB/s), io=1024KiB (1049kB), run=8-8msec
root@ubuntu:~/fio-dev/fio-canonical# ls -l test.0.0
-rw-r--r-- 1 root root 1048576 Jan 28 20:39 test.0.0
```

PR #1222, opened 2021-04-24
gettime: Fix compilation on non-Linux with pthread_getaffinity_np()

kusumi:
874d55e50c("os/os-linux: add pthread CPU affinity helper") and a few
commits after that broke compilation on non-Linux platforms which support
`pthread_getaffinity_np()`.

If the C code is to test `CONFIG_PTHREAD_GETAFFINITY`, it should just
directly use `pthread_getaffinity_np()` instead of definining a wrapper
that is not specific to pthread.

### DIFF
--- a/gettime.c
+++ b/gettime.c
@@ -679,7 +679,7 @@ int fio_monotonic_clocktest(int debug)
 	unsigned int i;
 	os_cpu_mask_t mask;
 
-#ifdef CONFIG_PTHREAD_GETAFFINITY
+#ifdef FIO_HAVE_GET_THREAD_AFFINITY
 	fio_get_thread_affinity(mask);
 #else
 	memset(&mask, 0, sizeof(mask));
--- a/os/os-aix.h
+++ b/os/os-aix.h
@@ -18,6 +18,12 @@
 
 #define FIO_USE_GENERIC_SWAP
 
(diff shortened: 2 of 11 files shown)


axboe:
The point of the wrapper is to abstract out that from the OS side of things, like how masks etc are handled in general. Your change violates that, and then mixes fio types with OS types. How does it currently break on non-linux?

kusumi:
> The point of the wrapper is to abstract out that from the OS side of things, like how masks etc are handled in general. Your change violates that, and then mixes fio types with OS types.

Then the code should be testing `FIO_HAVE_GET_THREAD_GETAFFINITY` with `fio_get_thread_affinity()` defined in all os/os-xxx.h headers.
I could do that too.

> How does it currently break on non-linux?

It's simply that only `os/os-linux.h` has `fio_get_thread_affinity()`.

axboe:
> It's simply that only os/os-linux.h has fio_get_thread_affinity().

Ah yes of course, that's silly...

The change you suggest is the right approach, would be great if you could do that.

kusumi:
Force pushed v2 patch.
I didn't define `fio_get_thread_affinity()` in `os/os.h`, but rather defined a copy in each header.
Windows may want to do this via Win32.

axboe:
Looks good, thanks.

PR #1229, opened 2021-05-12
fix fio2gnuplot to work with new logging format

larsks:
The logging format updates documented in 1a953d97 were never
propagated to fio2gnuplot, which since then has been failing with a
ValueError exception.

This commit explicits limits fio2gnuplot to only reading the first
4 columns in the log file.

### DIFF
--- a/tools/plot/fio2gnuplot
+++ b/tools/plot/fio2gnuplot
@@ -198,7 +198,7 @@ def compute_temp_file(fio_data_file,disk_perf,gnuplot_output_dir, min_time, max_
 			# Index will be used to remember what file was featuring what value
 			index=index+1
 
-			time, perf, x, block_size = line[1]
+			time, perf, x, block_size = line[1][:4]
 			if (blk_size == 0):
 				try:
 					blk_size=int(block_size)


tools/plot/fio2gnuplot
```diff
@@ -198,7 +198,7 @@ def compute_temp_file(fio_data_file,disk_perf,gnuplot_output_dir, min_time, max_
 			# Index will be used to remember what file was featuring what value
 			index=index+1
 
-			time, perf, x, block_size = line[1]
+			time, perf, x, block_size = line[1][:4]
```
ErwanAliasr1, line 201: Another option was to do time, perf, x, block_size,_

ErwanAliasr1 (commented):
Looks good to me

tools/plot/fio2gnuplot
larsks: @ErwanAliasr1 that will fail if any additional columns get added, which is why I went with this option.

ErwanAliasr1: fair enough ;)

ErwanAliasr1 (approved):

ErwanAliasr1:
@axboe this one is fine too, it fixes some code I wrote. I had this fix too to push and never did it.

PR #1232, opened 2021-05-25
android: add support for NDK sharedmem

DevriesL:
Android add support for NDK sharedmem since API level 26 and prohibit the directly use of ashmem since API level 29, so we can use sharedmem if targeting API level is higher than 26.

Please refer to the Android official document: [Behavior changes: apps targeting API 29+](https://developer.android.com/about/versions/10/behavior-changes-10#shared-memory)

### DIFF
--- a/os/os-android.h
+++ b/os/os-android.h
@@ -71,10 +71,14 @@
 #include <stdio.h>
 #include <linux/ashmem.h>
 #include <linux/shm.h>
+#include <android/api-level.h>
+#if __ANDROID_API__ >= __ANDROID_API_O__
+#include <android/sharedmem.h>
+#else
+#define ASHMEM_DEVICE	"/dev/ashmem"
+#endif
 #define shmid_ds shmid64_ds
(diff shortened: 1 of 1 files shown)


os/os-android.h
```diff
@@ -89,6 +93,19 @@ static inline int shmctl(int __shmid, int __cmd, struct shmid_ds *__buf)
...
 
+#if __ANDROID_API__ >= __ANDROID_API_O__
+static inline int shmget(key_t __key, size_t __size, int __shmflg)
+{
+	int fd;
+	char keybuf[11];
+
+	sprintf(keybuf, "%d", __key);
+
+	fd = ASharedMemory_create(keybuf, __size + sizeof(uint64_t));
```
axboe, line 104: Probably just make that:

```
return ASharedMemory_create(keybuf, __size + sizeof(uint64_t));
```

and get rid of 'fd'.

axboe:
Thanks!

PR #1237, opened 2021-06-09
configure: silence "pkg-config: not found"

kusumi:
Do what check_min_lib_version() does to not print the error message
when pkg-config doesn't exist.

```
DAOS File System (dfs) Engine no
./configure: pkg-config: not found
NFS engine                    no
```


sitsofe (approved):
LGTM. @axboe can we pull this one?

PR #1240, opened 2021-06-27
tools: Adding fiograph

ErwanAliasr1:
This tool generates graphviz graphs of any fio job file.
    
    Reading a fio file can be sometimes tricky when :
      - you have lots of jobs
      - jobs override some options
      - jobs have dependencies
    
    Understanding which jobs are running at the same time, which one waits
    for which is sometimes difficult and could easily lead to mistakes.
    It's pretty common to find jobs where people think jobs are sequential
    while some run in parallel.
    
    For ease the understanding of the scheduling and what's the
    configuration status (local + global variables), this tool will use some
    graphical helpers to enlighten things.
    
    The configuration file can be used to adjust the rendering but by default :
      - jobs are defined in a blue rounded box
      - jobs running at the same time are inside the same gray background
      - time & size variables are shown as a self arrow
      - dependencies between jobs are shown by up->down black arrows
      - specific engines options are written in blue (top of the job box)
      - standard options are printed in green (bottom of the job box)
      - pre/post actions are printed in red (head/tail of the job box)
      - numjobs is replaced by a 'x' multiplicator on the job name
    
    Several output formats can be defined via --format : list can be found
    here https://graphviz.org/docs/outputs/
    
    if --view option is used, the rendered file will be immediately shown.
    
    if --keep is used, the grapvhiz file will be kept.
    
    This commit also adds the rendering of the examples jobs.
    If a newcomer gets into the repository, they can immediately better
    understand what the jobs do.


axboe:
This is nifty! Thanks, pulled.

PR #1252, opened 2021-07-25
engines/exec: Code cleanup to remove leaks

ErwanAliasr1:
As per the coverty reports, there was some issues in my code :
- Some structures were not properly freed before returning.
- Some file descriptors were not properly closed
- Testing with 'if (!int)' isn't a good way to test if the value is negative

Signed-off-by: Erwan Velu <erwanaliasr1@gmail.com>


axboe:
Thanks!

Issue #1254, opened 2021-07-26
Undefined behavior due to the double releases on the lock td->io_u_lock?


anson-lo:
Dear developers, thank you very much for your checking! It seems there is an error in the method verify_async_thread. The lock may be unlocked twice that may lead to undefined behavior. Is that true?

https://github.com/axboe/fio/blob/6202c70d8d5cbdd3fb4bc23b96f691cbd25a327e/verify.c#L1407
https://github.com/axboe/fio/blob/6202c70d8d5cbdd3fb4bc23b96f691cbd25a327e/verify.c#L1414
https://github.com/axboe/fio/blob/6202c70d8d5cbdd3fb4bc23b96f691cbd25a327e/verify.c#L1420

```
static void *verify_async_thread(void *data)
{
	...;
		pthread_mutex_lock(&td->io_u_lock);

		while (flist_empty(&td->verify_list) &&
		       !td->verify_thread_exit) {
			ret = pthread_cond_wait(&td->verify_cond,
							&td->io_u_lock);
			if (ret) {
				pthread_mutex_unlock(&td->io_u_lock);  // release the lock and break
				break;
			}
		}

		flist_splice_init(&td->verify_list, &list);
		pthread_mutex_unlock(&td->io_u_lock); // release the lock again

		...;
}
```

Best,

sitsofe:
@anson-lo Hi,

This does indeed look like an error. Would you like to submit a pull request to fix it?

anson-lo:
@sitsofe Sure, I have created a PR, and please check it. Also, can you look at another issue #1253 I proposed?

PR #1255, opened 2021-07-31
Fix an error triggered by double releasing the lock

anson-lo:
Fix an error triggered by double releasing the lock. The patch is for issue#1254.

Signed-off-by: Anson Lo ycaibb@gmail.com

### DIFF
--- a/verify.c
+++ b/verify.c
@@ -1411,7 +1411,6 @@ static void *verify_async_thread(void *data)
 			ret = pthread_cond_wait(&td->verify_cond,
 							&td->io_u_lock);
 			if (ret) {
-				pthread_mutex_unlock(&td->io_u_lock);
 				break;
 			}
 		}


axboe:
Patch looks good, but the Signed-off-by goes in the commit message, not on github. The github interface is just for ease of use, full history/explanation/sign-off goes into the actual git commit (and hence git tree). Can you amend the commit and include that?

axboe:
And if you're doing that anyway, instead of writing it in the text, just add this line before your signed-off-by line:

Fixes: https://github.com/axboe/fio/issues/1254

and it'll reference it.

axboe:
That should be:

```
Fixes: https://github.com/axboe/fio/issues/1254
```

github turned that into a link...

anson-lo:
@anson-lo I created a new pull request, and please check it. Thanks!

PR #1256, opened 2021-08-01
Fix an error triggered by double releasing the lock

anson-lo:
Fix an error triggered by double releasing the lock. 
Fixes: https://github.com/axboe/fio/issues/1254.

Signed-off-by: Anson Lo ycaibb@gmail.com


axboe:
Looks good, pulled, thanks!

PR #1260, opened 2021-08-10
engines/dfs: add support for 1.3 DAOS API

johannlombardi:
A few changes were done to the pool connect and container open API
in DAOS 1.3+. UUID string or label are now passed via the API
instead of uuid_t structures. Change the dfs engine accordingly.

Signed-off-by: Johann Lombardi <johann.lombardi@intel.com>

### DIFF
--- a/HOWTO
+++ b/HOWTO
@@ -2561,11 +2561,11 @@ with the caveat that when used on the command line, they must come after the
 
 .. option:: pool=str : [dfs]
 
-	Specify the UUID of the DAOS pool to connect to.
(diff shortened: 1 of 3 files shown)


johannlombardi:
@axboe @sitsofe I think that the patch is ready for your consideration. Thanks in advance.

sitsofe:
The defines don't make this pretty but the changes look sensible.

@axboe looks pullable.

axboe:
Agree, let's do it.

johannlombardi:
Thanks for merging. I will clean up the defines when support for those older versions is dropped.

PR #1261, opened 2021-08-11
lib/seqlock: implement seqlock with C++ atomic if compiled with C++

tchaikov:
because some functions declared by <stdatomic.h> share the same names
with those declared by <atomic>, for instance `kill_dependency()` is
defined as a macro by <stdatomic.h>, while it is defined as a template
function in <atomic>.

this renders it impossible to compile an ioengine written in C++ if
its source file includes both <atomic> and <fio.h>. the latter includes
<stdatomic.h> via arch/arch.h. the compile error would look like:

In file included from ../src/test/fio/fio_ceph_objectstore.cc:26:
In file included from src/fio/fio.h:18:
In file included from src/fio/thread_options.h:8:
In file included from src/fio/gettime.h:7:
src/fio/lib/seqlock.h:21:9: error: expected ')'
                seq = atomic_load_acquire(&s->sequence);
                      ^
src/fio/arch/../lib/../arch/arch.h:47:32: note: expanded from macro 'atomic_load_acquire'
        atomic_load_explicit((_Atomic typeof(*(p)) *)(p),       \
                                      ^
src/fio/lib/seqlock.h:21:9: note: to match this '('

to address this issue, instead of using the functions in <stdatomic.h> to
implement seqlock, use the primitives offered by C++ standard library
if the source code is compiled using a C++ compiler.

Signed-off-by: Kefu Chai <tchaikov@gmail.com>


tchaikov:
@sitsofe @axboe hi, Sitsofe and Jens, could you please help review this change?

sitsofe:
@tchaikov wouldn't it be better to redefine things like atomic_add/atomic_sub etc. when compiling as C++? Then all the magic is confined to the same place...

tchaikov:
@sitsofe i updated the patch accordingly. but the type of `sequence` in `seqlock.h` is conditionalized on `defined(__cplusplus)`,  so i don't need to cast it to `std::atomic<unsigned int>` in `arch.h` to appease the argument-dependent lookup of the C++ compiler. but if you think it'd be better to to drop the change in `seqlock.h` completely. i will be happy to do so.

PR #1263, opened 2021-08-18
server: reopen standard streams to /dev/null

antroseco:
For some custom ioengines (see https://github.com/spdk/spdk/issues/1118), it's not trivial to suppress output to stdout and stderr, so they would write to some other file descriptor fio had opened - which is bad.
This change ensures that fd's 0, 1, and 2 (stdin, stdout, stderr) are always valid and can be used without any unintended consequences.

### DIFF
--- a/server.c
+++ b/server.c
@@ -2565,6 +2565,7 @@ static int write_pid(pid_t pid, const char *pidfile)
  */
 int fio_start_server(char *pidfile)
 {
+	FILE *file;
 	pid_t pid;
 	int ret;
 
@@ -2597,13 +2598,27 @@ int fio_start_server(char *pidfile)
 	setsid();
 	openlog("fio", LOG_NDELAY|LOG_NOWAIT|LOG_PID, LOG_USER);
 	log_syslog = true;
-	close(STDIN_FILENO);
-	close(STDOUT_FILENO);
-	close(STDERR_FILENO);
+
+	file = freopen("/dev/null", "r", stdin);
+	if (!file)
+		perror("freopen");
+
+	file = freopen("/dev/null", "w", stdout);
+	if (!file)
+		perror("freopen");
+
+	file = freopen("/dev/null", "w", stderr);
+	if (!file)
+		perror("freopen");
+
 	f_out = NULL;
 	f_err = NULL;
 
 	ret = fio_server();
+
+	fclose(stdin);
+	fclose(stdout);
+	fclose(stderr);
 
 	closelog();
 	unlink(pidfile);


axboe:
I guess that's fine. But can you re-format your commit message

1) Please keep it at 74 chars width at most
2) It needs a Signed-off-by at the end with your name and email

axboe:
Much better, doesn't hurt the eyes anymore :-)

axboe:
One more thing:

```
server.c: In function ‘fio_start_server’:
server.c:2600:9: warning: ignoring return value of ‘freopen’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
 2600 |         freopen("/dev/null", "r", stdin);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
server.c:2601:9: warning: ignoring return value of ‘freopen’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
 2601 |         freopen("/dev/null", "w", stdout);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
server.c:2602:9: warning: ignoring return value of ‘freopen’ declared with attribute ‘warn_unused_result’ [-Wunused-result]
 2602 |         freopen("/dev/null", "w", stderr);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

we need to do something about the error return, even if we don't really care about them here.

axboe:
Probably just something ala:

```
diff --git a/server.c b/server.c
index bbe5fb0c58cb..a5b893d29335 100644
--- a/server.c
+++ b/server.c
@@ -2565,6 +2565,7 @@ static int write_pid(pid_t pid, const char *pidfile)
  */
 int fio_start_server(char *pidfile)
 {
+	FILE *file;
 	pid_t pid;
 	int ret;
 
@@ -2597,9 +2598,15 @@ int fio_start_server(char *pidfile)
 	setsid();
 	openlog("fio", LOG_NDELAY|LOG_NOWAIT|LOG_PID, LOG_USER);
 	log_syslog = true;
-	freopen("/dev/null", "r", stdin);
-	freopen("/dev/null", "w", stdout);
-	freopen("/dev/null", "w", stderr);
+	file = freopen("/dev/null", "r", stdin);
+	if (!file)
+		perror("freopen");
+	file = freopen("/dev/null", "w", stdout);
+	if (!file)
+		perror("freopen");
+	file = freopen("/dev/null", "w", stderr);
+	if (!file)
+		perror("freopen");
 	f_out = NULL;
 	f_err = NULL;
  
```

antroseco:
Ugh sorry about that - fixed.

axboe:
Do we need to close them after fio_server() returns? Guess it doesn't really matter as we exit after that.

antroseco:
I suppose it is more correct. Might as well then.

axboe:
Perfect, thanks!

PR #1270, opened 2021-09-08
Code cleanup

ErwanAliasr1:
This PR was done after running scan-build to detect useless variable assignments and a null pointer issue.

### DIFF
--- a/engines/sg.c
+++ b/engines/sg.c
@@ -471,10 +471,9 @@ static enum fio_q_status fio_sgio_rw_doio(struct thread_data *td,
 			if (__io_u == io_u)
 				break;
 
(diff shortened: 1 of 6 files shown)


engines/sg.c
```diff
@@ -472,7 +472,6 @@ static enum fio_q_status fio_sgio_rw_doio(struct thread_data *td,
 				break;
 
 			if (io_u_sync_complete(td, __io_u)) {
-				ret = -1;
```
axboe, line 475: First I thought this might be a legit bug, but actually looks fine. You still need to remove the braces, though.

ErwanAliasr1: Just upload a new version with the braces.

axboe:
Thanks!

PR #1271, opened 2021-09-16
t/io_uring: Reporting bandwidth

ErwanAliasr1:
When performing tests at various block size, it's sometimes a bit
difficult to estimate if we reach the limit of the datapath.

This commit offer to simply prints the resulting bandwitdh of the IOPS
multiplied by the block size.

A typical output looks like :

  [user@hosŧ] t/io_uring -b512 -d128 -c32 -s32 -p1 -F1 -B1 -n4 /dev/nvme0n1
   ...
   IOPS=729856, BWPS=356 MB IOS/call=32/32, inflight=(105 119 108 109)

  [user@host] t/io_uring -b4096 -d128 -c32 -s32 -p1 -F1 -B1 -n4 /dev/nvme0n1
   ...
   IOPS=746368, BWPS=2915 MB IOS/call=32/31, inflight=(121 115 122 122)

In the 4K case, as for a PCI Gen3 product, we are clearly limited by the
bandwidth while in the 512 case we hit latency issues.

BWPS is expressed in MB/sec.


t/io_uring.c
```diff
@@ -740,8 +740,8 @@ int main(int argc, char *argv[])
 		} else
 			rpc = ipc = -1;
 		file_depths(fdepths);
-		printf("IOPS=%lu, IOS/call=%ld/%ld, inflight=(%s)\n",
-				this_done - done, rpc, ipc, fdepths);
+		printf("IOPS=%lu, BWPS=%lu MB, IOS/call=%ld/%ld, inflight=(%s)\n",
+				this_done - done, (this_done - done) * bs / 1048576, rpc, ipc, fdepths);
```
axboe, line 744: I'd change that so it reads:

BW=2915MiB/s

to make it really clear. Probably add a variable that does holds iops, ala:

iops = this_done - done;

and then use that for both IOPS and BW numbers.

ErwanAliasr1: I found odd to have IOPS=x and BW=y/sec while PS meant /sec ;)
That's why I tried to get a similar report IOPS/BWPS.
Ack for MiB.

axboe: IOPS is pretty commonly used and understood, BWPS is not ;-)

ErwanAliasr1: thought to same but tried ;)
Fixed.

PR #1272, opened 2021-09-16
t/one-core.sh: Adding script to run the one-core io benchmark

ErwanAliasr1:
Associated to fio, the t/io_uring test is used to compute the max IOPS a
single core can get.

Jens published several times the procedure he uses, but trying to
reproduce this setup is error-prone. It's easy to miss a configuration
and get a different result.

This script is about setting up a common setup to reproduce these runs.

From the fio directory, execute like the folliowing :

	[user@fio] t/one-core.sh /dev/nvme0n1 [other drives]
	system: Checking configuration
	Warning: For better performance, you should enable nvme poll queues by setting nvme.poll_queues=32 on the kernel commande line
	nvme0n1: set none as io scheduler
	nvme0n1: iostats set to 1. Disabling.
	nvme0n1: nomerge set to 0. Disabling.
	system: End of system checks
	io_uring: Running taskset -c 0,12 t/io_uring -b512 -d128 -c32 -s32 -p1 -F1 -B1 -n4 /dev/nvme0n1
        [...]
	IOPS=731008, BWPS=356 MB IOS/call=32/31, inflight=(108 127 126 106)

This script will take care of the following items:
- nvme poll queues
- io scheduler
- iostats
- nomerge
- finding the logical cores running on the first physical core
- calling t/io_uring with the proper parameters in 512 bytes fashion

Signed-off-by: Erwan Velu <erwanaliasr1@gmail.com>

### DIFF
--- a/t/one-core-peak.sh
+++ b/t/one-core-peak.sh
@@ -1,0 +1,199 @@
+#!/bin/bash
+
+args=$*
+first_cores=""
+taskset_cores=""
+first_cores_count=0
+nb_threads=4 #default from the benchmark
+
+fatal() {
+  echo "$@"
+  exit 1
+}
+
+hint() {
+  echo "Warning: $*"
+}
+
+info() {
+  item=$1
+  shift
+  echo "${item}: $*"
+}
+
+check_root() {
+  [[ ${EUID} -eq 0 ]] || fatal "You should be root to run this tool"
+}
+
+check_binary() {
+  # Ensure the binaries are present and executable
+  for bin in "$@"; do
+    if [ ! -x ${bin} ]; then
+      which ${bin} >/dev/null
+      [ $? -eq 0 ] || fatal "${bin} doesn't exists or is not executable"
+    fi
+  done
+}
+
+
+detect_first_core() {
+  # Detect which logical cpus belongs to the first physical core
+  # If Hyperthreading is enabled, two cores are returned
+  cpus=$(lscpu  --all -pSOCKET,CORE,CPU |grep "0,0")
+  for cpu in ${cpus}; do
+    IFS=','
+    # shellcheck disable=SC2206
+    array=(${cpu})
+    if [ ${first_cores_count} -eq 0 ]; then
+      first_cores="${array[2]}"
+    else
+      first_cores="${first_cores} ${array[2]}"
+    fi
+
+    first_cores_count=$((first_cores_count + 1))
+    unset IFS
+  done
+  [ ${first_cores_count} -eq 0 ] && fatal "Cannot detect first core"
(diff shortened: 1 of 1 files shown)


ByteHamster:
Thanks, I think that will help people who also try to reproduce the measurements. Some thoughts:

- How about also setting `/sys/block/nvmeX/queue/io_poll` to `1`? I don't remember if that got set to `1` because of the boot parameter or if I did that manually.
- How about setting the cpu scheduler to `performance`? `echo "performance" | tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor`?
- When iterating over the cpus, how about also making sure that they are available in `/sys/devices/system/node/nodeX/cpuY`, where X=`/sys/block/nvmeZ/device/numa_node`? Unfortunately, I can't test this currently because the machine is booted with Numa disabled.
- How about naming the file something like `t/io_uring_one_core.sh`? Makes it more clear that it belongs to `t/io_uring` but could also be potentially annoying because it will always show up in tab completion.
- Maybe also a little note could help? "You are running Kernel `uname -r`. Make sure to use the latest version for best results." I mean, it won't directly set anything but it could sensitize users of the script.
- Not really clean but `lspci -vv | grep -P "[0-9a-f]{2}:[0-9a-f]{2}\.[0-9a-f]|LnkSta:|LnkCap:" | grep -A 2 -i -P "(nvme|nvm express)"` would try to display the used and maximum PCI lanes of each SSD, for sensitizing users.

sitsofe (commented):
@ErwanAliasr1 Could you push this script through Shellcheck and see what you think of its comments?

ErwanAliasr1:
That was a first run to share the idea. I will improve it

Le ven. 17 sept. 2021 à 11:10, Sitsofe Wheeler ***@***.***> a
écrit :

ErwanAliasr1:
> @ErwanAliasr1 Could you push this script through Shellcheck and see what you think of its comments?
Fixed my IDE setup, only minor stuff. I'll push an update.

ErwanAliasr1:
Just pushed a new version that set the proper cpu scaling & idle governor and set io_poll to 1

ErwanAliasr1:
@ByteHamster Just push a version that prints the nvme config like : 
nvme0n1: MODEL=KCD6XVUL3T20 FW=GPK1 serial=11E0A016TX1R PCI=0000:44:00.0@16.0 GT/s PCIe IRQ=63 NUMA=4 CPUS=16-19,48-51 

Hope you'll like it

ErwanAliasr1:
Just added the processor, memory speed & kernel version

ErwanAliasr1:
@axboe unless you see other stuff to add here, I think this PR is ready

Issue #1093, opened 2020-09-28
loops has no impact when using do_verify=0


Seanistolen:
![image](https://user-images.githubusercontent.com/31502905/94406794-2102df00-01a5-11eb-9b4b-6bee346ac157.png)
Im trying to do sequential write on my drive for 2 time with the parameter **"loops="**, but it seems the job would exit with no error on half.

here is my script:
![image](https://user-images.githubusercontent.com/31502905/94407309-e3528600-01a5-11eb-9d0b-40ab9dac19ae.png)

but when I set size=10G or size=99%, the loops just worked well.

sitsofe:
@Seanistolen:

Can you cut the job file down to the bare minimum that still reproduces the problem (i.e. remove every line that doesn't have an impact, make every number as small as possible, use a single file etc.) and post the reduced job file here?

Seanistolen:
finally I found the impact, it's **do_verify=0.**
[global]
filename=/dev/nvme0n1
ioengine=libaio
the minimum workload is 👇
[precondition]
size=100%
loops=2
#do_verify=0
rw=write
iodepth=8
bs=128k
stonewall

sitsofe:
So enabling `do_verify=0` means that it doesn't do two loops of I/O? Just to check `/dev/nvme0n1` is definitely a device and not a "regular" file right? I'd also be surprised if stonewall was necessary in the above. Just to double check there's nothing related to fio mentioned in `dmesg`?

Seanistolen:
Sorry, stonewall is unnecessary, I was not sure whether **loops** should be combined with **stonewall** or not, then I removed stonewall and try again, the answer is:

- So enabling do_verify=0 means that it doesn't do two loops of I/O?
Yes
- Just to check /dev/nvme0n1 is definitely a device and not a "regular" file right?
It's a device, Samsung 980 Pro.
- Just to double check there's nothing related to fio mentioned in dmesg?
Nothing new in dmesg.

sitsofe:
I can reproduce this with the following:
```
% dd if=/dev/zero of=loop-doverify bs=16k count=1; fio --filename=loop-doverify --size=100% --loops=2 --bs=4k --stonewall --name=normal --gtod_reduce=1 --name=doverify0 --do_verify=0 
1+0 records in
1+0 records out
16384 bytes transferred in 0.000079 secs (207611712 bytes/sec)
[...]

PR #1278, opened 2021-09-28
Fix for issue#1093

ravisowmya:
This pull request address the issue from : https://github.com/axboe/fio/issues/1093
Root cause and fix:
* Issue: if --loops > 1 and --do_verify = 0, we are running one less iteration than the loop count (i.e if --loops=3 it runs 2 iterations as opposed to 3)
* As a part of main thread, we have keep_running that decrements the loops iteration count
* Additionally, as a part of the file management and IO setup, we clear the IO states as a part of this main loop
* The API used for clearing the IO states is clear_io_state and it is called twice within the main thread:
          * At the very beginning before IOs are performed
          * At the very end after all the IOs are processed and we are ready to invoke the next iteration
          * This second c;lear_io_state is skipped if do_verify=0 because of the continue statement in line 1928
* As a part of the clear_io_state, we have the following check
```
if (td->o.time_based || td->o.loops || td->o.do_verify)
  td->nr_done_files = 0;
```
* Since we skip the second call for clear_io_state, we rely on the the first call of this API to take care of IO/file management setup
* BUT for the very last loop, we would have decremented the loop count already in keep_running which results in td->o.loops to be 0. So, clear_io_state does not reset td->nr_done_files which results in a setup error in do_io and we exit early without actually performing any IOs
* Fix is to move the clear_io_state ahead of the continue statement BEFORE the loop count is decremented
* Ran multiple workloads with the above fix and observed write count is loops * size as expected


axboe:
I think the fix looks fine, but the commit message needs a bit of work. The format is generally:

1) Title of commit. Consider it a one-liner description of the problem
2) Empty line
3) Meat of the commit message. What you have here looks pretty good, but do format it at 74 chars wide at most. If not it's impossible to read in git log.
4) Empty line
5) Signed-off-by: Real Name <my@email.com>

axboe:
Also reference other commits in the repo to see how they are formatted.

ravisowmya:
@axboe : Please let me know if the title and commit message look good to you. Thank you!

axboe:
It looks the same? Did you forget to push it out?

ravisowmya:
You're right! I merged and pushed my commit message changes now. Thank you

axboe:
I see the changes now. But it's a mix now, instead of being one clean commit. Please rebase as one commit, I don't want to pollute the history with fixups that aren't strictly needed.

ravisowmya:
Squashed all commit under 1 commit. Thank you!

PR #1282, opened 2021-10-06
Some improvments around one-core-peak & t/io_uring

ErwanAliasr1:
- Reporting kernel config : BLK_CGROUP / BLK_WBT_MQ / RETPOLINE / PAGE_TABLE_ISOLATION
- Reporting the kernel cmdline
-  Adding a runtime option to t/io_uring

### DIFF
--- a/t/io_uring.c
+++ b/t/io_uring.c
@@ -118,6 +118,7 @@ static int do_nop = 0;		/* no-op SQ ring commands */
 static int nthreads = 1;
 static int stats = 0;		/* generate IO stats */
 static int aio = 0;		/* use libaio */
+static int runtime = 0;	/* runtime */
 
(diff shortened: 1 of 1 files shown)


t/io_uring.c
```diff
@@ -841,18 +842,22 @@ static struct submitter *get_submitter(int offset)
 	return ret;
 }
 
-static void sig_int(int sig)
+static void do_finish()
```
axboe, line 845: Maybe take a const char * argument to do_finish() and have it print

Exiting on signal

for signals, and

Exiting on timeout

for non-signal exits. Don't need to print the signal, just get rid of that. And with that change, my other comment would no longer apply - you're missing a void here.

PR #1289, opened 2021-10-17
Improving one-core-peak

ErwanAliasr1:


t/one-core-peak.sh
```diff
@@ -250,7 +257,18 @@ show_system() {
   fi
 }
 
+get_specific_options() {
+  # Add some io_uring benchmark features based on the kernel release
+  options=""
+  # Starting Kernel 5.16, we can enable the DMA pre-mapping
```
axboe, line 269: Never make these things dependent on kernel version, that breaks quite often as distros backport patches. I think for this particular feature, you'd just have to try, and revert to not using -D1 if it fails...

t/io_uring.c
```diff
@@ -1059,7 +1059,7 @@ static void usage(char *argv, int status)
 		" -b <int>  : Block size, default %d\n"
 		" -p <bool> : Polled IO, default %d\n"
 		" -B <bool> : Fixed buffers, default %d\n"
-		" -R <bool> : DMA map fixed buffers, default %d\n"
+		" -D <bool> : DMA map fixed buffers, default %d\n"
```
axboe, line 1062: Looks good, but commit missing a title.

t/one-core-peak.sh
ErwanAliasr1: done.

PR #1292, opened 2021-10-19
Clean up generated dependency makefiles in t/fuzz/

sweettea:
Since t/fuzz/ contains some c files, gcc generates their dependency makefiles in the same directory, and there is not currently a clean rule which addresses them. So this adds an appropriate clean entry to do so.


sweettea:
Sorry about that. Should have a proper title, linewrapped commit message, and signed-off-by line now.

Issue #1090, opened 2020-09-22
fio 3.22 client/server scenario in Windows 2019/SMB share scenario


ipvirg:
Anyone can direct me to documentation on deploying fio v3.22 client/server scenario for windows 2019 server environment? 
﻿fio -v
fio-3.22

fio does appear timing out when probing the SMB share -

fio --client=192.168.1.3 --runtime=30 .\fioconfig_64k_seq.fio
fio: client 192.168.1.3, timeout on cmd PROBE
fio: client 192.168.1.3 timed out

sitsofe:
@ipvirg Does everything work when you run a client and a server on the same machine and you use localhost/127.0.0.1?

sitsofe:
(additionally can you include the contents of `fioconfig_64k_seq.fio` here?)

ipvirg:
﻿Same timeout issue with local client 127.0.0.1 

PS C:\Users\fiouser.FIODOMAINr> fio --client=127.0.0.1 --runtime=30 .\fioconfig_64k_seq.fio
fio: client 127.0.0.1, timeout on cmd PROBE
fio: client 127.0.0.1 timed out
PS C:\Users\fiouser.FIODOMAIN\> more .\fioconfig_64k_seq.fio
[global]
name=fio-directory

directory=C\:\Datafiles
nrfiles=1

rw=readwrite
bs=64k

direct=1

time_based=1
iodepth=1

numjobs=1
group_reporting

size=1GB

ramp_time=20
randseed=50

[vm_prime]

Verified fio is running client/server:

Get-process | findstr fio
   2160      12   136452       5844       0.02   5984   2 fio
netstat -an |findstr 8765
  TCP    0.0.0.0:8765           0.0.0.0:0              LISTENING
﻿

sitsofe:
My faint memory is that fio's server mode needed `fork()` (which Windows doesn't do) and I'm not aware that has changed... Ah here we go: https://www.spinics.net/lists/fio/msg02565.html .

Implementing a thread server mode would be new feature work (would be fun to see though!).

sitsofe:
https://www.spinics.net/lists/fio/msg05248.html mentions the same `timeout on cmd PROBE` messages and in https://www.spinics.net/lists/fio/msg05262.html Jens replies:

> Windows doesn't support fork(), and the fio server implementation was written to use processes. It could be made threaded instead and switched to that, in which case it would work. But it would require a fair amount of work...

[...]

PR #1312, opened 2021-12-10
Windows server suport

blah325:
This fork adds two major changes.  

1) Basic support for --server on windows.  The implementation strategy is described here: 
[https://github.com/axboe/fio/issues/1090#issuecomment-984885752](url)

2)  Added a new windows only IO engine option called “no_completion_thread”.    

Without this option, Windows FIO creates a completion polling thread for each worker thread.  This also requires an event queue for the completion thread to forward completions to the worker thread.  There are two problems with this.  

The first is that it creates a lot of overhead for the OS to manage twice as many threads as well as the overhead of the event queue.  Using no_completion_thread=1 provides notably better IOP/s with “cpus_allowed_policy=split”.   

The second is that it kind of cheats on the ‘num_jobs’ argument because it allowed two threads(and hence two host cores) to be generated for each worker thread unless cpus_allows_split is used to constrain both the "start" and "completion" threads to a given core.   This allowed for unusually high values, depending on how the OS scheduler assigned those threads.

no_completion_thread=1 also has the benefit of aligning the windowsaio engine flow with that of linuxaio.  I didn't want to break others' dependencies or change their results so this was added as a new engine option instead of a new default behavior.

Thanks

### DIFF
--- a/stat.c
+++ b/stat.c
@@ -3052,9 +3052,10 @@ void add_sync_clat_sample(struct thread_stat *ts, unsigned long long nsec)
 	add_stat_sample(&ts->sync_stat, nsec);
 }
 
-static void add_lat_percentile_sample(struct thread_stat *ts,
-				      unsigned long long nsec,
-				      enum fio_ddir ddir, enum fio_lat lat)
+static inline void add_lat_percentile_sample(struct thread_stat *ts,
+					     unsigned long long nsec,
+					     enum fio_ddir ddir,
+					     enum fio_lat lat)
 {
 	unsigned int idx = plat_val_to_idx(nsec);
 	assert(idx < FIO_IO_U_PLAT_NR);
@@ -3062,10 +3063,10 @@ static void add_lat_percentile_sample(struct thread_stat *ts,
 	ts->io_u_plat[lat][ddir][idx]++;
 }
 
-static void add_lat_percentile_prio_sample(struct thread_stat *ts,
-					   unsigned long long nsec,
-					   enum fio_ddir ddir,
-					   bool high_prio)
+static inline void add_lat_percentile_prio_sample(struct thread_stat *ts,
+						  unsigned long long nsec,
+						  enum fio_ddir ddir,
+						  bool high_prio)
 {
 	unsigned int idx = plat_val_to_idx(nsec);
 


engines/windowsaio.c
```diff
@@ -322,34 +343,67 @@ static int fio_windowsaio_getevents(struct thread_data *td, unsigned int min,
 		end_count = start_count + (t->tv_sec * 1000) + (t->tv_nsec / 1000000);
 	}
 
-	do {
-		io_u_qiter(&td->io_u_all, io_u, i) {
-			if (!(io_u->flags & IO_U_F_FLIGHT))
-				continue;
+	if (o->no_completion_thread) {
```
axboe, line 346: Please split this into two functions, ala:

```
if (o->no_completion_thread)
    return fio_windowsaio_getevents_nothread(...);
return fio_windowaio_getevents_thread();
```

or something like that.

optgroup.h
```diff
@@ -116,6 +117,7 @@ enum opt_category_group {
 	FIO_OPT_G_FILESTAT	= (1ULL << __FIO_OPT_G_FILESTAT),
 	FIO_OPT_G_LIBCUFILE	= (1ULL << __FIO_OPT_G_LIBCUFILE),
 	FIO_OPT_G_DFS		= (1ULL << __FIO_OPT_G_DFS),
+	FIO_OPT_G_WINDOWSAIO= (1ULL << __FIO_OPT_G_WINDOWSAIO),
```
axboe, line 120: missing space?

axboe:
I'll look over this in more detail, but please follow the general commit message format. The description of each change should be in the commit message, not in the github description. The git tree is the source of truth, not github.

You also need a Signed-off-by line in each commit. Just check out existing commits in the repo for what the format is, ask if you have any questions.

server.c
```diff
@@ -261,7 +302,20 @@ static int fio_recv_data(int sk, void *buf, unsigned int len, bool wait)
 		flags = OS_MSG_DONTWAIT;
 
 	do {
-		int ret = recv(sk, p, len, flags);
+		int ret;
+#ifdef _WIN32
```
axboe: This should not be win32 specific, I'd suggest just making it so in general.

server.c
```diff
@@ -1300,6 +1457,250 @@ static int get_my_addr_str(int sk)
 	return 0;
 }
 
+#ifdef _WIN32
```
axboe, line 1449: All of this should be in windows specific files. Generally, any use of ifdef in the C code, if it can be avoided, should be avoided.

server.h
```diff
@@ -226,9 +226,11 @@ extern void fio_server_send_gs(struct group_run_stats *);
 extern void fio_server_send_du(void);
 extern void fio_server_send_job_options(struct flist_head *, unsigned int);
 extern int fio_server_get_verify_state(const char *, int, void **);
+extern bool fio_server_poll_single_fd(int fd, short events, int timeout);
 
 extern struct fio_net_cmd *fio_net_recv_cmd(int sk, bool wait);
 
+
```
axboe, line 233: spurious change?

engines/windowsaio.c
blah325: Updated per comments.

ammarfaizi2:
@blah325 before adding more commits. JFYI, the commit message should be word-wrapped at 72 chars. Also, you need a `Signed-off-by` tag with your real name and email at the end of your commit message.

optgroup.h
blah325: Added tab between name and '=' to match other entries.

server.c
blah325: I ended up having to revert this change and move to a more client-specific change in fio_drain_client_text().  I did update the final change so that it is now for all builds.

server.h
blah325: fixed

axboe:
Once you are ready for another review, please collapse the patches into reviewable commits again.

server.c
blah325: The big chunk of code noted here is now moved.  It didn't reduce the number of ifdef significantly but they are smaller chunks.  If you have any ideas on how to avoid the remaining ifdefs, I can look into it.

blah325:
> Once you are ready for another review, please collapse the patches into reviewable commits again.

Can you point me to a reference on how to do that?

axboe:
> Can you point me to a reference on how to do that?

Not sure if it's the best resource out there (I just googled it), but here's one:

https://levelup.gitconnected.com/how-to-squash-git-commits-9a095c1bc1fc

blah325:
> > Can you point me to a reference on how to do that?
> 
> Not sure if it's the best resource out there (I just googled it), but here's one:
> 
> https://levelup.gitconnected.com/how-to-squash-git-commits-9a095c1bc1fc

Thanks.  That was helpful. I think I finally managed to rebase and squash.

axboe:
Did you push it out? Because the tree on github still has a bunch of commits.

axboe:
You don't want any merges, you want a clean history with the changes separated out nicely.

ammarfaizi2:
@blah325 after rebase, don't do `git pull` even if the git says you may want to pull, because it will generate a merge commit like that (non fast forward pull).

You may fail to push after rebase, in that case, keep pushing with `--force` flag to update the remote HEAD.

blah325:
Thanks for the pointers on how to properly rebase/squash.  Please take a look again.

axboe:
I think you squashed too much now, the three commits are now in one commit! What I didn't want is a set of changes and then commits to address review comments, potentially leaving broken bits in the middle. But it should be three separate commits.

axboe:
Also note that it still seems to fail for the most basic of tests:

https://ci.appveyor.com/project/axboe/fio/builds/41906829/job/ddg86tgu559o6an9

jr325:
> Also note that it still seems to fail for the most basic of tests:
> 
> https://ci.appveyor.com/project/axboe/fio/builds/41906829/job/ddg86tgu559o6an9

I neglected to add the list terminator when adding the windowsaio options structure.  The Cygwin and Linux compiles happened to layout structures such that iterating beyond the last entry accessed zeroed data that appeared to be a null name string (terminating iteration).  In the clang compile, theelement beyond the last valid entry was not zeroed data so it attempted to add it attempted to process as another options entry; eventually resulting in a seg fault.  Fix was simply to add the proper list terminator.

blah325:
> Also note that it still seems to fail for the most basic of tests:
> 
> https://ci.appveyor.com/project/axboe/fio/builds/41906829/job/ddg86tgu559o6an9

I neglected to add the list terminator when adding the windowsaio options structure. The Cygwin and Linux compiles happened to layout structures such that iterating beyond the last entry accessed zeroed data that appeared to be a null name string (terminating iteration). In the clang compile, the element beyond the last valid entry was not zeroed data so it attempted to process it as another options entry; eventually resulting in a seg fault. Fix was simply to add the proper list terminator.

ammarfaizi2:
@blah325 Do not squash everything into a single commit. This should be split into 3 commits, each commit should only do one thing.

Note:
Each commit should not break the build and tests. What was asked actually says "don't leave a WIP commit that breaks the build or tests", so please only squash the unfinished work commits, not submitting all of them into one commit.

blah325:
> @blah325 Do not squash everything into a single commit. This should be split into 3 commits, each commit should only do one thing.
> 
> Note: Each commit should not break the build and tests. What was asked actually says "don't leave a WIP commit that breaks the build or tests", so please only squash the unfinished work commits, not submitting all of them into one commit.

Ok.  I'll revert to multiple commits.

PR #1317, opened 2021-12-16
Windows server support and performance improvement

blah325:
This fork has two high level changes

1) Windows --server support
2) New command line option to align windowsaio with linuxaio's threading model and improve IOP/s.

### DIFF
--- a/engines/windowsaio.c
+++ b/engines/windowsaio.c
@@ -11,6 +11,7 @@
 #include <errno.h>
 
 #include "../fio.h"
+#include "../optgroup.h"
 
 typedef BOOL (WINAPI *CANCELIOEX)(HANDLE hFile, LPOVERLAPPED lpOverlapped);
 
@@ -33,6 +34,26 @@ struct windowsaio_data {
 struct thread_ctx {
 	HANDLE iocp;
 	struct windowsaio_data *wd;
+};
+
+struct windowsaio_options {
+	struct thread_data *td;
+	unsigned int no_completion_thread;
+};
+
+static struct fio_option options[] = {
+	{
+		.name	= "no_completion_thread",
+		.lname	= "No completion polling thread",
+		.type	= FIO_OPT_STR_SET,
+		.off1	= offsetof(struct windowsaio_options, no_completion_thread),
+		.help	= "Use to avoid separate completion polling thread",
+		.category = FIO_OPT_C_ENGINE,
+		.group	= FIO_OPT_G_WINDOWSAIO,
+	},
+	{
+		.name	= NULL,
+	},
 };
 
 static DWORD WINAPI IoCompletionRoutine(LPVOID lpParameter);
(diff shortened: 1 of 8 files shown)


rubenk:
I have no comments about the code itself but I just want to say that I tested this PR on Windows 2016 and 2019 and get a nice speed improvement, almost in the order of what I get with diskspd.exe, so thank you for writing this.

engines/windowsaio.c
```diff
@@ -353,6 +429,16 @@ static int fio_windowsaio_getevents(struct thread_data *td, unsigned int min,
...
 }
 
+static int fio_windowsaio_getevents(struct thread_data *td, unsigned int min,
+				    unsigned int max, const struct timespec *t)
+{
+	struct windowsaio_options *o = td->eo;
+
+	if (o->no_completion_thread)
+    	return fio_windowsaio_getevents_nothread(td, min, max, t);
+	return fio_windowaio_getevents_thread(td, min, max, t);
```
ammarfaizi2, line 439: Wrong indentation in the first return.

```suggestion
	if (o->no_completion_thread)
		return fio_windowsaio_getevents_nothread(td, min, max, t);
	return fio_windowaio_getevents_thread(td, min, max, t);
```

engines/windowsaio.c
```diff
@@ -302,10 +325,64 @@ static struct io_u* fio_windowsaio_event(struct thread_data *td, int event)
...
+			(t != NULL && timeout_expired(start_count, end_count)))
+			break;
+	} while (1);
+	return dequeued;
+}
+
+/* dequeue completion entrees creates by separate IoCompletionRoutine thread */
+static int fio_windowaio_getevents_thread(struct thread_data *td, unsigned int min,
+				    unsigned int max, const struct timespec *t) {
+
```
ammarfaizi2, line 385: Wrong brace placement. For functions, they have the opening brace at the beginning of the next line.

```suggestion
static int fio_windowaio_getevents_thread(struct thread_data *td, unsigned int min,
					  unsigned int max, const struct timespec *t)
{
```

ammarfaizi2 (changes requested):
Minor style fixups.

engines/windowsaio.c
```diff
@@ -91,29 +113,30 @@ static int fio_windowsaio_init(struct thread_data *td)
...
+			}
+			if (rc || wd->iothread == NULL)
+				rc = 1;
 		}
-
-		if (rc || wd->iothread == NULL)
-			rc = 1;
 	}
 
 	return rc;
```
ammarfaizi2, line 142: Instead of using `rc` variable + many "if" statements like this, I think it would be much more concise to use `goto` statements to handle the failure cases. `goto` also reduces deeply nested "if" which is bad.

ammarfaizi2 (changes requested):

engines/windowsaio.c
blah325: This change set only added the new "if" on line 116.  This caused the existing "if rc" checks to get indented but otherwise, they match the existing code.   This function(fio_windowsaio_init) is only called during init so overhead of the additional if checks should be minimal.  Can we defer the refactor to another change set?

engines/windowsaio.c
blah325: fixed

engines/windowsaio.c
blah325: fixed

axboe:
I pulled this in - I still think there's work to do here, and I do wish the windows server and windows engine work had been separate pull requests (you should never mix them up like that). But let's not hold it up any longer.

PR #1322, opened 2021-12-23
GitHub actions i686, small t/io_uring fixes

vincentkfu:
Jens, please consider these patches.

They fix the 32-bit GitHub Actions build problem and re-enable testing for the 32-build. (Latency measurement for t/io_uring is broken on 32-bit platforms, though).

Also, I fixed a couple t/io_uring issues found while I was investigating the build problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
         - linux-gcc
         - linux-clang
         - macos
+        - linux-i686-gcc
         include:
         - build: linux-gcc
           os: ubuntu-20.04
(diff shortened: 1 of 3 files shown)


axboe:
This one:

t/io_uring: use only the lower 32 bits as array index

has me a little puzzled, since it is explicitly using the lower 32-bits. With your change, it's the lower 16-bits.

Rest of the patches seem fine to me.

vincentkfu:
You are right. I miscounted. I have updated the branch with that patch dropped.

axboe:
Thanks!

sitsofe:
@vincentkfu nice work on the i686 issue! I got as far as working out it was due to PHP deps from the sury PPA but I didn't know how to fix it.

PR #1330, opened 2022-01-28
Fix sphinx documentation and have a pretty README on GitHub

vincentkfu:
Happily there is a nice solution to the problem @sitsofe noticed with #1328. All we have to do is rename README to README.rst. Then, we'll accomplish these two aims

- Have a pretty README on GitHub (which was the original aim of #1328)
- Still have the README included in the sphinx documentation


axboe:
Thanks Vincent!

PR #1342, opened 2022-02-15
Fix :<nr> suffix with random read/write causing 0 initial offset

aggieNick02:
When using the :<nr> suffix with random reads or writes, the initial
offset would be set to 0 for the first nr-1 operations. This happened
because td->ddir_seq_nr was initialized to the specified option value,
when it needs to always be initialized to 1, so that the first call to
get_next_offset leads to choosing a new random offset for the first nr
operations.

### DIFF
--- a/init.c
+++ b/init.c
@@ -1576,7 +1576,14 @@ static int add_job(struct thread_data *td, const char *jobname, int job_add_num,
 	td->ts.sig_figs = o->sig_figs;
 
 	init_thread_stat_min_vals(&td->ts);
-	td->ddir_seq_nr = o->ddir_seq_nr;
+
+	/*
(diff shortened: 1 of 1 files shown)


init.c
```diff
@@ -1576,7 +1576,12 @@ static int add_job(struct thread_data *td, const char *jobname, int job_add_num,
 	td->ts.sig_figs = o->sig_figs;
 
 	init_thread_stat_min_vals(&td->ts);
-	td->ddir_seq_nr = o->ddir_seq_nr;
+
+	//td->>ddir_seq_nr needs to be initialized to 1, NOT o->ddir_seq_nr,
```
axboe, line 1580: Please use proper C comments. For a multi-line comment, the fio style is:

```
/*
 * Here's a line
 * Here's another
*/
```

aggieNick02: Oops, sorry about that. Updated the PR.

PR #1353, opened 2022-02-28
Option to record alternate_epoch but keep zero-based timestamps

aggieNick02:
Allow the option to set and record alternate_epoch but leave the
timestamps in all log files zero-based. This lets a user record the time
at which a job began with the desired clock_id while keeping nice
zero-based timestamps in the log files. (The zero-based timestamps are
easier to look at visually and also make the log files smaller, and
timestamps against the desired clock_id can be reconstructed by the user
if desired.)

The option log_alternate_eopch_clock_id is renamed to
alternate_epoch_clock_id. The new options relating to alternate_epoch
have not been in a released version of fio, so I expect this renaming
should be ok.


vincentkfu:
I will take a closer look at this later but we typically try to maintain backward compatibility (unless there's a compelling reason otherwise) . Can you have the new preferred option name and the original one be synonyms?  There is an `alias` field in `struct fio_option` that you can use for this.

vincentkfu:
Also please add to the bottom of your commit message:

Fixes: https://github.com/axboe/fio/issues/1544

vincentkfu:
As far as I can see it should be possible to accomplish what you need by simply adding a new key in the JSON output that records when a job started. Then there is no need to change existing options. Would this work for you?

Consider an option called something like `job_start_time_clock_id` and the default could be either disabled or something else sensible. If this is enabled there would just be a new job start time key in the JSON output for each job.

aggieNick02:
Thanks much for looking at this. I totally get not wanting to change option names - back when I wrote the PR the `log_alternate_eopch_clock_id ` option was brand new and hadn't been in a tagged release of fio, but that's definitely not the case anymore.

What you described with `job_start_time_clock_id` sounds great and solves the problem well. It leaves the existing `log_alternate_epoch_clock_id` alone. I'll look at if it makes most sense for the default `job_start_time_clock_id` option to either not log anything, or use a sensible clock_id. I'll try to get the PR together quickly.

Very much appreciate the time and the feedback and idea. Thanks again!

aggieNick02:
Suggested modified approach is up: https://github.com/axboe/fio/pull/1617

PR #1359, opened 2022-03-09
fixed memory leaking in the parent process detected by ASAN

dpronin:
- fixed memory leaking in the parent process detected by ASAN

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -2432,7 +2432,10 @@ reap:
 							strerror(ret));
 			} else {
 				pid_t pid;
+				struct fio_file **files;
 				dprint(FD_PROCESS, "will fork\n");
+				files = td->files;
+				read_barrier();
 				pid = fork();
 				if (!pid) {
 					int ret;
@@ -2441,6 +2444,9 @@ reap:
 					_exit(ret);
 				} else if (i == fio_debug_jobno)
 					*fio_debug_jobp = pid;
+				// freeing previously allocated memory for files
+				// this memory freed MUST NOT be shared between processes, only the pointer itself may be shared within TD
+				free(files);
 				free(fd);
 				fd = NULL;
 			}


dpronin:
Observation before applying these changes:

One job section:

```

bash> ./fio examples/aio-read.fio -x file1

...

==954068==ERROR: LeakSanitizer: detected memory leaks

...

Direct leak of 8 byte(s) in 1 object(s) allocated from:
    #0 0x345cdd in malloc (/home/dpronin/Development/workspace/fio/fio+0x345cdd)
    #1 0x40e60a in dup_files /home/dpronin/Development/workspace/fio/filesetup.c:2034:14

...

```

Four job sections:

```

bash> ./fio examples/aio-read.fio

...

==955058==ERROR: LeakSanitizer: detected memory leaks

...

Direct leak of 32 byte(s) in 4 object(s) allocated from:
    #0 0x345cdd in malloc (/home/dpronin/Development/workspace/fio/fio+0x345cdd)
    #1 0x40e60a in dup_files /home/dpronin/Development/workspace/fio/filesetup.c:2034:14

...

```

This occurs because the parent process that has its own copy of files does not free them properly before pointer 'td->files' located in shared memory is cleared by a child's process

backend.c
```diff
@@ -2432,7 +2432,10 @@ static void run_threads(struct sk_out *sk_out)
 							strerror(ret));
 			} else {
 				pid_t pid;
+				struct fio_file **files;
 				dprint(FD_PROCESS, "will fork\n");
+				files = td->files;
+				read_barrier();
```
dpronin, line 2438: @axboe, what do you think of this line? Is it reasonable for me to insert `read_barrier` quite after reading `td->files` and before `fork`? I thought that is would not be odd and does not hurt anything but ensure here reading files before forking. But I assume that forking implies full memory barrier, doesn't it?

axboe: If a read_barrier() is needed for that, something is horribly broken.

dpronin: @axboe, read_barrier() is needed in my opinion for a generic way to say to a compiler that 'td->files' must be fetched before getting forked to prevent the compiler from possible reordering of instructions. It is required because after forking we unlikely by may get td->files to be NULL generally speaking, because upon forking the parent process may unlikely but sleep until the child is done. That is why I have introduced 'read_barrier' just to avoid such a rarely faced but possible situation

PR #1360, opened 2022-03-10
Substituted tabs by spaces in help page of configure script

dpronin:
- configure script refactoring, tabs are replaced by spaces in the help page of configure script


axboe:
Some formatting notes on the commit message:

- First line is a title. Don't prefix it with '-'
- Then empty line
- The actual commit message
- Then empty line
- Signed-off-by

Your commits seem to use the '-' as a prefix, please get rid of that. Note that all lines should be line wrapped at 72-74 chars, no long lines. Those are hard to read in git log.

dpronin:
@axboe, done

PR #1361, opened 2022-03-10
Improvements in dup_files function

dpronin:
- cleared allocation of td->files when duplicating files
- call 'assert', before iterating over original thread_data, that prevents possible segmentation fault when duplicating files


axboe:
Same message here on commit message formatting.

dpronin:
@axboe, done

PR #1362, opened 2022-03-10
opportunity for a user to compile with ASAN by giving the option to the configure script

dpronin:
- possible enabling ASAN for the compiler when calling 'configure' srript using '--enable-asan' option

### DIFF
--- a/configure
+++ b/configure
@@ -248,6 +248,8 @@ for opt do
   ;;
   --disable-dfs) dfs="no"
   ;;
+  --enable-asan) asan="yes"
+  ;;
   --help)
     show_help="yes"
     ;;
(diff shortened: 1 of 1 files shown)


axboe:
> I have no idea why it has failed with linux-clang

Unfortunately a few of the tests have spurious failures, I don't think there's anything to worry about here and it's not related to clang.

axboe:
Same message on commit message formatting.

vincentkfu:
> > I have no idea why it has failed with linux-clang
> 
> Unfortunately a few of the tests have spurious failures, I don't think there's anything to worry about here and it's not related to clang.

I have seen one other recent failure with https://github.com/axboe/fio/blob/master/t/jobs/t0010-b7aae4ba.fio but I can't find the test run anymore and the failure of course has nothing to do with this PR.

PR #1365, opened 2022-03-11
fix warnings if NDEBUG is enabled

dpronin:
Observation before changes: if CFLAGS is given with -DNDEBUG we will see that everything that is checked within `assert` call is discarded since `assert` is a macro that turns into nothing if NDEBUG is given while compilation

I sophisticated using these unused variables by dummy casting them to void

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -1600,7 +1600,7 @@ static void *thread_main(void *data)
 	uint64_t bytes_done[DDIR_RWDIR_CNT];
 	int deadlock_loop_cnt;
 	bool clear_state;
-	int res, ret;
+	int ret;
 
 	sk_out_assign(sk_out);
 	free(fd);
@@ -1931,13 +1931,23 @@ static void *thread_main(void *data)
 	 * another thread is checking its io_u's for overlap
 	 */
 	if (td_offload_overlap(td)) {
-		int res = pthread_mutex_lock(&overlap_check);
-		assert(res == 0);
+		int res;
+
+		res = pthread_mutex_lock(&overlap_check);
+		if (res) {
+			td->error = errno;
+			goto err;
+		}
 	}
 	td_set_runstate(td, TD_FINISHING);
 	if (td_offload_overlap(td)) {
+		int res;
+
 		res = pthread_mutex_unlock(&overlap_check);
-		assert(res == 0);
+		if (res) {
+			td->error = errno;
+			goto err;
+		}
 	}
 
 	update_rusage_stat(td);
--- a/helper_thread.c
+++ b/helper_thread.c
@@ -1,4 +1,7 @@
+#include <errno.h>
 #include <signal.h>
+#include <stdio.h>
+#include <string.h>
 #include <unistd.h>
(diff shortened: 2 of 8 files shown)


axboe:
A lot of these should just properly handle errors, rather than be lazy check with an assert. I just fixed up t/dedupe to do that, I hope it can be inspiration for the others.

Also please use descriptive titles for commit messages too. Phase 1 and 2 doesn't really mean anything. Why is it 2 patches? It looks like one is addressing core code, the other test programs. Which means that that should be reflected in the title instead.

dpronin:
@axboe, done, will it do?

axboe:
No, I don't think you read and understood my comment. My point is that rather than work-around this with ugly

```
(void) ret;
```

additions, the cases that are actually valid errors and should be passed back should be fixed, rather than just papered over by silencing a warning.

backend.c
```diff
@@ -1931,12 +1931,14 @@ static void *thread_main(void *data)
 	 * another thread is checking its io_u's for overlap
 	 */
 	if (td_offload_overlap(td)) {
-		int res = pthread_mutex_lock(&overlap_check);
+		int const res = pthread_mutex_lock(&overlap_check);
+		(void)res;
```
dpronin, line 1935: @axboe, what do you suggest I should do here and in similar places? abort if non-zero return?

dpronin:
@axboe, check it out

vincentkfu:
Instead of fprintf(stderr, ...) fio uses log_err() for error messages.

dpronin:
@vincentkfu, done

dpronin:
@axboe, are we going to merge it in?

axboe:
I'm doing a release soon and prefer not to have further changes that aren't strictly needed for a release. So please ping me a bit after 3.30 is released and we'll take a look.

dpronin:
@axboe as you wish

dpronin:
@axboe ping

dpronin:
@axboe may we proceed with this PR, I am waiting for your remarks right in the review

helper_thread.c
```diff
@@ -126,7 +126,7 @@ static void submit_action(enum action a)
 
 	ret = write_to_pipe(helper_data->pipe[1], &data, sizeof(data));
 	if (ret != 1) {
-		fprintf(stderr, "failed to write action into pipe, err %i:%s", errno, strerror(errno));
+		log_err("failed to write action into pipe, err %i:%s", errno, strerror(errno));
```
axboe, line 129: Didn't you just add this one in patch 1?

dpronin: Clarify, please, I didn't get

axboe: You add this fprintf() in the first patch, and then a later patch changes it. Just use log_err() in the first patch, and you then have one less place to fix up?

axboe: Either that, or do the fprintf -> log_err change first. Then it's already consistent by the time you add your addition.

dpronin: @axboe got you. done

dpronin:
@axboe is there anything else we should change before merging?

PR #1368, opened 2022-03-15
Properly encode engine flags in thread_data::flags

albertofaria:
We have 16 engine flags and an 18-bit shift, so cast engine flags to
unsigned long long before shifting to avoid dropping
FIO_ASYNCIO_SYNC_TRIM and FIO_NO_OFFLOAD.

Also make thread_data::flags unsigned long long to ensure it fits all
flags even when longs are 32 bit, and fix TD_ENG_FLAG_MASK.

Signed-off-by: Alberto Faria <afaria@redhat.com>

### DIFF
--- a/fio.h
+++ b/fio.h
@@ -184,7 +184,7 @@ struct zone_split_index {
  */
 struct thread_data {
 	struct flist_head opt_list;
-	unsigned long flags;
+	unsigned long long flags;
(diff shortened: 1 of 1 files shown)


axboe:
Thanks, good catch. There really should've been a build error for this to catch mistakes of this nature...

Issue #1372, opened 2022-03-21
why time_based=1 and file_service_type=sequential only rewrite the first file


sjp00556:
**Please acknowledge the following before creating a ticket**

- [x]  I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
time_based=1 and file_service_type=sequential only rewrite the first file and other files are all empty.  if remove file_service_type=sequential param, all files will be written

**Environment**: centos7

**fio version**: 3.27 & 3.20

**Reproduction steps**

```
# job
[global]

#128k write
[job]
name=128Kwrite1
direct=1
ioengine=psync
directory=/home/data/bench/fio/
rw=write
file_service_type=sequential
bs=128k
filesize=128k
nrfiles=10
numjobs=1
time_based=1
runtime=5
```

![screenshot-20220321-125349](https://user-images.githubusercontent.com/6017354/159215761-c828e85d-0d22-4cc8-b79b-28bc2d211ff0.png)

vincentkfu:
Here is a possible fix:
```
diff --git a/io_u.c b/io_u.c
index 806ceb77..98a44576 100644
--- a/io_u.c
+++ b/io_u.c
@@ -355,7 +355,7 @@ static int get_next_seq_offset(struct thread_data *td, struct fio_file *f,
         * and invalidate the cache, if we need to.
         */
        if (f->last_pos[ddir] >= f->io_size + get_start_offset(td, f) &&
-           o->time_based) {
+           o->time_based && o->nr_files == 1) {
                f->last_pos[ddir] = f->file_offset;
                loop_cache_invalidate(td, f);
        }
```
In a job without `time_based`, `get_next_seq_offset` will eventually produce an offset that is larger than the file's real size and this will trigger a move to the next file. But in a `time_based` job when `get_next_seq_offset` reaches the end of the first file it just wraps around. Here's some debug output illustrating what happens:
```
file     4449  get file test.0.0, ref=1
io       4449  fill: io_u 0x55edd0640380: off=0xf000,len=0x1000,ddir=1,file=test.0.0
io       4449  prep: io_u 0x55edd0640380: off=0xf000,len=0x1000,ddir=1,file=test.0.0
io       4449  queue: io_u 0x55edd0640380: off=0xf000,len=0x1000,ddir=1,file=test.0.0
io       4449  complete: io_u 0x55edd0640380: off=0xf000,len=0x1000,ddir=1,file=test.0.0
[...]

PR #1375, opened 2022-03-24
io_u: produce bad offsets for some time_based jobs

vincentkfu:
Allow get_next_seq_offset to produce bad offsets for time_based jobs
when fio is accessing more than one file. Otherwise fio will not skip to
the next file once the current one is done.

Fixes: https://github.com/axboe/fio/issues/1372

Signed-off-by: Vincent Fu <vincent.fu@samsung.com>

I've run all of our automated tests and this patch introduces no issues. So I think this fix is unlikely to cause any regressions.


axboe:
Thanks for fixing this one, Vincent!

PR #1379, opened 2022-04-02
actions-full-test.sh, removed sudo from the script

dpronin:
the script might be called with 'sudo' from the call site, if required


dpronin:
@axboe ping

axboe:
Will this break CI? sudo as root should be a no-op, so really no harm in leaving it in there even if I agree with the change in principle.

dpronin:
@axboe as you might see CI has passed, 6 checks have passed successfully. My point has been not using 'sudo' if not required, but if it is required a call site may run scripts within `sudo` calling `sudo ci/actions-full-test.sh`

PR #1380, opened 2022-04-02
added sublime text specific files in .gitignore

dpronin:
ignore by git index sublime specific files


dpronin:
@axboe ping

ammarfaizi2:
I think this is a personal text editor problem rather than fio's. There is no relevance between fio and sublime, it is a personal text editor choice.

Commit message should explain why such a change is necessary, and you don't explain that here?

dpronin:
OK, understood, closing the PR

PR #1385, opened 2022-04-10
fixed jobs eta memory leak

dpronin:
Launch fio built with ASAN enabled, for example, by this way:
```
./fio --max-jobs=4 --output=fio-output.json --output-format=json+ --filename=fio_jsonplus_clat2csv.test --ioengine=libaio --time_based --runtime=3s --size=1M --slat_percentiles=1 --clat_percentiles=1 --lat_percentiles=1 --thread=1 --name=test1 --rw=randrw --name=test2 --rw=read --name=test3 --rw=write --thread=1
```

Borrowed from one tests the project has got

We can see many memory leaks, in particular:

```
Direct leak of 573 byte(s) in 3 object(s) allocated from:
    #0 0x345e0d in malloc (/home/dpronin/Development/workspace/fio/fio+0x345e0d)
    #1 0x414499 in get_jobs_eta /home/dpronin/Development/workspace/fio/eta.c:690:7
```

This PR fixes this leak

### DIFF
--- a/eta.c
+++ b/eta.c
@@ -3,6 +3,7 @@
  */
 #include <unistd.h>
 #include <string.h>
+#include <stdlib.h>
 #ifdef CONFIG_VALGRIND_DEV
 #include <valgrind/drd.h>
 #else
@@ -707,10 +708,10 @@ void print_thread_status(void)
 	size_t size;
 
 	je = get_jobs_eta(false, &size);
-	if (je)
+	if (je) {
 		display_thread_status(je);
-
-	free(je);
+		free(je);
+	}
 }
 
 void print_status_init(int thr_number)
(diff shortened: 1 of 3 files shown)


dpronin:
@axboe ping

dpronin:
@axboe ping

axboe:
The free_jobs_eta() helper seems rather pointless, as no extra work needs to be done outside of just calling free() on it. If it had reference counting or could come from a different backing, maybe, but as it stands just add the missing free() call instead?

dpronin:
@axboe it is just a habit of mine to define functions with reverse meaning for deallocation. If you want me to just call 'free' tell me one more time please to let me know that this thing is principal for you as a maintainer of the project

axboe:
Not sure why I have to say it twice? IMHO it's a useless helper.

dpronin:
@axboe not sure why you ended your first comment with a question mark having made it to be a question, didn't you? If you want someone to do something I do not think you should go with question, otherwise the other side would think you like to discuss

dpronin:
@axboe done

PR #1387, opened 2022-04-10
fixed bunch of memory leaks in json constructor

dpronin:
fixed memory leak produced by not freed string given by 'strdup' in
'json_object_add_value_string' function. Just removed 'strdup' and made
string in 'json_value' as immutable

### DIFF
--- a/json.h
+++ b/json.h
@@ -81,8 +81,13 @@ static inline int json_object_add_value_string(struct json_object *obj,
 	struct json_value arg = {
 		.type = JSON_TYPE_STRING,
 	};
+	union {
+		const char *a;
+		char *b;
+	} string;
 
-	arg.string = strdup(val ? : "");
+	string.a = val ? val : "";
+	arg.string = string.b;
 	return json_object_add_value_type(obj, name, &arg);
 }
 
--- a/t/run-fio-tests.py
+++ b/t/run-fio-tests.py
@@ -546,9 +546,10 @@ class FioJobTest_iops_rate(FioJobTest):
             return
 
         iops1 = self.json_data['jobs'][0]['read']['iops']
-        iops2 = self.json_data['jobs'][1]['read']['iops']
-        ratio = iops2 / iops1
         logging.debug("Test %d: iops1: %f", self.testnum, iops1)
+        iops2 = self.json_data['jobs'][1]['read']['iops']
+        logging.debug("Test %d: iops2: %f", self.testnum, iops2)
+        ratio = iops2 / iops1
         logging.debug("Test %d: ratio: %f", self.testnum, ratio)
 
         if iops1 < 950 or iops1 > 1050:


json.h
```diff
@@ -15,7 +15,7 @@ struct json_value {
 	union {
 		long long integer_number;
 		double float_number;
-		char *string;
+		char const *string;
 		struct json_object *object;
 		struct json_array *array;
 	};
```
ammarfaizi2, line 21: I suggest having one more extra member for the "JSON add method" purpose, here we can have `char const *const_str;` instead of changing `char *string`, so it's still compatible with other parts that are using it.

 
```
	union {
		long long integer_number;
		double float_number;
		char *string;
		char const *const_str;
		struct json_object *object;
		struct json_array *array;
	};
```

And then to avoid `strdup`, you can assign to `const_str`.

dpronin: As you might have noticed there has been a `type` field designating what a field in the union should be used. How are you going to distinguish between `string` and `const string`? Do you want to offer a new enum item for the field `type`?

dpronin: @axboe Actually, I cannot understand why you have denied my changes about freeing string that has been duplicated. `strdup` is used there because there is a mutable string in `struct json_value`, and this is a part of this json library inside FIO. If I remove `strdup` I cannot assign immutable string to mutable, you know, and casting away `const` looks to me illegal, but if you insist on doing so, OK, we can do this ugly approach. On the other hand @ammarfaizi2 offers to introduce a new field `const string`, what looks as a workaround for the sake of workaround, although it is acceptable, but the latter requires to introduce a new enumeration item for the field `type` in `struct json_value`. We need your decision then

ammarfaizi2: > As you might have noticed there has been a `type` field designating what a field in the union should be used. How are you going to distinguish between `string` and `const string`? Do you want to offer a new enum item for the field `type`?

OK, if you follow that that `json_object_add_value_type()` with argument type string, it doesn't actually change anything, so the `strdup()` isn't necessary, I suggest the following changes:
```diff
diff --git a/json.h b/json.h
index d9824263..ec87751b 100644
--- a/json.h
+++ b/json.h
@@ -81,8 +81,13 @@ static inline int json_object_add_value_string(struct json_object *obj,
 	struct json_value arg = {
 		.type = JSON_TYPE_STRING,
 	};
+	union {
+		const char *a;
+		char *b;
+	} p;
 
-	arg.string = strdup(val ? : "");
+	p.a = val ? val : "";
+	arg.string = p.b;
 	return json_object_add_value_type(obj, name, &arg);
 }
 

```
which is much simpler and faster.

Previously, you said it's illegal, but it's actually not. Because the callee doesn't change anything. Comment?

dpronin: I tend to agree to accept this snippet of code. My point has been to get rid of memory leak in JSON

dpronin:
@ammarfaizi2 @axboe take a look, guys, I updated the changes

ammarfaizi2:
> @ammarfaizi2 @axboe take a look, guys, I updated the changes

Hmmm... it seems it breaks the CI, the appveyor failed. Not sure what's going on there.

dpronin:
> > @ammarfaizi2 @axboe take a look, guys, I updated the changes
> 
> Hmmm... it seems it breaks the CI, the appveyor failed. Not sure what's going on there.

Windows' specific tests using cygwin failed

from run-fio-tests.py:

```
iops1 = self.json_data['jobs'][0]['read']['iops']
        iops2 = self.json_data['jobs'][1]['read']['iops']
        ratio = iops2 / iops1
        logging.debug("Test %d: iops1: %f", self.testnum, iops1)
        logging.debug("Test %d: ratio: %f", self.testnum, ratio)

        if iops1 < 950 or iops1 > 1050:
            self.failure_reason = "{0} iops value mismatch,".format(self.failure_reason)
            self.passed = False

        if ratio < 6 or ratio > 10:
            self.failure_reason = "{0} iops ratio mismatch,".format(self.failure_reason)
            self.passed = False
```

According to logs iops ratio is 25.616238, meaning that iops2 is about 39,041642258, the last condition `if ratio < 6 or ratio > 10` came true:

```
DEBUG:root:Test 11: return code: 0
DEBUG:root:Test 11: skipped 2 lines decoding JSON data
DEBUG:root:Test 11: iops1: 1000.100000
DEBUG:root:Test 11: ratio: 25.616238
DEBUG:root:Test 11: stderr:
DEBUG:root:Test 11: stdout:
Test 11 FAILED:  iops ratio mismatch, t0011-5d2788d5.fio
```

dpronin:
I've pushed updated logging of these crucial values of iops1, iops2, ration, let's wait for the results

dpronin:
@ammarfaizi2 The 11th test seems to have passed giving the results:

```
...
DEBUG:root:Test 11: return code: 0
DEBUG:root:Test 11: skipped 2 lines decoding JSON data
DEBUG:root:Test 11: iops1: 1000.100000
DEBUG:root:Test 11: iops2: 8001.700000
DEBUG:root:Test 11: ratio: 8.000900
Test 11 PASSED t0011-5d2788d5.fio
DEBUG:root:Test 12: return code: 0
...
```

What I can conclude that the test has some sort of fluctuations in the better occasion, in the worst scenario it is a kind of race condition somewhere we have had so far

dpronin:
@axboe I requesting you to pay attention and give your verdict

ammarfaizi2 (approved):

vincentkfu:
Test 11 fails occasionally. Most of time time it passes but there is a problem with the `flow` option.

PR #1399, opened 2022-05-21
pmemblk.c: fix one logic bug - read always with write

guoanwu:
logic issue,if read success and return 0, then pmemblk_write called as well:
if (io_u->ddir == DDIR_READ &&
  0 != pmemblk_read(pmb->pmb_pool, buf, off)) {
	io_u->error = errno;
	break;
} else if (0 != pmemblk_write(pmb->pmb_pool, buf, off)) {
	io_u->error = errno;
	break;
}

### DIFF
--- a/engines/pmemblk.c
+++ b/engines/pmemblk.c
@@ -375,10 +375,11 @@ static enum fio_q_status fio_pmemblk_queue(struct thread_data *td,
 		off /= pmb->pmb_bsize;
 		len /= pmb->pmb_bsize;
 		while (0 < len) {
-			if (io_u->ddir == DDIR_READ &&
-			   0 != pmemblk_read(pmb->pmb_pool, buf, off)) {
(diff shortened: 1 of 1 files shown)


axboe:
Thanks, change looks good. But can you please ensure to put a proper Signed-off-by in there as well?

Would be nice to fix the:

if (0 != foo)

and particularly the

while (0 < len)

style as well as a separate change, that's not the fio style and particularly the latter is unreadable compared to the straight forward

while (len >0)

guoanwu:
got it and will add one space after the key word.

axboe:
> got it and will add one space after the key word.

Did you push out the wrong thing? Still missing the signed-off-by.

guoanwu:
Thanks, I see. will update soon.

guoanwu:
already add the sign off information in the commit。

PR #1411, opened 2022-06-15
Init file_cache to invalid (maj, min)

useche:
In the very unlikely case that the trace was taken from a device with
major and minor zeroes, the cache will wrongly hit the first time. We
are hitting this problem when trying to replay generated traces with
major and minor zeroes. Initializing with ~0U fixes this problem.

Signed-off-by: Luis Useche useche@gmail.com

### DIFF
--- a/blktrace.c
+++ b/blktrace.c
@@ -442,7 +442,10 @@ err:
 bool read_blktrace(struct thread_data* td)
 {
 	struct blk_io_trace t;
-	struct file_cache cache = { };
+	struct file_cache cache = {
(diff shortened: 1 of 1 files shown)


axboe:
This:

```
Signed-off-by: Luis Useche [useche@gmail.com](mailto:useche@gmail.com)
```

needs to go into the commit, not necessarily in here (though that doesn't hurt).

useche:
Done

axboe:
Thanks!

PR #1419, opened 2022-07-06
server: only do cpu_to_le64() on io_sample_data member if iolog is histogram

tuan-hoang1:
Signed-off-by: Ulrich Weigand <ulrich.weigand@de.ibm.com>
Signed-off-by: Tuan Hoang <tuan.hoang1@ibm.com>

### DIFF
--- a/server.c
+++ b/server.c
@@ -2284,7 +2284,8 @@ int fio_send_iolog(struct thread_data *td, struct io_log *log, const char *name)
 			struct io_sample *s = get_sample(log, cur_log, i);
 
 			s->time		= cpu_to_le64(s->time);
-			s->data.val	= cpu_to_le64(s->data.val);
+			if (log->log_type != IO_LOG_TYPE_HIST)
+				s->data.val	= cpu_to_le64(s->data.val);
 			s->__ddir	= __cpu_to_le32(s->__ddir);
 			s->bs		= cpu_to_le64(s->bs);
 


tuan-hoang1:
Hi @axboe, I also think there is also a similar check in client.c as well (see below). What do you think ? 
```
diff --git a/client.c b/client.c
index 605a3ce5..37da74bc 100644
--- a/client.c
+++ b/client.c
@@ -1702,7 +1702,8 @@ static struct cmd_iolog_pdu *convert_iolog(struct fio_net_cmd *cmd,
                        s = (struct io_sample *)((char *)s + sizeof(struct io_u_plat_entry) * i);
 
                s->time         = le64_to_cpu(s->time);
-               s->data.val     = le64_to_cpu(s->data.val);
+               if (ret->log_type != IO_LOG_TYPE_HIST)
+                       s->data.val     = le64_to_cpu(s->data.val);
                s->__ddir       = __le32_to_cpu(s->__ddir);
                s->bs           = le64_to_cpu(s->bs);
                s->priority     = le16_to_cpu(s->priority);

```

axboe:
Can you put an actual commit message in there? A good commit message should explain WHY the change is being made, but there's nothing in there I can't just learn from looking at the change itself.

tuan-hoang1:
> Can you put an actual commit message in there? A good commit message should explain WHY the change is being made, but there's nothing in there I can't just learn from looking at the change itself.

I just updated. Thanks !

axboe:
Great, thanks! That makes sense. Agree on the client side fwiw - if we need this on the server side, client side should do the same. I'm going to merge this one, please send one for the client code as well. Thanks.

PR #1425, opened 2022-07-19
Fixed misplaced goto in http.c

giubacc:
In http.c:fio_http_queue function, when the user specifies rw=write and if curl_easy_perform fails then the control reaches line 565 and incorrectly prints: "WARNING: Only DDIR_READ/DDIR_WRITE/DDIR_TRIM are supported!".
Fix to this consists in moving statement: "goto err" at the end of the block as already has been done for trim/read blocks.

Signed-off-by: Giuseppe Baccini <giuseppe.baccini@suse.com>


vincentkfu:
I flubbed the workflow when I merged this. The actual merge is here: https://github.com/axboe/fio/commit/4af67ef2b7b0091419c4de9cbceda13d4c64db63

Issue #1424, opened 2022-07-18
Issues creating single file with create_serialize=0


weberc-ntap:
**Please acknowledge the following before creating a ticket**

- [ X ] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**

I'm looking into some issues when running fio and have noticed some unusual behavior.  The intent is to write a very large file using a large number of jobs in parallel.  For this reason, I'm using create_serialize=0.  However, this results in setup_files() getting called in parallel as well.  This in turn calls extend_file() which does unlink and open/create operations to attempt to resize the file. When run in parallel, it's deleting and creating the file repeatedly, so you're not guaranteed to get a file of the appropriate size. 

Original FIO command of interest:
`fio --output-format=normal --rw=write --fallocate=none --group_reporting=1 --disable_lat=1 --disable_clat=1 --disable_slat=1 --startdelay=5 --ramp_time=3 --time_based=0 --invalidate=0 --create_serialize=0 --size=4G --nrfiles=1 --filename=/home/user/fio_testdir/cw1 --offset_increment=4G --blocksize=128K --name=socket0 --numjobs=80
`
Command I'm using to recreate:
`./fio --output-format=normal  --name=global --rw=write --create_only=0 --fallocate=none --group_reporting=1 --disable_lat=1 --disable_clat=1 --disable_slat=1 --startdelay=5 --ramp_time=3 --time_based=0 --invalidate=0 --create_serialize=0 --size=1M --nrfiles=1 --filename=/home/user/fio_testdir/cw1 --offset_increment=1M --blocksize=128K --name=socket0 --numjobs=64
`
In my example below, I'm expecting a file of 64MB -- I ended up with a file of 60MB in size.  

```
Starting 64 processes
file     1933568 setup files
file     1933565 setup files
file     1933571 setup files
file     1933566 setup files
file     1933569 setup files
file     1933568 get file size for 0x7f59e73c0890/0//home/user/fio_testdir/cw1
file     1933571 get file size for 0x7f59e73c0e30/0//home/user/fio_testdir/cw1
file     1933565 get file size for 0x7f59e73c02f0/0//home/user/fio_testdir/cw1
file     1933567 setup files
file     1933570 setup files
socket0: Laying out IO file (1 file / 59MiB)
socket0: Laying out IO file (1 file / 62MiB)
file     1933566 get file size for 0x7f59e73c04d0/0//home/user/fio_testdir/cw1
[...]

PR #1431, opened 2022-07-25
Fix multithread setup issues when operating on a single shared file

weberc-ntap:
Proposal to fix axboe/fio#1424

I added some changes to perform file creation/extension when running with create_serialize=0 and nr_files=1.  setup_files() will still be done in parallel after the threads are kicked off later, but some early, single-threaded file creation will be done now.  This will put things in a good enough state that steup_files() should not need to unlink and create new files, thus avoiding the original issue.

I did find that there is still an issue when using --fallocate=none as an argument.  This results in a file of size 0, and setup_files() will still later try to perform the setup in parallel.  Using fallocate=native or fallocate=truncate both appear to be viable and desirable alternatives to fallocate=none.  

The combination of these changes and fallocate=none/truncate fix issue #1424 for me.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -2314,8 +2314,25 @@ static void run_threads(struct sk_out *sk_out)
 	for_each_td(td, i) {
 		print_status_init(td->thread_number - 1);
 
-		if (!td->o.create_serialize)
+		if (!td->o.create_serialize) {
+			/*
+			 *  When operating on a single rile in parallel,
+			 *  perform single-threaded early setup so that
+			 *  when setup_files() does not run into issues
+			 *  later.
+			*/
+			if (!i && td->o.nr_files==1) {
+				if (setup_shared_file(td)) {
+					exit_value++;
+					if (td->error)
+						log_err("fio: pid=%d, err=%d/%s\n",
+							(int) td->pid, td->error, td->verror);
+					td_set_runstate(td, TD_REAPED);
+					todo--;
+				}
+			}
 			continue;
+		}
 
 		if (fio_verify_load_state(td))
 			goto reap;
--- a/filesetup.c
+++ b/filesetup.c
@@ -143,7 +143,7 @@ static int extend_file(struct thread_data *td, struct fio_file *f)
 	if (unlink_file || new_layout) {
 		int ret;
 
-		dprint(FD_FILE, "layout unlink %s\n", f->file_name);
+		dprint(FD_FILE, "layout %d unlink %d %s\n", new_layout, unlink_file, f->file_name);
 
 		ret = td_io_unlink_file(td, f);
 		if (ret != 0 && ret != ENOENT) {
@@ -197,6 +197,9 @@ static int extend_file(struct thread_data *td, struct fio_file *f)
 			}
 		}
 	}
+
+
+	dprint(FD_FILE, "fill file %s, size %llu\n", f->file_name, (unsigned long long) f->real_file_size);
 
 	left = f->real_file_size;
 	bs = td->o.max_bs[DDIR_WRITE];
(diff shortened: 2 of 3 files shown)


axboe:
Please turn this into an actual patch series. It's fine to have multiple commits, even encouraged, but nobody wants to see a patch add debug output for testing you presumably did, then cleanup patches that remove it.

axboe:
Each patch should be able to stand alone. It's fine to build up to a feature by adding code that helps implement it, in fact it's encouraged as it generally makes the actual functional change smaller and easier to review. But each patch should be useful by itself, and not just remove cruft added by other patches.

Each patch should also result in a compilable and runnable end result, as otherwise bisection doesn't work as intended.

axboe:
Each commit message should also explain WHY a change is being made, and additionally it should follow the general fio commit style guidelines:

1) First line is a commit title, a descriptive one-liner for the change
2) Empty line
3) Commit message itself. As mentioned, this explains why the change is useful.
4) Empty line
5) Signed-off-by: Real Name <real@email.com>

See other commits in the repo for hints on what a commit should look like. The last few from Vincent are good examples.

Also note that you should aim for clean lines at 72-74 chars wide. This ensures it is nicely readable when running git log.

weberc-ntap:
Fair enough, thank you for the feedback and instructions. Let me clean this up the proper way and resubmit.
________________________________

weberc-ntap:
Hopefully this looks better now. Please let me know if there are any additional changes you would like to see here.

backend.c
```diff
@@ -2314,8 +2314,25 @@ static void run_threads(struct sk_out *sk_out)
...
 
-		if (!td->o.create_serialize)
+		if (!td->o.create_serialize) {
+			/*
+			 *  When operating on a single rile in parallel,
+			 *  perform single-threaded early setup so that
+			 *  when setup_files() does not run into issues
+			 *  later.
+			*/
+			if (i==0 && td->o.nr_files==1) {
```
axboe, line 2324: if (!i && td->o.nr_files == 1)

filesetup.c
```diff
@@ -1078,6 +1081,45 @@ static bool create_work_dirs(struct thread_data *td, const char *fname)
 	return true;
 }
 
+int setup_shared_file(struct thread_data *td)
+{
+	struct fio_file *f;
+	uint64_t file_size;
+	int err=0;
```
axboe, line 1088: int err = 0;

filesetup.c
```diff
@@ -1078,6 +1081,45 @@ static bool create_work_dirs(struct thread_data *td, const char *fname)
...
+	if (f == NULL) {
+		log_err("fio: NULL shared file\n");
+		return -1;
+	}
+
+	file_size = thread_number * td->o.size;
+	dprint(FD_FILE, "shared setup %s real_file_size=%llu, desired=%llu\n", 
+			f->file_name, (unsigned long long)f->real_file_size, (unsigned long long)file_size);
+
+	if(f->real_file_size < file_size) {
```
axboe, line 1108: Space after ```if```

filesetup.c
```diff
@@ -1078,6 +1081,45 @@ static bool create_work_dirs(struct thread_data *td, const char *fname)
...
+		dprint(FD_FILE, "fio: extending shared file\n");
+		f->real_file_size = file_size;
+		err = extend_file(td, f);
+		if(!err) {
```
axboe, line 1112: Space after if

axboe:
From a quick look, much better. Just a few style comments for now, I'll take a closer look in a bit.

backend.c
weberc-ntap: Fixed

filesetup.c
weberc-ntap: Fixed

filesetup.c
weberc-ntap: Fixed

filesetup.c
weberc-ntap: Fixed

weberc-ntap:
Apologies for the formatting issues. I should have caught those. I updated the commit with fixes.

PR #1439, opened 2022-08-09
lib/rand: Enhance __fill_random_buf using the multiple random seed

sungup:
The __fill_random_buf fills a buffer using the random 8byte integer for
write operation. But, this mechanism is depend on the CPU performance
and could not reach the max performance on the PCIe Gen5 devices.
I have tested 128KB single worker sequential write on PCIe Gen5 NVMe,
but it cannot reach write throughput 6.0GB/s.

So, I have reviewed the __fill_random_buf and focused the multiplier
dependency to generate the random number. So, I have changed
__fill_random_buf using the multiple-random-seed to reduce the
dependencies bewteen generating and assinging in the small data
filling loop.

I'll attach detail analysis result in the PR of this branch.

Signed-off-by: Sungup Moon <sungup.moon@samsung.com>

### DIFF
--- a/configure
+++ b/configure
@@ -116,6 +116,10 @@ has() {
   type "$1" >/dev/null 2>&1
 }
 
+num() {
+  echo "$1" | grep -P -q "^[0-9]+$"
+}
+
 check_define() {
   cat > $TMPC <<EOF
 #if !defined($1)
@@ -174,6 +178,7 @@ libnfs=""
 xnvme=""
 libzbc=""
 dfs=""
+seed_buckets=""
 dynamic_engines="no"
 prefix=/usr/local
 
@@ -255,6 +260,8 @@ for opt do
   ;;
   --enable-asan) asan="yes"
   ;;
+  --seed-buckets=*) seed_buckets="$optarg"
+  ;;
   --help)
     show_help="yes"
     ;;
@@ -302,6 +309,7 @@ if test "$show_help" = "yes" ; then
   echo "--dynamic-libengines    Lib-based ioengines as dynamic libraries"
   echo "--disable-dfs           Disable DAOS File System support even if found"
   echo "--enable-asan           Enable address sanitizer"
+  echo "--seed-buckets=         Number of seed buckets for the refill-buffer"
   exit $exit_val
 fi
 
@@ -3273,6 +3281,15 @@ if test "$disable_tcmalloc" != "yes"; then
   fi
 fi
 print_config "TCMalloc support" "$tcmalloc"
+if ! num "$seed_buckets"; then
+  seed_buckets=4
+elif test "$seed_buckets" -lt 2; then
+  seed_buckets=2
+elif test "$seed_buckets" -gt 16; then
+  seed_buckets=16
+fi
+echo "#define CONFIG_SEED_BUCKETS $seed_buckets" >> $config_host_h
+print_config "seed_buckets" "$seed_buckets"
 
 echo "LIBS+=$LIBS" >> $config_host_mak
 echo "GFIO_LIBS+=$GFIO_LIBS" >> $config_host_mak
--- a/lib/rand.c
+++ b/lib/rand.c
@@ -95,7 +95,7 @@ void init_rand_seed(struct frand_state *state, uint64_t seed, bool use64)
 		__init_rand64(&state->state64, seed);
 }
 
-void __fill_random_buf(void *buf, unsigned int len, uint64_t seed)
+void __fill_random_buf_small(void *buf, unsigned int len, uint64_t seed)
 {
 	uint64_t *b = buf;
 	uint64_t *e = b  + len / sizeof(*b);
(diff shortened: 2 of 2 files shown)


sungup:
Hi, @axboe, please help to review this PR.

Here is a PR for the __fill_random_buf function. Our team are testing the new
PCIe generation device, but master version FIO cannot reach the device
specification about the sequential write performance.

I have already reviewed the latest __fill_random_buf and random-fill-aes
branch, but there are some problem to improve that function

1. Latest version __fill_random_buf show nice performance than before, but it
   has not solve the instruction dependencies of MUL. So, that function can be
   more improved if reduce the MUL dependencies.
2. random-fill-aes function is very goot performance on the Intel systems. But
   AMD system cannot shows performance gain like Intel CPU, and aes feature is
   not supported on the ARM architecture.

Current master version use MUL instruction while generating random number, but
MUL is known as a multiple-cycle instruction and the latency is longer than
ADD/SUB instruction. I assumed that single random-seed is cause of CPU stall,
and I have changed the function uses multiple-random seed. Also to prove my
assumtion, I compared the performance between the current __fill_random_buf,
fill_random_buf_aes, and PR requested version. To compare each algorithm, I
create simple c/c++ main function and calls 8 Million times get enough result.

Tested systems are Intel Skylake, AMD Epyc CPU, Apple M1 Max, and Ampere ARM
CPU. Following table is the result of my benchmark.


| Elapsed Time(sec) | master |  random-fill-aes | 2buckets | 4buckets | 8buckets | 16buckets |
|:------------------|-------:|-----------------:|---------:|---------:|---------:|----------:|
| Intel Icelake     | 131.0  | 65.5             | 93.7     | 55.0     | 54.8     | 54.9      |
| AMD Epyc          | 125.2  | 83.8             | 63.0     | 48.0     | 52.6     | 42.3      |
| ARM (M1 Max)      | 128.9  | Not supported    | 68.4     | 36.6     | 25.8     | 23.1      |
| ARM (Ampere)      | 183.1  | Not supported    | 137.6    | 137.5    | 137.6    | 137.5     |

As a result, random-fill-aes shows improvement on Intel/AMD, but multiple-seed
using over 4 buckets shows better result than using aes. Also, M1 Max shows the
most reduced overhead using the multiple-seed buckets.

axboe:
This is great, I did ponder the dependency when I did the aes work, but didn't do anything about it. This pull request is much appreciated! And I agree that it's nice to have the generic filler be as fast as it can be as well.

Any chance you'd want to play with the aes variant too, to kill the dependency?

I'm going to tag a new release very shortly, and I think it'd be prudent to pull this one after making the release.

vincentkfu:
Hello Sungup, here are some comments:

- please use C-style comments (/* ... */ instead of //)
- If the number of buckets will be a build-time option, it would be better to have it as an option in the configure script and `echo "#define SEED_BUCKETS $VALUE" > $config_host_h` instead of having it configured via the Makefile.

I did some quick tests and found big improvements with your patch running commands similar to

```
./fio --name=test --ioengine=null --rw=write --bs=128k --size=100g --refill=1
```
Below are bandwidth measurements on Intel Skylake:

| block size | PR | fio-3.30-97-g6e44f |
| :- | -: | -: |
| 4K | 11800 MiB/s | 6572 MiB/s |
| 128K | 24100 MiB/s | 9572 MiB/s |

sungup:
Hello @axboe and @vincentkfu 
Thank you for your reply about this PR. :-)

I have change the code following the comments from Vincent, but I'm not familiar the configure script.
Please check the changes is right.

configure
```diff
@@ -3273,6 +3281,15 @@ if test "$disable_tcmalloc" != "yes"; then
...
 fi
 print_config "TCMalloc support" "$tcmalloc"
+if ! num "$seed_buckets"; then
+  seed_buckets=4
+elif test "$seed_buckets" -lt 2; then
+  seed_buckets=2
+elif test "$seed_buckets" -gt 16; then
+  seed_buckets=16
+fi
+echo "#define SEED_BUCKETS $seed_buckets" >> $config_host_h
```
axboe, line 3291: Looks fine to me, but I think we should call this one CONFIG_SEED_BUCKETS to match with the other entries in configure.

axboe:
I ran this on my laptop (m1 pro running linux), and I get:

4K: 5611MB/sec -> 12.1GB/sec
128k: 7620MB/sec -> 28.4GB/sec

with the default 4 buckets.

Let's rename SEED_BUCKETS to CONFIG_SEED_BUCKETS and let's please squash the two patches into a single patch. You can do that while doing the rename.

Once that's done, we can pull this in! Thanks for doing this work.

configure
sungup: I've changed to the CONFIG_SEED_BUCKETS. But, line length is little bit longer in `lib/rand.c`, change also code format in __fill_random_buf.

sungup:
Thank you for review.

I have squashed the commits and push into a single patch.
This command based squashing is my first time, and it is very good experience time to me.

I'm so happy this PR is helpful to you!

axboe:
Thanks for doing this work! Now pulled.

Issue #1443, opened 2022-08-16
fio doesn't exit on an I/O error with `--continue_on_error=none`


Hi-Angel:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**

Documentation says *(in short)*: `continue_on_error=none — Exit on any I/O or verify errors.` However, if a disk returns an IO error, FIO silently ignores it even if `--continue_on_error=none` is passed.

For steps-to-reproduce here I use a dmsetup's `error` device, and I confirmed it actually returns IO errors by using a `cat /dev/mapper/test > /dev/null` and seeing cat would fail with an IO error.

**Environment**: Ubuntu 18.04

**fio version**: latest master `fio-3.31-11-g4deb` and an older `fio-3.1`.

**Reproduction steps**

1. Create a ram disk as `sudo modprobe rd rd_nr=1 rd_size=$(( 100 * 1024 ))`
2. Create a virtual device that would sporadically return errors as `sudo dmsetup create test --table '0 123 error /dev/ram0 0 10 2'`
3. Run fio as `fio --name=temp-fio --rw=randwrite --rwmixwrite=30 --filename=/dev/mapper/test --time_based --runtime=1h --exitall_on_error --continue_on_error=none`
4. Run `journalctl -f` and wait for a message like `kernel: Buffer I/O error on dev dm-281, logical block 70, lost async page write` to appear

**Expected**

`fio` would exit immediately upon seeing the error

**Actual**

`fio` would continue running until it's interrupted or the time runs out.

axboe:
The error might not be directly visible to fio, if it's an async buffered write. The write most likely completed just fine in terms of fio, in that the write(2) (or similar) system call returned the full size of the write and not an error. You'd only see the error when you later fsync'ed the file descriptor. So I don't think there's much fio could do here. If you set ```---direct=1``` as an option it'd likely look different.

continue_on_error=none is also the default behavior. fio will always exit on any IO error, unless it's explicitly told not to.

Hi-Angel:
Oh, this is cool! I confirm that `--direct=1` works. Yeah, your explanation makes sense, thank you!

I'll send a clarification to the documentation section shortly then

PR #1444, opened 2022-08-16
doc: clarify that I/O errors may go unnoticed without direct=1

Hi-Angel:
Fixes: https://github.com/axboe/fio/issues/1443

### DIFF
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -3927,6 +3927,13 @@ Error handling
 	appended, the total error count and the first error. The error field given
 	in the stats is the first error that was hit during the run.
 
+	Note: a write error from the device may go unnoticed by fio when using
+	buffered IO, as the write() (or similar) system call merely dirties the
+	kernel pages, unless :option:`sync` or :option:`direct` is used. Device IO
+	errors occur when the dirty data is actually written out to disk. If fully
+	sync writes aren't desirable, :option:`fsync` or :option:`fdatasync` can be
+	used as well. This is specific to writes, as reads are always synchronous.
+
 	The allowed values are:
 
 		**none**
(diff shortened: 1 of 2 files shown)


Hi-Angel:
> I think it should also mention that some variant of fsync or fdatasync would also help

Hmm, not sure I got that… fsync() and fdatasync() are libc functions. I don't really see how information about C functions might be useful to a user of the `fio` application.

> and that this issue is specific to writes as reads will always be sync.

Oh, thanks! I actually was wondering if it works for reads or not. In the end I decided to go on the safe side by being more abstract with a "buffered I/O", but yeah, now that you clarified that I will be more explicit. Thank you!

axboe:
fsync and fdatasync are also fio job options.

Hi-Angel:
Oh, I see! Sorry, thanks, will do

Hi-Angel:
Done, please see if you want anything changed

axboe:
I would probably phrase it as:

Note: a write error from the device may go unnoticed by fio when using buffered IO, as the write() (or similar) system call merely dirties the kernel pages, unless :option:`sync` or :option:`direct` is used. Device IO errors occur when the dirty data is actually written out to disk. If fully sync writes aren't desirable, :option:`fsync` or :option:`fdatasync` can be used as well. This is specific to writes, as reads are always synchronous.

with the right kind of option syntax. And double check that the two sections are the same in fio.1 and the HOWTO, they were different in your patch. Thanks!

Hi-Angel:
Thanks, will do

> And double check that the two sections are the same in fio.1 and the HOWTO, they were different in your patch. Thanks!

But they aren't. Here's a screenshot of the diff — nothing except the syntactic specifics of the two formats is different:
Screenshot_20220817_014444

Hi-Angel:
Done!

axboe:
Thanks!

Hi-Angel:
Thank you! The [docs on the site](https://fio.readthedocs.io/en/latest/fio_doc.html) didn't change though, do they have to be updated separately?

vincentkfu:
https://fio.readthedocs.io/en/latest/ is maintained by someone else

PR #1446, opened 2022-08-18
Fix logic determining if operating on single shared file

### LINKED ISSUE #1442: td_io_open_file: Assertion `f->fd == -1' failed
When running a posixaio job, with the following job parameters, my first thread consistently fails after the '0 byte' file is created.

```
root@ft-loadgen03:/mnt/hcsf/nfs# fio nfs_fio.txt
fio: time_based requires a runtime/timeout setting
nfs_load: (g=0): rw=write, bs=(R) 8192B-8192B, (W) 8192B-8192B, (T) 8192B-8192B, ioengine=posixaio, iodepth=1
...
fio-3.31-3-g9dc5
Starting 20 processes
fio: ioengines.c:517: td_io_open_file: Assertion `f->fd == -1' failed.
fio: pid=1276951, got signal=6[eta 00m:00s]

```

**Job Definition**
```
[global]
time_based=1
startdelay=5
exitall_on_error=0
create_serialize=0
directory=/mnt/nfs/
group_reporting
clocksource=gettimeofday
ioengine=posixaio
disk_util=0
iodepth=1
create_on_open=1

[nfs_load]
rw=write
bs=8k
numjobs=20
direct=0
size=1TB
stonewall
```

Any assistance would be much appreciated.

Thank you!

weberc-ntap:
nrfiles=1 is not a valid decision point to determine whether
or not fio is operating on a single shared file. If numjobs>1
and filename is NOT specified, each job will create a single
non-shared file. add_file() is called differently when filename
is specified on the command line. If filename is specified,
then add_file() is called once from an options callback and
td->files_size==1. If filename is NOT specified, and each job
creates its own file, then td->files_size>=2 as filenames are
dynamically added. Change the logic to call setup_shared_file()
so that it is only called wheen td->files_size==1.

Signed-off-by: Chris Weber <weberc@netapp.com>

This fixes #1442 

I tested using the fio.txt job options:
```
[global]
time_based=1
startdelay=5
exitall_on_error=0
create_serialize=0
directory=/home/user/fiotest
group_reporting
clocksource=gettimeofday
ioengine=posixaio
disk_util=0
iodepth=1
create_on_open=1

[nfs_load]
rw=write
bs=8k
numjobs=20
direct=0
size=1M
stonewall
```

I also verified that our use case of a single-shared file still behaves as desired.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -2316,12 +2316,16 @@ static void run_threads(struct sk_out *sk_out)
 
 		if (!td->o.create_serialize) {
 			/*
-			 *  When operating on a single rile in parallel,
+			 *  When operating on a single file in parallel,
 			 *  perform single-threaded early setup so that
 			 *  when setup_files() does not run into issues
-			 *  later.
+			 *  later. As files are added by add_job(), a
+			 *  single shared file will end up with
+			 *  files_size==1. Jobs dynamically creating
+			 *  files will end up with files_size>=2, even
+			 *  if there is only a single file per job.
 			*/
-			if (!i && td->o.nr_files == 1) {
+			if (!i && td->files_size == 1) {
 				if (setup_shared_file(td)) {
 					exit_value++;
 					if (td->error)


vincentkfu:
@jhardydev @crtierney42 can you see if this fixes the issue for you?

vincentkfu:
@weberc-ntap Would something like the following resolve the original issue for you?

```
fio --name=test --create_serialize=0 --filesize=16M --io_size=1M --nrfiles=1 --filename=test --offset_increment=1M --numjobs=16 --group_reporting --fallocate=none
```

It replaces `--size=1M --numjobs=16` with `--filesize=16M --io_size=1M --numjobs=16`

Also, there is a problem with the reliance on `thread_number` in https://github.com/axboe/fio/pull/1431. An invocation that defines additional jobs after the one that needs special handling produces a file that is too large:

```
./fio-canonical/fio-tip --name=test --create_serialize=0 --size=1M --nrfiles=1 --filename=test --offset_increment=1M --numjobs=16 --group_reporting --fallocate=none --debug=file --create_only=1 --name=test2 --ioengine=null --filesize=1T --numjobs=4 --create_only=1 --group_reporting
```

The above produces a 20MiB file because of the four additional null ioengine jobs. 

```
...
fio-3.31-16-gd33c
Starting 20 processes
file     122853 get file size for 0x7fa13185d110/0/test
file     122853 shared setup test real_file_size=0, desired=20971520
file     122853 fio: extending shared file
file     122853 layout 1 unlink 0 test
file     122853 open file test, flags 241
file     122853 truncate file test, size 20971520
file     122853 fill file test, size 20971520
file     122853 get file size for 0x7fa13185d110/0/test
file     122853 shared setup new real_file_size=20971520
...
```

Also an invocation where the job that needs special handling isn't the first one:

```
./fio-canonical/fio-tip --name=test2 --ioengine=null --filesize=1T --numjobs=4 --create_only=1 --group_reporting --name=test --create_serialize=0 --size=1M --nrfiles=1 --filename=test --offset_increment=1M --numjobs=16 --group_reporting --fallocate=none --debug=file --create_only=1
```

doesn't trigger the special setup because thread 0 is the only one considered for special handling.

Finally, I think there needs to be a check for offset_increment as well because an invocation that omits offset_increment

```
./fio-canonical/fio-tip --name=test --create_serialize=0 --size=1M --nrfiles=1 --filename=test --numjobs=16 --group_reporting --fallocate=none --debug=file --create_only=1
```

should not trigger the special setup and should have all 16 threads writing to a 1MiB file.

weberc-ntap:
I would be open to adding the extra `filesize` option, but I feel like there should be a way to do this without it as well.  I'd like to take one more look quick today to see if it can be done without adding a new option.

@crtierney42 Thoughts on adding a new `filesize` option?

@vincentkfu For the final case you pointed out above, I think the code to determine the file size should be fixed:
`file_size = thread_number * td->o.size;`
Should instead be closer to what the the other code does:
`file_size = ((thread_number-1) * td->o.offset_increment) + td->o.size;`

vincentkfu:
> I would be open to adding the extra `filesize` option, but I feel like there should be a way to do this without it as well. I'd like to take one more look quick today to see if it can be done without adding a new option.
> 
> @crtierney42 Thoughts on adding a new `filesize` option?

The `filesize` option already exists and requires no code changes. Please try the command I suggested and let me know if what it does meets your requirements. If it does we can update the documentation to say that `create_serialize=0` may not work as expected if multiple jobs are writing to the same file.

> 
> @vincentkfu For the final case you pointed out above, I think the code to determine the file size should be fixed: `file_size = thread_number * td->o.size;` Should instead be closer to what the the other code does: `file_size = ((thread_number-1) * td->o.offset_increment) + td->o.size;`

The problem with this proposal is that `thread_number` appears to be the total number of jobs to be run in a single fio invocation. Some might be writing to the file of interest and so may not. Probably something like `numjobs` is what you need.

I think calculating `file_size` this way is very tricky to get right because there are so many different cases to account for.

I am hoping that `filesize` and `io_size` will do the trick for you and then we can just revert https://github.com/axboe/fio/pull/1431.

weberc-ntap:
I understand now. Let me check and report back.

weberc-ntap:
Things look pretty good for these two cases:
```
./fio --output-format=normal --name=global --rw=write --create_only=1 --fallocate=truncate --group_reporting=1 --disable_lat=1 --disable_clat=1 --disable_slat=1 --startdelay=5 --ramp_time=3 --time_based=0 --invalidate=0 --create_serialize=0 --filesize=200M --io_size=1M --nrfiles=1 --filename=/home/user/fio_testdir/cw1 --offset_increment=1M --blocksize=128K --name=socket0 --numjobs=200 --debug=file
```

```
./fio --output-format=normal --name=global --rw=write --create_only=1 --fallocate=truncate --group_reporting=1 --disable_lat=1 --disable_clat=1 --disable_slat=1 --startdelay=5 --ramp_time=3 --time_based=0 --invalidate=0 --create_serialize=0 --filesize=200M --io_size=1M --nrfiles=1 --filename=/home/user/fio_testdir/cw1 --offset_increment=1M --blocksize=128K --name=socket0 --numjobs=200 --debug=file
```

I still see the parallel unlink/create behavior in the read case:
```
./fio --output-format=normal --name=global --rw=read --create_only=0 --fallocate=truncate --group_reporting=1 --disable_lat=1 --disable_clat=1 --disable_slat=1 --startdelay=5 --ramp_time=3 --time_based=0 --invalidate=0 --create_serialize=0 --filesize=200M --io_size=1M --nrfiles=1 --filename=/home/user/fio_testdir/cw1 --offset_increment=1M --blocksize=128K --name=socket0 --numjobs=200 --debug=file
```

Note that this only happens if I delete the file between each step.  The normal use-case would be to run the three steps back-to-back without deleting the file in-between.  In that case, things appear to work OK.  

I would propose reverting the offending patches to fix the regressions for the other cases you identified.  

If the read case above gives future issues, I can revisit this again.

PR #1448, opened 2022-08-21
Avoid refilling io buffers if applicable to reduce CPU utilization

ningzheng-x:
This commit is aimed to avoid refilling io buffers in write scenarios by setting existing
parameter refill_buffers properly to reduce CPU utilization if applicable.

In a non-read case, there is no need to refill io buffers every time when data dedupe or
data verify is not set. We only need to refill io buffers for the 1st time, and reuse these
buffers in the following rounds. By doing this, we can significantly reduce the CPU
utilization, which could avoid CPU being the bottleneck when testing multiple SSDs
in one single server. This is of more importance when we turn to PCIe Gen4 or Gen5
interface and set a specific compression ratio.

Ning Zheng <ningzhengrpi@163.com>

### DIFF
--- a/init.c
+++ b/init.c
@@ -900,6 +900,8 @@ static int fixup_options(struct thread_data *td)
 		}
 	}
 
+	o->io_buffer_filled = 0;
+	
 	/*
 	 * Using a non-uniform random distribution excludes usage of
 	 * a random map
--- a/init.c
+++ b/init.c
@@ -887,16 +887,21 @@ static int fixup_options(struct thread_data *td)
 	/*
 	 * For fully compressible data, just zero them at init time.
 	 * It's faster than repeatedly filling it. For non-zero
-	 * compression, we should have refill_buffers set. Set it, unless
-	 * the job file already changed it.
+	 * compression, we will have refill_buffers set if either
+	 * data dedupe or verify is set. Set it, unless the job file
+	 * already changed it.
 	 */
 	if (o->compress_percentage) {
 		if (o->compress_percentage == 100) {
 			o->zero_buffers = 1;
 			o->compress_percentage = 0;
 		} else if (!fio_option_is_set(o, refill_buffers)) {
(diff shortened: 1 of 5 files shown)


init.c
```diff
@@ -887,16 +887,21 @@ static int fixup_options(struct thread_data *td)
...
+	 * already changed it.
 	 */
 	if (o->compress_percentage) {
 		if (o->compress_percentage == 100) {
 			o->zero_buffers = 1;
 			o->compress_percentage = 0;
 		} else if (!fio_option_is_set(o, refill_buffers)) {
-			o->refill_buffers = 1;
-			td->flags |= TD_F_REFILL_BUFFERS;
+			if (td_read(td) || o->dedupe_percentage != 100) {
```
axboe: I really don't get this. Why would you set refill_buffers for something that is a read? That'd just fill buffers only to get rewritten by whatever is read.

io_u.c
```diff
@@ -2235,6 +2235,14 @@ void fill_io_buffer(struct thread_data *td, void *buf, unsigned long long min_wr
 		unsigned long long left = max_bs;
 		unsigned long long this_write;
 
+		if (td_write(td) && o->dedupe_percentage == 100 &&
+		    o->verify == VERIFY_NONE) {
+			if (o->io_buffer_filled >= o->iodepth) {
```
axboe: parenthesis not needed here.

axboe:
Comments on why this change makes sense aside, your git history is now somewhat of a mess. We don't want a bunch of reverts and then a retry of the change done on top of that, please appropriately squash and rebase it.

axboe:
Commits also need proper commit messages. Basically all the stuff you put on github belongs in the git tree, not on github. That includes the signed off by. Please check other commits in the tree for how they are done. I'm going to close this one, you may open a new pull request when these things have been fixed up.

PR #1450, opened 2022-08-23
Add wait for handling SIGBREAK

bjpaupor:
When closing a command prompt window or terminating it using something
like the taskkill command, each child process (such as a running FIO
workload) is sent a SIGBREAK signal. Once those child processes have
responded to that signal, Windows terminates them if they're still
executing.

This change has the main thread to wait for others to exit when handling
a SIGBREAK signal, such that each job will still have time to wrap-up
and give stats before the entire program terminates.

Signed-off-by: Brandon Paupore <brandon.paupore@wdc.com>

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -90,6 +90,25 @@ static void sig_int(int sig)
 	}
 }
 
+#ifdef WIN32
+static void sig_break(int sig)
+{
+	struct thread_data *td;
+	int i;
+
+	sig_int(sig);
+
+	/**
+	 * Windows terminates all job processes on SIGBREAK after the handler
+	 * returns, so give them time to wrap-up and give stats
+	 */
+	for_each_td(td, i) {
+		while (td->runstate < TD_EXITED)
+			sleep(1);
+	}
+}
+#endif
+
 void sig_show_status(int sig)
 {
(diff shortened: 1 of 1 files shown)


jeffreyalien (commented):
This change was made so the FIO output data could be captured when taskkill was used to end FIO jobs running.  Without the fix, the output data file would be empty.

vincentkfu:
I can't get this feature to work. In one window I run

```
fio.exe --name=test --ioengine=null --time_based --runtime=60s --filesize=1M --thread=1 --numjobs=2 --group_reporting --output=test
```

In another window I run:

```
taskkill /IM fio.exe /F
```

The job stops but the output file is empty. How are you getting this to work?

Also, fio doesn't spawn separate processes when run under Windows. So I don't understand why the commit message mentions child processes.

bjpaupor:
@vincentkfu Yeah a forceful taskkill won't allow for things to wrap up before stopping, this enables the user to taskkill the command prompt window itself without the /F option and reliably get output from FIO.

To terminology, Windows Task Manager lists a running FIO workload under Processes, specifically under the parent "Windows Command Processor" there would be "fio.exe".

PR #1452, opened 2022-08-24
Enable CPU affinity support on Android

bvanassche:
This patch enables the --cpumask=, --cpus_allowed= and
--cpus_allowed_policy= fio options.


axboe:
Thanks Bart!

PR #1455, opened 2022-08-25
io_uring: dereference pthread_t value from pthread_self

kraj:
__init_rand64 takes 64bit value and srand48 takes unsigned 32bit value,
pthread_t is opaque type and some libcs ( e.g. musl ) do not define them
in plain old data types and ends up with errors

| t/io_uring.c:809:32: error: incompatible pointer to integer conversion passing 'pthread_t' (aka 'struct __pthread *') to parameter of type 'uint64_t' (aka 'unsigned long') [-Wint-conver
sion]
|         __init_rand64(&s->rand_state, pthread_self());
|                                       ^~~~~~~~~~~~~~

Signed-off-by: Khem Raj <raj.khem@gmail.com>

2. Empty second line
4. Empty line

### DIFF
--- a/t/io_uring.c
+++ b/t/io_uring.c
@@ -799,15 +799,14 @@ static int submitter_init(struct submitter *s)
 	int i, nr_batch, err;
 	static int init_printed;
 	char buf[80];
-
 	s->tid = gettid();
 	printf("submitter=%d, tid=%d, file=%s, node=%d\n", s->index, s->tid,
 							s->filename, s->numa_node);
 
 	set_affinity(s);
 
-	__init_rand64(&s->rand_state, pthread_self());
-	srand48(pthread_self());
+	__init_rand64(&s->rand_state, s->tid);
+	srand48(s->tid);
 
 	for (i = 0; i < MAX_FDS; i++)
 		s->files[i].fileno = i;


t/io_uring.c
```diff
@@ -799,15 +799,15 @@ static int submitter_init(struct submitter *s)
...
+	pthread_t tid = pthread_self();
 	s->tid = gettid();
 	printf("submitter=%d, tid=%d, file=%s, node=%d\n", s->index, s->tid,
 							s->filename, s->numa_node);
 
 	set_affinity(s);
 
-	__init_rand64(&s->rand_state, pthread_self());
-	srand48(pthread_self());
+	__init_rand64(&s->rand_state, *((unsigned long long int*) &tid));
```
axboe, line 809: How does this work if sizeof(pthread_t) != sizeof(unsigned long long)?

axboe:
I'd just use s->tid instead and get rid of the pthread_self().

t/io_uring.c
kraj: right, it seems to be ok when pthread_t size is smaller than uint64_t, but it will have aliasing issues due to the way its subverting type system :(

axboe:
Thanks!

Issue #1457, opened 2022-08-28
Fio can silently drop log entries when using the log compression option


aggieNick02:
**Please acknowledge the following before creating a ticket**

- [X] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
The log compression option is powerful, allowing per I/O logging with reduced memory usage. Unfortunately, it can result in dropped log entries. Using the null ioengine seems to make the issue reproduce pretty reliably.

The number of entries dropped varies, and can vary between different logs in a single run. Turning off compression makes the issue go away.

Turning the `log_compression` parameter down, while perhaps not something one would do in production, seems to make the problem more pervasive, plus gives a repro that is much faster to run. In the repro below I've reduced `size` from a typical `1000G` and `log_compression` from a typical `10M` by a factor of 1024 to `1000M` and `10K`, respectively

**Environment**: Ubuntu 20.04.4 LTS

**fio version**: fio-3.31-31-ga5a2 (head of master, also reproduces in fio-3.30

**Reproduction steps**
Create a new directory and cd to it. Then run:

```
fio --filename=/dev/nvme1n1 --name=job --per_job_logs=0 --write_bw_log=fio_0 --ioengine=null --direct=1 --size=1000M --rw=write --bs=128K --iodepth=1 --log_offset=1 --log_store_compressed=1 --log_compression=10K
```

Notice total I/O (all writes) is 8000

Now inflate the bw log, and then count the number of lines to see how many I/O were recorded:

```
fio --inflate-log=fio_0_bw.log.fz > fio_0_bw.log
wc -l fio_0_bw.log
```

and observe the output:
```
3904 fio_0_bw.log
```

With the offset logged, it makes it easy to see where a drop has happened. It actually happened at this exact spot two times in a row, interestingly.
```
6, 5461333333, 1, 131072, 402391040, 0
6, 5242880000, 1, 131072, 402522112, 0
9, 4369066666, 1, 131072, 939524096, 0
9, 2114064516, 1, 131072, 939655168, 0
```

aggieNick02:
Bug does not appear to occur if `--log_store_compressed=1` is omitted.

(Storing compressed is a really nice option - files are much smaller, and each one can be inflated and then gzip'ed in parallel to end up with reasonably sized log files that are in a friendly format for use by other tools.)

vincentkfu:
[...]

PR #1458, opened 2022-09-01
Fix fio silently dropping log entries when using log_compression

aggieNick02:
This fixes issue 1457: https://github.com/axboe/fio/issues/1457

It would be great if we could get a new fio release (3.33) that includes this. :-)

The fix is to just add a workqueue_flush to iolog_flush so that any log
chunks being compressed in the workqueue get flushed before we compress
and write the entries in the current active logs. Without this, log
entries in the workqueue can be omitted or be written out of order.

We also add a new test that ensures the log file has the right number of
entries in the correct order when using log_compression, both with and
without the store_log_compressed option.

There is still some cleanup to do with respect to the current
documentation and comments for the log_compression option, as it does not
affect the size of chunks, but instead is effectively a boolean that
controls whether chunks are compressed and removed. I'll leave that for
a separate PR.

CONFIRMED
2. Empty second line
4. Empty line

### DIFF
--- a/iolog.c
+++ b/iolog.c
@@ -1574,14 +1574,14 @@ void iolog_compress_exit(struct thread_data *td)
  * Queue work item to compress the existing log entries. We reset the
  * current log to a small size, and reference the existing log in the
  * data that we queue for compression. Once compression has been done,
- * this old log is freed. If called with finish == true, will not return
- * until the log compression has completed, and will flush all previous
- * logs too
+ * this old log is freed. Will not return until the log compression
+ * has completed, and will flush all previous logs too
  */
 static int iolog_flush(struct io_log *log)
 {
 	struct iolog_flush_data *data;
 
+	workqueue_flush(&log->td->log_compress_wq);
 	data = malloc(sizeof(*data));
 	if (!data)
 		return 1;
--- a/t/log_compression.py
+++ b/t/log_compression.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python3
+#
+# log_compression.py
+#
+# Test log_compression and log_store_compressed. Uses null ioengine.
+# Previous bugs have caused output in per I/O log files to be missing
+# and/or out of order
+#
+# Expected result: 8000 log entries, offset starting at 0 and increasing by bs
+# Buggy result: Log entries out of order (usually without log_store_compressed)
+# and/or missing log entries (usually with log_store_compressed)
+#
+# USAGE
+# python log_compression.py [-f fio-executable]
+#
+# EXAMPLES
+# python t/log_compression.py
+# python t/log_compression.py -f ./fio
+#
+# REQUIREMENTS
+# Python 3.5+
+#
+# ===TEST MATRIX===
+#
+# With log_compression=10K
+# With log_store_compressed=1 and log_compression=10K
+
+import os
+import sys
+import platform
+import argparse
+import subprocess
+
+
+def parse_args():
+    """Parse command-line arguments."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-f', '--fio',
+                        help='path to fio executable (e.g., ./fio)')
+    return parser.parse_args()
+
+
+def run_fio(fio,log_store_compressed):
+    fio_args = [
+        '--name=job',
+        '--ioengine=null',
+        '--filesize=1000M',
+        '--bs=128K',
+        '--rw=write',
+        '--iodepth=1',
+        '--write_bw_log=test',
+        '--per_job_logs=0',
+        '--log_offset=1',
+        '--log_compression=10K',
+        ]
+    if log_store_compressed:
+        fio_args.append('--log_store_compressed=1')
+
+    subprocess.check_output([fio] + fio_args)
+
+    if log_store_compressed:
+        fio_inflate_args = [
+            '--inflate-log=test_bw.log.fz'
+            ]
+        with open('test_bw.from_fz.log','wt') as f:
+            subprocess.check_call([fio]+fio_inflate_args,stdout=f)
+
+def check_log_file(log_store_compressed):
+    filename = 'test_bw.from_fz.log' if log_store_compressed else 'test_bw.log'
+    with open(filename,'rt') as f:
+        file_data = f.read()
+    log_lines = [x for x in file_data.split('\n') if len(x.strip())!=0]
+    log_ios = len(log_lines)
+
+    filesize = 1000*1024*1024
+    bs = 128*1024
+    ios = filesize//bs
+    if log_ios!=ios:
+        print('wrong number of ios ({}) in log; should be {}'.format(log_ios,ios))
+        return False
+
+    expected_offset = 0
+    for line_number,line in enumerate(log_lines):
+        log_offset = int(line.split(',')[4])
+        if log_offset != expected_offset:
+            print('wrong offset ({}) for io number {} in log; should be {}'.format(
+                log_offset, line_number, expected_offset))
+            return False
+        expected_offset += bs
+    return True
+
+def main():
+    """Entry point for this script."""
+    args = parse_args()
+    if args.fio:
+        fio_path = args.fio
+    else:
+        fio_path = os.path.join(os.path.dirname(__file__), '../fio')
+        if not os.path.exists(fio_path):
+            fio_path = 'fio'
+    print("fio path is", fio_path)
+
+    passed_count = 0
+    failed_count = 0
+    for log_store_compressed in [False, True]:
+        run_fio(fio_path, log_store_compressed)
+        passed = check_log_file(log_store_compressed)
+        print('Test with log_store_compressed={} {}'.format(log_store_compressed,
+            'PASSED' if passed else 'FAILED'))
+        if passed:
+            passed_count+=1
+        else:
+            failed_count+=1
+
+    print('{} tests passed, {} failed'.format(passed_count, failed_count))
+
+    sys.exit(failed_count)
+
+if __name__ == '__main__':
+    main()
+
--- a/t/run-fio-tests.py
+++ b/t/run-fio-tests.py
@@ -1124,6 +1124,14 @@ TEST_LIST = [
         'success':          SUCCESS_DEFAULT,
         'requirements':     [],
     },
+    {
+        'test_id':          1012,
+        'test_class':       FioExeTest,
+        'exe':              't/log_compression.py',
+        'parameters':       ['-f', '{fio_path}'],
+        'success':          SUCCESS_DEFAULT,
+        'requirements':     [],
+    },
 ]
 
 


axboe:
There's no reason why we can't just cut a 3.33 with the fix. Let's do that in about a weeks time.

vincentkfu:
`fio --inflate-log` seems to be broken in Windows (unrelated to this patch).

aggieNick02:
> `fio --inflate-log` seems to be broken in Windows (unrelated to this patch).

Yikes. I'm surprised that running without `log_store_compressed` doesn't have the same problem (since it also hits the same functions, albeit with slightly different parameters.) The biggest difference I notice is `gz_hdr` in `inflate_chunk` is `true` for `--inflate-log` but `false` for immediate inflation, so that `z_stream_init` adds `32` to `wbits`. I might be missing something, but I would be surprised if removing the add of 32 broke anything, and it may fix windows. (The tests I just added still pass after removing it.)

Also kinda scary is that running fio (with the reproduction code for this issue) results in multiple `log_err`:
```
fio: deflate log (-5)
fio: deflate log (-5)
fio: deflate log (-5)
```
The output looks ok on first examination, but those are errors from the call to zlib's `deflate`, and send us to the error case in `gz_work`.

vincentkfu:
I did some quick troubleshooting on Windows and the results suggest that Windows fio actually produces corrupt compressed log files. So `--inflate-log` may be working ok but the output with `log_store_compressed=1` is corrupt.

In a cygwin terminal window:
```
./fio/fio.exe --name=test --ioengine=null --size=1M --log_store_compressed=1 --write_bw_log=test1M --per_job_log=0 --log_offset=1
fio: this platform does not support process shared mutexes, forcing use of threads. Use the 'thread' option to get rid of this warning.
test: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=null, iodepth=1
fio-3.32-12-ge577
Starting 1 thread
 
test: (groupid=0, jobs=1): err= 0: pid=7416: Thu Sep 1 12:58:04 2022
  read: IOPS=256k, BW=1000MiB/s (1049MB/s)(1024KiB/1msec)
    clat (nsec): min=28, max=965, avg=37.87, stdev=58.24
     lat (nsec): min=60, max=1631, avg=71.24, stdev=97.94
    clat percentiles (nsec):
     |  1.00th=[   30],  5.00th=[   30], 10.00th=[   31], 20.00th=[   33],
     | 30.00th=[   34], 40.00th=[   34], 50.00th=[   35], 60.00th=[   35],
     | 70.00th=[   35], 80.00th=[   35], 90.00th=[   36], 95.00th=[   37],
     | 99.00th=[   38], 99.50th=[   68], 99.90th=[  964], 99.95th=[  964],
     | 99.99th=[  964]
   bw (  MiB/s): min= 4145, max=142857, per=100.00%, avg=116961.08, stdev=10205.06, samples=256
  lat (nsec)   : 50=99.22%, 100=0.39%, 1000=0.39%
     submit    : 0=0.0%, 4=0.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
     complete  : 0=0.0%, 4=0.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
     issued rwts: total=256,0,0,0 short=0,0,0,0 dropped=0,0,0,0
 
   READ: bw=1000MiB/s (1049MB/s), 1000MiB/s-1000MiB/s (1049MB/s-1049MB/s), io=1024KiB (1049kB), run=1-1msec

$ fio --inflate-log test1M_bw.log.fz
fio: failed inflating log: -3
```

-3 is Z_DATA_ERROR

aggieNick02:
> I did some quick troubleshooting on Windows and the results suggest that Windows fio actually produces corrupt compressed log files. So `--inflate-log` may be working ok but the output with `log_store_compressed=1` is corrupt.

That's really interesting. I'm at a bit of a loss as to how everything is working when `log_store_compressed=0` just because all `log_store_compressed=1` does is take the already compressed chunks and push them to a file. 😕 

I also tried taking a compressed *.fz file from a run on linux and bringing it over to windows to inflate it. It worked without error, which further backs up your suggestion that the fio is producing corrupt *.fz files on windows.

(As an aside, it probably isn't the best idea to inflate *.fz files on a different machine, just because the files are made of compressed chunks of in-memory structures, so endian issues *could* come into play. But the experiment is good for showing inflation works on windows.)

(Further, it would be awesome (and hopefully I/we can add one day) to have an option to compress a protobuf version of the log file entries. You could then open and process per I/O log files from any machine and any language without having to worry about inflating (can take a long time and makes big files), in-memory format changes to the log structure, endian-ness,etc.)

aggieNick02:
I was doing a little more investigation. I'm not 100% sure the inflate code is in the clear anymore. I could actually see there being problems with both.

Anyways, I was looking at the results of using fio on windows to inflate the linux run. It looks sane for a while, then it doesn't:
```
27, 595781818, 1, 131072, 984743936, 0
27, 834853503, 1, 131072, 984875008, 0
27, 970903703, 1, 131072, 985006080, 0
27, 560136752, 1, 131072, 8601993216, 0
0, 1769472, 135266304, 140737488420864, 16486077346873344, 132
452984832, 17315143742390272, 16777216, 4220435800799576064, 115964116992, 0
0, -9223372032559808512, 0, 29686813950010, 9513871805006217216, 0
```
I'll see what I can figure out this afternoon. If there's not a quick and obvious fix, it might make sense to disable log compression on windows until the issue or issues are resolved.

aggieNick02:
The following isn't always happening, but it may be a sign of the underlying bug, and is definitely wrong. Adding --debug=compress, I've had small runs like this:
```
fio.exe --name=job --per_job_logs=0 --write_bw_log=fio_0 --ioengine=null --direct=1 --size=5120K --rw=write --bs=128K --iodepth=1 --log_offset=1 --log_compression=10K --debug=compress --log_store_compressed=1
...
compress 688   deflate input size=1600, seq=1, log=fio_0_bw.log.fz
compress 688   seq=1, chunk=297
compress 688   deflated to size=297
compress 688   log write chunk size=297, seq=1
```
and then I go check the file size on disk, and it is different - 299 bytes.

I'll also see corruption when the size on disk matches, but the fact that the size sometimes doesn't is definitely wrong.

aggieNick02:
I got cygwin installed so I can build and print-debug on windows. The behavior I'm seeing is wacky.
I added debug code to print out the binary contents of `ic->buf` in `inflate_gz_chunks`, and made the run extra small. Fio's debug output indicated that it wrote a single chunk of size 337 bytes. But the file on disk is 339 bytes.

Comparing the buffer contents to the file contents, the 2 spurious bytes are the `0D` bytes indicated in the picture below, oddly both interrupting `C1 80 0A` after the `80` and before the `0A`. I guess I'll now know these by heart - `0A` is `\n`, `0D` is `\r`.

image

So linux line-endings are being converted to windows line-endings. I'm not a C person so I had to go look things up, but we are opening the output file in text mode, which is causing this. So that explains the problem when the file sizes are different.

There's also the issue from a couple posts above:
```
27, 595781818, 1, 131072, 984743936, 0
27, 834853503, 1, 131072, 984875008, 0
27, 970903703, 1, 131072, 985006080, 0
27, 560136752, 1, 131072, 8601993216, 0
0, 1769472, 135266304, 140737488420864, 16486077346873344, 132
452984832, 17315143742390272, 16777216, 4220435800799576064, 115964116992, 0
0, -9223372032559808512, 0, 29686813950010, 9513871805006217216, 0
```
and that happens when file size equals buffer size, so something else is going on there. I'll wrap up this post and investigate that separately.

vincentkfu:
Great work!

axboe:
I second that! As soon as I read 0d/0a I knew exactly what it was, but teasing that out is the important part. Looking forward to your fix.

aggieNick02:
Thanks.

I think I've got the same-size "corruption" figured out as well. I had an `.fz` file from windows that would not inflate on windows, but inflated fine (and sanely) on linux. I'm guessing read tries to do the same kind of tricks write does w.r.t text files and line endings, and the file had a `0D`. Changing the file open in the inflate to `'rb'` got it working.

Annoyingly, while debugging that, I had one time where my debugging output of the buffer contents was missing a line (I was printing a byte per line). That is, `len` was 338, the file size was 338, but my dprints indicated there should have been 337 bytes, with an `85` byte missing from the dprints at offset 327...

None of that makes *any* sense - it's literally a for loop set to iterate 338 times that prints 337 - I've triple checked it though, and that's what happened. I don't think it's related at all to these issues, and wish I hadn't seen it so I could sleep better. ;-)

axboe:
Fresh eyes are always better - but yes, it can be tough to walk away in situations like that.

aggieNick02:
Hmm, I'm wondering if the fio live status could cause a print to be missed like that. That seems like the most plausible explanation... this is the end of the file on disk:
```
d5 7d 85 28 77 6b 7f eb 0f c6 4c 57 f9
```
and the tail of my odd output, with the comment about the missing byte location manually added:

```
compress 3116  d5
compress 3116  7d
<!!! - missing 85 byte - !!!>
compress 3116  28
compress 3116  77
compress 3116  6b
compress 3116  7f
compress 3116  eb
compress 3116  c6
compress 3116  4c
compress 3116  57
compress 3116  f9
compress 3116  log write chunk size=338, seq=1
!!!ok

job: (groupid=0, jobs=1): err= 0: pid=3116: Fri Sep 2 21:49:38 2022
  write: IOPS=40.0k, BW=5000MiB/s (5243MB/s)(5120KiB/1msec); 0 zone resets
    clat (nsec): min=18, max=568, avg=48.57, stdev=85.83
     lat (nsec): min=452, max=13169, avg=1030.15, stdev=1982.19
    clat percentiles (nsec):
     |  1.00th=[   18],  5.00th=[   19], 10.00th=[   21], 20.00th=[   24],
     | 30.00th=[   26], 40.00th=[   28], 50.00th=[   31], 60.00th=[   34],
     | 70.00th=[   41], 80.00th=[   44], 90.00th=[   51], 95.00th=[   57],
     | 99.00th=[  572], 99.50th=[  572], 99.90th=[  572], 99.95th=[  572],
     | 99.99th=[  572]
   bw (  MiB/s): min=72362, max=4129032, per=-41943.04%, avg=2191439.91, stdev=1241030.99, samples=40
  lat (nsec)   : 20=5.00%, 50=80.00%, 100=10.00%, 250=2.50%, 750=2.50%
     submit    : 0=0.0%, 4=0.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
     complete  : 0=0.0%, 4=0.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
     issued rwts: total=0,40,0,0 short=0,0,0,0 dropped=0,0,0,0

  WRITE: bw=5000MiB/s (5243MB/s), 4096MiB/s-5000MiB/s (4295MB/s-5243MB/s), io=5120KiB (5243kB), run=1-1msec
```

PR #1468, opened 2022-09-09
Change log_compression option from int to bool

aggieNick02:
The log_compression option no longer has an effect on when a chunk of log entries is compressed in the background, but instead functions as a boolean for whether or not to perform the compression. It was also used to decide decompression chunk size after a job, but we can just use a fixed chunk size as is done for the fio --inflate-log option.

As such, change the option from int to bool to avoid confusion. Note that any existing usage of this option will need to be updated to specify a bool instead of an int.

CONFIRMED
2. Empty second line
4. Empty line

### DIFF
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -3786,18 +3786,18 @@ Measurements and reporting
 	entry as well as the other data values. Defaults to 0 meaning that
 	offsets are not present in logs. Also see `Log File Formats`_.
 
-.. option:: log_compression=int
+.. option:: log_compression=bool
 
 	If this is set, fio will compress the I/O logs as it goes, to keep the
-	memory footprint lower. When a log reaches the specified size, that chunk is
-	removed and compressed in the background. Given that I/O logs are fairly
-	highly compressible, this yields a nice memory savings for longer runs. The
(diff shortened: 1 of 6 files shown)


aggieNick02:
Oops, you'd think I'd follow my own commit warning. I'll update the test and repush. Sorry about that.

axboe:
> Note that any existing usage of this option will need to be updated to specify a bool instead of an int.

That is a non-starter, you can't just break existing job files like that.

aggieNick02:
> > Note that any existing usage of this option will need to be updated to specify a bool instead of an int.
> 
> That is a non-starter, you can't just break existing job files like that.

Ah, ok. Would you be ok with deprecation? If so, any input on what the new boolean option would be called? Maybe `log_do_compression`?

axboe:
Or maybe ```log_compression_on```? Something like that, and then you can have that toggle the other setting too, if that makes it more handy that way.

aggieNick02:
> Or maybe `log_compression_on`? Something like that, and then you can have that toggle the other setting too, if that makes it more handy that way.

Sorry, which other setting?

aggieNick02:
Opened a new PR with deprecation instead.

PR #1471, opened 2022-09-21
Add log_compression_on option and deprecate log_compression

aggieNick02:
(Second attempt at this. First one just eliminated the old option. This keeps the old option but deprecates, allowing existing jobs to continue working.)

The log_compression option no longer has an effect on when a chunk of log entries is compressed in the background, but instead functions as a boolean for whether or not to perform the compression. It was also used to decide decompression chunk size after a job, but we can just use a fixed chunk size as is done for the fio --inflate-log option.

As such, deprecate the current option (to keep existing jobs working), and add a new bool option named log_compression_on.

CONFIRMED

2. Empty second line
4. Empty line


aggieNick02:
another pr addressed the issues discussed here

PR #1487, opened 2022-11-13
Improve the configure script and os/os.h

bvanassche:


ankit-sam (approved):
Changes looks good, I am to able to see the error with the flag.

PR #1489, opened 2022-11-17
Add cpu/op and CPU usr/sys ms to json

useche:
Add cpu/op and CPU usr/sys ms to json

Having these times might be useful to calculate any events waited by the CPU time (e.g. CPU/op).

Signed-off-by: Luis Useche <useche@gmail.com>

### DIFF
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -4187,6 +4187,7 @@ group) the output looks like::
 	  lat (msec)   : 2=4.16%, 4=1.84%, 10=4.90%, 20=11.33%, 50=5.37%
 	  lat (msec)   : 100=0.65%
 	  cpu          : usr=0.27%, sys=0.18%, ctx=12072, majf=0, minf=21
+	  cpu/op       : 42.70 (cpu usec/op)
 	  IO depths    : 1=85.0%, 2=13.1%, 4=1.8%, 8=0.1%, 16=0.0%, 32=0.0%, >=64=0.0%
 	     submit    : 0=0.0%, 4=100.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
 	     complete  : 0=0.0%, 4=100.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
@@ -4260,6 +4261,11 @@ writes in the example above).  In the order listed, they denote:
 		finally the number of major and minor page faults. The CPU utilization
 		numbers are averages for the jobs in that reporting group, while the
 		context and fault counters are summed.
+**cpu/op**
+                Average CPU time per operation. The CPU time includes user and
+                system CPU times for this thread. All operations in this thread
+                are included: read, write, and trims. The unit of the metric is
+                the CPU time in microseconds per operation (usec/op).
 
 **IO depths**
 		The distribution of I/O depths over the job lifetime.  The numbers are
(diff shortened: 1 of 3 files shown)


ammarfaizi2:
The second commit is missing a `Signed-off-by` tag.

useche:
> 

Done

vincentkfu:
It seems like (CPU ms) / op can actually be calculated based on the values already reported in the normal output. So having fio do this calculation and reporting the result is merely for convenience. Can you provide a more detailed justification explaining why it is useful to report this value?

The label in the normal output is "cpu msec/op" but the actual calculation is:

`cpu_per_op = (double) ts->total_complete / (usr_cpu + sys_cpu);`

These two are inconsistent and the denominator is the sum of two percentages.

If you are adding a field to the normal output please add the same field to the JSON output. I think it's fine for the JSON output to have more information than the normal output but it shouldn't have less.

With changes in the output fio.1 and HOWTO should be updated as well.

useche:
Hi Vincent. Thanks for taking a look. 

> It seems like (CPU ms) / op can actually be calculated based on the values already reported in the normal output. So having fio do this calculation and reporting the result is merely for convenience. Can you provide a more detailed justification explaining why it is useful to report this value?

Although the total number of I/Os is available, the CPU time in milliseconds is not. The CPU in milliseconds is necessary to correctly make this calculation.

> The label in the normal output is "cpu msec/op" but the actual calculation is:
> 
> `cpu_per_op = (double) ts->total_complete / (usr_cpu + sys_cpu);`
> 
> These two are inconsistent and the denominator is the sum of two percentages.

You are right. I'm not sure what happened. The code that makes sense is up now.

> If you are adding a field to the normal output please add the same field to the JSON output. I think it's fine for the JSON output to have more information than the normal output but it shouldn't have less.

Done.

> With changes in the output fio.1 and HOWTO should be updated as well.

Done

vincentkfu:
> Hi Vincent. Thanks for taking a look.
> 
> > It seems like (CPU ms) / op can actually be calculated based on the values already reported in the normal output. So having fio do this calculation and reporting the result is merely for convenience. Can you provide a more detailed justification explaining why it is useful to report this value?
> 
> Although the total number of I/Os is available, the CPU time in milliseconds is not. The CPU in milliseconds is necessary to correctly make this calculation.
> 

CPU milliseconds can be obtained by multiplying runtime by the sum of the user and system CPU usage percentages.

useche:
> > Hi Vincent. Thanks for taking a look.
> > > It seems like (CPU ms) / op can actually be calculated based on the values already reported in the normal output. So having fio do this calculation and reporting the result is merely for convenience. Can you provide a more detailed justification explaining why it is useful to report this value?
> > 
> > 
> > Although the total number of I/Os is available, the CPU time in milliseconds is not. The CPU in milliseconds is necessary to correctly make this calculation.
> 
> CPU milliseconds can be obtained by multiplying runtime by the sum of the user and system CPU usage percentages.

You are right, we can obtain the CPU time using this calculation. Then, it will be just for convenience for one time runs. I was more interested in the json one for automation, but we can get the CPU milliseconds with this calculation too. With that, it sounds like these patches are not that useful. If you think that too, we may abandon these patches.

vincentkfu:
I don't have any specific objection to adding this value but if you can obtain what you need with no code changes, then I see no compelling need for these patches. Moreover, it seems like fio's output is already complicated enough that it was difficult for a motivated person to find the runtime. In that case we should be wary of making it even more complicated.

PR #1497, opened 2023-01-20
tools/fiograph: accommodate job files not ending in .fio

vincentkfu:
For job files not ending in .fio, fiograph will overwrite the job file with a graphviz script and then delete it if --keep is not specified.

Fix this by creating temporary files to contain the graphviz script and image file. Then rename or delete the script and image file as directed by the user specified options.

Signed-off-by: Vincent Fu <vincent.fu@samsung.com>

### DIFF
--- a/tools/fiograph/fiograph.py
+++ b/tools/fiograph/fiograph.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python3
+import uuid
+import time
 import errno
 from graphviz import Digraph
 import argparse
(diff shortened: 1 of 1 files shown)


vincentkfu:
@ErwanAliasr1 do you have any thoughts about this?

The `time.sleep(1)` is not nice to have and the file that is opened with `--view` will have a strange name but this is a bug that we have to fix.

ErwanAliasr1 (commented):
I see two possible way to fix this : 

- making a fault if job file is not ending by fio
- Using a tempfile with the filename as prefix : https://docs.python.org/3/library/tempfile.html

I'd prefer the 2nd option which sounds more flexible for users, wdyt ?

vincentkfu:
I pushed a slightly modified version of the patch here: https://git.kernel.dk/cgit/fio/commit/?id=13bc47b65d32df6fd212c4687c7fd29b4ab7c09d

Feel free to follow up if you have ideas for improvements.

ErwanAliasr1:
Sorry I didn't came back to you :(

Le lun. 23 janv. 2023 à 21:15, Vincent Kang Fu ***@***.***> a
écrit :

PR #1498, opened 2023-01-21
Fix libaio engine with nowait option

aegiryy:
This PR fixed an issue when `nowait` option is used for the libaio engine.
Specifically, the completion events can contain an EAGAIN error, which
will fail fio in the current implementation.  This change requeues the I/O
request.

### DIFF
--- a/io_u.c
+++ b/io_u.c
@@ -2069,6 +2069,10 @@ static void io_completed(struct thread_data *td, struct io_u **io_u_ptr,
 				icd->error = ret;
 		}
 	} else if (io_u->error) {
+		if (io_u->error == EAGAIN) {
+			requeue_io_u(td, io_u_ptr);
+			return;
+		}
 		icd->error = io_u->error;
 		io_u_log_error(td, io_u);
 	}


axboe:
This should be handled in the libaio engine directly.

Also, your commit formatting is all wrong. Please look at other commits on how to do it, lines are too long and you're missing a signed-off-by line with your real name and email.

axboe:
It's also not clear if this is the right thing to do. If nowait is used, then IO is issued with RWF_NOWAIT. If you get an EAGAIN error from that, it doesn't mean that a retry under the same conditions will succeed. What it means is that "I cannot do this in a non-blocking manner, you should retry the operation without this flag from a context where it is safe to do so". If you just requeue the IO, you'll be retrying it in the same manner as before. You could get tons of EAGAIN, or even perpetual ones, if you just blindly retry.

axboe:
Now, for storage devices, the EAGAIN is most likely because we ran out of requests, in which case a retry will probably eventually succeed. But it may take tons of retries to get there, where you're just burning CPU meanwhile.

For other types of reads, the EAGAIN will indeed be fatal and no amount of retries will help that.

So in short, I don't think your patch is correct.

aegiryy:
The situation I hit is that the kernel runs out of the hardware tags when queueing the request, so retrying the same operation will eventually succeed (but it may take many retries as you stated).  Is there a situation where retry makes no progress?  If so, can you point out what it might be?  The current implementation of "nowait" option will simply report an error and then exit, which I believe is not the desirable behavior either?

axboe:
Yep, that's exactly the one case I outlined. So it will work for that case, but in a bad way as it may retry a ton of times before succeeding.

One case where it would not work at all and just sit in a loop repeatedly retrying without ever succeeding, is if you do buffered reads or write with libaio. That may not make a lot of sense as libaio doesn't support buffered io (it'll just be sync), but it'll break with your patch as fio will just requeue and retry without ever making progress.

'nowait' does what it's supposed to, RWF_NOWAIT based IO. That has limitations and requirements that fio can't always meet, I'm not convinced that working around them is sensical. And at least the current patch is known broken for several cases.

aegiryy:
Thanks for your elaboration!  I wasn’t expecting buffered I/O with libaio, but you made a good point here.  On the other hand, RWF_NOWAIT based I/O requires the handling of EAGAIN, which is not properly done in the current implementation (and hence is incomplete IMHO).  Do you have any suggestion on how to handle it in a more principled manner?

axboe:
I think what you're looking for isn't currently possible. You would need to expand the 'nowait' parameter to be a ```FIO_OPT_STR``` type with "0"/"1" being posvals that you can set now, and then add a "retry" option as well that then tells the libaio engine to requeue without RWF_NOWAIT set if it sees EAGAIN. The reason that has to be a new parameter value is that you could potentially break existing jobs that use 'nowait' if they rely on getting EAGAIN if the IO cannot be done in a nonblocking fashion. The change you currently have isn't compatible with that, nor would a change that didn't set RWF_NOWAIT on a retry be.

xinyangge-db:
> The reason that has to be a new parameter value is that you could potentially break existing jobs that use 'nowait' if they rely on getting EAGAIN if the IO cannot be done in a nonblocking fashion.

Interesting.  Is this the only current use case of "nowait"?  I asked because, if no EAGAIN is encountered, then it doesn't make a difference whether "nowait" is specified or not.

Anyway, based on what you described, handling EAGAIN on RWF_NOWAIT is not as valuable as I originally thought.  Please feel free to close this PR.  Thanks for the discussion!

aegiryy:
> The reason that has to be a new parameter value is that you could potentially break existing jobs that use 'nowait' if they rely on getting EAGAIN if the IO cannot be done in a nonblocking fashion.

Interesting.  Is probing whether the IO can be done in a nonblocking fashion the only use case of "nowait"?  I asked because, if no EAGAIN is encountered, then it doesn't make a difference whether "nowait" is specified or not.  (but if EAGAIN is actually hit, the benchmark will always fail)

Anyway, based on what you described, handling EAGAIN on RWF_NOWAIT is not as valuable as I originally thought.  Please feel free to discard the PR.  Thanks for the discussion!

PR #1505, opened 2023-02-04
Clarify verify_backlog wont verify data on reads issued by mixed work…

horshack-dpreview:
…loads

The `verify_backlog` option provides for continual verification of data written during the workload, as contrasted with post-workload verifications. For mixed read/write workloads, any reads generated by the workload itself will not verify the data read, even for reads falling on blocks that have been written by the workload. Only reads generated by `verify_backlog` will verify the data read. This nuance isn't obvious to users who haven't looked at the source code and so may overestimate the level of data integrity checking being employed for their workloads. I've added a sentence to the HOWTO to explain this distinction.

Tangentially related, `verify_backlog_batch` HOWTO states "if verify_backlog_batch is larger than verify_backlog, some blocks will be verified more than once." I don't believe this is correct - when a verify is issued the associated write io_u is pulled from history and there is no provision to repeat the verifying read. I'd like to remove this sentence.


horshack-dpreview:
Upon further review the verify feature has additional non-obvious configuration-specific nuances that also affect how much data is verified. Rather than muddy up each option-specific description with these details perhaps it's better that I write a preamble in the verification section to capture them all? Or perhaps a completely separate data integrity HOWTO?

axboe:
Data integrity testing is something that people often get wrong, hence I do think that a section on how to do it in the HOWTO would be potentially very useful.

Issue #1501, opened 2023-01-30
Sequential benchmarks are cpu-constrained due to single-threaded nature of jobs


horshack-dpreview:
**Description of the new feature**
Wasn't sure whether to file this as an issue or feature request. From reading the documentation it's my understanding `fio` only supports a single process/thread per job for I/O submission and completion handling. For fast devices such as NVMe this limits the potential IOPs since the workload can be processor core-constrained. The existing solution is to specify multiple concurrent jobs, as each will be run in a separate process/thread. This is fine for random workloads since there is no dependency or correlation of block addresses/offsets between jobs. However this isn't viable for sequential workloads because with each job using its own sequential block addresses/offsets, the device sees a pseudo-sequential workload that has duplicate/overlapping block addresses/offsets. This causes temporally-close requests to be serviced out of the device's cache rather than from its storage media, yielding inaccurate results in terms of what the device is actually capable of. For example, if we have a two-job 64KB sequential read workload, the device will see the following requests from `fio`:
```
Read @ LBA 00K, size = 64K [from job #1]
Read @ LBA 00K, size = 64K [from job #2]
Read @ LBA 64K, size = 64K [from job #1]
Read @ LBA 64K, size = 64K [from job #2]
etc...
```
Would you consider of of the following two solutions? One is to add support for multiple threads per job, with the threads sharing a single sequential block/offset variable. An alternate solution would be to implement this using the existing multiple-job mechanism and providing an option to share/sync the block/offset variable across the processes of the jobs.

If I'm overlooking an existing solution to this scenario then I apologize in advance.

vincentkfu:
Work `io_submit_mode=offload` work for your use case? It's limited to synchronous ioengines. Also consider running with `--gtod_reduce=1` to reduce the amount of work fio does for each I/O. There is also `t/io_uring` which can be faster but is not as flexible as fio.

horshack-dpreview:
@vincentkfu Thanks for the reply and great suggestion on using the offload mechanism. I've started to play with it now. The offload docs say "this option cannot reliably be used with async IO engines." Any insight into what issues I'd face using an async engine with offload?

horshack-dpreview:
@vincentkfu, Here are the results using a matrix of single-threaded inline vs offline I/O submission as compared to multi-threaded, both with and without gtod_reduce. I also threw in an ioengine comparison.
```
IOPs    iodepth     numjobs     io_submit_mode  ioengine    gtod_reduce

[...]

PR #1508, opened 2023-02-06
Improve IOPs 50% by avoiding clock sampling when rate options not used

horshack-dpreview:
Profiling revealed thread_main() is spending 50% of its time in calls to utime_since_now() from rate_ddir(). This call is only necessary if the user specified a rate option for the job. A conditional was added to avoid the call if !should_check_rate(). See this link for details and profiling data:

https://github.com/axboe/fio/issues/1501#issuecomment-1418327049

### DIFF
--- a/io_u.c
+++ b/io_u.c
@@ -785,7 +785,15 @@ static enum fio_ddir get_rw_ddir(struct thread_data *td)
 	else
 		ddir = DDIR_INVAL;
 
-	td->rwmix_ddir = rate_ddir(td, ddir);
+	if (!should_check_rate(td)) {
+		/*
+		 * avoid time-consuming call to utime_since_now() if rate checking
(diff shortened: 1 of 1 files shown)


axboe:
Nice catch! But let's move that check into get_rw_ddir() - and when you do, please include braces when you have comments in an if statement. Makes it easier to read.

axboe:
Room for some cleanups in that area, but that's not impacted by your change. Thanks!

horshack-dpreview:
@axboe, Thanks! When you have time can you comment on my original issue of `fio` not scaling for sequential workloads and my proposed solutions? Of the two I think sharing a common sequential offset between jobs would be less impactful. Link:

https://github.com/axboe/fio/issues/1501

PR #1510, opened 2023-02-07
fio: steadystate: allow for custom check interval

cloehle:
Allow for a different steady state check interval than 1s with a new --ss_interval parameter.

Steady state is reached when the steady state condition (like slope) is true when comparing the last windows (set with --ss_dur). The actual values for this comparison is currently calculated for a 1s interval during the window.
This is especially problematic for slow random devices, where the values do not converge for such a fine granularity.
Letting the user set this solves this problem, although requires them figuring out an appropriate value themselves.

--ss=iops:5% --ss_dur=120s should reproduce this for some slow SD card.
Then adding like --ss_interval=20s may let it converge.


init.c
```diff
@@ -73,6 +73,9 @@ bool write_bw_log = false;
 bool read_only = false;
 int status_interval = 0;
 
+int ss_check_interval = STEADYSTATE_MSEC;
+
```
axboe, line 77: One extra empty line here

thread_options.h
```diff
@@ -211,6 +211,7 @@ struct thread_options {
 	fio_fp64_t ss_limit;
 	unsigned long long ss_dur;
 	unsigned long long ss_ramp_time;
+	unsigned long long ss_check_interval;
```
axboe, line 214: This needs a FIO_SERVER_VER bump, and cconv.c additions.

axboe:
Minor nits, but you also need to update the HOWTO and fio man page for this option. It can't be undocumented.

cloehle:
Thanks would you like commits on top or rebase / force push for additions? Merged PRs seem ambiguous regarding this.

axboe:
I don't want fixups on top for a pull request, so please just amend what you already have.

helper_thread.c
```diff
@@ -281,7 +281,7 @@ static void *helper_thread_main(void *data)
 		},
 		{
 			.name = "steadystate",
-			.interval_ms = steadystate_enabled ? STEADYSTATE_MSEC :
+			.interval_ms = steadystate_enabled ? ss_check_interval :
```
vincentkfu, line 284: In the documentation note that this forces all jobs in a single fio invocation to have the same `ss_check_interval`.

Probably we should have a check for this in `fixup_options()` to make sure that all jobs with steady state detection enabled are using the same interval.

options.c
```diff
@@ -5185,6 +5185,20 @@ struct fio_option fio_options[FIO_MAX_OPTS] = {
...
 	},
+        {
+		.name   = "steadystate_check_interval",
+		.lname  = "Steady state check interval",
+		.alias  = "ss_interval",
+		.parent	= "steadystate",
+		.type   = FIO_OPT_STR_VAL_TIME,
+		.off1   = offsetof(struct thread_options, ss_check_interval),
+		.help   = "Polling interval for the steady state check (too low means steadystate will not converge)",
+		.def    = "0",
```
vincentkfu, line 5196: Consider making the default 1 and just getting rid of STEADYSTATE_MSEC

vincentkfu (changes requested):
There is also a line in steadystate.c with the hard-coded value 1000000L preceded by the comment "Begin recording data one second after ss->ramp_time has elapsed".

This should be changed to record the first IOPS/BW value `ss_check_interval` seconds past the end of `ss->ramp_time`.

vincentkfu:
Please check the CI failures. Apologies for the poor state of the steady state test script. It has been on my agenda to improve it but I haven't had the time.

It would be nice to have some sort of simple test for this feature as well.

helper_thread.c
cloehle: I can add that if you want, but that should apply to the other steadystate options too, right? (and currently there is no fixup for those)

vincentkfu: In `td_steadystate_init` is a check for the other options for consistency within a reporting group. Conceptually it's reasonable to run multiple jobs concurrently and allow them to have different stopping criteria.

`ss_check_interval` has to be a global option because it affects the helper thread that triggers the steady state check for all running jobs.

cloehle:
Please review.
I added a test for the parameter, but of course it doesn't test what this parameter is actually works, just that setting it doesn't mess anything up.
Personally I could use (on my dedicated hardware):
`./fio --name=test --thread --ioengine=null --size=1G --rw=randread --rwmixread=100 --group_reporting --numjobs=1 --time_based --runtime=2000 --ss_dur=90 --ss=iops:1% --ss_ramp=10 --debug=steadystate --ss_interval=30`
(ss attained)
vs:
`./fio --name=test --thread --ioengine=null --size=1G --rw=randread --rwmixread=100 --group_reporting --numjobs=1 --time_based --runtime=600 --ss_dur=90 --ss=iops:1% --ss_ramp=10 --debug=steadystate`
(ss not attained for me, so jobtime=runtime)
But of course that won't really work depending on what the CI is running on.

I can't really think of anything right now that doesn't involve real hardware or adding an ioengine just for this test, but if you insist on a feature test I will give it some more thought or add a slowwithspikesioengine.

cloehle:
Is the failed AppVeyor CI a false positive? ratio looks fine to me? Some float conversion issue?

init.c
```diff
@@ -73,6 +73,8 @@ bool write_bw_log = false;
 bool read_only = false;
 int status_interval = 0;
 
+int ss_check_interval = 1000;
```
vincentkfu, line 76: The new `thread_options` member is unsigned but this is signed. Wouldn't it better to make this unsigned as well for the sake of consistency?

Also, there are multiple places where this is divided by or multiplied by 1000. Consider making this `ss_check_interval_us` so that the units are clear and we can avoid some arithmetic.

HOWTO.rst
```diff
@@ -3826,6 +3826,13 @@ Steady state
 	collection for checking the steady state job termination criterion. The
 	default is 0.  When the unit is omitted, the value is interpreted in seconds.
 
+.. option:: steadystate_check_interval=time, ss_interval=time
```
vincentkfu, line 3829: Please also update the description for `ss_dur` to no longer state that performance will be measured once per second.

HOWTO.rst
```diff
@@ -3826,6 +3826,13 @@ Steady state
...
+
+	The values suring the rolling window will be collected with a period
```
vincentkfu, line 3831: during instead of suring. Also I think it would be useful to clarify that if `ss_dur`=60s and the check interval is 6s only 10 measurements will be collected.

vincentkfu (changes requested):

vincentkfu:
> Is the failed AppVeyor CI a false positive? ratio looks fine to me? Some float conversion issue?

The test failure is unrelated to your patch.

init.c
vincentkfu: I would move this declaration next to where `steadystate_enabled` is declared in steadystate.c in order to have all of this in the same place.

cloehle:
Thanks for the review and please re-review, hopefully I addressed everything you mentioned.
I did not change ss_check_interval to ss_check_interval_us although it is mostly used as such, mainly because rate_time is in ms and the calculation already looks weird enough, but if you prefer it anyway, I can.

cloehle:
My bad, I assumed I could get around rebasing since there was no conflict but the for_each_td refactor of course broke the build. Fixed now

axboe:
> My bad, I assumed I could get around rebasing since there was no conflict but the for_each_td refactor of course broke the build. Fixed now

No worries, it could also have been fixed up during merging. But for this particular case, as the CI will run on the merge AND it merges cleanly, then it's easier to just rebase it.

steadystate.c
```diff
@@ -312,6 +319,7 @@ int td_steadystate_init(struct thread_data *td)
 		ss->dur = o->ss_dur;
 		ss->limit = o->ss_limit.u.f;
 		ss->ramp_time = o->ss_ramp_time;
+		ss_check_interval = o->ss_check_interval / 1000L;
```
axboe, line 322: Same question here, can o->ss_check_interval ever be < 1000?

cloehle:
Added a ss_check_interval >=1s check to address Jens' comments.
Generally there is no use for lower values, if 1s works for you then use that, if not increase ss_check_interval until you get a converging steadystate, would be my argument.

vincentkfu:
I think there's still something missing as I applied the patch to fio 3.34, ran with `ss_dur=10` and `ss_interval=2s` but ended up with 10 (instead of 5) IOPS and bw measurements:

```
./fio --name=test --filesize=1M --steadystate=iops_slope:1% --ss_dur=10s --ss_interval=2s --ioengine=null --time_based=1 --runtime=30s --output-format=json
{
  "fio version" : "fio-3.34-2-g0f4e",
  "timestamp" : 1678836467,
  "timestamp_ms" : 1678836467417,
  "time" : "Tue Mar 14 19:27:47 2023",
  "jobs" : [
    {
      "jobname" : "test",
      "groupid" : 0,
      "error" : 0,
      "eta" : 0,
      "elapsed" : 31,
      "job options" : {
        "name" : "test",
        "filesize" : "1M",
        "steadystate" : "iops_slope:1%",
        "steadystate_duration" : "10s",
        "steadystate_check_interval" : "2s",
        "ioengine" : "null",
        "time_based" : "1",
        "runtime" : "30s"
      },
...
      "steadystate" : {
        "ss" : "iops_slope:1.000000%",
        "duration" : 10,
        "attained" : 0,
        "criterion" : "56.055508%",
        "max_deviation" : 0.000000,
        "slope" : 2467159.921212,
        "data" : {
          "bw_mean" : 32119104108,
          "iops_mean" : 7841577,
          "iops" : [
            8020594,
            7908355,
            7930974,
            7880179,
            7864311,
            7797434,
            7774733,
            7772509,
            7846982,
            7793071
          ],
          "bw" : [
            32852355383,
            32392624675,
            32485271552,
            32277213511,
            32212221624,
            31938292703,
            31845306368,
            31836198912,
            32141240930,
            31920418947
          ]
        }
      }
    }
  ]
}
```

HOWTO.rst
```diff
@@ -3822,16 +3822,24 @@ Steady state
 .. option:: steadystate_duration=time, ss_dur=time
 
 	A rolling window of this duration will be used to judge whether steady state
-	has been reached. Data will be collected once per second. The default is 0
-	which disables steady state detection.  When the unit is omitted, the
-	value is interpreted in seconds.
+	has been reached. Data will be collected every ss_check_interval.
```
vincentkfu, line 3825: This same change needs to also be made to fio.1

init.c
```diff
@@ -981,6 +981,26 @@ static int fixup_options(struct thread_data *td)
...
 
+	for_each_td(td2) {
+		if (td->o.ss_check_interval != td2->o.ss_check_interval) {
+			log_err("fio: conflicting ss_check_interval: %llu and %llu, must be globally equal\n",
+					td->o.ss_check_interval, td2->o.ss_check_interval);
+			ret |= 1;
+		}
+	} end_for_each();
+	if (td->o.ss_dur && td->o.ss_check_interval / 1000L < 1000) {
+		log_err("fio: ss_check_interval %llu must be at least 1s,\n",
```
vincentkfu, line 992: drop the comma at the end of the error message or replace it with a period. might as well include the usec units in the error message as well since `fio: ss_check_interval 100000 must be at least 1s,` is confusing with no units

init.c
```diff
@@ -981,6 +981,26 @@ static int fixup_options(struct thread_data *td)
...
+				td->o.ss_check_interval);
+		ret |= 1;
+
+	}
+	if (td->o.ss_dur && (td->o.ss_dur % td->o.ss_check_interval != 0 || td->o.ss_dur <= td->o.ss_check_interval)) {
+		log_err("fio: ss_duration %llu must be multiple of ss_check_interval %llu,\n",
```
vincentkfu, line 998: same comment as above

cloehle:
Thanks Vincent, I addressed your comments.

vincentkfu:
I merged this at https://github.com/axboe/fio/commit/90e678ba69ab9bbbfb2136f2d84a7224b72a61cb and messed it up by forgetting to add my sign-off or a merge commit. Sorry about that, @axboe.

Issue #1492, opened 2022-11-28
FIO version 3.33 for Linux sometimes generates erroneous READ statistics during testing of WRITE with verify


george-eldridge:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
About 5% of the time FIO 3.33 for Linux will generate erroneous READ statistics during testing of WRITE with verify. The time of the READ test will be much less than the time of the WRITE test. The amount of data reported for the READ test is 1GB less than amount of data reported for the WRITE test. The READ throughput numbers that are reported are unrealistically high.

**Environment**:
System Information:
Operating System:     Red Hat Enterprise Linux 8.6 (Ootpa)
Kernel Version:       4.18.0-372.9.1.el8.x86_64
Python Version:       3.9.12

**fio version**: fio-3.33

**Reproduction steps**
Create 20 jobs that are 15 minutes each and alternate between READ (rw=read) and WRITE (rw=write) with verify enabled and various block sizes
Run the jobs and collect the output. In some cases of the WRITE test (1 or 2 out of the 20 jobs) there will be erroneous READ statistics produced along with the WRITE statistics.

Sample Output:
====================================================================================================
Run count: 20
11/24/2022, 11:30:17
====================================================================================================
nvme1n1_job_00: (g=0): rw=write, bs=(R) 64.0KiB-64.0KiB, (W) 64.0KiB-64.0KiB, (T) 64.0KiB-64.0KiB, ioengine=libaio, iodepth=8
fio-3.33
Starting 1 thread

nvme1n1_job_00: (groupid=0, jobs=1): err= 0: pid=606650: Thu Nov 24 11:45:17 2022
  read: IOPS=1770k, BW=108GiB/s (116GB/s)(1089GiB/10080msec)
    slat (nsec): min=2520, max=374951, avg=2925.48, stdev=432.99
    clat (usec): min=30, max=26553, avg=337.09, stdev=130.97
     lat (usec): min=34, max=26556, avg=340.01, stdev=130.96
    clat percentiles (usec):
     |  1.00th=[  145],  5.00th=[  153], 10.00th=[  182], 20.00th=[  239],
     | 30.00th=[  265], 40.00th=[  289], 50.00th=[  318], 60.00th=[  351],
     | 70.00th=[  379], 80.00th=[  433], 90.00th=[  506], 95.00th=[  578],
[...]

PR #1514, opened 2023-02-10
Read stats for backlog verifies not reported for time-expired work…

horshack-dpreview:
…loads

When verify_backlog is used on a write-only workload with a runtime= value and the runtime expires before the workload has written its full dataset, the read stats for the backlog verifies are not reported, resulting in a stat result showing only the workload writes (ie, the "read:" results section is completely missing from fio's stats output)

The logic in thread_main() fails to call update_runtime() for DDIR_READ because the existing call to update_runtime() for DDIR_READ on write-only workloads is currently only done after do_verify() is complete, which wont be called in this scenario because td->terminate is true due to the expiration of the runtime.

I'm not up to speed on fio's stats logic so please review this fix carefully. For example, it appears to be ok to call update_runtime() multiple times for a given DDIR_* value, as this is already done for mix read-write workloads that have verify_backlog enabled - once for the backlog reads and again for any remaining verify reads performed by do_verify().

Issue link: https://github.com/axboe/fio/issues/1515

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -1919,7 +1919,8 @@ static void *thread_main(void *data)
 			}
 		} while (1);
 
-		if (td_read(td) && td->io_bytes[DDIR_READ])
+		if (td->io_bytes[DDIR_READ] && (td_read(td) ||
+			((td->flags & TD_F_VER_BACKLOG) && td_write(td))))
 			update_runtime(td, elapsed_us, DDIR_READ);
 		if (td_write(td) && td->io_bytes[DDIR_WRITE])
 			update_runtime(td, elapsed_us, DDIR_WRITE);


horshack-dpreview:
A secondary issue associated with the same root cause of the issue above is incorrect read stats results for the case of a non-timed execution. For example:

```
sudo modprobe brd rd_size=1024000 rd_nr=1
sudo fio -name=test -rw=write -filename=/dev/ram0 -verify=xxhash -verify_backlog=1 -bs=512b -size=100M
```

Yields a read throughput stat of 97 GB/s. That's because the backlog verifies aren't being sampled for the stats but the reads for the post-workload verify are (`do_verify`), but there are no verifies to perform in `do_verify` because  the backlog completed them all, thus the read stats are wildly inaccurate since the completion time of do_verify() is near zero.

My change fixes this secondary issue but the existing sampling does bring up a question. Right now the stats for both the writes and verify-reads issued by `do_io` are using the same `thread_main` `td->start` but that's not really accurate since the writes and reads are sharing the same time quantum. For example, if it takes 1 second to complete the write-only workload with `verify_backlog` issuing reads during the workload the stats will calculate the writes and reads over the same 1 second interval, thus the report stats for each are 1/2 of real-time. In a sense the results are accurate since that's the effective throughput of each when having to share time with the other.

kawasaki:
While I looked into the issue #1492, I found this PR. I took a look in the fix code and it looks good to me. I also tried to run fio with the fix and confirmed that it avoids the problem the issue #1492 reports. I think the fix worth merging to the upstream.

Before merging the fix, I suggest to improve the commit message:
- The last paragraph "I'm not up to speed..." is not appropriate to leave in the commit log.
- Instead of "Issue link:" tag, "Link:" tag will be consistent with other fio commits. ref: [commit 0574e8c](https://github.com/axboe/fio/commit/0574e8c3b2b47e1e2564c2f50ea0b6f2629f2e48)

kawasaki:
> My change fixes this secondary issue but the existing sampling does bring up a question. Right now the stats for both the writes and verify-reads issued by `do_io` are using the same `thread_main` `td->start` but that's not really accurate since the writes and reads are sharing the same time quantum. For example, if it takes 1 second to complete the write-only workload with `verify_backlog` issuing reads during the workload the stats will calculate the writes and reads over the same 1 second interval, thus the report stats for each are 1/2 of real-time. In a sense the results are accurate since that's the effective throughput of each when having to share time with the other.

Though this is a question to the fio maintainers, here I note my opinion: it should be fine to share the same runtime between write and verify-read for verify_backlog case, because write and read already share the same runtime for rw=randrw case. The runtime is mixed, but it is better than no report or weird numbers.

horshack-dpreview:
> > My change fixes this secondary issue but the existing sampling does bring up a question. Right now the stats for both the writes and verify-reads issued by `do_io` are using the same `thread_main` `td->start` but that's not really accurate since the writes and reads are sharing the same time quantum. For example, if it takes 1 second to complete the write-only workload with `verify_backlog` issuing reads during the workload the stats will calculate the writes and reads over the same 1 second interval, thus the report stats for each are 1/2 of real-time. In a sense the results are accurate since that's the effective throughput of each when having to share time with the other.
> 
> Though this is a question to the fio maintainers, here I note my opinion: it should be fine to share the same runtime between write and verify-read for verify_backlog case, because write and read already share the same runtime for rw=randrw case. The runtime is mixed, but it is better than no report or weird numbers.

I agree that reasonable but opaque results are preferable to incorrect or no results.

IMO the complication is how these stats can convey two similar pieces of information - 1) Amount of data transferred over a  period of total time (wall time) and 2) How fast data was transferred over a period of active transfer time. \# 2 implies a measure of a device's performance whereas \# 1 often has just workload observation utility. Complicating this further is potential cross-over between the two measures since one can affect the other even when both aren't actively transferring - for example, write-back writes consuming internal device bandwidth when performing delayed-flushing while servicing ongoing host reads. Since fio can be perceived principally as a benchmarking tool, users likely assume \# 2 even in situations where \# 1 makes more sense when examined.

kawasaki:
I can agree with your comments on the performance measurement comparison (#1 data transferred over total time vs #2 data transferred over active transfer time) and the verify_backlog option adds that misunderstanding possibility. But even with normal verify, I don't think that the performance numbers are suitable for benchmarking. Workloads with verify are different from real use cases. And the reported numbers do not show storage device performance due to the verify-read overhead. Having said that, I think your fix is still valuable to make fio reporting consistent across various combinations of options.

vincentkfu:
Thanks for fixing this.

PR #1516, opened 2023-02-10
Clarify documentation for runtime parameter

horshack-dpreview:
I realize this is highly subjective but I think the description of the runtime parameter could be made a bit more precise. I misinterpreted its meaning after reading the doc and only learned of my mistake by trial and error using fio.  Either I'm just slow or the description could use just a little more precision :)

### DIFF
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -686,10 +686,12 @@ Time related parameters
 
 .. option:: runtime=time
 
-	Tell fio to terminate processing after the specified period of time.  It
-	can be quite hard to determine for how long a specified job will run, so
-	this parameter is handy to cap the total runtime to a given time.  When
-	the unit is omitted, the value is interpreted in seconds.
+	Limit runtime. The test will run until it completes the configured I/O
+	workload or until it has run for this specified amount of time, whichever
+	occurs first. It can be quite hard to determine for how long a specified
+	job will run, so this parameter is handy to cap the total runtime to a
+	given time.  When the unit is omitted, the value is interpreted in
+	seconds.
 
 .. option:: time_based
 
--- a/fio.1
+++ b/fio.1
@@ -471,10 +471,12 @@ See \fB\-\-max\-jobs\fR. Default: 1.
 .SS "Time related parameters"
 .TP
 .BI runtime \fR=\fPtime
-Tell fio to terminate processing after the specified period of time. It
-can be quite hard to determine for how long a specified job will run, so
-this parameter is handy to cap the total runtime to a given time. When
-the unit is omitted, the value is interpreted in seconds.
+Limit runtime. The test will run until it completes the configured I/O
+workload or until it has run for this specified amount of time, whichever
+occurs first. It can be quite hard to determine for how long a specified
+job will run, so this parameter is handy to cap the total runtime to a
+given time.  When the unit is omitted, the value is interpreted in
+seconds.
 .TP
 .BI time_based
 If set, fio will run for the duration of the \fBruntime\fR specified


ankit-sam (commented):
Hi @horshack-dpreview its still confusing to me and probably for others also. Is there a configuration where fio exits before the specified runtime?
I however see that ```thinktime```, ```thinktime_iotime```  can increase the overall ```runtime```, as the fio job maybe sleeping when the ```runtime``` is over. But this needs to be fixed for ```thinktime```.

I tried other non time related options such as ```verify``` etc and there was no change to the runtime.

There are however other time related parameters such as ```ramp_time``` and ```startdelay``` that can affect the total time.

Overall I think documentation for ```runtime``` needs to clarify few things:
1. ```runtime``` should always be used in conjunction with option ```time_based```.
2. ```runtime``` is not the total time for fio run, as total time includes ```runtime, ramp_time, startdelay``` or other time related parameters.

horshack-dpreview:
@ankit-sam, `runtime` caps the execution of the workload the specified time. You can verify this by using a very small value with a large workload config - the test will stop before the workload completes. There may be code paths that don't correctly honor the `runtime` value but those would need to be fixed individually.

As coded, `runtime` would be better named as `runtime_max`, and `runtime` with `time_based` renamed as `runtime_min`. I'm not suggesting these changes for this particular PR.

ankit-sam:
Limit runtime. The test will run until it completes the configured I/O workload or until it has run for this specified amount of time, whichever occurs first.

You have updated the doc to this. When runtime is specified even for small configured workload which gets completed before runtime, fio will and should not terminate. It should continue for the entire runtime duration, like by starting from the begining.

Can you share output log for any such case, where fio exits before the runtime? We should fix that if that's the case.

horshack-dpreview:
> Limit runtime. The test will run until it completes the configured I/O workload or until it has run for this specified amount of time, whichever occurs first.
> 
> You have updated the doc to this. When runtime is specified even for small configured workload which gets completed before runtime, fio will and should not terminate. It should continue for the entire runtime duration, like by starting from the begining.
> 
> Can you share output log for any such case, where fio exits before the runtime? We should fix that if that's the case.

It's working as designed. Specifying a `runtime=x` without `time_based` is supposed to limit the runtime to `x` regardless of workload - that's described in the original documentation. Specifying `runtime=x` with `time_based` will run for the full time specified, regardless of workload, performing multiple iterations of the workload if necessary to reach the time.

Examples:
Setup to create 1GB ramdisk at /dev/ram0: `modprobe brd rd_size=1024000 rd_nr=1`

Cap runtime to 1 second (workload otherwise takes 10 seconds on my system):
`sudo fio -name=test --filename=/dev/ram0 -rw=write -size=1GB -bs=512b -runtime=1`

Run for a minimum of 60 seconds:
`sudo fio -name=test --filename=/dev/ram0 -rw=write -size=1GB -bs=512b -runtime=60 -time_based`

ankit-sam:
> Overall I think documentation for `runtime` needs to clarify few things:
> 
> 1. `runtime` should always be used in conjunction with option `time_based`.
> 2. `runtime` is not the total time for fio run, as total time includes `runtime, ramp_time, startdelay` or other time related parameters.

This is why I mentioned it in the begining, that for fio option runtime the documentation needs to be updated and these two things should go in the document, and everyone should always use ```runtime=x``` in conjunction with ```time_based```.
What you are trying to add is
```The test will run until it completes the configured I/O workload or until it has run for this specified amount of time, whichever occurs first.```
is not right and it will basically create more confusion.

horshack-dpreview:
> > Overall I think documentation for `runtime` needs to clarify few things:
> > 
> > 1. `runtime` should always be used in conjunction with option `time_based`.
> > 2. `runtime` is not the total time for fio run, as total time includes `runtime, ramp_time, startdelay` or other time related parameters.
> 
> This is why I mentioned it in the begining, that for fio option runtime the documentation needs to be updated and these two things should go in the document, and everyone should always use `runtime=x` in conjunction with `time_based`. What you are trying to add is `The test will run until it completes the configured I/O workload or until it has run for this specified amount of time, whichever occurs first.` is not right and it will basically create more confusion.

Sorry but I'm not understanding your point. `runtime` without `time_based` operates exactly as I described. The existing documentation for `time_based` explains how `runtime` operates differently when it is also specified.

ankit-sam:
> Examples: Setup to create 1GB ramdisk at /dev/ram0: `modprobe brd rd_size=1024000 rd_nr=1`
> 
> Cap runtime to 1 second (workload otherwise takes 10 seconds on my system): `sudo fio -name=test --filename=/dev/ram0 -rw=write -size=1GB -bs=512b -runtime=1`
> 
> Run for a minimum of 60 seconds: `sudo fio -name=test --filename=/dev/ram0 -rw=write -size=1GB -bs=512b -runtime=60 -time_based`

Ah I got it now, Sorry my bad for creating a fuss. I should have tried with the same configuration as you described or with a bigger size.
What I tried was a qemu emulated NVMe device
```fio --name=test --filename=/dev/nvme0n1 --bs=4k --rw=write --size=1G --runtime=1```
fio log:
```WRITE: bw=1364MiB/s (1430MB/s), 1364MiB/s-1364MiB/s (1430MB/s-1430MB/s), io=1024MiB (1074MB), run=751-751msec```
that test got over in 751ms, and I instantly thought that it will always be based on size (1GB in this case), I should have tried with a bigger size.

horshack-dpreview:
@ankit-sam, No worries, glad we're on the same page. Another way to slow down the I/Os is using a smaller block size, and also using random so I/Os wont be coalesced (aka "plugged") like they are for sequential.

axboe:
Change looks fine, but fio.1 needs a similar treatment.

horshack-dpreview:
@axboe, I've updated fio.1

Issue #1520, opened 2023-02-18
fio: ioengines.c:346: td_io_queue: Assertion `res == 0' failed.


horshack-dpreview:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
"fio: ioengines.c:346: td_io_queue: Assertion 'res == 0' failed" when `-serialize_overlap=1`, `io_submit_mode=offload`, and `verify=crc32` are used in combination

**Environment**: Ubuntu 22.04.1 LTS
**fio version**: 3.33 (tip)

**Reproduction steps**

1. `modprobe brd rd_size=1024000 rd_nr=1` (creates ramdisk)
2. `fio -name=test -verify=crc32 -thread -size=16K -bs=512 -iodepth=1 -serialize_overlap 1 -filename=/dev/ram0 -rw=write -numjobs=1 -io_submit_mode=offload`

PR #1521, opened 2023-02-18
ioengines.c:346: td_io_queue: Assertion `res == 0' failed

horshack-dpreview:
Assertion in ioengines.c::td_io_queue() fails for pthread_mutex_unlock() on overlap_check mutex when serialize_overlap=1, io_submit_mode=offload, and verify=<any> are used together.

backend.c::fio_io_sync() invokes td_io_queue(), which expects the caller to have ownership of the overlap_check mutex when serialize_overlap and offloading are configured, as part of the overlap-check interlock with IO_U_F_FLIGHT. The mutex is not acquired for this path because it's not an I/O requiring an overlap check.

The fix is to refine the conditional that triggers td_io_queue() to release the overlap_check mutex. Rather than using broad config options, the conditional now uses a new io_u flag named IO_U_F_OVERLAP_LOCK, which is only set for the offload worker thread path that acquires the mutex.

Link: https://github.com/axboe/fio/issues/1520

### DIFF
--- a/io_u.h
+++ b/io_u.h
@@ -21,6 +21,7 @@ enum {
 	IO_U_F_TRIMMED		= 1 << 5,
 	IO_U_F_BARRIER		= 1 << 6,
 	IO_U_F_VER_LIST		= 1 << 7,
+	IO_U_F_OVERLAP_LOCK	= 1 << 8,
 };
 
 /*
--- a/ioengines.c
+++ b/ioengines.c
@@ -341,9 +341,10 @@ enum fio_q_status td_io_queue(struct thread_data *td, struct io_u *io_u)
 	 * started the overlap check because the IO_U_F_FLIGHT
 	 * flag is now set
 	 */
-	if (td_offload_overlap(td)) {
+	if (io_u->flags & IO_U_F_OVERLAP_LOCK) {
 		int res = pthread_mutex_unlock(&overlap_check);
 		assert(res == 0);
+		io_u_clear(td, io_u, IO_U_F_OVERLAP_LOCK);
 	}
 
 	assert(fio_file_open(io_u->file));
--- a/rate-submit.c
+++ b/rate-submit.c
@@ -47,6 +47,7 @@ retry:
 		assert(res == 0);
 		goto retry;
 	}
+	io_u_set(td, io_u, IO_U_F_OVERLAP_LOCK);
 }
 
 static int io_workqueue_fn(struct submit_worker *sw,


axboe:
This is causing a crash, and I should've done a better job reviewing it. The rate-submit.c is nonsensical. Did you test this at all? I'm going to revert the change for now.

axboe:
Just run:

./fio t/jobs/t0013.fio

and it'll insta-crash.

horshack-dpreview:
> Just run:
> 
> ./fio t/jobs/t0013.fio
> 
> and it'll insta-crash.

Reproduced. I don't know how I added that line to rate-submit.c without seeing the obvious terminating value issue for `td`. I'm sorry.

I was more perplexed how my test runs didn't catch it. Turns out it only faults when `numjobs` is an even multiple of 8, which none of my test runs are. Digging into it, `for_each_td` will only yield an ending `td` value of NULL when the number of threads allocated is an even multiple of `JOBS_PER_SEG (8)`, since that's the only scenario where the `thread_segment[n+1]` will have a NULL value for `threads`:

https://github.com/axboe/fio/blob/9943adfb8adf50e6f54ab6aff613c99a3a75d224/fio.h#L587-L593

To prevent others from making my same careless mistake I'd like to propose changing `for_each_td` to make it impossible to use the terminating value of `td`. This can be accomplished by 1) moving the scope of `td` inside `for_each_td` or 2) explicitly assigning `td` to NULL at the end of the loop. The first proposal is cleaner and more efficient but touches a lot of files - the latter has only one touch but is less efficient. Please let me know what you think.

axboe:
Option 1 seems nicer. Even if efficiency isn't THAT important so we could tolerate just clearing it, it'd still just crash rather than be safe by design.

horshack-dpreview:
> t/jobs/t0013.fio

Thanks. I wanted to give both `td` and `i` the same macro-scope treatment but in C/C++ if you declare any type within the for() loop initializer all variables must be of that type.  For example, the following is invalid:
```
#define for_each_td(td, i)  \
    for (int (i) = 0, struct thread_data *(td) = &segments[0].threads[0]; (i) < (int) thread_number; (i)++, (td) = tnumber_to_td((i)))
```
Even if we wanted to only restrict the scope of `td` the following is invalid too, since i is not the same type as `td` declared:
```
#define for_each_td(td, i)  \
    for (struct thread_data *(td) = &segments[0].threads[0], int (i) = 0; (i) < (int) thread_number; (i)++, (td) = tnumber_to_td((i)))
```

To work around this i must be initialized outside the for() loop initializer:
```
#define for_each_td(td, i)  \
    (i) = 0;                \
    for (struct thread_data *(td) = &segments[0].threads[0]; (i) < (int) thread_number; (i)++, (td) = tnumber_to_td((i)))
```
Which is fine but not beautiful. Next issue: Every snippet which uses `for_each_td` will get a `-Wunused-variable` warning because the outer-scope version of `td` is not used in most of fio's code - it's only declared for use in `for_each_td`. The cleanest solution would be to remove all existing local declarations of `td` intended for use in `for_each_td` (and the loop index variable too while we're at it). But that means we now have to use a hard-coded variable name for the construct, since C/C++ doesn't provide a way to pass a variable name to a macro and have that macro declare a variable with that name...if the name doesn't already exist (which it does in the outer scope for the current implementation). The solution is to always use the name `td` within the `for_each_td` block, which is fine except in 4 places where the code uses `td2` instead, since `td` is used as the function argument name in those cases. For those we'd have to rename that existing `td` to something else, which seems manageable.

It's doable but comes down to style and whether the juice is worth the squeeze. The alternative is the original option 2, where we assure `td` ends with a value of NULL so we at least catch the usage via a segfault. To implement that I would modify `tnumber_to_td()` to range check and return NULL for a thread # that is n+1 the # of threads.

horshack-dpreview:
@axboe, Here is the corrected fix for the original issue. Would you like me to create a new pull request or can you pull it from it from my branch?

Link: https://github.com/horshack-dpreview/fio/commit/aac9c033fe45827f166a893a20e7a7ea183e0205

horshack-dpreview:
@axboe, I've implemented the discussed changes to `for_each_td()`:

https://github.com/axboe/fio/pull/1531

Issue #1526, opened 2023-02-24
FIO reports verify: bad header rand_seed, in the RW mode when iodepth > 1


KuribohG:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**

FIO reports verify error in the RW mode when iodepth > 1.

After debugging, I found this may be the reason: Under RW mode, there may be some IO requests that must be requeued, if there is a read/write switching when iodepth > 1. The `rand_seed` field in `io_u` changed when requeue (which I think is not desirable), but fio is still using the original `rand_seed` for verifying.

I'm not sure if this problem is related to https://github.com/axboe/fio/issues/1503.

**Environment**:  Ubuntu 20.04

**fio version**:  3.32

**Reproduction steps**

```
[random-readers]
group_reporting
rw=rw
do_verify=1
verify=crc32c
ioengine=vsync
threads=1
numjobs=1
blocksize=128k
size=1024k
nrfiles=1
filename_format=testsync.$jobnum
iodepth=4
iodepth_batch_submit=4
iodepth_batch_complete_min=4
iodepth_batch_complete_max=4
```
Run fio with the configuration above, and the bug will reproduce consistently.

horshack-dpreview:
@KuribohG  Thanks for your submission. I was able to reproduce your issue using the current tip of master. As you discovered, the issue occurs when an I/O request gets requeued. This occurs because of `iodepth_batch_complete_max` - when that number of I/Os have been submitted (and waiting to complete), the next I/O generated results in an internal `FIO_Q_BUSY` before submission, which causes the I/O to be queued for later submission. When that I/O is later dequeued and executed, the  logic in `backend.c::do_io()` calls `verify.c::populate_verify_io_u()`, but `populate_verify_io_u()` was already called once when the I/O was first generated and requeued with `FIO_Q_BUSY` - this causes the verify logic's random seed to advance an additional time (generated by `verify.c::fill_verify_pattern()`, which causes a later verify read I/O to fail because the seeds will be out of sync. Here's the relevant source:

https://github.com/axboe/fio/blob/b5904c0d7434a49770cdb90eada1c724f0f7fe4e/backend.c#L703-L708
https://github.com/axboe/fio/blob/b5904c0d7434a49770cdb90eada1c724f0f7fe4e/verify.c#L50-L63

[...]

PR #1527, opened 2023-02-26
FIx "verify bad_hdr rand_seed" for requeued I/Os

horshack-dpreview:
On configurations that can cause I/Os to be internally requeued from FIO_Q_BUSY such as '--iodepth_batch_complete_max', and the workload has verify enabled, the subsequent verification of the data fails with a bad verify rand_seed because the pattern for the I/O is generated twice for the same I/O, causing the seed to become out of sync when the verify is later performed. The seed is generate twice because do_io() handles the I/O twice, first when it originates the I/O and again when it later gets the same I/O back from get_io_u() after it's is pulled from the requeue list, which is where the first submission landed due to the workload reaching '--iodepth_batch_complete_max'.

The fix is for do_io() to track when it has generated the verify pattern for an I/O via a new io_u flag 'IO_U_F_PATTERN_DONE', avoiding a second call to populate_verify_io_u() when that flag is detected.

Link: https://github.com/axboe/fio/issues/1526

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -1040,8 +1040,11 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
 		}
 
 		if (io_u->ddir == DDIR_WRITE && td->flags & TD_F_DO_VERIFY) {
-			io_u->numberio = td->io_issues[io_u->ddir];
-			populate_verify_io_u(td, io_u);
+			if (!(io_u->flags & IO_U_F_PATTERN_DONE)) {
+				io_u_set(td, io_u, IO_U_F_PATTERN_DONE);
+				io_u->numberio = td->io_issues[io_u->ddir];
+				populate_verify_io_u(td, io_u);
+			}
 		}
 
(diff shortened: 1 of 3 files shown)


axboe:
> Note IO_U_F_PATTERN_DONE is set to 1<<9 instead of the next available 1<<8 because there is already a commit pending that uses 1<<8 (IO_U_F_OVERLAP_LOCK).

Don't do that, it's easy to resolve at merge time. And what if that other commit never goes in? Additionally, it should not go into the commit message - it's fine in here, but it doesn't belong in the commit as that comment means nothing later one when someone is reading the code.

horshack-dpreview:
@axboe , Got it, thanks. Wasn't sure about the tolerance for merge conflicts on the project. I've made the corrections you suggested.

axboe:
I don't mind merging at all, and I'm always a bit frustrated at people incessantly rebasing on master even when there are no conflict.

PR #1529, opened 2023-02-28
make operation error:make error on 'uos' system

Cuelive:
make operation error:make error on 'uos' system

Blktrace file make operation error ’undefined reference to `major'‘ Problems on uos system

liuyafei@uniontech.com

### DIFF
--- a/blktrace.c
+++ b/blktrace.c
@@ -5,6 +5,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <errno.h>
+#include <sys/sysmacros.h>
 
 #include "flist.h"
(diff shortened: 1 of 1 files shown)


axboe:
Still missing a few bits... Here's what your commit message should look like:

```
blktrace: fix compilation error on the uos system

When compiling on uos, it fails with an undefined reference to 'major'. Fix this by
including the correct header for it.

Signed-off-by: Missing Name <liuyafei@uniontech.com>
```

and replace "Missing Name" with your real name.

Cuelive:
@axboe Commit modified, please help me merge, thanks you !!

PR #1530, opened 2023-03-02
Introducing new SPDK based IO engine

mariuszbarczak:


jimharris:
How much does this differ from the plugin in the SPDK tree?

https://github.com/spdk/spdk/tree/master/examples/bdev/fio_plugin

Note that SPDK has two fio plugins in its tree, one for the bdev layer (similar to this one) but also one that operates at the SPDK nvme driver level

https://github.com/spdk/spdk/tree/master/examples/nvme/fio_plugin

I'd suggest naming this one spdk_bdev to differentiate it from the nvme plugin (even if the nvme plugin doesn't go into upstream fio right now).

mariuszbarczak:
IMHO, SPDK is popular and valuable that a direct IO engine inside the fio benchmark executable would make sense. However, given Jim's comments about the two preexisting IO engines and this being a third one, I'd like to withdraw this pull request and work with Jim and SPDK community to determine the best path forward.

The differences are:  
- The functionality is directly integrated and no need to pre-load the engine using dynamic library 
- CPU statistics (SPDK top) is available for this version of plugin (it uses SPDK app API) 
- It is possible to run SPDK configuration which requires many CPU cores

jimharris:
> IMHO, SPDK is popular and valuable that a direct IO engine inside the fio benchmark executable would make sense. However, given Jim's comments about the two preexisting IO engines and this being a third one, I'd like to withdraw this pull request and work with Jim and SPDK community to determine the best path forward.

Thanks @mariuszbarczak.  I agree.  A lot of work has gone into the SPDK fio plugin in its repository, it gets tested heavily in SPDK CI and regular performance reports.  It seems better to modify that one to meet any unmet requirements you found, rather than creating a brand new one.  There are also features like zones that you get with the existing SPDK plugin.

> The differences are:
> 
> * The functionality is directly integrated and no need to pre-load the engine using dynamic library

The existing SPDK fio plugins could also be directly integrated into fio if there's a strong pull for it.  The plugin posted here is not unique in this regard. 

> * CPU statistics (SPDK top) is available for this version of plugin (it uses SPDK app API)

Existing SPDK plugin doesn't use the SPDK app framework because it wants to directly submit IO to SPDK from the fio pthreads.  We've found this to be a better measurement of not just performance but also CPU utilization efficiency of the measured workload, compared to funneling I/O to separate polled mode pthreads spawned by SPDK.

Let's discuss in SPDK channels how to enable the `spdk_top` application with the SPDK fio plugin.

> * It is possible to run SPDK configuration which requires many CPU cores

Please file issue in SPDK GitHub database with details on your configuration.

PR #1531, opened 2023-03-02
Refactor for_each_td() to catch inappropriate td ptr reuse

horshack-dpreview:
I recently introduced a bug caused by reusing a struct thread_data *td after the end of a for_each_td() loop construct.

Link: https://github.com/axboe/fio/pull/1521#issuecomment-1448591102

To prevent others from making this same mistake, this commit refactors for_each_td() so that both the struct thread_data * and the loop index variable are placed inside their own scope for the loop. This will cause any reference to those variables outside the for_each_td() to produce an undeclared identifier error, provided the outer scope doesn't already reuse those same variable names for other code within the routine (which is fine because the scopes are separate).

Because C/C++ doesn't let you declare two different variable types within the scope of a for() loop initializer, creating a scope for both struct thread_data * and the loop index required explicitly declaring a scope with a curly brace. This means for_each_td() includes an opening curly brace to create the scope, which means all uses of for_each_td() must now end with an invocation of a new macro named end_for_each_td() to emit an ending curly brace to match the scope brace created by for_each_td():

	for_each_td(td, i) {
		while (td->runstate < TD_EXITED)
			sleep(1);
	}
	end_for_each_td()

The alternative is to end every for_each_td() construct with an inline curly brace, which is off-putting since the implementation of an extra opening curly brace is abstracted in for_each_td():

	for_each_td(td, i) {
		while (td->runstate < TD_EXITED)
			sleep(1);
	}}

Most fio logic only declares "struct thread_data *td" and "int i" for use in for_each_td(), which means those declarations will now cause -Wunused-variable warnings since they're not used outside the scope the refactored for_each_td(). Those declarations have been removed.

Implementing this change caught a latent bug in eta.c::calc_thread_status() that accesses the ending value of struct thread_data *td after the end of for_each_td(), now manifesting as a compile error, so working as designed :) That bug is corrected as part of this commit.

### DIFF
--- a/engines/libblkio.c
+++ b/engines/libblkio.c
@@ -283,16 +283,14 @@ static bool possibly_null_strs_equal(const char *a, const char *b)
  */
 static int total_threaded_subjobs(bool hipri)
 {
-	struct thread_data *td;
-	unsigned int i;
 	int count = 0;
 
-	for_each_td(td, i) {
+	for_each_td(td) {
 		const struct fio_blkio_options *options = td->eo;
 		if (strcmp(td->o.ioengine, "libblkio") == 0 &&
 		    td->o.use_thread && (bool)options->hipri == hipri)
 			++count;
-	}
+	} end_for_each();
 
 	return count;
 }
--- a/backend.c
+++ b/backend.c
@@ -93,19 +93,16 @@ static void sig_int(int sig)
 #ifdef WIN32
 static void sig_break(int sig)
 {
-	struct thread_data *td;
-	int i;
-
 	sig_int(sig);
 
 	/**
 	 * Windows terminates all job processes on SIGBREAK after the handler
 	 * returns, so give them time to wrap-up and give stats
 	 */
-	for_each_td(td, i) {
+	for_each_td(td) {
 		while (td->runstate < TD_EXITED)
 			sleep(1);
-	}
+	} end_for_each();
 }
 #endif
 
@@ -2056,15 +2053,14 @@ err:
 static void reap_threads(unsigned int *nr_running, uint64_t *t_rate,
 			 uint64_t *m_rate)
 {
-	struct thread_data *td;
 	unsigned int cputhreads, realthreads, pending;
-	int i, status, ret;
+	int status, ret;
 
 	/*
 	 * reap exited threads (TD_EXITED -> TD_REAPED)
 	 */
 	realthreads = pending = cputhreads = 0;
-	for_each_td(td, i) {
+	for_each_td(td) {
 		int flags = 0;
 
 		if (!strcmp(td->o.ioengine, "cpuio"))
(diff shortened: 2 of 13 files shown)


axboe:
I did notice the bug in eta.c as well when looking at this. But this should be fixed upfront, don't mix it in with the for_each_td() change, as that one needs to stand alone and me mostly mechanical.

Looks good to me, but I'd get rid of passing in 'i' to for_each_td() as well. It serves zero purposes at this point, and the callers only care about iterating over 'td' anyway, not some random index value. So keep that hidden inside for_each_td() as __i or something like that.

I do like the split with the end_for_each_td(), that's still readable and something that folks have come across before as well. But please put them on the same line as the closing brace:

```
for_each_td(td) {
        do_stuff(td);
} end_for_each();
```

and just make it end_for_each() rather than include 'td' in the naming there too.

horshack-dpreview:
> I did notice the bug in eta.c as well when looking at this. But this should be fixed upfront, don't mix it in with the for_each_td() change, as that one needs to stand alone and me mostly mechanical.

I agree and would usually never mix two tangentially-related fixes in a single commit. In this instance the logic in eta.c as written doesn't compile with the refactored `for_each_td()` and I didn't want to break the build. I went ahead and removed the fix and commented out that segment of the conditional so that it'll compile (with a comment to fix). I'll submit a separate fix for that.

> 
> Looks good to me, but I'd get rid of passing in 'i' to for_each_td() as well. It serves zero purposes at this point, and the callers only care about iterating over 'td' anyway, not some random index value. So keep that hidden inside for_each_td() as __i or something like that.

There are 19 instances of logic that use the "i" loop variable. I also wanted to consider the possibility for potential nested use for `for_each_td()`. That said, I agree it's cleaner to declare the loop index internally, so I went ahead and made the change. I named it `__td_index` for clarity and changed the 19 instances to use it instead of "i" (or whatever name they specified for each particular loop).

> 
> I do like the split with the end_for_each_td(), that's still readable and something that folks have come across before as well. But please put them on the same line as the closing brace:
> 
> ```
> for_each_td(td) {
>         do_stuff(td);
> } end_for_each();
> ```
> 
> and just make it end_for_each() rather than include 'td' in the naming there too.

Done

axboe:
> I'll submit a separate fix for that once you merge this pull - they depend on each other so can't do it in advance.

How so? You can fix the eta issue first, then do the for_each_td() change after that.

axboe:
Looks like engines/libblkio.c was missed in this conversion.

engines/libblkio.c
```diff
@@ -287,12 +287,12 @@ static int total_threaded_subjobs(bool hipri)
 	unsigned int i;
```
axboe, line 287: You forgot to kill 'i'.

horshack-dpreview: > You forgot to kill 'i'.

I was trying to get libblkio to build on my system to compile and test this change but it keeps hanging during the ``meson compile -C build`, so I'm having to push without being able to compile. Just removed `td` and `i`.

horshack-dpreview:
Fix for bug found in eta.c::calc_thread_status() by refactoring separated into the following pull request:

Link: https://github.com/axboe/fio/pull/1535

Issue #1537, opened 2023-03-05
fio doesn't seem to keep up with timestamps while replaying


lokeshjaliminche:
**Description of the bug:**
I am trying to replay a trace file with fio. However, fio doesn't seem
to keep up with the timestamps and runs very slowly. Any ideas on how
to achieve this?

I am attaching job file and trace file below

**Environment**: 
Ubuntu Linux 5.4.0-139-generic 

**fio version**: 
 fio --version
fio-3.33-109-gb5904

**Reproduction steps**
[global]
direct=1

[job_1]
numjobs=1
ioengine=libaio
replay_skip=write
read_iolog=./replay_2.log

Tracefile:
[replay_2.log.tar.gz](https://github.com/axboe/fio/files/10891364/replay_2.log.tar.gz)

horshack-dpreview:
@lokeshjaliminche, I was able to reproduce your replay delays. By default, fio reproduces the timestamp between I/Os when replaying and from checking your log I see some very large delays between some I/Os. Like this (1st column is #usec from start):
```
20 /dev/ram0 write 488197214720 4096
21 /dev/ram0 read 482637861376 24576
22 /dev/ram0 read 494470889984 12288
23 /dev/ram0 read 487460954624 65536
1000001 /dev/ram0 read 489065263616 53248
1000002 /dev/ram0 read 489065320960 24576
1000003 /dev/ram0 write 70974357504 4096
1000004 /dev/ram0 write 483084784128 8192
1000005 /dev/ram0 read 489065345536 32768
1000006 /dev/ram0 read 489065452032 16384
1000007 /dev/ram0 read 489065382400 65536
1 /dev/ram0 read 497433108992 4096
1000001 /dev/ram0 write 480025637376 4096
1 /dev/ram0 read 481019101696 4096
1000001 /dev/ram0 write 480244576768 4096
1000002 /dev/ram0 write 482193912320 4096
1000003 /dev/ram0 write 481894437376 4096
1000004 /dev/ram0 write 480241426944 8192
1000005 /dev/ram0 write 480241496576 4096
1000006 /dev/ram0 write 482194211328 4096
1000007 /dev/ram0 write 493699576320 4096
1000008 /dev/ram0 write 48205594624 4096
1000009 /dev/ram0 write 491778695680 4096
1000010 /dev/ram0 write 480180539904 4096
1 /dev/ram0 read 21969521664 37888
```
Can you post the original workload config that generated this log?

[...]

PR #1540, opened 2023-03-09
Fix I/O replay timing

horshack-dpreview:
fio supports replaying I/Os from a previous invocation, captured either via blktrace or via `write_iolog`. That support includes simulating the timing of the I/Os, accomplished by using the I/O timestamps from the log to insert delays in the submission path if necessary to match the pacing of the previous session. The logic to calculate the elapsed time between I/Os during replay is not accounting for the time spent in submitting the I/O to the kernel; this causes large discrepancies in playback timing. Playback can take 2x or more longer than the original workload due to the progressively-larger drift from excessive delays.

This change also includes fixes to additional issues in the same logic, specifically not adjusting to the expected vs actual sleep time for delays based on the post-adjusted delay value vs the original delay value. Another change is to keep a cumulative adjustment factor for both the I/O-to-I/O time (when it's larger than the log delay) and for the expected vs actual sleep time for delays.

Link: https://github.com/axboe/fio/issues/1537#issuecomment-1461092644


horshack-dpreview:
Update commit message to give a more thorough explanation of the issue and fix.

horshack-dpreview:
@vincentkfu, This PR has been languishing for a while. Is there anything preventing it from being merged?

vincentkfu:
Please fix the build failures. I am on vacation now but this is on my list of things to look at after I return to work next week.

PR #1545, opened 2023-03-24
fdp cleanup and trim support for io_uring_cmd

ankit-sam:
This series has 2 commits:

1. For fdp remove the usage of expensive modulo operation as the same can be achieved with addition and wrapping around available reclaim unit handles.

2. Add support for trim operation to the io_uring_cmd ioengine. Print ZBD zone reset stats for trim operation.

### DIFF
--- a/engines/io_uring.c
+++ b/engines/io_uring.c
@@ -24,6 +24,7 @@
 #include "../lib/types.h"
 #include "../os/linux/io_uring.h"
 #include "cmdprio.h"
+#include "zbd.h"
 #include "nvme.h"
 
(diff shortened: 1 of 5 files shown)


ankit-sam:
Tagging @kawasaki as the second commit touches ZBD part.

vincentkfu:
These patches look fine to me. If there is a need we can eventually replace the IOCTL with a write zeroes command with the trim bit set. This would allow trims to be async commands.

If Jens doesn't beat me to it I will plan to merge this after I return from vacation.

Issue #1502, opened 2023-01-31
randrepeat is not overridden by randseed flag when FIO_RAND_BLOCK_OFF is initialized.


freeeborn:
fixup_options has following code:

         /*
	 * If randseed is set, that overrides randrepeat
	 */
         if (fio_option_is_set(o, rand_seed))
		o->rand_repeatable = 0;

https://github.com/axboe/fio/blob/7cbbc6459d9b5f9d687ab55ef5bfe3ae86a99d05/init.c#L929

However, random seeds are initialized before fixup:

	if (setup_random_seeds(td)) {
		td_verror(td, errno, "setup_random_seeds");
		goto err;
	}

	if (fixup_options(td))
		goto err;

https://github.com/axboe/fio/blob/7cbbc6459d9b5f9d687ab55ef5bfe3ae86a99d05/init.c#L1536

this causes --randseed to be ignored for block offsets unless --randrepeat=0 is also specified

if (td->o.rand_repeatable)
		td->rand_seeds[FIO_RAND_BLOCK_OFF] = FIO_RANDSEED * td->thread_number;

https://github.com/axboe/fio/blob/7cbbc6459d9b5f9d687ab55ef5bfe3ae86a99d05/init.c#L1068

which causes sequence of block offsets for loads with randwrite and io_size does not depend on randseed.

vincentkfu:
The description for `randseed` says that it overrides `randrepeat`. So you have correctly identified a bug. Can you confirm that calling `fixup_options()` before seeding the random number generators resolves the issue?

https://github.com/vincentkfu/fio/tree/1502

The random number seed code would benefit from some cleanups and it seems as if `allrandrepeat` overrides both `randseed` and `randrepeat` although this is not documented.

PR #1546, opened 2023-03-27
RFC clean up random seed options

vincentkfu:
I'm looking for feedback on these two questions:
- Is this cleanup worth the breakage it might cause users?
- If the answer to the above question is yes, is there a better way to deprecate `allrandrepeat`?

Here is what the documentation says about fio's three random seed-related options:

```
randrepeat=bool
Seed the random number generator used for random I/O patterns in a predictable way so the pattern is repeatable across runs. Default: true.

allrandrepeat=bool
Seed all random number generators in a predictable way so results are repeatable across runs. Default: false.

randseed=int
Seed the random number generators based on this seed value, to be able to control what sequence of output is being generated. If not set, the random sequence depends on the [randrepeat](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-randrepeat) setting.
```

The description for `randrepeat` is imprecise, but combined with the description for `allrandrepeat` users might imagine that `randrepeat` chooses a predetermined random seed only for generating random offsets whereas `allrandrepeat` chooses a predetermined random seed for all of the random number generators fio uses. 

It turns out, however, that `randrepeat` actually selects predetermined random seeds for *all* of fio's random number generators. So `allrandrepeat` is superfluous. This patch series refactors the code to make it easier to follow, makes `allrandrepeat` an alias of `randrepeat`, and adds a test script to guard against future breakage.

This is the new behavior:

- make `randrepeat` and `allrandrepeat` synonyms.
- use 0x52780A68ADBC0089 (=0x89 * 0x9e370001 * 0x9e370001 * 0x9e370001 * 0x9e370001) as the base random seed
  - this is the current default base random seed used when `randrepeat`=1
  - this changes the default base random seed used when `allrandrepeat`=1 which was `FIO_RANDSEED` (0xb1899bedUL)
- allow `randseed` to override both `randrepeat` and `allrandrepeat`
  - the documentation implies that that `randseed` overrides `randrepeat` but this [hasn't been](https://github.com/axboe/fio/issues/1502) the case
  - currently `allrandrepeat` overrides `randseed`. So changing this might lead to headaches for users depending on current behavior.

The current behavior is this:

| | `randrepeat`=0 | `randrepeat`=1 |
| - | - | - |
| `allrandrepeat`=0 | | |
| - `randseed` unset | 1. fio uses `/dev/urandom` to generate random seeds | 2. fio uses default `randseed` for seeds but uses `thread_number*FIO_RANDSEED` for `rand_seeds[FIO_RAND_BLOCK_OFF]` |
| - `randseed` set | 3. fio uses supplied `randseed` to generate random seeds | 4. fio uses supplied `randseed` to generate random seeds but uses `thread_number*FIO_RANDSEED` for `rand_seeds[FIO_RAND_BLOCK_OFF]` |
| | | |
| `allrandrepeat`=1 | | |
| - `randseed` unset | 5. fio uses `FIO_RANDSEED` to generate random seeds | 6. fio uses `FIO_RANDSEED` to generate random seeds but uses `thread_number*FIO_RANDSEED` for `rand_seeds[FIO_RAND_BLOCK_OFF]` |
| - `randseed` set | 7. fio uses `FIO_RANDSEED` to generate random seeds | 8. fio uses `FIO_RANDSEED` to generate random seeds but uses `thread_number*FIO_RANDSEED` for `rand_seeds[FIO_RAND_BLOCK_OFF]` |

This change may cause users headaches if they rely on the current behavior of these options but I think we have to bite the bullet and make the change because the current state is untenable.

With this patch series the behavior becomes:

| | `[all]randrepeat`=0 | `[all]randrepeat`=1 |
| - | - | - |
| - `randseed` unset | 1. fio uses `/dev/urandom` or equivalent to generate random seeds | 2. fio uses default `randseed` for seeds |
| - `randseed` set | 3. fio uses supplied `randseed` to generate random seeds | 4. fio uses supplied `randseed` to generate random seeds |


t/random_seed.py
```diff
@@ -0,0 +1,384 @@
+#!/usr/bin/env python3
+"""
+# random_seed.py
+#
+# Test fio's random seed options.
+#
+# - make sure that randseed overrides randrepeat and allrandrepeat
+# - make sure that seeds differ across invocations when [all]randrepeat=0 and randseed is not set
+# - make sure that seeds are always the same when [all]randrepeat=1 and randseed is not set
+#
```
ankit-sam, line 10: I think you can also add the usage, examples and requirements similar to what t/readonly.py, t/strided.py and other python based scripts.

Otherwise looks good, all the tests passed.

ankit-sam (commented):
hey @vincentkfu I only have 2 comments, other than that the changes look good to me.

t/random_seed.py
vincentkfu: Done. Thanks.

vincentkfu:
These patches were applied with light edits to the last two commit messages beginning with https://github.com/axboe/fio/commit/64b2ba1460687332817655ede202166386af938b

PR #1551, opened 2023-04-03
Increase waiting time for the started threads (100ms -> 1sec)

parkch0708:
Increase waiting time for the started threads (100ms -> 1sec)

When the device's Identify CMD takes mor than 100ms, Random Read Performance is dropped. Increasing waiting time is helpful for this issue.

Signed-off-by: Chanhyun Park <chanhyun0708@gmail.com>

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -2510,7 +2510,7 @@ reap:
 			if (mtime_since_now(&this_start) > JOB_START_TIMEOUT)
 				break;
 
-			do_usleep(100000);
+			do_usleep(1000000);
 
 			for (i = 0; i < this_jobs; i++) {
 				td = map[i];


axboe:
The commit format is not correct, but more importantly, isn't this just something that is solved by using a bit of ramp time?

parkch0708:
> The commit format is not correct, but more importantly, isn't this just something that is solved by using a bit of ramp time?

Hello. ramp_time option is not helpful for this issue. Where can I see the commit format? If necessary, I will re-commit & new pull request or change commit message. Thanks.

axboe:
Then maybe let's make it smarter, rather than just make it 10x the existing value. That seems like a bit of a bad work around, when there's a chance to actually make it better. Perhaps using pthread_barrier or something like that.

For the commit format, just look at the git repo? There are lots of commits in there.

parkch0708:
> Then maybe let's make it smarter, rather than just make it 10x the existing value. That seems like a bit of a bad work around, when there's a chance to actually make it better. Perhaps using pthread_barrier or something like that.
> 
> For the commit format, just look at the git repo? There are lots of commits in there.

Actually increasing around 200ms can be helpful. Is that OK?

PR #1552, opened 2023-04-04
backend: increase waiting time for the started threads from 100ms to 300ms

parkch0708:
backend: increase waiting time for the started threads from 100ms to 300ms

When the device's Identify CMD takes mor than 100ms, Random Read Performance is dropped. Increasing waiting time is helpful for this issue.

Signed-off-by: Chanhyun Park <chanhyun0708@gmail.com>

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -2510,7 +2510,7 @@ reap:
 			if (mtime_since_now(&this_start) > JOB_START_TIMEOUT)
 				break;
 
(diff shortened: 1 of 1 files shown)


axboe:
I already told you NOT to go down the route of random time delay additions. What happens when some device takes 350msec now? Please take my advice and fix this properly.

PR #1556, opened 2023-04-07
engines/libaio: fix io_getevents min/max events arguments

lrumancik:
If the io_getevents system call is interrupted, it can succeed while returning less than the min_nr requested events. If this occurs during the final reaping of events, libaio will call io_getevents again wtih the same minimum events argument. Since some of the events have already been reaped, this results in a hang. To fix this, update the arguments for io_getevents based on the previously seen events.


axboe:
Can you take a look at the io_uring engine too and see if the same issue is there?

lrumancik:
Sure, will take a look

lrumancik:
In io_uring, the actual_min is updated so the hang should not occur. The max events is not updated based on previously seen events. I assume this should be updated as well? If so, I can create a patch for this.

axboe:
Yep please do, thanks!

PR #1570, opened 2023-05-09
use 'min' macro to find out next value of actual_min in libaio

dpronin:
Small update to make algorithm of computing next value of actual_min is clearer


ankit-sam:
This change looks good to me.

dpronin:
@axboe is there something wrong in the PR?

PR #1576, opened 2023-05-19
Improve zoned block device support

bvanassche:


bvanassche:
@floatious Do you perhaps want to help with reviewing this pull request?

zbd.c
```diff
@@ -821,6 +822,8 @@ static int parse_zone_info(struct thread_data *td, struct fio_file *f)
 						     PTHREAD_MUTEX_RECURSIVE);
 			p->start = z->start;
 			p->capacity = z->capacity;
+			if (capacity != z->capacity)
+				capacity = zone_size + 1;
```
floatious, line 826: capacity = zone_size + 1 is not the most easy thing to understand.
It could probably also overflow (although having a zone size that is UINT64_MAX is not realistic,
but perhaps some static code analyzers catches it.)

Considering that this function is not in any fast path, and seldom called.
How about a simple "bool identical_zone_cap = true;"
and simply set it to false "if (capacity != z->capacity)".

Final if can then be just "if (identical_zone_cap) dprint()

bvanassche:
> Do we usually have ; or , when printing multiple things? (nr_zones=%d; errno=%d)

It seems like both are used frequently:
```
$ git grep -nH 'print.*,' | wc -l; git grep -nH 'print.*;' | wc -l
   1113
    910
```

zbd.c
bvanassche: Done :-)

floatious (approved):

floatious (approved):

PR #1583, opened 2023-06-23
Support restricting the number of active zones

bvanassche:


damien-lemoal:
Bart, can you send the patching to the mailing list ? Easier to review :)

kawasaki:
I'm aware of that when zoned block devices have the max_active_zones limit, fio zonemode=zbd behavior is not ideal: when the zones in closed conditions exist at fio job start, those zones are not selected for write target. Also, write to other zones may cause max_active_zones limit error. It looks that the Bart's commit is trying to address them by adding the new option "max_active_zones".

Actually, I'm preparing a patch to address the issues in a different way: it will modify the existing option max_open_zones when the zoned block device has the max_active_zones limit. In such a case, fio can handle zones in closed status as write target. AFAIK, ZNS drives often have same max_active_zones limit value as max_open_zones, then default value of the max_open_zones option value taken from the device will works good.

I wonder which way is the better for the users: different two options "max_open_zones" and "max_active_zones"? Or single "max_open_zones" option with max_active_zones support? We can continue the discussion on the fio list.

PR #1584, opened 2023-06-26
zbd: Support finishing zones on Android

bvanassche:
BLKFINISHZONE is missing from older versions of the Android NDK header files. Hence, define BLKFINISHZONE if it has not been defined and detect at runtime whether or not the kernel supports finishing zones.


damien-lemoal:
Bart, can you send the patches to the fio list please ? That makes testing and review easier :)

bvanassche:
See also commit 8e2b81b854286f32eae7951a434dddebd968f9d5.

PR #1600, opened 2023-07-26
iolog.c: fix inaccurate clat when replay trace

min22:
When do timestamp replay with high qd it will only reap the completed commands when the qd reach the max iodepth, the commands probably are finished long ago before command completion handling.

Fix is to use io_u_queued_complete instead of just usec_sleep in iolog_delay

Signed-off-by: Kookoo Gu <Zhimin.Gu@solidigm.com>

### DIFF
--- a/iolog.c
+++ b/iolog.c
@@ -82,8 +82,8 @@ static void iolog_delay(struct thread_data *td, unsigned long delay)
 {
 	uint64_t usec = utime_since_now(&td->last_issue);
 	unsigned long orig_delay = delay;
-	uint64_t this_delay;
 	struct timespec ts;
+	int ret = 0;
 
 	if (delay < td->time_offset) {
 		td->time_offset = 0;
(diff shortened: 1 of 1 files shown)


iolog.c
```diff
@@ -97,13 +97,13 @@ static void iolog_delay(struct thread_data *td, unsigned long delay)
...
 
 	fio_gettime(&ts, NULL);
-	while (delay && !td->terminate) {
-		this_delay = delay;
-		if (this_delay > 500000)
-			this_delay = 500000;
 
-		usec_sleep(td, this_delay);
-		delay -= this_delay;
+	while(delay && !td->terminate) {
```
axboe, line 101: Style: missing space between ```while``` and parens start.

min22: Thank you for the catch!  I have fixed it in the new commit.

axboe: I don't see the new commit?

PR #1601, opened 2023-07-27
diskutil.h: fix missing headers wanted by the header

dpronin:
diskutil.h requires 3 more headers to fulfill several types therein  without having to rely on headers hopefully included before this one


dpronin:
@axboe I do not think that IOPS could have been affected by these changes, I assume there are fluctuations those might exist and sometimes are revealed

PR #1602, opened 2023-07-27
helper_thread.h: fix missing header and types forward declaration

dpronin:
- missing headers should be included at the places where they are certainly used, so include <stdbool.h>
- helper_thread_create() function requires two structures to be declared, so declare in advance `struct fio_sem` and `struct sk_out`

### DIFF
--- a/helper_thread.h
+++ b/helper_thread.h
@@ -3,6 +3,9 @@
 
 #include <stdbool.h>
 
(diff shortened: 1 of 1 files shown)


dpronin:
@axboe again, changes do not seem to affect tests so they fail with `iops ratio mismatch`

PR #1605, opened 2023-07-28
io_uring engine: 'atomic_load_relaxed' instead of 'atomic_load_acquire'

dpronin:
motivation here is that we do not have here any explicit READ dependency
 on atomic load because actually we just need in these places only
 operation to perform atomically without any explicit barriers given by
 memory model


dpronin:
@axboe I have no idea how changes in `io_uring` engine might influence `null` engine according to the results of FAILED tests

axboe:
Test 10 is known flaky, so not related and nothing to worry about.

PR #1606, opened 2023-07-28
use 'const' and 'static' where it is required

dpronin:
protect variables from programmer's point of view and hide several by  'static' variables

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -2494,10 +2494,7 @@ reap:
 							strerror(ret));
 			} else {
 				pid_t pid;
-				void *eo;
 				dprint(FD_PROCESS, "will fork\n");
-				eo = td->eo;
-				read_barrier();
 				pid = fork();
 				if (!pid) {
 					int ret;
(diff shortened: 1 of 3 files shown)


dpronin:
@axboe are there any ideas why `constness` and/or `static` of some variables/constants and/or parameters may affect runtime tests? I don't have one

vincentkfu:
I don't know why but your changes make the test jobs segfault on windows.

dpronin:
I presume, because I used to be a developer under windows, `static const` variables Windows compiler places in read-only segment section in binary that is mapped within PAGE with READ ONLY protection, and probably, there is some code that attempts to write to this section/mapped page

vincentkfu:
Actually the Windows failures started with the double free fix. So this patch is not the cause.

dpronin:
> Actually the Windows failures started with the double free fix. So this patch is not the cause.

Is there any opportunity to check these tests without launching up Windows as a virtual machine? Is there an emulator of cygwin under Linux?

vincentkfu:
None that I know of.

dpronin:
It's a pity

PR #1607, opened 2023-07-30
use 'const' where it is required

dpronin:
protect variables and parameters from programmer's point of view with 'constness'


dpronin:
@axboe merge in?

PR #1617, opened 2023-08-24
Record job start time obtained via clock_gettime to fix time pain points

### LINKED ISSUE #1544: Epoch issues

**Description of the bug:**
By default, fio log files use timestamps based on an epoch that is the beginning of the job. If you have multiple jobs, especially with some jobs waiting on other jobs before starting, this means you have different epochs and the times in the logs for each job cannot be compared/ordered.

Using `log_alternate_epoch` solves this problem as the logs all then share a common epoch. Unfortunately, it introduces a new problem. That is, the start time for each job relative to this common epoch is not known. So the gap between a job starting and the first log entry is indeterminate (although typically small, 0-1ms). (Less critically, `log_alternate_epoch` also increases log file size because the number of characters for each log entry timestamp increases.)

The shortcomings of each approach can be resolved by using the beginning of job epoch for log files but recording the start time of each job against a common (alternate) epoch.

I put something like this together in this PR from a while back: https://github.com/axboe/fio/pull/1353/commits

Back then I was more concerned about correlating log entries against other events on the same machine, and the PR was aimed at simplifying log files while still being able to know time against a machine-based epoch. It was nice to have but not super important (imho).

But as I've started to work with fio experiments with multiple jobs, having timekeeping shortcomings both with the normal epoch and `log_alternate_epoch` has led me to the same solution (the PR) to a different problem.

I was hoping the PR could be reconsidered or maybe some other solution brainstormed? I'm happy to implement a solution, whether it looks like the PR or something completely different.
**Environment**: 
All

**fio version**:
All

**Reproduction steps**
A simple fio run with per sample logging, each job waiting on the previous, illustrates the issues

aggieNick02:
Add a new key in the json per-job output, clock_gettime_job_start, that records the job start time obtained via a call to clock_gettime using the clock_id specified by the clock_id option. This allows times of fio jobs and log entries to be compared/ordered against each other and against other system events recorded against the same clock_id.

Rename the alternate_epoch_clock_id option to clock_id, as now this clock_id serves two purposes. The primary purpose is to be the clock_id for recording clock_gettime_job_start. The secondary purpose is to be the clock_id used if log_alternate_epoch is specified, in which case each log file timestamp is based on the epoch specified by clock_id. (Each such timestamp is obtained by taking the traditional zero-based timestamps and adding clock_gettime_job_start to them.)

We also make log_unix_epoch an official alias of log_alternate_epoch, instead of maintaining both redundant options.

Fixes #1544


HOWTO.rst
```diff
@@ -4103,22 +4103,24 @@ Measurements and reporting
...
+    timestamps.
 
 .. option:: log_alternate_epoch_clock_id=int
 
-	Specifies the clock_id to be used by clock_gettime to obtain the alternate epoch
-	if either log_unix_epoch or log_alternate_epoch are true. Otherwise has no
-	effect. Default value is 0, or CLOCK_REALTIME.
+    Backwards compatible alias for clock_id.
+
+.. option:: clock_id=int
```
vincentkfu, line 4119: The name `clock_id` is not specific enough and risks being confused with the option `clocksource`. How about something like `log_clock_id`?

HOWTO.rst
```diff
@@ -4103,22 +4103,24 @@ Measurements and reporting
...
+
```
vincentkfu, line 4120: There's a redundant mention of clock_gettime in the first sentence below.

stat.c
```diff
@@ -3044,11 +3046,13 @@ static void __add_log_sample(struct io_log *iolog, union io_sample_data data,
...
 		struct io_sample *s;
+		bool add_alternate_epoch;
 
 		s = get_sample(iolog, cur_log, cur_log->nr_samples);
 
 		s->data = data;
-		s->time = t + (iolog->td ? iolog->td->alternate_epoch : 0);
+		add_alternate_epoch = (iolog->td != NULL) && iolog->td->o.log_alternate_epoch;
+		s->time = t + (add_alternate_epoch ? iolog->td->clock_gettime_job_start : 0);
 		io_sample_set_ddir(iolog, s, ddir);
```
vincentkfu, line 3056: I think the following would be easier for a human to read:
```
s->time = t;
if (iolog->td && iolog->td->o.log_alternate_epoch)
    s->time += iolog->td->clock_gettime_job_start;
```

vincentkfu:
I don't see any significant problems here. You can manually test the client/server code changes by checking to see if things work as expected in client/server mode.

What happens to `clock_gettime_job_start` when `group_reporting` is enabled? Currently it seems that the key is just dropped which is probably fine, but this should be mentioned in the documentation. The alternative would be to list the different start times for each of the jobs that comprise the group but we can save that for a future enhancement.

.HOWTO.rst.swp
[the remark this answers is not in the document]
aggieNick02: Thanks, will remove

HOWTO.rst
aggieNick02: Gotcha. It used to be `log_alternate_epoch_clock_id`. The only problem with `log_clock_id` is that it now affects the job start time recorded in the json, which is independent of the `log_alternate_epoch` setting which controls if the log files use times based on this (needs better name) `clock_id`.

How about `alternate_epoch_clock_id`, and then I rename the `clock_gettime_job_start` variable and json entry to `alternate_epoch_job_start` ? I think that may work pretty well if it sounds ok to you.

HOWTO.rst
aggieNick02: Oops, thanks, I'll fix that.

aggieNick02:
> What happens to clock_gettime_job_start when group_reporting is enabled? Currently it seems that the key is just dropped which is probably fine, but this should be mentioned in the documentation. The alternative would be to list the different start times for each of the jobs that comprise the group but we can save that for a future enhancement.

I believe it will actually just record the first start time in the group. It's definitely not ideal. There's currently code in there that looks to do the same thing for `kb_base`, `unit_base`, `sig_figs`, `thread_number`, `pid`. For `kb_base` and `unit_base` there's even code that logs warnings if `kb_base` or `unit_base` is different for different jobs. But for all of those fields it does appear to just be first entry wins.

I agree it would be a nice enhancement to clean this up. A list of each set of values for each group entry probably makes sense, but I'm guessing the old values would need to be kept at the top-level too for backwards-compatibility purposes.

I could just not output the job start time if group reporting is on, if you think that is better.

aggieNick02:
> I don't see any significant problems here. You can manually test the client/server code changes by checking to see if things work as expected in client/server mode.

Thanks, I'll do this before I post an updated PR with the suggested corrections/changes.

vincentkfu:
> > What happens to clock_gettime_job_start when group_reporting is enabled? Currently it seems that the key is just dropped which is probably fine, but this should be mentioned in the documentation. The alternative would be to list the different start times for each of the jobs that comprise the group but we can save that for a future enhancement.
> 
> I believe it will actually just record the first start time in the group. It's definitely not ideal. There's currently code in there that looks to do the same thing for `kb_base`, `unit_base`, `sig_figs`, `thread_number`, `pid`. For `kb_base` and `unit_base` there's even code that logs warnings if `kb_base` or `unit_base` is different for different jobs. But for all of those fields it does appear to just be first entry wins.
> 
> I agree it would be a nice enhancement to clean this up. A list of each set of values for each group entry probably makes sense, but I'm guessing the old values would need to be kept at the top-level too for backwards-compatibility purposes.
> 
> I could just not output the job start time if group reporting is on, if you think that is better.



> > What happens to clock_gettime_job_start when group_reporting is enabled? Currently it seems that the key is just dropped which is probably fine, but this should be mentioned in the documentation. The alternative would be to list the different start times for each of the jobs that comprise the group but we can save that for a future enhancement.
> 
> I believe it will actually just record the first start time in the group. It's definitely not ideal. There's currently code in there that looks to do the same thing for `kb_base`, `unit_base`, `sig_figs`, `thread_number`, `pid`. For `kb_base` and `unit_base` there's even code that logs warnings if `kb_base` or `unit_base` is different for different jobs. But for all of those fields it does appear to just be first entry wins.
> 
> I agree it would be a nice enhancement to clean this up. A list of each set of values for each group entry probably makes sense, but I'm guessing the old values would need to be kept at the top-level too for backwards-compatibility purposes.
> 
> I could just not output the job start time if group reporting is on, if you think that is better.

When I did some testing yesterday I thought I saw the job start time disappear with group reporting on but I just tried it again and you are right, the key is still there. It would be nice to do something sane when group reporting is enabled but for now a warning about this situation in the documentation should be enough since this is an issue that fio doesn't handle well in general.

HOWTO.rst
aggieNick02: I'm happy to do that (completely separate option and `struct thread_data` member). I started out planning to do that but then ended up the way I submitted because, well, the options are more related than they seem from the documentation side. (You may already be aware of this, but just in case.) Here's the background.

Before I started using fio, there was a `log_unix_epoch` option. The docs say it logs Unix timestamps, and this is *kindof* true. What it actually does is remember the Unix timestamp when the job started (obtained via `clock_gettime`) and then add the elapsed job time (obtained based on the configured `clocksource`) to obtain a Unix-format timestamp for each entry. This is actually different than what you'd get by calling `clock_gettime` to get the Unix timestamp for each I/O operation, since Unix time is not monotonic and is affected by NTP.

Then I added `log_alternate_epoch`, which is just like `log_unix_epoch` but lets you use any `clock_id` instead of `CLOCK_REALTIME` (used for `log_unix_epoch`). (Interesting aside - in the case of using `CLOCK_MONOTONIC_RAW`, it turns out that the job start time plus the elapsed job time *can* actually be the same as what you'd get calling `clock_gettime` for each I/O,)

Anyways, that's why job start time ends up so "tied" to the options for log file timestamps. But there is no reason the log file timestamps and the recorded job start time have to share the same clock_id.

If I had a time machine, I'd stop any of the `log` options from being created, and the log files would always record time elapsed since job started, and a single `job_start_time` would be the only mechanism for recording time using another `clock_id`, since that is what is really happening in the implementation. The "alternate" timestamps in the log files are really an illusion, if that makes sense.

But I know that can't happen. Maybe all of this is really just another good argument for what you suggested, as it keeps the (imho) log timestamps and recording job start time completely separate in terms of implementation, so either one could be revamped in the future without having to worry about the other.

Just let me know what you think is best (just make it two separate things or try to come up with an option name for both purposes); I'm happy to do it either way.

Thanks for listening to my rambling thoughts. I

aggieNick02:
> When I did some testing yesterday I thought I saw the job start time disappear with group reporting on but I just tried it again and you are right, the key is still there. It would be nice to do something sane when group reporting is enabled but for now a warning about this situation in the documentation should be enough since this is an issue that fio doesn't handle well in general.

Sounds good. I'll add a doc warning with the next PR.

HOWTO.rst
vincentkfu: Thanks for the recap. I think it would be cleaner to just have an entirely separate start time feature.

aggieNick02: Cool, I'll do that then. Thanks for the feedback!

PR #1621, opened 2023-09-01
Record job start time to fix time pain points

### LINKED ISSUE #1544: Epoch issues

aggieNick02:
Add a new key in the json per-job output, job_start, that records the job start time obtained via a call to clock_gettime using the clock_id specified by the new job_start_clock_id option. This allows times of fio jobs and log entries to be compared/ordered against each other and against other system events recorded against the same clock_id.

Also make log_unix_epoch an official alias of log_alternate_epoch, instead of maintaining both redundant options.

Add a note to the documentation for group_reporting about how there are several per-job values for which only the first job's value is recorded in the json output format when group_reporting is enabled.

Fixes #1544

### DIFF
--- a/client.c
+++ b/client.c
@@ -956,6 +956,7 @@ static void convert_ts(struct thread_stat *dst, struct thread_stat *src)
 	dst->error		= le32_to_cpu(src->error);
 	dst->thread_number	= le32_to_cpu(src->thread_number);
 	dst->groupid		= le32_to_cpu(src->groupid);
+	dst->job_start		= le64_to_cpu(src->job_start);
 	dst->pid		= le32_to_cpu(src->pid);
 	dst->members		= le32_to_cpu(src->members);
 	dst->unified_rw_rep	= le32_to_cpu(src->unified_rw_rep);
--- a/server.c
+++ b/server.c
@@ -1706,6 +1706,7 @@ void fio_server_send_ts(struct thread_stat *ts, struct group_run_stats *rs)
 	p.ts.error		= cpu_to_le32(ts->error);
 	p.ts.thread_number	= cpu_to_le32(ts->thread_number);
 	p.ts.groupid		= cpu_to_le32(ts->groupid);
+	p.ts.job_start		= cpu_to_le64(ts->job_start);
 	p.ts.pid		= cpu_to_le32(ts->pid);
 	p.ts.members		= cpu_to_le32(ts->members);
 	p.ts.unified_rw_rep	= cpu_to_le32(ts->unified_rw_rep);
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -755,6 +755,10 @@ Time related parameters
 	calls will be excluded from other uses. Fio will manually clear it from the
 	CPU mask of other jobs.
 
+.. option:: job_start_clock_id=int
+   The clock_id passed to the call to `clock_gettime` used to record job_start
+   in the `json` output format. Default is 0, or CLOCK_REALTIME.
+
 
 Target file/device
 ~~~~~~~~~~~~~~~~~~
@@ -3965,6 +3969,13 @@ Measurements and reporting
 	per-job, use :option:`group_reporting`. Jobs in a file will be part of the
 	same reporting group, unless if separated by a :option:`stonewall`, or by
 	using :option:`new_group`.
+
+    NOTE: When :option: `group_reporting` is used along with `json` output,
+    there are certain per-job properties which can be different between jobs
+    but do not have a natural group-level equivalent. Examples include
+    `kb_base`, `unit_base`, `sig_figs`, `thread_number`, `pid`, and
+    `job_start`. For these properties, the values for the first job are
+    recorded for the group.
 
 .. option:: new_group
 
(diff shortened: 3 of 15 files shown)


aggieNick02:
Unrelated to the content of the PR - in testing client/server, I spun my wheels for a bit before I realized that command-line options from the client don't really seem to work - instead, everything needs to be in a job file. Any idea if this is known/expected? The docs make it look like command-line options are supported, but it doesn't seem like it. Wondering if it's a bug or if the docs need fixing.

client.c
```diff
@@ -956,6 +956,7 @@ static void convert_ts(struct thread_stat *dst, struct thread_stat *src)
 	dst->error		= le32_to_cpu(src->error);
 	dst->thread_number	= le32_to_cpu(src->thread_number);
 	dst->groupid		= le32_to_cpu(src->groupid);
+	dst->job_start = le64_to_cpu(src->job_start);
```
vincentkfu, line 959: Can you line this up like the other lines?

server.c
```diff
@@ -1706,6 +1706,7 @@ void fio_server_send_ts(struct thread_stat *ts, struct group_run_stats *rs)
 	p.ts.error		= cpu_to_le32(ts->error);
 	p.ts.thread_number	= cpu_to_le32(ts->thread_number);
 	p.ts.groupid		= cpu_to_le32(ts->groupid);
+	p.ts.job_start = cpu_to_le64(ts->job_start);
```
vincentkfu, line 1709: Please line this up like the other lines

vincentkfu:
The code changes all look reasonable. The only issue I see is that there is a single patch doing two things (adding the new key and making two options synonyms of each other). Could you separate these into two distinct patches?

vincentkfu:
> Unrelated to the content of the PR - in testing client/server, I spun my wheels for a bit before I realized that command-line options from the client don't really seem to work - instead, everything needs to be in a job file. Any idea if this is known/expected? The docs make it look like command-line options are supported, but it doesn't seem like it. Wondering if it's a bug or if the docs need fixing.

This sounds like something we should fix, although it's not clear whether we should change the documentation to match fio's behavior or vice versa.

client.c
aggieNick02: Sure, sorry about that.

server.c
aggieNick02: Sure, sorry about that.

aggieNick02:
> The code changes all look reasonable. The only issue I see is that there is a single patch doing two things (adding the new key and making two options synonyms of each other). Could you separate these into two distinct patches?

Sure, I can definitely do this if you want. The one wrinkle in doing so is that there will be conflicts as the changes touch some of the same lines. Do you want me to first do the new key and then wait until that PR is merged and then do the synonyms?

aggieNick02:
> > Unrelated to the content of the PR - in testing client/server, I spun my wheels for a bit before I realized that command-line options from the client don't really seem to work - instead, everything needs to be in a job file. Any idea if this is known/expected? The docs make it look like command-line options are supported, but it doesn't seem like it. Wondering if it's a bug or if the docs need fixing.
> 
> This sounds like something we should fix, although it's not clear whether we should change the documentation to match fio's behavior or vice versa.

I'll take a quick look and see if I can see what might be going wrong.

vincentkfu:
> > The code changes all look reasonable. The only issue I see is that there is a single patch doing two things (adding the new key and making two options synonyms of each other). Could you separate these into two distinct patches?
> 
> Sure, I can definitely do this if you want. The one wrinkle in doing so is that there will be conflicts as the changes touch some of the same lines. Do you want me to first do the new key and then wait until that PR is merged and then do the synonyms?

I would just have this one PR include both patches. Either order is fine.

aggieNick02:
> I would just have this one PR include both patches. Either order is fine.

Ah, ok. I just updated the PR with two separate commits. Thanks!

PR #1623, opened 2023-09-01
Add basic error checking to parsing nr from rw=randrw:<nr>, etc

### LINKED ISSUE #1622: Basic error checking needed in parsing nr from rw=randrw:<nr>, etc

**Description of the bug:**
Parsing of nr in rw=randrw:<nr> etc silently fails on bad input due to atoi use

**Environment**: All

**fio version**: Latest

**Reproduction steps**
Pass 'rw=randrw:foo' on command line. fio will run fine, I/O will be sequential

aggieNick02:
Previously this was parsed by just doing atoi(<nr>). This returns 0 or has undefined behavior in error cases.

Silently getting a 0 for nr is not great. In fact, 0 (or less) should likely not be allowed for nr; while the code handles it, the effective result is that the randomness is gone - all I/O becomes sequential. It makes sense to prohibit 0 as an nr value in the random case.

We leverage str_to_decimal to do our parsing instead of atoi. It isn't perfect, but it is a lot more resilient than atoi, and used in other similar places. We can then return an error when parsing fails, and also return an error when the parsed numeric value is outside of the ranges that can be stored in the unsigned int used for nr, along with when nr is 0.

Fixes #1622

### DIFF
--- a/options.c
+++ b/options.c
@@ -596,9 +596,21 @@ static int str_rw_cb(void *data, const char *str)
 	if (!nr)
 		return 0;
 
-	if (td_random(td))
-		o->ddir_seq_nr = atoi(nr);
-	else {
+	if (td_random(td)) {
+		long long val;
+
+		if (str_to_decimal(nr, &val, 1, o, 0, 0)) {
+			log_err("fio: randrw postfix parsing failed\n");
+			free(nr);
+			return 1;
+		}
+		if ((val <= 0) || (val > UINT_MAX)) {
+			log_err("fio: randrw postfix parsing out of range\n");
+			free(nr);
+			return 1;
+		}
+		o->ddir_seq_nr = (unsigned int) val;
+	} else {
 		long long val;
 
 		if (str_to_decimal(nr, &val, 1, o, 0, 0)) {


aggieNick02:
Hi @vincentkfu , I was wondering if you might take a quick look at this and let me know what you think. I've been burnt by this before so I think moving away from atoi is good. I can change it to allow 0 even though it doesn't make sense if we are worried there could be clients intentionally passing 0.

options.c
```diff
@@ -597,7 +597,22 @@ static int str_rw_cb(void *data, const char *str)
 		return 0;
 
 	if (td_random(td))
-		o->ddir_seq_nr = atoi(nr);
+	{
```
axboe, line 600: Braces go on the same line as the if. You have an odd mix here.

options.c
```diff
@@ -597,7 +597,22 @@ static int str_rw_cb(void *data, const char *str)
...
+		long long val;
+
+		if (str_to_decimal(nr, &val, 1, o, 0, 0)) {
+			log_err("fio: randrw postfix parsing failed\n");
+			free(nr);
+			return 1;
+		}
+		if ((val <= 0) || (val>UINT_MAX))
```
axboe, line 608: spacing, val > UINT_MAX

options.c
```diff
@@ -597,7 +597,22 @@ static int str_rw_cb(void *data, const char *str)
...
+		{
+			log_err("fio: randrw postfix parsing out of range\n");
+			free(nr);
+			return 1;
+		}
+		o->ddir_seq_nr = (unsigned int) val;
+	}
```
axboe, line 615: else and closing brace go on the same line.

axboe:
Some style fixups needed, but I agree with moving away from atoi().

aggieNick02:
Sorry about the brace/space issues. I think I have those corrected now. Thanks!

PR #1628, opened 2023-09-13
Fix typo in helper_thread.c and run block_signals(void) function in linux only

shailevi23:
The typo fix will now compile block_signals(void) correctly.

Compiling the code with the typo fix of "#ifdef CONFIG_PTHREAD_SIGMASK" will show that ret is unknown, hence, we are adding the line "int ret;".
This fix of typo will fix some other bugs I've encountered.

Signed-off-by: Shai Levy <shailevy23@gmail.com>

### DIFF
--- a/helper_thread.c
+++ b/helper_thread.c
@@ -106,9 +106,11 @@ static int read_from_pipe(int fd, void *buf, size_t len)
 
 static void block_signals(void)
 {
-#ifdef HAVE_PTHREAD_SIGMASK
+#ifdef CONFIG_PTHREAD_SIGMASK
 	sigset_t sigmask;
 
+	int ret;
+	
 	ret = pthread_sigmask(SIG_UNBLOCK, NULL, &sigmask);
 	assert(ret == 0);
 	ret = pthread_sigmask(SIG_BLOCK, &sigmask, NULL);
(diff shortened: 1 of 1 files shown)


vincentkfu:
Please fix the Windows msys2 build error. It looks like pthread_sigmask [may be defined to be a noop in msys2](https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/1370463485.10191.15.camel%40erik-laptop/#msg31004088). Consider changing the test in the configure script to trigger the same warning that is in the build failure to fail to detect pthread_sigmask when it is defined to be a noop.

axboe:
@vincentkfu I think maybe we just delete the code? It's clearly been disabled forever and we haven't seen any signal related issues there.

vincentkfu:
> @vincentkfu I think maybe we just delete the code? It's clearly been disabled forever and we haven't seen any signal related issues there.

I haven't seen any signal related issues either although the commit message does mention some bugs resolved by this fix. Without any details of these bugs I don't have strong feelings either way.

shailevi23:
@axboe 
I had a bug which was fixed adding this lines,
Hence, the signal was sent and blocked so please don’t delete it.

PR #1632, opened 2023-09-14
fix typo in helper_thread.c and run block_signals(void) function in linux only.

shailevi23:
fix typo in helper_thread.c and run block_signals(void) function in linux only.

Compiling the code with the typo fix of "#ifdef CONFIG_PTHREAD_SIGMASK" will show that ret is unknown, hence, we are adding the line "int ret;". This fix of typo will fix some other bugs I've encountered.

Also, sigmask works in linux environment, so we will if defined it to linux only.

Signed-off-by: Shai Levy shailevy23@gmail.com.

### DIFF
--- a/helper_thread.c
+++ b/helper_thread.c
@@ -106,13 +106,16 @@ static int read_from_pipe(int fd, void *buf, size_t len)
 
 static void block_signals(void)
 {
-#ifdef HAVE_PTHREAD_SIGMASK
-	sigset_t sigmask;
+#ifdef CONFIG_PTHREAD_SIGMASK
+	#if defined(__linux__)
+		sigset_t sigmask;
 
-	ret = pthread_sigmask(SIG_UNBLOCK, NULL, &sigmask);
-	assert(ret == 0);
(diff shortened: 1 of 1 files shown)


vincentkfu:
Your new patch will prevent the code from compiling on platforms like BSD and Solaris.

Please follow the suggestion I made in [your earlier PR](https://github.com/axboe/fio/pull/1628). Evidence suggests that the Windows msys2 build failed because on that platform `pthread_sigmask` is defined as a noop which triggered the unused variable warning for `sigmask`.

The `configure` script detects the existence of `pthread_sigmask` by trying to compile this program:

```
#include <stddef.h> /* NULL */
#include <signal.h> /* pthread_sigmask() */
int main(void)
{
  return pthread_sigmask(0, NULL, NULL);
}
```

Change that program to also trigger the same warning so that compilation also fails. This will make `CONFIG_PTHREAD_SIGMASK` undefined for the Windows msys2 build and then your original fix will pass all of our testing.

PR #1636, opened 2023-09-19
Update docs to clarify how to pass job options in client mode

### LINKED ISSUE #1629: Docs should indicate that client/server mode only supports defining jobs via job files, not command line arguments

**Description of the bug:**
Fio documentation needs updating to clarify that when running in client/server mode, job parameters must be passed via a job file, not via the command line. I'm happy to put up a PR with doc changes making this clear.

We just need verification that this is intended and not a bug - based on my looking at the code, I'm pretty confident, but if Jens could confirm, that would be great.

**Environment**: Applies to all

**fio version**: Latest

**Reproduction steps**
After playing a bit and exploring the code, I'm pretty certain that, when running in client/server mode, any job parameters passed on the client command line are discarded. The only way to convey job configuration to the server is via job files.

Code exploration details:
To be sent to the server as a command-line option, the option must be marked with `FIO_CLIENT_FLAG` - see the beginning of `parse_cmd_line` in `init.c` . Job-specific command line options are not marked as such, and they are thus parsed on the client in `parse_cmd_line` in `init.c` in the `FIO_GETOPT_JOB` case.

So the only way job-specific command line options could be sent to the client is by sending the populated thread options struct from the client to the server, as far as I can tell. On the server side when the options are received, `convert_thread_options_to_cpu` must be called, and the only caller of that function is` handle_update_job_cmd` in `server.c`. The only sender to the server of that net command (`FIO_NET_CMD_UPDATE_JOB`) is `fio_client_update_options` called from `goptions.c`, which I think is a graphical interface fio can have?

aggieNick02:
When run in client mode, fio does not pass any job options specified on the command line to the fio server. When run in client mode, all job options must be specified via local or remote job files. Update the docs to indicate this to avoid end-user confusion.

Fixes #1629

### DIFF
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -5105,6 +5105,9 @@ is the connect string, and `remote-args` and `job file(s)` are sent to the
 server. The `server` string follows the same format as it does on the server
 side, to allow IP/hostname/socket and port strings.
 
+Note that all job options must be defined in job files when running fio as a
+client. Any job options specified in `remote-args` will be ignored.
+
 Fio can connect to multiple servers this way::
 
     fio --client=<server1> <job file(s)> --client=<server2> <job file(s)>
--- a/fio.1
+++ b/fio.1
@@ -4838,6 +4838,9 @@ is the connect string, and `remote\-args' and `job file(s)' are sent to the
 server. The `server' string follows the same format as it does on the server
 side, to allow IP/hostname/socket and port strings.
 .P
+Note that all job options must be defined in job files when running fio as a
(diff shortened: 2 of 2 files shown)


vincentkfu:
For posterity:

The macOS CI job failed with runner image macos-12 v20230916.1: https://github.com/axboe/fio/actions/runs/6241938619/job/16944993465

I re-ran it and the job succeeded with runner image macos-12 v20230901.1: https://github.com/axboe/fio/actions/runs/6241938619/job/16945544582

Presumably v20230916.1 was problematic and GitHub switched things back behind the scenes.

PR #1639, opened 2023-09-29
verify: Fix the bad pattern block offset value

ipylypiv:
We offset `buf` by `header_size` for pattern verification. Add `header_size` to the mismatched `buf` offset to get the correct block offset value.


axboe:
Looks good, thanks.

PR #1641, opened 2023-10-03
Handle 32-bit overflows in disk utilization stats

stilor:
Linux prints [1] some of the values reported in block device's `stat` sysfs file as 32-bit unsigned integers. fio interprets them as 64-bit integers when reading that sysfs file and performs further arithmetics on them in 64-bits. If the reported value overflows during fio run, a huge bogus value is reported in the "disk utilization" block instead.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/block/genhd.c#n962

### DIFF
--- a/diskutil.c
+++ b/diskutil.c
@@ -103,8 +103,8 @@ static void update_io_tick_disk(struct disk_util *du)
 
 	fio_gettime(&t, NULL);
 	dus->s.msec += mtime_since(&du->time, &t);
-	memcpy(&du->time, &t, sizeof(t));
-	memcpy(&ldus->s, &__dus.s, sizeof(__dus.s));
+	du->time = t;
+	ldus->s = __dus.s;
 }
 
 int update_io_ticks(void)
--- a/diskutil.c
+++ b/diskutil.c
@@ -77,6 +77,23 @@ static int get_io_ticks(struct disk_util *du, struct disk_util_stat *dus)
 	return ret != 10;
 }
 
+static uint64_t safe_32bit_diff(uint64_t nval, uint64_t oval)
+{
+	/* Linux kernel prints some of the stat fields as 32-bit integers. It is
+	 * possible that the value overflows, but since fio uses unsigned 64-bit
+	 * arithmetic in update_io_tick_disk(), it instead results in a huge
+	 * bogus value being added to the respective accumulating field. Just
+	 * in case Linux starts reporting these metrics as 64-bit values in the
+	 * future, check that overflow actually happens around the 32-bit
+	 * unsigned boundary; assume overflow only happens once between
+	 * successive polls.
+	 */
+	if (oval <= nval || oval >= (1ull << 32))
+		return nval - oval;
+	else
+		return (1ull << 32) + nval - oval;
+}
+
 static void update_io_tick_disk(struct disk_util *du)
 {
 	struct disk_util_stat __dus, *dus, *ldus;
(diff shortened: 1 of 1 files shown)


diskutil.c
```diff
@@ -77,6 +77,24 @@ static int get_io_ticks(struct disk_util *du, struct disk_util_stat *dus)
 	return ret != 10;
 }
 
+static uint64_t safe_32bit_diff(uint64_t nval, uint64_t oval)
+{
+	/* Linux kernel prints some of the stat fields as 32-bit integers. It is
+	 * possible that the value overflows, but since fio uses unsigned 64-bit
+	 * arithmetics in update_io_tick_disk(), it instead results in a huge
```
szaydel, line 84: Suggest changing from `arithmetics` to `arithmetic`, sorry not grammar police. :)

diskutil.c
```diff
@@ -77,6 +77,24 @@ static int get_io_ticks(struct disk_util *du, struct disk_util_stat *dus)
...
+	 * bogus value being added to the respective accumulating field. Just
+	 * in case Linux starts reporting these metrics as 64-bit values in the
+	 * future, check that overflow actually happens around the 32-bit
+	 * unsigned boundary; assume overflow only happens once between
+	 * successive polls.
+	 */
+	if (oval <= nval || oval >= (1ull << 32)) {
```
ankit-sam, line 91: No need for curly braces for this `if else` condition

ankit-sam (commented):
@stilor please fix the build failure

vincentkfu:
Please change the link in the commit message to one that doesn't require login credentials.

stilor:
> @stilor please fix the build failure

This seems to be a GCC bug. I minimized the testcase to https://godbolt.org/z/nMEGzfW7Y and it still produces this error. Reported https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111696 to GCC maintainers with a minimal testcase derived from diskutil.c.

For now, disabled the warning using GCC diagnostic pragmas around the code triggering the spurious warning. Alternatively, if you prefer I could add a test for affected compilers in configure and if the compiler fails the test, add -Wno-stringop-overflow (to prevent such spurious warnings in all files.

diskutil.c
```diff
@@ -96,14 +113,21 @@ static void update_io_tick_disk(struct disk_util *du)
...
+	dus->s.ticks[0] += safe_32bit_diff(__dus.s.ticks[0], ldus->s.ticks[0]);
+	dus->s.ticks[1] += safe_32bit_diff(__dus.s.ticks[1], ldus->s.ticks[1]);
+	dus->s.io_ticks += safe_32bit_diff(__dus.s.io_ticks, ldus->s.io_ticks);
+	dus->s.time_in_queue +=
+			safe_32bit_diff(__dus.s.time_in_queue, ldus->s.time_in_queue);
 
 	fio_gettime(&t, NULL);
 	dus->s.msec += mtime_since(&du->time, &t);
 	memcpy(&du->time, &t, sizeof(t));
+
```
bvanassche, line 125: Please remove this `#pragma` and change the memcpy() calls in this function into assignments. There is no reason to use memcpy() here.

bvanassche (changes requested):

diskutil.c
stilor: Done.

diskutil.c
```diff
@@ -96,15 +113,16 @@ static void update_io_tick_disk(struct disk_util *du)
...
-	dus->s.time_in_queue += (__dus.s.time_in_queue - ldus->s.time_in_queue);
+	dus->s.ticks[0] += safe_32bit_diff(__dus.s.ticks[0], ldus->s.ticks[0]);
+	dus->s.ticks[1] += safe_32bit_diff(__dus.s.ticks[1], ldus->s.ticks[1]);
+	dus->s.io_ticks += safe_32bit_diff(__dus.s.io_ticks, ldus->s.io_ticks);
+	dus->s.time_in_queue +=
+			safe_32bit_diff(__dus.s.time_in_queue, ldus->s.time_in_queue);
 
 	fio_gettime(&t, NULL);
 	dus->s.msec += mtime_since(&du->time, &t);
-	memcpy(&du->time, &t, sizeof(t));
```
bvanassche, line 106: Please consider moving this change into a separate patch.

stilor: Done.

PR #1645, opened 2023-10-16
fix typo in helper_thread.c and add sigmask var to configure script in order to handle msys2.

shailevi23:
fix typo in helper_thread.c and add sigmask var to configure to detect the existence of pthread_sigmask.

Compiling the code with the typo fix of "#ifdef CONFIG_PTHREAD_SIGMASK" will show that ret is unknown, hence, we are adding the line "int ret;". This fix of typo will fix some other bugs I've encountered.
In case of a msys2, configure script handles it, it triggers the same warning that is in the build failure to fail to detect pthread_sigmask when it is defined to be a noop.

Signed-off-by: Shai Levy [shailevy23@gmail.com](mailto:shailevy23@gmail.com).

### DIFF
--- a/configure
+++ b/configure
@@ -864,7 +864,8 @@ cat > $TMPC <<EOF
 #include <signal.h> /* pthread_sigmask() */
 int main(void)
 {
-  return pthread_sigmask(0, NULL, NULL);
+  sigset_t sigmask;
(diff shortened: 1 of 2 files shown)


shailevi23:
@vincentkfu Thanks, I've changed the commit messages as you asked.

PR #1648, opened 2023-10-18
fix assert failed when timeout during call rate_ddir

zqs-Oppenauer:
fix assert failed when timeout during call rate_ddir

Adding DDIR_TIMEOUT in enum `fio_ddir`, and `rate_ddir` returns it when fio timeouts. `set_io_u_file` will directly break out of the loop, and `fill_io_u` won't be called again, which causes assert to fail in `rate_ddir`, because td->rwmix_ddir is DDIR_INVAL.

 Signed-off-by: QingSong Zhu <zhuqingsong.0909@bytedance.com>

### DIFF
--- a/io_u.c
+++ b/io_u.c
@@ -717,7 +717,7 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
 		 * check if the usec is capable of taking negative values
 		 */
 		if (now > td->o.timeout) {
-			ddir = DDIR_INVAL;
+			ddir = DDIR_TIMEOUT;
 			return ddir;
 		}
 		usec = td->o.timeout - now;
@@ -726,7 +726,7 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
 
 	now = utime_since_now(&td->epoch);
 	if ((td->o.timeout && (now > td->o.timeout)) || td->terminate)
-		ddir = DDIR_INVAL;
+		ddir = DDIR_TIMEOUT;
 
 	return ddir;
 }
@@ -951,7 +951,7 @@ static int fill_io_u(struct thread_data *td, struct io_u *io_u)
 
 	set_rw_ddir(td, io_u);
 
-	if (io_u->ddir == DDIR_INVAL) {
+	if (io_u->ddir == DDIR_INVAL || io_u->ddir == DDIR_TIMEOUT) {
 		dprint(FD_IO, "invalid direction received ddir = %d", io_u->ddir);
 		return 1;
 	}
@@ -1419,6 +1419,10 @@ static long set_io_u_file(struct thread_data *td, struct io_u *io_u)
 		put_file_log(td, f);
 		td_io_close_file(td, f);
 		io_u->file = NULL;
+
+		if (io_u->ddir == DDIR_TIMEOUT)
+			return 1;
+
 		if (td->o.file_service_type & __FIO_FSERVICE_NONUNIFORM)
 			fio_file_reset(td, f);
 		else {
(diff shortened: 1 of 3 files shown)


axboe:
This needs a MUCH better description in the actual commit message. There's nothing in there, you left it all on the github pull request message which won't go into the tree.

axboe:
Caveat - haven't attempted to reproduce this, but can't we just do:

```
diff --git a/io_u.c b/io_u.c
index 07e5bac5ef52..27c4d55fa4b7 100644
--- a/io_u.c
+++ b/io_u.c
@@ -792,7 +792,7 @@ static enum fio_ddir get_rw_ddir(struct thread_data *td)
 		 * https://github.com/axboe/fio/issues/1501#issuecomment-1418327049
 		 */
 		td->rwmix_ddir = ddir;
-	} else
+	} else if (ddir != DDIR_INVAL)
 		td->rwmix_ddir = rate_ddir(td, ddir);
 	return td->rwmix_ddir;
 }
```

and be done with it?

axboe:
Or probably this is the better way, same kind of thing:

```
diff --git a/io_u.c b/io_u.c
index 07e5bac5ef52..ea016c90e0e2 100644
--- a/io_u.c
+++ b/io_u.c
@@ -792,7 +792,7 @@ static enum fio_ddir get_rw_ddir(struct thread_data *td)
 		 * https://github.com/axboe/fio/issues/1501#issuecomment-1418327049
 		 */
 		td->rwmix_ddir = ddir;
-	} else
+	} else if (ddir_rw(ddir))
 		td->rwmix_ddir = rate_ddir(td, ddir);
 	return td->rwmix_ddir;
 }
```

zqs-Oppenauer:
> This needs a MUCH better description in the actual commit message. There's nothing in there, you left it all on the github pull request message which won't go into the tree.
> 
> You also need the Signed-off-by in there, no in the git pull request.

Thanks very much for your 

> This needs a MUCH better description in the actual commit message. There's nothing in there, you left it all on the github pull request message which won't go into the tree.
> 
> You also need the Signed-off-by in there, no in the git pull request.

Thank you very much for your reminder. I will add  description and Signed-off-by in the commit message.

axboe:
I did ponder whether DDIR_TIMEOUT would be a solution, since it's much cleaner than hacking in odd return values and needing to check it twice. That provides a way to pass this information back inline, rather than out-of-band. I think that proposed patch looks MUCH better than the original one.

zqs-Oppenauer:
> I did ponder whether DDIR_TIMEOUT would be a solution, since it's much cleaner than hacking in odd return values and needing to check it twice. That provides a way to pass this information back inline, rather than out-of-band. I think that proposed patch looks MUCH better than the original one.

Thanks very much, I have already updated this commit according to the proposed patch and added the description and Signed-off-by in the commit message.

io_u.c
```diff
@@ -1419,6 +1419,11 @@ static long set_io_u_file(struct thread_data *td, struct io_u *io_u)
 		put_file_log(td, f);
 		td_io_close_file(td, f);
 		io_u->file = NULL;
+
+		if (io_u->ddir == DDIR_TIMEOUT) {
```
axboe, line 1423: No braces necessary here.

axboe:
Looks much better, just one minor style issue. Outside of that, please line wrap commit messages at around 72 chars. Long lines are hard to read in git log. With those two things tweaked, this should be ready to go in.

axboe:
Once I triggered CI, it failed. The cases where we have a switch based on ddir, you need to also add DDIR_TIMEOUT to the DDIR_INVAL cases.

axboe:
Thanks!

Issue #1655, opened 2023-10-23
fio --enghelp: infinitely says 'nfs'


vt-alt:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
`fio -i` infinitely says `nfs`.

**Environment**: ALT Linux Sisyphus

**fio version**: `fio-3.36` compiled with with `--dynamic-libengines`. Without `--dynamic-libengines` there is no repeating. 

**Reproduction steps**
* Install `fio` and `libaio` with any other ext engine, tested with `fio-engine-http` and `fio-engine-nbd`, then run `fio -i`:
```
$ fio --enghelp
Available IO engines:
        io_uring
        io_uring_cmd
        sg
        mtd
        gfapi_async
        gfapi
        splice
        e4defrag
        falloc
        posixaio
        nfs
        nfs
        nfs
        nfs
        nfs
        nfs
        nfs
        nfs
        nfs
        nfs
        nfs
...
```
`fio --enghelp=http` and `fio --enghelp=nbd` still work.

If additional engines are not installed `nfs` printed once and internal engine list is correct.

If only `fio-libaio.so` is installed `nfs` is (incorrectly) listed as the last engine (i.e. printing stops after `nfs`).

GDB backtrace after ^C while `nfs` is printing:
```
(gdb) bt
#0  0x00007ffff7d39610 in __GI___libc_write (fd=1, buf=0x555555763500, nbytes=5) at ../sysdeps/unix/sysv/linux/write.c:26
#1  0x00007ffff7cbbf85 in _IO_new_file_write (f=0x7ffff7e125c0 <_IO_2_1_stdout_>, data=0x555555763500, n=5) at fileops.c:1181
#2  0x00007ffff7cba2b4 in new_do_write (fp=0x7ffff7e125c0 <_IO_2_1_stdout_>,
    data=0x555555763500 "\tnfs\n\nelete\n engines:\n:\tfffff\nCpus_allowed_list:\t0-19\nMems_allowed:\t00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,00000000,000000"..., to_do=to_do@entry=5)
    at /usr/src/debug/glibc-2.38.0.27.750a45a783-alt1/libio/libioP.h:1030
[...]

PR #1659, opened 2023-10-25
nfs: Fix incorrect engine registering for '--enghelp' list

vt-alt:
`ioengine` from `nfs` (internal) engine is incorrectly exported thus overriding its value in constructor callbacks of other external engines, that are used for registering engine for listing with `--enghelp`.

Because flist is unsafe to double adding it also making `engine_list` to become corrupt and causing infinite loop or abnormal list termination when printing engine list.

Issue: https://github.com/axboe/fio/issues/1655
Fixes: 9326926b ("NFS engine")

---
It seems that `ioengine` for internal engines should not be exported and should be always `static`.

The change is tested downstream on ALT Linux.

### DIFF
--- a/engines/nfs.c
+++ b/engines/nfs.c
@@ -308,7 +308,7 @@ static int fio_libnfs_close(struct thread_data *td, struct fio_file *f)
 	return ret;
 }
 
-struct ioengine_ops ioengine = {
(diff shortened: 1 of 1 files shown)


vincentkfu:
Thanks

PR #1662, opened 2023-10-31
crct10: use isa-l for crc if available

ankit-sam:
isa-l provides fast implementation for various polynomials.
This will be only used for end to end data protection, and has a significant impact on performance.

For ISA_L please check: https://github.com/intel/isa-l

### DIFF
--- a/configure
+++ b/configure
@@ -189,6 +189,7 @@ libiscsi="no"
 libnbd="no"
 libnfs=""
 xnvme=""
+isal=""
 libblkio=""
 libzbc=""
 dfs=""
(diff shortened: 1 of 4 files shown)


crc/crct10dif_common.c
```diff
@@ -24,6 +24,17 @@
...
  */
 
+#ifdef CONFIG_LIBISAL
+#include <isa-l/crc.h>
+
+extern unsigned short fio_crc_t10dif(unsigned short crc,
+				     const unsigned char *buffer,
+				     unsigned int len)
+{
+	return (crc16_t10dif(crc, buffer, len));
```
axboe, line 34: extra parens here

configure
```diff
@@ -2684,6 +2688,19 @@ if test "$xnvme" != "no" ; then
 fi
 print_config "xnvme engine" "$xnvme"
 
+##########################################
```
axboe, line 2691: I'd feel better if this actually tried to compile something using crc16_t10dif(crc, buffer, len) just so we know it works.

ankit-sam: Thanks for reviewing quickly, please have a look now.

PR #1673, opened 2023-11-20
engines/xnvme: only include entry-header ('libxnvme.h')

safl:
This changes how the xNVMe fio io-engine consumes the xNVMe library by only including the library-entry header "libxnvme.h".

From version 0.7.0 the xNVMe API headers are refactored to drop header guards on the individual headers and abide by the idiom of "headers must not include other headers".

The exception is the library-entry header "libxnvme.h". The library-entry-header includes all headers provided with xNVMe, which is a convenient approach to consuming the library. One where, in case the API namespace grows or shrinks, then the xNVMe fio io-engine need not change how it includes xNVMe.

However, since fio has consumed the main-entry header and individual headers, xNVMe has held back on removing the guards on _nvm, _zns, and _spec to avoid breaking the xNVMe fio engine. They will eventually be deprecated. Thus, this change to consume xNVMe in the manner intended from version v0.7.0 and onwards.

### DIFF
--- a/engines/xnvme.c
+++ b/engines/xnvme.c
@@ -10,10 +10,6 @@
 #include <stdlib.h>
 #include <assert.h>
 #include <libxnvme.h>
-#include <libxnvme_libconf.h>
-#include <libxnvme_nvm.h>
-#include <libxnvme_znd.h>
-#include <libxnvme_spec_fs.h>
 #include "fio.h"
 #include "zbd_types.h"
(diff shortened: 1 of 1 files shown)


axboe:
Patch looks fine, but color me skeptical on removing header guards for xnvme. Not my project so don't really care, but seems like a bad idea and you're just going to be fielding silly issues because of it.

PR #1689, opened 2023-12-15
Fix a few issues found by scanner

preichl:
Following issues were found by Red Hat's OpenScanHub:

fio-3.35/engines/rdma.c:279:3: warning[deadcode.DeadStores]:
	Value stored to 'ret' is never read

fio-3.35/server.c:1884:3: warning[deadcode.DeadStores]:
	Value stored to 'extended_buf_wp' is never read

fio-3.35/engines/http.c:642: var_tested_neg:
	Assigning: "r" = a negative value.
fio-3.35/engines/http.c:714: assign: Assigning: "io_u->error" = "r". fio-3.35/engines/http.c:715: negative_returns: "io_u->error" is passed
	to a parameter that cannot be negative.

I suggest changing the data type of io_u->error from uint to int since it appears that negative values are being assigned to it.

### DIFF
--- a/engines/http.c
+++ b/engines/http.c
@@ -640,7 +640,6 @@ static enum fio_q_status fio_http_queue(struct thread_data *td,
 	char url[1024];
 	long status;
 	CURLcode res;
-	int r = -1;
 
 	fio_ro_check(td, io_u);
 	memset(&_curl_stream, 0, sizeof(_curl_stream));
@@ -712,7 +711,7 @@ static enum fio_q_status fio_http_queue(struct thread_data *td,
 	log_err("WARNING: Only DDIR_READ/DDIR_WRITE/DDIR_TRIM are supported!\n");
 
 err:
-	io_u->error = r;
+	io_u->error = -1;
 	td_verror(td, io_u->error, "transfer");
 out:
 	curl_slist_free_all(slist);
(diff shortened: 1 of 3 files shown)


axboe:
The changes looks fine, but the commit title:

Fix a few issues found by scanner

is pretty bad as it means absolutely nothing. What scanner? What issues? I strongly suspect this should be 3 separate commits, not one bundled up.

preichl:
Thank you for comments, please see 2nd version.

io_u.h
```diff
@@ -81,7 +81,7 @@ struct io_u {
 	struct io_piece *ipo;
 
 	unsigned long long resid;
-	unsigned int error;
+	int error;
```
axboe, line 84: Changes like this HAS to be done very carefully, and not just because some static checker has a blurb about it. The error field is positive values, did you consider that perhaps it's that one use case that may be broken? That would be worthwhile fixing up.

axboe:
That io_u->error change is highly suspect imho. The other 3 patches look fine. I'd suggest you drop that one for now and we can get the other 3 applied, and then take a deeper look at if this is an inconsistent use of the error field in http instead.

axboe:
Pulled with that dropped, and then fixed up the error in http.

PR #1706, opened 2024-01-25
fix reported defect from coverity scan

ankit-sam:
Fix the two Null pointer dereferences issue reported by Coverity scan

Null pointer dereferences  (FORWARD_NULL)
Dereferencing null pointer "l->td"

Null pointer dereferences  (REVERSE_INULL)
Null-checking "p->td" suggests that it may be null, but it has already been dereferenced on all paths leading to the check.

### DIFF
--- a/iolog.c
+++ b/iolog.c
@@ -862,7 +862,12 @@ void setup_log(struct io_log **log, struct log_params *p,
 		l->log_ddir_mask = LOG_OFFSET_SAMPLE_BIT;
 	if (l->log_prio)
 		l->log_ddir_mask |= LOG_PRIO_SAMPLE_BIT;
-	if (l->td->o.log_max == IO_LOG_SAMPLE_BOTH)
+	/*
+	 * The bandwidth-log option generates agg-read_bw.log,
+	 * agg-write_bw.log and agg-trim_bw.log for which l->td is NULL.
+	 * Check if l->td is valid before dereferencing it.
+	 */
+	if (l->td && l->td->o.log_max == IO_LOG_SAMPLE_BOTH)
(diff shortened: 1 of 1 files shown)


iolog.c
```diff
@@ -862,7 +862,7 @@ void setup_log(struct io_log **log, struct log_params *p,
 		l->log_ddir_mask = LOG_OFFSET_SAMPLE_BIT;
 	if (l->log_prio)
 		l->log_ddir_mask |= LOG_PRIO_SAMPLE_BIT;
-	if (l->td->o.log_max == IO_LOG_SAMPLE_BOTH)
+	if (l->td && l->td->o.log_max == IO_LOG_SAMPLE_BOTH)
```
axboe, line 865: This is a work-around just to appease the checker, not a huge fan. Either p->td is always valid, and in which case l->td is also always valid, or it's valid to call setup_log() with NULL p->td and checking is needed. Decide which one it is and do the fix based on that.

ankit-sam: Yes `setup_log()` gets called with NULL p->td for `agg-read_bw.log`, `agg-write_bw.log` and `agg-trim_bw.log`. So this check is required.

I think this is the reason we have the same check few lines above.
`if (l->td && l->td->o.io_submit_mode != IO_MODE_OFFLOAD)`

ankit-sam: Done, please take a look.

axboe: Did you push it? Don't see it.

ankit-sam: Ah my mistake, I only updated the commit message previously. Its should be done now.

axboe:
Thanks!

Issue #1725, opened 2024-02-18
ZBD: assertion when check offset valid during reset zone


LinXpy:
**Please acknowledge the following before creating a ticket**

- [y ] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
* During io testing with percentage size on the ZBD device, fio asserted when check offset valid: assertion when check offset valid during reset zone
<img width="508" alt="assert-is-valid-offset" src="https://github.com/axboe/fio/assets/18701082/6f1327da-3e70-4af6-98f5-986626553bde">

**Environment**:  5.15.0-53-generic #59~20.04.1-Ubuntu SMP Thu Oct 20 15:10:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

**fio version**: fio-3.36 and old versions

**Reproduction steps**
* reproduce cmd: fio -filename=/dev/nvme0n1 -ioengine=libaio -direct=1 -iodepth=32 -thread -name=fio_name -numjobs=1 -group_reporting -verify_backlog=16777216 -cpus_allowed=1 -rw=write -zonemode=zbd -significant_figures=10 -bs=128K -do_verify=1 -verify_pattern=0x5A -offset=53% -max_open_zones=16 -size=20%
* necessary to reproduce the issue (https://stackoverflow.com/help/mcve ) --> use the percentage value for the param "offset" & "size" is easy to reproduce.

axboe:
Since you already root caused the problem, why not just send a patch or PR to fix it?

LinXpy:
> Since you already root caused the problem, why not just send a patch or PR to fix it?

I'm not sure about the detailed procedure for requesting PR to fio repo, is there any guidance doc? Thanks.

vincentkfu:
Here is a guide that seems reasonable: https://opensource.com/article/19/7/create-pull-request-github

LinXpy:
@vincentkfu  thanks for the guide link, i followed it and requested a PR: https://github.com/axboe/fio/pull/1728 for the fix.

PR #1728, opened 2024-02-24
zbd: fix assertion issue during reset zone when using percentage offset

LinXpy:

zbd: fix assertion issue during reset zone when using percentage offset

Add a new member for structure fio_file to record the new start offset which is ever updated, especially in zbd test, the file offset maybe ever adjusted according to the zone's boundary and bs alignment when using percentage offset value.

Signed-off-by: LinXpy <paynefd1222@163.com>

### DIFF
--- a/file.h
+++ b/file.h
@@ -101,6 +101,7 @@ struct fio_file {
 	uint64_t real_file_size;
 	uint64_t file_offset;
 	uint64_t io_size;
+	uint64_t start_offset_update;
 
 	struct fio_ruhs_info *ruhs_info;
 
--- a/filesetup.c
+++ b/filesetup.c
@@ -1173,6 +1173,7 @@ int setup_files(struct thread_data *td)
 	need_extend = 0;
 	for_each_file(td, f, i) {
 		f->file_offset = get_start_offset(td, f);
+		f->start_offset_update = f->file_offset;
 
 		/*
 		 * Update ->io_size depending on options specified.
--- a/libfio.c
+++ b/libfio.c
@@ -122,6 +122,9 @@ void clear_io_state(struct thread_data *td, int all)
 	for_each_file(td, f, i) {
 		fio_file_clear_done(f);
 		f->file_offset = get_start_offset(td, f);
+		if (f->file_offset != f->start_offset_update) {
+			f->file_offset = f->start_offset_update;
+		}
 	}
 
 	/*
--- a/zbd.c
+++ b/zbd.c
@@ -698,6 +698,7 @@ static bool zbd_zone_align_file_sizes(struct thread_data *td,
 			 new_offset);
 		f->io_size -= (new_offset - f->file_offset);
 		f->file_offset = new_offset;
+		f->start_offset_update = new_offset;
 	}
 
 	z = zbd_offset_to_zone(f, f->file_offset + f->io_size);


vincentkfu:
@kawasaki Can you take a look?

kawasaki:
Thanks for the report. I tried the following command line, and observed an assertion failure.

```
 sudo ./fio --name=w --filename=/dev/nvme0n1 --zonemode=zbd --direct=1 --rw=write --size=2z --offset=1% --loop=2                                                                         
w: (g=0): rw=write, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
fio-3.36-89-g6ed0
Starting 1 process
/dev/nvme0n1: rounded up offset from 10739712 to 12582912
/dev/nvme0n1: rounded down io_size from 6545408 to 4194304
fio: zbd.c:109: zone_lock: Assertion `f->min_zone <= nz && nz < f->max_zone' failed.
fio: pid=102122, got signal=6

w: (groupid=0, jobs=1): err= 0: pid=102122: Tue Feb 27 19:15:58 2024
  lat (usec)   : 250=71.97%, 500=20.70%, 750=7.23%
  lat (msec)   : 2=0.10%
     issued rwts: total=0,1024,0,0 short=0,0,0,0 dropped=0,0,0,0

  WRITE: bw=0B/s (0B/s), 0B/s-0B/s (0B/s-0B/s), io=0B (0B), run=248-248msec

Disk stats (read/write):
  nvme0n1: ios=125/1024, sectors=4408/8192, merge=0/0, ticks=13/187, in_queue=200, util=46.13%
fio: file hash not empty on exit
```

I guess this is the issue that @LinXpy faced. If not, please let me know. In general, this assertion failure is observed when the write workload fulfills two conditions:

1. --offset is not aligned to zone boundaries
2. the workload is repeated with --loops or --time_based

I'm not sure if such workloads are meaningful. They request repeated writes not at write pointers, which are not expected. @LinXpy, if you have any background to run such workload, please share.

If users really want to run such workloads, I think the fix by @LinXpy is essentially correct. I think it still needs brush up to have better variable name and to reduce changes out of zbd.c. Of note is that the change will add a branch in clear_io_state() and add 64 bit to struct fio_file. I hope this additional overhead is acceptable, but before such discussion, I would like to consider if such workload (repeated write to offset unaligned zone boundaries) is really useful.

LinXpy:
Hi, @kawasaki, the failure scene you saw seems a little diff from mine, for the background and the reproduce fio cmd, you can refer to the ticket: https://github.com/axboe/fio/issues/1725.

In my test, i used the below cmd, and can 100% reproduce the failure.
`fio -filename=/dev/nvme0n1 -ioengine=libaio -direct=1 -iodepth=32 -thread -name=fio_name -numjobs=1 -group_reporting -verify_backlog=16777216 -cpus_allowed=1 -rw=write -zonemode=zbd -significant_figures=10 -bs=128K -do_verify=1 -verify_pattern=0x5A -offset=53% -max_open_zones=16 -size=20%`

In my test, the pertentage value for offset is also a must factor to reproduce the failure, but my test didn't run multi loops(only 1 loop) and didn't do time_based test, it just want to fill some data area.

kawasaki:
@LinXpy Thanks for sharing the workload detail. I was able to recreate the assertion failure. I noticed that you the workload has verify options. Actually, the verify feature has similar effect as loop or time_based option: ref [reset_io_counters()](https://github.com/axboe/fio/blob/2dddfd35396e2f7a1bb06cc7c92aa1e283be084e/libfio.c#L110). I disabled the verify option, then I saw the assertion failure disappears. So, this looks a corner case for me: all of zonemode=zbd, percentage offset and verify option are specified, the assertion failure happens. As I noted, percentage offset is not meaningful to zoned block devices. So I suggest not to use the three options together, rather than your fix which adds the overhead.

One lesson I learnt is that the assertion message is not informative. I'm thinking to replace the assertion failure message with more meaningful error message to guide review the offset alignment.

LinXpy:
@kawasaki , that makes sense, i agree. The percentage offset/size for ZBD device is not so meaningful, but there is possibility of someone will use the percentage offset/size for ZBD device, so if a good meaningful error message to guide the user to use a specific value instead of percentage value to avoid the assertion failure, that'll be very helpful!

Issue #1735, opened 2024-03-19
Fio iolog replay does not report disk stats corrrectly


yygcode:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
Fio iolog replay does not report disk stats corrrectly, and the `utils` always be zero. And It can be fixed by the patch:

```
$ git diff
diff --git a/iolog.c b/iolog.c
index f52a9a80..c065e421 100644
--- a/iolog.c
+++ b/iolog.c
@@ -812,6 +812,8 @@ bool init_iolog(struct thread_data *td)
        if (!ret)
                td_verror(td, EINVAL, "failed initializing iolog");
 
+       init_disk_util(td);
+
        return ret;
 }
```

**Environment**: 
```
[yanyg@x1{192.168.50.132} ~/git/fio ] <2024-03-19 23:43:15>
$ uname -a
Linux x1 6.3.0 #1 SMP PREEMPT_DYNAMIC Mon May  1 13:09:40 CST 2023 x86_64 GNU/Linux
[yanyg@x1{192.168.50.132} ~/git/fio ] <2024-03-19 23:43:36>
```

**fio version**: 
```
$ ./build/fio --version
fio-3.36
[yanyg@x1{192.168.50.132} ~/git/fio ] <2024-03-19 23:43:54>
$ git log --oneline -1
7d6c99e9 (HEAD -> master, origin/master, origin/HEAD) docs: fix documentation for rate_cycle
```

**Reproduction steps**

Step1: Generates iolog

```
#! /bin/bash

gen_iolog()
{
    local dev="./test.file"
    local val=""
    echo "fio version 2 iolog"
    echo "$dev add"
    echo "$dev open"
    SIZE=$((1024 * 1024))
    val=$(for ((x = 0; x < SIZE; x += 512)); do
              echo "$dev write $x 512"
          done)

    val="$val
$val
$val
$val"
    echo "$val"
    echo "$dev close"
}

gen_iolog > test.iolog
```

Write above scripts to gen_iolog.sh and run it.

Step2: Write config
```
$ cat test.cfg 
; TEST IOLOG
[global]
direct=1
filename=./test.file
filesize=128M
ioengine=libaio

[job1]
rw=write
iodepth=1
runtime=10
read_iolog=./test.iolog
```

Step3: Run fio
```
$ sudo ./build/fio_orig test.cfg 
[...]

PR #1736, opened 2024-03-20
iolog: fix disk stats issue

yygcode:
iolog: fix disk stats issue
    
In the iolog replay scenario, the disk util in the td structure is
not initialized, resulting in the disk stats not being correctly
updated.
    
Fixes: https://github.com/axboe/fio/issues/1735
    
Signed-off-by: Yong Gang <yygcode@gmail.com>

### DIFF
--- a/iolog.c
+++ b/iolog.c
@@ -588,6 +588,8 @@ int init_iolog(struct thread_data *td)
 	if (ret)
 		td_verror(td, EINVAL, "failed initializing iolog");
 
+	init_disk_util(td);
+
 	return ret;
 }
 


vincentkfu:
- Please use your real name in the sign off: Signed-off-by: `REAL NAME <yygcode@gmail.com>`
- Also you can include above the sign off a line with "Fixes: https://github.com/axboe/fio/issues/1735"
- Finally, the first line of the commit message should omit [PATCH] and instead be something like `iolog: fix disk stats issue`

yygcode:
> * Please use your real name in the sign off: `Signed-off-by: REAL NAME <yygcode@gmail.com>`
> * Also you can include above the sign off a line with `Fixes: Link-to-the-issue-you-filed`
> * Finally, the first line of the commit message should omit [PATCH] and instead be something like `iolog: fix disk stats issue`

Thanks. All comments have been resolved.

PR #1741, opened 2024-03-23
Fix: 12 typos in xxhash.c

RoboSchmied:
fix typos in xxhash.c


axboe:
This is an import from an external source, I think it's better to leave typos as-is to remain consistent.

Issue #1739, opened 2024-03-22
Created files should be cleaned up at the end of the run


edigaryev:
Problem: when invoking `fio --name=test --size=1MB` in a given directory, it leaves a file named `test.0.0` in that directory and this file is neither (1) removed automatically nor (2) there's an option to remove it.

PR #1742, opened 2024-03-25
docs: use "delete" term instead of "unlink", which is less common

edigaryev:
This way people would have an easier time discovering this command-line option, see https://github.com/axboe/fio/issues/1739.

### DIFF
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -971,13 +971,13 @@ Target file/device
 
 .. option:: unlink=bool
 
-	Unlink the job files when done. Not the default, as repeated runs of that
+	Unlink (delete) the job files when done. Not the default, as repeated runs of that
 	job would then waste time recreating the file set again and again. Default:
 	false.
 
 .. option:: unlink_each_loop=bool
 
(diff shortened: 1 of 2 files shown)


axboe:
I'm fine adding "delete" in there as well, but don't get rid of unlink. It may not mean anything to you, but it's a very commonly used word and in fact fully reflects what is being done here. As you may have noticed, it's what the option is called as well.

edigaryev:
>I'm fine adding "delete" in there as well, but don't get rid of unlink.

See https://github.com/axboe/fio/pull/1742/commits/4efb3c69db5bf39f74a357feeed1c79f8fb0ae25.

axboe:
I think you missed some of your changes, and please squash them into one commit.

edigaryev:
> I think you missed some of your changes, and please squash them into one commit.

Good catch, see https://github.com/axboe/fio/pull/1742/commits/3fe133a51a9c51052388a7b6a1cd8ff0a67746b6.

Mxdon123 (approved):

PR #1767, opened 2024-05-30
iolog, dataplacement: fix coverity scan defect

parkvibes:
### **iolog: fix Error handling issues (NEGATIVE_RETURNS)**

CID 494151: Error handling issues (NEGATIVE_RETURNS) @ io_u.c:1877 in get_io_u()
This patch removes negative returns from dp_init() to ensure
its value can be properly consumed by td_verror()

Signed-off-by: Hyunwoo Park <dshw.park@samsung.com>
 
### **iolog: fix Null pointer dereferences (FORWARD_NULL)**

CID 494150: Null pointer dereferences (FORWARD_NULL) @ iolog.c:148 in ipo_special()
This patch removes the possibility of null pointer dereferencing(io_u->file)
throughout the call stack of get_io_u() → read_iolog_get() → dp_fill_dspec_data()

Signed-off-by: Hyunwoo Park <dshw.park@samsung.com>

### DIFF
--- a/iolog.c
+++ b/iolog.c
@@ -145,7 +145,7 @@ static int ipo_special(struct thread_data *td, struct io_piece *ipo)
 				int dp_init_ret = dp_init(td);
 
 				if (dp_init_ret != 0) {
-					td_verror(td, dp_init_ret, "dp_init");
+					td_verror(td, abs(dp_init_ret), "dp_init");
 					return -1;
 				}
 			}
--- a/iolog.c
+++ b/iolog.c
@@ -236,14 +236,14 @@ int read_iolog_get(struct thread_data *td, struct io_u *io_u)
 						io_u->buflen, io_u->file->file_name);
 			if (ipo->delay)
 				iolog_delay(td, ipo->delay);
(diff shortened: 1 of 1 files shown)


vincentkfu:
Please change the first lines of the commit messages to describe the actual issue instead of just writing that the issue was identified by coverity.

vincentkfu:
For the negative `errno` issue I would much rather see code that does something like what `zbd.c:zbd_init_zone_info()` does:

```
        ret = zbd_create_zone_info(td, file);
        if (ret < 0)
                td_verror(td, -ret, "zbd_create_zone_info() failed");
```

Just negate the value sent to `td_verror()`. This involves fewer code changes than your patch.

I do not know the full history of returning negative `errno` values but this is a common pattern: https://stackoverflow.com/questions/36845866/history-of-using-negative-errno-values-in-gnu

parkvibes:
As `dp_init()` can return a positive value, I made `dp_init_ret` an absolute value using `abs(dp_init_ret)`

PR #1768, opened 2024-06-04
FDP plids option: accept ranges and clarify documentation

vincentkfu:
Feedback is welcome.

### DIFF
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -2538,12 +2538,17 @@ with the caveat that when used on the command line, they must come after the
 
 .. option:: plids=str, fdp_pli=str : [io_uring_cmd] [xnvme]
 
-        Select which Placement IDs (streams) or Placement ID Indices (FDP) this
-        job is allowed to use for writes. For FDP by default, the job will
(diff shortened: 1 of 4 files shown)


vincentkfu:
The last patch increases the maximum number of placement IDs a job can specify from 16 to 128. This increases the size of `struct thread_options` from 3288 bytes to 3480 bytes. It's not great to add 192 more bytes to this structure for an option that will only see occasional use but I can't think of a better way to do this.

vincentkfu:
Merged:

https://git.kernel.dk/cgit/fio/commit/?id=4518667d4242ddbd804302d0f1a758ede1208edf
https://git.kernel.dk/cgit/fio/commit/?id=95db41f1e106742cc515c9b8d599fd49f792fc2a
https://git.kernel.dk/cgit/fio/commit/?id=612da74406e8f62fd6e129a8391f59fd3ffb7ee3
https://git.kernel.dk/cgit/fio/commit/?id=695ab5c21d44a004b8f90bcc38b9331bc2d8feb7

PR #1775, opened 2024-06-13
For nfs engine, treat randwrite the same as write when opening file

panxiao2014:
When use nfs engine, only write mode can work if file doesn't exist before test starts. This code change just made randwrite behaves the same as write. So both write and randwrite could work in scenario where file doesn't exist before test start.

Note: other than write and randwrite mode, nfs engine would report error when test start. This is because fio would not layout the file before test start. This also indicate that currently nfs engine requires user to make sure files are ready before test when choose mode such as read, randread and readwrite.

This requires more code changes. So now just add one line code change to let randwrite can work.

### DIFF
--- a/engines/nfs.c
+++ b/engines/nfs.c
@@ -280,7 +280,7 @@ static int fio_libnfs_open(struct thread_data *td, struct fio_file *f)
 	nfs_data = calloc(1, sizeof(struct nfs_data));
 	nfs_data->options = options;
 
-	if (td->o.td_ddir == TD_DDIR_WRITE)
+	if (td_write(td))
 		flags |= O_CREAT | O_RDWR;
 	else
 		flags |= O_RDWR;


panxiao2014:
Hi @vincentkfu , I saw there is an error whn checking build on cygwin platform: multiple definition of `strtoll' found. I did not touch this part of code. Could you please check if there is additional task that I should do before submitting PR?

Thanks.

vincentkfu:
That failure is due to changes in an external library and unrelated to your changes.

panxiao2014:
Thanks Vincent. We are using fio for our NFS server test. The change here could make our test script automation works easier. Please help to check if this code change could be promoted.

engines/nfs.c
```diff
@@ -280,7 +280,7 @@ static int fio_libnfs_open(struct thread_data *td, struct fio_file *f)
 	nfs_data = calloc(1, sizeof(struct nfs_data));
 	nfs_data->options = options;
 
-	if (td->o.td_ddir == TD_DDIR_WRITE)
+	if (td->o.td_ddir == TD_DDIR_WRITE || td->o.td_ddir == TD_DDIR_RANDWRITE)
```
axboe, line 283: This should just use:

```
if (td_write(td))
```

it's just a basic bug in the NFS engine, that it checks specifically for == TD_DDIR_WRITE rather than & TD_DDIR_WRITE.

panxiao2014:
Thanks Jens. I have updated the code change and run some test to verify it. Please review.

axboe:
Looks fine, but please squash that into one commit. It's a bit pointless to have a suboptimal fix, and then a fixup on top.

PR #1777, opened 2024-06-22
Flush enhancements and fixes

minwooim:
This patchset introduces enhancements of fsync/fdatasync operations and
fixes.  The first patch added support for NVMe FLUSH commands in
io_uring_cmd ioengine.  The second one fixes issuing fsync/fdatasync
even if the previous command is READ.  As documentation says, fsync
should sync the file after every N WRITE commands issued.  The third one
fixes not issuing fsync/fdatasync in trimwrite worklload.  The last one
is a trivial patch to show the numbe of fsync/fdatasync in the stat
report.

Minwoo Im (4):
  io_uring: Add support FLUSH command
  io_u: ensure fsync only after write(s)
  io_u: Support fsync for --rw=trimwrite
  stat: Show number of DDIR_SYNC samples

 engines/io_uring.c | 1 +
 engines/nvme.c     | 5 +++++
 engines/nvme.h     | 1 +
 fio.h              | 9 +++++++++
 io_u.c             | 4 ++--
 ioengines.c        | 1 +
 stat.c             | 3 ++-
 7 files changed, 21 insertions(+), 3 deletions(-)


engines/io_uring.c
```diff
@@ -39,6 +39,7 @@ enum uring_cmd_write_mode {
 	FIO_URING_CMD_WMODE_UNCOR,
 	FIO_URING_CMD_WMODE_ZEROES,
 	FIO_URING_CMD_WMODE_VERIFY,
+	FIO_URING_CMD_WMODE_FLUSH,
```
vincentkfu, line 42: Why is this added?

minwooim: Uhh... my bad.  It's something remaining from the previous approach. Will remove it!

minwooim:
> For the second patch, it looks like what's happening is that fio issues a write, followed by a sync, then a read, then (since `last_was_sync` is false) another sync.

If `last_ddir_issued` are going to like:
```
last_ddir_issued = DDIR_WRITE
last_ddir_issued = DDIR_SYNC
last_ddir_issued = DDIR_READ
```

then `if (should_fsync(td) && td->last_ddir_issued != DDIR_READ)` should be false and another sync will not happen.

> 
> For the second patch, why don't you also remove `last_was_sync` when adding `last_ddir_issued`?
> 
> In `should_fsync` replace the check of `last_was_sync` by checking if `last_ddir_issued` is false.

If you meant `last_ddir_issued` is not `DDIR_SYNC`, got your point.

> 
> Also please rename `last_ddir` to `last_ddir_completed`.

Okay.

> 
> Can you create a test case for this?

Alright, I'll work on the mixed workload and create test cases using the issued rwts section of the report.

fio.h
```diff
@@ -629,7 +630,7 @@ static inline bool multi_range_trim(struct thread_data *td, struct io_u *io_u)
 
 static inline bool should_fsync(struct thread_data *td)
 {
-	if (td->last_was_sync)
+	if (td->last_ddir_issued == DDIR_SYNC)
```
vincentkfu, line 633: With ` last_was_sync` we had:

```
td->last_was_sync = ddir_sync(io_u->ddir);
```

So this should use `ddir_sync()` instead of checking for equality with `DDIR_SYNC` because we actually have three different sync DDIRs.

engines/io_uring.c
vincentkfu: Still needs to be removed

vincentkfu:
For the `io_u: ensure fsync only after write(s)` patch can you split it into two patches? One with the rename `last_ddir` to `last_ddir_completed`, and one with the bug fix.

Finally, the test cases look good but can you just add them to `nvmept.py`? It doesn't make sense to have a new script for these four test cases.

engines/io_uring.c
minwooim: Will remove it!

fio.h
minwooim: Agreed! will fix it. Thanks.

minwooim:
> For the `io_u: ensure fsync only after write(s)` patch can you split it into two patches? One with the rename `last_ddir` to `last_ddir_completed`, and one with the bug fix.

Sure, I will.

> Finally, the test cases look good but can you just add them to `nvmept.py`? It doesn't make sense to have a new script for these four test cases.

Okay.

vincentkfu:
Don'[t forget  to remove the superfluous item from `uring_cmd_write_mode`.

minwooim:
@vincentkfu ,

Applied review comments, Thanks for your time.

ioengines.c
```diff
@@ -466,6 +462,7 @@ enum fio_q_status td_io_queue(struct thread_data *td, struct io_u *io_u)
 		}
 	}
 
+	td->last_ddir_issued = ddir;
```
vincentkfu, line 465: I did not notice this in the previous version of your patch but if `ret == FIO_Q_BUSY` then the I/O was not actually issued. Assignment of `last_ddir_issued` should go in the two places where `last_was_sync` was set.

libfio.c
```diff
@@ -101,7 +101,6 @@ static void reset_io_counters(struct thread_data *td, int all)
 
 	td->zone_bytes = 0;
 
-	td->last_was_sync = false;
```
vincentkfu, line 104: I think we never should have set `last_was_sync` to false here because when ramp time is over we want to continue to sync at the appropriate point, so there is no need to modify the value for `last_ddir_issued`. Agree?

ioengines.c
minwooim: Ah, indeed.  Let me fix this up!

libfio.c
minwooim: If you meant we don't need to modify the value of `last_ddir_issued` in this place, yes, I agree.

vincentkfu:
Applied, thanks.

PR #1779, opened 2024-06-28
fio: fix formats under MIPS64/PPC

neheb:
__SANE_USERSPACE_TYPES__ needs to be defined to get consistent formats on all platforms.

### DIFF
--- a/engines/nvme.h
+++ b/engines/nvme.h
@@ -7,6 +7,8 @@
 #ifndef FIO_NVME_H
 #define FIO_NVME_H
 
+#define __SANE_USERSPACE_TYPES__
+
 #include <linux/nvme_ioctl.h>
 #include "../fio.h"
 


axboe:
Please ensure the commit message is properly formatted. Break lines at 72-74 chars, and include a signed-off-by line. Just see other commits in the repo if in doubt.

engines/nvme.h
```diff
@@ -7,6 +7,8 @@
 #ifndef FIO_NVME_H
 #define FIO_NVME_H
 
+#define __SANE_USERSPACE_TYPES__
```
axboe, line 10: This should probably go in the Makefile for mips64, not in this specific location?

sitsofe:
Hello @neheb:

Would you be able to make the changes that Jens suggested?

neheb:
not sure how to go about that.

axboe:
Just do ```git --amend``` while on your branch with that commit, and edit the commit message. Then ```git push --force``` afterwards to overwrite the commit.

neheb:
It's not that. I don't see where platform specific defines are located or should be.

axboe:
There are none, but it'd be easy to add in the top level Makefile for when fio is being compiled with mips64 as the target. You could also probably just as well add it in ```arch/arch-mips.h``` as anything would end up including that. Probably the better choice.

Edit: and the ppc counterparts too.

neheb:
Did that.

axboe:
Still missing a signed-off-by, see any other commit in the tree.

axboe:
Is this fine to set for 32-bit mips and ppc too? Presumably yes and it will only impact the 64-bit variants? Since you're going to be amending the commit message anyway, maybe add a bit more of an explanation of why this is needed.

axboe:
And finally, does this need an ifndef around it? In case of nested included.

neheb:
added ifndef. and yes it's no-op on 32-bit.

axboe:
I'll mention this one again: Still missing a signed-off-by, see any other commit in the tree. When that's done, this should be good to go. Also see the other previous comment https://github.com/axboe/fio/pull/1779#issuecomment-3217693807

neheb:
done.

PR #1785, opened 2024-07-10
Support platform QNX OS

huangweiliang:
Compile with SDP7.1:
./configure --cc=aarch64-unknown-nto-qnx7.1.0-gcc --disable-shm make

Function is verfied by UFS device read and write test.

Signed-off-by: Huang weiliang <weller.huang@us.bosch.com>


huangweiliang:
need recreate it

PR #1786, opened 2024-07-10
Support platform QNX OS

huangweiliang:
Compile with SDP7.1:
./configure --cc=aarch64-unknown-nto-qnx7.1.0-gcc --disable-shm make

Function is verfied by UFS device read and write test.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -118,12 +118,20 @@ static void set_sig_handlers(void)
 
 	memset(&act, 0, sizeof(act));
 	act.sa_handler = sig_int;
+	#if defined(__QNX__)
+	act.sa_flags = SA_NOCLDSTOP;
+	#else
 	act.sa_flags = SA_RESTART;
+	#endif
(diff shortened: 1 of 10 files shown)


axboe:
Thanks for doing this work. Some notes:

1) Please look at commit formatting in the repo in general, most notably you need a proper signed-off-by line in there.

2) All the SA_NOCLDSTOP ifdefs are pretty ugly and also ill formatted. ifdefs go at the start of the line. Please follow coding style. I think this would be much cleaner if you had a generic OS_SA_RESTART, and put a:

```
#ifndef OS_SA_RESTART
#define OS_SA_RESTART SA_RESTART
#endif
```
in os/os.h and just have os-qnx.h define it to something else.

3) Ditto the shm/mman changes. Again this mostly looks like it was hacked into existence rather than some thought put into it. Not very maintainable.

huangweiliang:
I will close this pull request and create new one.
Thanks for your comment.

PR #1789, opened 2024-07-15
Suppress unused variable warning in helper_thread.c

r-barnes:
Allows `-Wunused-but-set-variable` to pass in opt mode. Adds what appears to be a missing assert.

2. Empty second line
4. Empty line

Reminders:

### DIFF
--- a/helper_thread.c
+++ b/helper_thread.c
@@ -110,7 +110,7 @@ static void block_signals(void)
 	sigset_t sigmask;
 
 	int ret;
-	(void)ret; // Suppress unused variable warning
+	(void)ret; /* Suppress unused variable warning */
 
 	ret = pthread_sigmask(SIG_UNBLOCK, NULL, &sigmask);
 	assert(ret == 0);


helper_thread.c
```diff
@@ -110,10 +110,12 @@ static void block_signals(void)
 	sigset_t sigmask;
 
 	int ret;
+	(void)ret; // Suppress unused variable warning
```
vincentkfu, line 113: Is this line necessary with the assert added below?

vincentkfu:
I've built with `./configure --extra-cflags=Wall` and confirmed that `CONFIG_PTHREAD_SIGMASK` is defined but cannot trigger the warning that you report.

Our CI builds with `-Werror` and this warning does not appear.

What did you do to trigger this warning?

vincentkfu:
Ok. The convention in our codebase is to use `/* ... */` for comments. Please make this change to your patch.

vincentkfu:
These are small changes but I can't let them go in this way.

How about this:

Patch 1: add missing assert only
Patch 2: add `(void) ret;` to suppress unused variable warning

Please make sure the commit messages for each patch accurately describe the purpose of the change.

sitsofe:
@r-barnes would you be able to make the changes that Vincent suggested? Thanks!

sitsofe:
Closing this as we never heard back from @r-barnes.

@r-barnes if you ever have the time feel free to pick this one back up.

PR #1790, opened 2024-07-15
FDP fixes

ankit-sam:
The current way of fdp ruhs initialization limited it to a maximum of 128 ruhs descriptors. Fio allows a total of 128 placement id 
indices at a time, but for devices supporting more the 128 ruhs, users should have the ability to pass any 128 placement id index / indices (0 to max number of ruhs - 1). This series fixes that.

### DIFF
--- a/dataplacement.c
+++ b/dataplacement.c
@@ -43,11 +43,8 @@ static int fdp_ruh_info(struct thread_data *td, struct fio_file *f,
 static int init_ruh_info(struct thread_data *td, struct fio_file *f)
 {
 	struct fio_ruhs_info *ruhs, *tmp;
+	uint32_t nr_ruhs;
 	int i, ret;
-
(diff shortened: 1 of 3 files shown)


vincentkfu:
I applied the patches instead of merging the PR because I slightly edited the commit message for this first patch.

https://git.kernel.dk/cgit/fio/commit/?id=56d12245953b4ae147784db2e77b1609197df4a0
https://git.kernel.dk/cgit/fio/commit/?id=70ae781d13f2daa6fe905bb42498c264e4b5aaeb
https://git.kernel.dk/cgit/fio/commit/?id=0e3c0070ddf75779c2ef840f6c80f49da9690bdf

vincentkfu:
I will take a closer look but I think this change will allow us to get rid of `FDP_MAX_RUHS`.

PR #1791, opened 2024-07-17
Small fdp fixes

ankit-sam:
This series has 3 patches
1. Removing FDP_MAX_RUHS, as now we are fetching all the ruhs reported by the device, and then parsing the ones requested as per the placement ID indices.
2. log_info to log_err for all the error messages.
3. Incrementing FIO_IOOPS_VERSION, for any external ioengine which may be using fdp_fetch_ruhs.


vincentkfu:
Looks good. Thanks.

PR #1799, opened 2024-08-06
Fix fsync absence when `fsync=N` parameter specified 

RomanSofyin:
When `fsync=N` parameter specified and the final IO block issued happens to be Nth the fsync should follow it is missed. This update checks if it needs to loop again in `do_io` to issue that final fsync when all the IO work is done.

Signed-off-by: Roman Sofin roma.sofin@gmail.com


backend.c
```diff
@@ -1000,6 +1000,7 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
 
 	while ((td->o.read_iolog_file && !flist_empty(&td->io_log_list)) ||
 		(!flist_empty(&td->trim_list)) || !io_issue_bytes_exceeded(td) ||
+		time_to_ddir_sync(td) ||
```
axboe, line 1003: This won't compile, there's no definition for this.

backend.c
```diff
@@ -1032,8 +1033,8 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
 		 */
 		if (bytes_issued >= total_bytes &&
 		    !td->o.read_iolog_file &&
-		    (!td->o.time_based ||
-		     (td->o.time_based && td->o.verify != VERIFY_NONE)))
+		    ((!td->o.time_based && !time_to_ddir_sync(td)) ||
```
axboe, line 1036: Ditto

fio.h
```diff
@@ -628,7 +628,7 @@ static inline bool multi_range_trim(struct thread_data *td, struct io_u *io_u)
 	return false;
 }
 
-static inline bool should_fsync(struct thread_data *td)
+static inline bool should_fsync(const struct thread_data *td)
```
axboe, line 631: Seems like a random change that's not related to the fix.

io_u.c
```diff
@@ -742,6 +742,30 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
 	return ddir;
 }
 
+static inline enum fio_ddir get_sync_ddir(const struct thread_data *td)
+{
+	if (should_fsync(td))
```
axboe, line 747: { goes with this line, please observe the surrounding style and match it.

io_u.c
```diff
@@ -742,6 +742,30 @@ static enum fio_ddir rate_ddir(struct thread_data *td, enum fio_ddir ddir)
...
+	{
+		if (td->o.fsync_blocks && td->io_issues[DDIR_WRITE] &&
+			!(td->io_issues[DDIR_WRITE] % td->o.fsync_blocks))
```
axboe, line 750: Should align with above, again observe surrounding style and match it. Same for the rest of this function.

io_u.c
```diff
@@ -755,19 +779,8 @@ static enum fio_ddir get_rw_ddir(struct thread_data *td)
 	 * See if it's time to fsync/fdatasync/sync_file_range first,
 	 * and if not then move on to check regular I/Os.
 	 */
-	if (should_fsync(td) && td->last_ddir_issued == DDIR_WRITE) {
-		if (td->o.fsync_blocks && td->io_issues[DDIR_WRITE] &&
```
axboe, line 759: This literally shows you the style, which is what you moved to a helper.

io_u.c
```diff
@@ -755,19 +779,8 @@ static enum fio_ddir get_rw_ddir(struct thread_data *td)
...
-
-		if (td->o.fdatasync_blocks && td->io_issues[DDIR_WRITE] &&
-		    !(td->io_issues[DDIR_WRITE] % td->o.fdatasync_blocks))
-			return DDIR_DATASYNC;
-
-		if (td->sync_file_range_nr && td->io_issues[DDIR_WRITE] &&
-		    !(td->io_issues[DDIR_WRITE] % td->sync_file_range_nr))
-			return DDIR_SYNC_FILE_RANGE;
-	}
+	if ((ddir = get_sync_ddir(td)) != DDIR_INVAL)
```
axboe, line 782: Assignment and check should be separate lines.

```
ddir = get_sync_ddir(td);
if (ddir != DDIR_INVAL)
    ....
```

RomanSofyin (commented):
Addressed the issues.

fio.h
axboe: It's random in that it doesn't seem related to your changes, other than that you made get_sync_ddir() take a constant 'td' and then this one needs it too. But it's not related to the fix.

Making 'td' const in more spots is a worthwhile endeavor, but should be a separate thing from a fix.

axboe:
> Addressed the issues.

Please squash the two commits into one, we should not have fix-up commits in the history. Once you do, I'll take a look again.

RomanSofyin:
Hi @axboe, can you look at this again?

axboe:
You still have the should_fsync() change in there. If you want to have it take a const td, then please do a prep patch with just that. Like I said initially, it's not related to your fix at all outside of you adding a new caller that needs it done. That's by definition a prep patch, should not be included with a fix.

RomanSofyin:
I created PR #1805 which I guess we merge first, then I update the current one in case there will be a conflict.

axboe:
Since the change depends on the other one, please just have both commits in a single pr.

RomanSofyin:
Should be fine now. If so, I'll close #1805.

RomanSofyin:
@axboe does this look fine?

sitsofe (approved):
@vincentkfu, @axboe This LGTM.

RomanSofyin:
I see [build-containers (oracle, oraclelinux, 9, x86_64)](https://github.com/axboe/fio/actions/runs/17201959961/job/48799117181?pr=1799#logs) has failed, is this smth expected?

sitsofe:
@RomanSofyin I kicked off a rebuild of that job and it passed so it seems like it was a transient error.

io_u.h
```diff
@@ -169,6 +169,8 @@ bool queue_full(const struct thread_data *);
 int do_io_u_sync(const struct thread_data *, struct io_u *);
 int do_io_u_trim(struct thread_data *, struct io_u *);
 
+bool ddir_is_sync(const struct thread_data *);
```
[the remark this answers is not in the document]
RomanSofyin: As you suggested I could call `get_sync_ddir()` only if `should_fsync()` is true:
```c
bool ddir_is_sync(const struct thread_data *td)
{
	return should_fsync(td) && get_sync_ddir(td) != DDIR_INVAL;
}
```

But I don't quite get how you want me to make `ddir_is_sync` static. Since `ddir_is_sync` gets called from `backend.c` it needs to be either in `backend.c` or in `io_u.h` to have it static. But `ddir_is_sync` calls `should_fsync` and `get_sync_ddir` which are static as well so can't be referred from anywhere but `io_u.c`.

Can you please clarify this?

backend.c
```diff
@@ -1032,8 +1033,8 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
 		 */
 		if (bytes_issued >= total_bytes &&
 		    !td->o.read_iolog_file &&
-		    (!td->o.time_based ||
-		     (td->o.time_based && td->o.verify != VERIFY_NONE)))
+		    ((!td->o.time_based && !ddir_is_sync(td)) ||
```
sitsofe, line 1036: If we're a time based job won't we have already exited due to the `runtime_exceeded` check further up? Further, wouldn't we want to avoid checking that we need to add a trailing fsync in the case that we are a non-time based job but still had a non-zero runtime limit (but again wouldn't we have already exited by now)?

io_u.h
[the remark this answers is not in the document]
sitsofe: @RomanSofyin I've chatted to @axboe and it should be sufficient to guard the _call sites_ to `ddir_is_sync()` by a `should_fsync()` pre-check.

backend.c
```diff
@@ -1000,6 +1000,7 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
 
 	while ((td->o.read_iolog_file && !flist_empty(&td->io_log_list)) ||
 		(!flist_empty(&td->trim_list)) || !io_issue_bytes_exceeded(td) ||
+		ddir_is_sync(td) ||
```
sitsofe, line 1003: Perhaps move this to be the last case as it's hopefully less common.

backend.c
RomanSofyin: `runtime_exceeded` condition hits if `timeout` (aka `runtime`) limit of a job is exceeded, no matter if it's a time based job or not. So if a job gets this line of code, it's allowed to do all its stuff.

As I see this, we can't avoid checking that because for a non-time based (i.e. "size based") job we terminate the loop when the number of bytes requested is exceeded even if there is an fsync call to issue left.

RomanSofyin:
@sitsofe a CI job seems to hang again, can you rerun it please? [link](https://github.com/axboe/fio/actions/runs/17545119959/job/49913252706 )

Also would you mind looking at the changes I delivered in 7b7f18c and tell if it's sufficient?

sitsofe:
@RomanSofyin 

> Also would you mind looking at the changes I delivered in https://github.com/axboe/fio/commit/7b7f18c1e33c7223788bfd42490d673150b30079 and tell if it's sufficient?

I'm currently on holiday but I'll take a proper look when they're over.

PR #1804, opened 2024-08-12
io_uring: Add .errdetails to parse CQ status

minwooim:
Background
 - fio normally prints out the strerr and errno value when facing errors.  In case of io_uring_cmd ioengine with --cmd_type=nvme, io_u->error represents the CQ entry status code type and status code which should be parsed as a NVMe error value rather than errno.

In io_u error failure condition, it prints out parsed CQ entry error status values with SCT(Status Code Type) and SC(Status Code).  The print will be like the following example:

  fio: io_uring_cmd: /dev/ng0n1: cq entry status (sct=0x00; sc=0x04)

If --cmd_type!=nvme, it prints out generic status code like below:

  fio: io_uring_cmd: /dev/<devnode>: status=0x4

### DIFF
--- a/engines/io_uring.c
+++ b/engines/io_uring.c
@@ -554,6 +554,39 @@ static struct io_u *fio_ioring_cmd_event(struct thread_data *td, int event)
 	return io_u;
 }
 
+static char *fio_ioring_cmd_errdetails(struct thread_data *td,
+				       struct io_u *io_u)
+{
+	struct ioring_options *o = td->eo;
+	unsigned int sct = (io_u->error >> 8) & 0x7;
+	unsigned int sc = io_u->error & 0xff;
+#define MAXERRDETAIL 1024
+#define MAXMSGCHUNK 128
+	char *msg, msgchunk[MAXMSGCHUNK];
+
+	msg = calloc(1, MAXERRDETAIL);
+	strcpy(msg, "io_uring_cmd: ");
+
+	snprintf(msgchunk, MAXMSGCHUNK, "%s: ", io_u->file->file_name);
+	strlcat(msg, msgchunk, MAXERRDETAIL);
+
+	if (o->cmd_type == FIO_URING_CMD_NVME) {
+		strlcat(msg, "cq entry status (", MAXERRDETAIL);
+
+		snprintf(msgchunk, MAXMSGCHUNK, "sct=0x%02x; ", sct);
+		strlcat(msg, msgchunk, MAXERRDETAIL);
+
+		snprintf(msgchunk, MAXMSGCHUNK, "sc=0x%02x)", sc);
+		strlcat(msg, msgchunk, MAXERRDETAIL);
+	} else {
+		/* Print status code in generic */
+		snprintf(msgchunk, MAXMSGCHUNK, "status=0x%x", io_u->error);
+		strlcat(msg, msgchunk, MAXERRDETAIL);
+	}
+
+	return msg;
+}
+
 static int fio_ioring_cqring_reap(struct thread_data *td, unsigned int events,
 				   unsigned int max)
 {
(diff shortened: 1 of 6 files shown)


engines/io_uring.c
```diff
@@ -554,6 +554,46 @@ static struct io_u *fio_ioring_cmd_event(struct thread_data *td, int event)
...
+	strcpy(msg, "io_uring_cmd: ");
+
+	snprintf(msgchunk, MAXMSGCHUNK, "%s: ", io_u->file->file_name);
+	strlcat(msg, msgchunk, MAXERRDETAIL);
+
+	/*
+	 * Check the filename to figure out cmd_type since we don't have
+	 * thread_data here.
+	 */
+	ret = sscanf(io_u->file->file_name, "/dev/ng%dn%d", &ctrlid, &nsid);
```
vincentkfu, line 577: Can you just use `strstr` to check if the filename begins with `/dev/ng`? Then there is no need to have the `ctrlid` and `nsid` variables which are ultimately unused.

engines/io_uring.c
```diff
@@ -554,6 +554,46 @@ static struct io_u *fio_ioring_cmd_event(struct thread_data *td, int event)
...
+	if (ret == 2) {
+		strlcat(msg, "cq entry status (", MAXERRDETAIL);
+
+		snprintf(msgchunk, MAXMSGCHUNK, "sct=0x%02x; ", sct);
+		strlcat(msg, msgchunk, MAXERRDETAIL);
+
+		snprintf(msgchunk, MAXMSGCHUNK, "sc=0x%02x", sc);
+		strlcat(msg, msgchunk, MAXERRDETAIL);
+
+		strlcat(msg, ")", MAXERRDETAIL);
```
vincentkfu, line 587: Can you remove this `strlcat` and just add the `)` to the `snprtinf` call above this one?

vincentkfu:
This looks fine. Just two small changes requested.

engines/io_uring.c
minwooim: Okay!

engines/io_uring.c
minwooim: Okay!

minwooim:
> This looks fine. Just two small changes requested.

Applied all the changes, Thanks for the review.

engines/io_uring.c
```diff
@@ -554,6 +554,41 @@ static struct io_u *fio_ioring_cmd_event(struct thread_data *td, int event)
...
+	strcpy(msg, "io_uring_cmd: ");
+
+	snprintf(msgchunk, MAXMSGCHUNK, "%s: ", io_u->file->file_name);
+	strlcat(msg, msgchunk, MAXERRDETAIL);
+
+	/*
+	 * Check the filename to figure out cmd_type since we don't have
+	 * thread_data here.
+	 */
+	if (strstr(io_u->file->file_name, "/dev/ng")) {
```
axboe, line 575: This is pretty terrible, to be honest. You cannot impute type of device based on the name of the device special node. It's perfectly valid to have different names for these. This will need a rethink to be workable.

vincentkfu: I can think of two alternatives:

1. Change `.errdetails` to also take a `struct thread_data` argument. `td` gives you access to the ioengine options which you can use to find the `cmd_type`. You would need to bump `FIO_IOOPS_VERSION`.
2. Add a pointer to `td` to `struct io_u`.

The first choice seems like the better approach to me.

axboe: First choice definitely the better one. And yes, it's not a problem to change the prototype of the arguments. Just remember to change ```FIO_IOOPS_VERSION``` as well.

minwooim: Updated! Thanks,

vincentkfu:
Please update the commit message for the patch adding `errdetials` to the `io_uring_cmd` ioengine. It no longer uses the filename to determine `cmd_type`.

Also you might as well squash the first two patches that change the function prototype and bump the version since they need to go together.

minwooim:
> Please update the commit message for the patch adding `errdetials` to the `io_uring_cmd` ioengine. It no longer uses the filename to determine `cmd_type`.
> 
> Also you might as well squash the first two patches that change the function prototype and bump the version since they need to go together.

Updated, Thanks for your review.

axboe:
This looks good to me now.

PR #1805, opened 2024-08-12
Make `should_fsync()` taking a constant.

RomanSofyin:
Make `should_fsync()` safer by using a pointer to a constant thread_data structure instance.
This update is mostly to address the need of `get_sync_ddir()` function (a caller of `should_fsync()`) to operate with a `thread_data` instance as a constant. See PR #1799 for details.

Signed-off-by: Roman Sofin roma.sofin@gmail.com


sitsofe:
We can close this as it will be done as part of #1799 .

PR #1809, opened 2024-08-23
[RFC] Non-repeating file_service_type: roundrobin1 and random1

koct9i:
These servicing modes does not repeat file until completion previous operation.

( i.e. when nrfiles > iodepth fio should chose files which has no in-flight I/O )

Streaming IO-engines might not handle several requests to one "file" at once.
Or such workload might not match any real scenarios.

Signed-off-by: Konstantin Khlebnikov <koct9i@gmail.com>

Reminders:

 // Ok, will do later if this change is ok

### DIFF
--- a/file.h
+++ b/file.h
@@ -35,6 +35,7 @@ enum fio_file_flags {
 	FIO_FILE_axmap		= 1 << 7,	/* uses axmap */
 	FIO_FILE_lfsr		= 1 << 8,	/* lfsr is used */
 	FIO_FILE_smalloc	= 1 << 9,	/* smalloc file/file_name */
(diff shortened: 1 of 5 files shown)


vincentkfu:
I think the general idea is fine. I will take a closer look later but here are some general comments:

- Documentation needs to be updated
- Please choose better names than `random1` and `roundrobin1`
- With these options it does not make sense to have `iodepth > nrfiles`. Perhaps fio should emit an error message in this situation. This could be added in `fixup_options()`.
- The code would be more readable to have a helper to check `service_type == RANDOM || service_type == RANDOM1`

Issue #1801, opened 2024-08-09
With '--build-static' enabled, rdma-core is failed to probe by configure


HuangShumin:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
fio-3.37 cannot be built with static rdma engine, even with libibverbs-dev/librdmacm-dev static libraries installed.

**Environment**: Debian 12.5

**fio version**: fio-3.37

**Reproduction steps**
```
apt install build-essential devscripts dpkg-dev equivs fakeroot lintian quilt
mk-build-deps -i -r fio
apt install libnl-3-dev libnl-route-3-dev libibverbs-dev librdmacm-dev
tar xzvf fio-3.37.tar.gz && cd fio-fio-3.37
./configure --build-static && make -j $(nproc) && ./fio --enghelp | grep rdma
# shoud output rdma engine here
```

**Troubleshooting**
After I modify configure script contents as below:
```
##########################################
# See if we need to build a static build
if test "$build_static" = "yes" ; then
  CFLAGS="$CFLAGS -ffunction-sections -fdata-sections"
  LDFLAGS="$LDFLAGS -lnl-3 -lnl-route-3 -libverbs -lrdmacm -lrpma -static -Wl,--gc-sections"
else
  build_static="no"
fi
print_config "Static build" "$build_static"
```
configure is able to probe static library of rdma, but it's reported with undefined errors by gcc ld as
```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/librdmacm.a(cma.c.o): in function `check_abi_version_nl_cb':
(.text+0x4fb): undefined reference to `nlmsg_hdr'
/usr/bin/ld: (.text+0x514): undefined reference to `nlmsg_parse'
/usr/bin/ld: (.text+0x54c): undefined reference to `nla_get_u64'
/usr/bin/ld: (.text+0x562): undefined reference to `nla_get_string'
/usr/bin/ld: (.text+0x594): undefined reference to `nla_get_u64'
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/librdmacm.a(cma.c.o): in function `ucma_query_route':
(.text+0x1255): undefined reference to `ibv_copy_path_rec_from_kern'
...
```

[...]

PR #1818, opened 2024-09-24
Fix configure with --build-static which enables static RDMA in #1801

HuangShumin:
Fix configure with --build-static option which enables static RDMA in Issue #1801

This pull request to probe static libibverbs/librdmacm library. It's neccessary to enable static RDMA engine build with netlink libnl-3 and libnl-route-3.

Signed-off-by: Huang Shumin longway68@qq.com

### DIFF
--- a/configure
+++ b/configure
@@ -1005,8 +1005,9 @@ int main(int argc, char **argv)
   return pd != NULL;
 }
 EOF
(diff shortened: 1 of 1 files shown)


vincentkfu:
Merged

https://git.kernel.dk/cgit/fio/commit/?id=f37bdd4c06d8528e4064713d004626d5d1276d08

Issue #1820, opened 2024-09-26
Add an option to open with O_EXCL


keithr-mext:
On linux, it can be difficult to determine if a drive is in use by other parts of the system, since it can be mounted in a non-default namespace or directly opened by an application. ([This page](https://unix.stackexchange.com/questions/111779/how-to-find-out-easily-whether-a-block-device-or-a-part-of-it-is-mounted-someh) describes some of the many ways a drive can be in use that can be difficult to find.) However, opening a block device with the` O_EXCL` flag causes the kernel to return `EBUSY` if the drive is in use.

From open(2):

In general, the behavior of O_EXCL is undefined if it is used without O_CREAT.  There is one exception: on Linux 2.6 and later, O_EXCL can be used without O_CREAT if pathname refers to a block device.  If the block device is in use by the system (e.g., mounted), open() fails with the error EBUSY.

There should be an option to tell fio to use this flag, similar to the `direct` option.

axboe:
Fio already checks if the device is mounted, and will fail a write workload unless ```allow_mounted_write``` has been set to 1. It defaults to 0. ```O_EXCL``` doesn't really bring much otherwise, as it will not prevent an exclusive open if someone else already has the device opened read+write, EXCEPT if that was also done with ```O_EXCL```. And vice versa, an ```O_EXCL``` open will also not prevent another read/write open.

keithr-mext:
I just clobbered a device by running fio on the base device (/dev/nvme1n1) while some partitions were mounted (the system was rebooted and linux shuffled the block device numbers; before the reboot nvme0 was the root disk and nvme1 was my test disk; I didn't notice that they had been swapped by the reboot.) Fio's mounted device check didn't catch that, because I gave it /dev/nvme1n1 rather than /dev/nvme1n1p1 as the filename; the mounted check would have worked in the latter case. Opening with O_EXCL would have failed with EBUSY because of the mounted partitions.

axboe:
That's a good point. I guess we could just augment the writeable open check with setting O_EXCL as well. That should provide more coverage, would've caught your case.

keithr-mext:
I have a fork in which I'm working on fixing this; what I've got currently is a `exclusive-open` option, but if it would be better to handle it a different way, such as making it always use O_EXCL for write workloads, let me know.

axboe:
[...]

Issue #1382, opened 2022-04-06
Make mount checks a bit smarter if possible


facundofc:
The `allow_mounted_write` flag (which defaults to 0) provides a safety net for those accidentally running a write job to a mounted device. Essentially, if the job specifies a `filename` which is a mounted device and the job is a writing one, it won't be run.

What it won't check is if the `filename` is the parent device of a mounted partition. So if you have say `/dev/sda` with two (mounted) partitions, `/dev/sda1` and `/dev/sda2`, a job including

```
filename=/dev/sda
rw=write
```

will just go ahead and run, much likely corrupting `sda`'s partitions.

I believe improving the check logic in order to include cases like that would be useful, since I can't think of any case in which you'd actually want to run such a test.

zwhitchcox:
I run tests on `/dev/sda` to test disk i/o on flash storage. Not sure if that is the intended use case though.

axboe:
> I run tests on /dev/sda to test disk i/o on flash storage. Not sure if that is the intended use case though.

Certainly it is. And that will work fine, but I suspect you don't do that while having a partition on /dev/sda mounted as a file system. If you just do read-only workloads like that, you're fine. But as /dev/sda is a superset of /dev/sda1, for example, writing to /dev/sda would end up clobbering data on /dev/sda1 too. The ```allow_mounted_write``` check doesn't check for that, it just checks that if you use /dev/sda as the filename, it'll fail by default if /dev/sda itself is mounted. If you do have partitions and one of them is mounted, the current check won't catch that.

zwhitchcox:
Oh ok, gotcha. That makes sense

Issue #1007, opened 2020-06-04
abort if dest device is mounted


curu:
some user may copy fio test command from the internet and set -filename=/dev/vdb, which  contains valuable data, and would be  destroied after running fio.

we should abort if -filename is set to a device file and the device or partition is already mounted, prevent possible data loss.

sitsofe:
Before we go further I can't stress the following enough: I recommend people should not access raw disks/block devices unless they KNOW they can afford to irrecoverably lose the data on it.

@curu I'm afraid this (people destroying their data by accident) can happen and there's only so much that can be done. I would think it would be better for all those giving the example on the internet to update their text to include a warning that says: "don't run this command on disks device that has data you can't afford to irrecoverably lose".

However back to your original point: notice what you used as your example - `/dev/vdb`. It's very unlikely people mount a filesystem over the whole "disk" block device. Usually you partition it the disk and then mount the filesystem atop that partition (e.g. the filesystem would be on `/dev/vdb1`). If you look in https://github.com/axboe/fio/blob/9caaf02b06deecf0b0a13c16b6d59dddc64b8c35/lib/mountcheck.c#L11 you'll see code that actually checks for this later (you're accessing the direct device (e.g. partition) the filesystem is on) BUT it can't handle the former (accessing the **device the partition is on** where the partition has a mounted filesystem) because the filesystem is at least once removed from it! Further the OS itself has made a design decision - OSes like FreeBSD and Windows don't let you write to raw disk if you have ANY partitions on it (and this is enforced by the kernel) whereas Linux will so it's the case that on Linux you can do something dangerous and nothing will stop you. Note this doesn't apply just to fio, even tools like `dd` can destroy your data and that doesn't even attempt to warn in any situation.

I'm going to close this because I don't think it's easy to do better than what we currently have without a huge amount of work (what happens when it's a filesystem on top a device mapper device on top of a partition on top of a block device?). However, if someone submits a patch that makes the checks even more robust we'll definitely review it (maybe https://github.com/karelzak/util-linux/blob/master/misc-utils/lsblk.c will give some ideas?).

PR #1822, opened 2024-09-28
linux: use O_EXCL when opening a file for write

keithr-mext:
On linux systems, opening a device with O_EXCL will fail with EBUSY if the drive is in use.  For example, if a partition is mounted on /dev/nvme1n1p1, the current mounted filesystem check will allow running write workloads on /dev/nvme1n1 even if allow_mounted_write is zero.  Opening /dev/nvme1n1 with O_EXCL will fail with EBUSY in that case.  Add O_EXCL when opening a file with O_WRONLY or O_RDWR on linux systems.

Fixes: https://github.com/axboe/fio/issues/1820
Signed-of-by: Keith Reynolds <keith.reynolds@mext.ai>

### DIFF
--- a/engines/nvme.c
+++ b/engines/nvme.c
@@ -819,7 +819,14 @@ int fio_nvme_reset_wp(struct thread_data *td, struct fio_file *f,
 	/* If the file is not yet opened, open it for this function. */
 	fd = f->fd;
 	if (fd < 0) {
-		fd = open(f->file_name, O_RDWR | O_LARGEFILE);
+		int flags = O_RDWR | O_LARGEFILE;
+
+#ifdef FIO_USE_O_EXCL
+		if (!td->o.allow_mounted_write)
+			flags |= O_EXCL;
+#endif
+
+		fd = open(f->file_name, flags);
 		if (fd < 0)
 			return -errno;
 	}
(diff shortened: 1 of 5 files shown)


keithr-mext:
The changes to xnvme will require a similar change to use O_EXCL in that library; I'm planning to submit one to them. Regardless of that, passing the correct access bits for the type of I/O fio will be doing seems better to me.

sitsofe:
@keithr-mext I _really_ like the idea of this pull request because I'm sure it would saved people a lot of heartache. A quick of search of the fio bug tracker shows at least the following requests for something like this:
- https://github.com/axboe/fio/issues/1007
- https://github.com/axboe/fio/issues/1382

and I'm sure there are many cases where accidents have happened and data destroyed.

However even if this pull request were to be rebased my concern is that jobs like:
```
fio --name=exclusive_dev  --rw=write --numjobs=2 --filename=/dev/vdz
fio --name=exclusive_file --rw=write --numjobs=2 --filename=/tmp/fio.tmp --size=4k
```
go on to fail with "Device or resource busy" / "File exists". If you can think of a way round this I'd be all for this being controlled by [`allow_mounted_write`](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-allow_mounted_write).

@axboe To sidestep the above would you be OK if this were turned into a seperate option even if it wasn't possible to turn it on by default?

PR #1873, opened 2025-03-02
docs: update docs for verify_mode=compare of io_uring_cmd

minwooim:
Add missing limitation of verify_mode=compare in io_uring_cmd ioengine.  Data verification with NVMe COMPARE command has been introduced in Commit 6170d92a61da ("io_uring: Support Compare command for verification") and this should have documented COMPARE command only supports in case of data pattern verification.

The two more options should be with --verify_mode=compare.
	verify_mode=compare
	verify=pattern
	verify_pattern=<pattern>

2. Empty second line
4. Empty line

Reminders:

### DIFF
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -2896,7 +2896,9 @@ with the caveat that when used on the command line, they must come after the
                 **read**
                         Use Read commands for data verification
                 **compare**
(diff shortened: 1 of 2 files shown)


vincentkfu:
Please update the HOWTO as well.

PR #1883, opened 2025-03-12
Improve flushing pagecache on darwin

Developer-Ecosystem-Engineering:
The current behavior of flushing the pagecache via `posix_fadvise(POSIX_FADV_DONTNEED)` is not a valid operation on macOS. 

This can result in adverse outcomes when measuring performance. The change will instead flush via `mmap` and `msync`, which better reflects the desired behavior on macOS

### DIFF
--- a/os/os-mac.h
+++ b/os/os-mac.h
@@ -17,6 +17,8 @@
 #include "../arch/arch.h"
 #include "../file.h"
 
+#include "mac/posix.h"
+
 #define FIO_USE_GENERIC_INIT_RANDOM_STATE
 #define FIO_HAVE_GETTID
 #define FIO_HAVE_CHARDEV_SIZE
@@ -117,3 +119,5 @@ static inline bool os_cpu_has(cpu_features feature)
 }
 
 #endif
+
+#define CONFIG_POSIX_FADVISE
--- a/Makefile
+++ b/Makefile
@@ -270,6 +270,7 @@ ifeq ($(CONFIG_TARGET_OS), HP-UX)
 endif
 ifeq ($(CONFIG_TARGET_OS), Darwin)
   LIBS	 += -lpthread -ldl
+  SOURCE += os/mac/posix.c
 endif
 ifneq (,$(findstring CYGWIN,$(CONFIG_TARGET_OS)))
   SOURCE += os/windows/cpu-affinity.c os/windows/posix.c os/windows/dlls.c
--- a/os/mac/posix.c
+++ b/os/mac/posix.c
@@ -1,0 +1,82 @@
+#include <errno.h>
+#include <stdint.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/param.h>
+
+#include "../../log.h"
+
+#include "posix.h"
+
+#define MMAP_CHUNK_SIZE		(16LL * 1024 * 1024 * 1024)
+
+/*
+ * NB: performance of discard_pages() will be slower under Rosetta.
+ */
+static int discard_pages(int fd, off_t offset, off_t len)
+{
+	/* Align offset and len to page size */
+	long pagesize = sysconf(_SC_PAGESIZE);
+	long offset_pad = offset % pagesize;
+	offset -= offset_pad;
+	len += offset_pad;
+	len = (len + pagesize - 1) & -pagesize;
+
+	while (len > 0) {
+		int saved_errno;
+		size_t mmap_len = MIN(MMAP_CHUNK_SIZE, len);
+		void *addr = mmap(0, mmap_len, PROT_NONE, MAP_SHARED, fd,
+				  offset);
+
+		if (addr == MAP_FAILED) {
+			saved_errno = errno;
+			log_err("discard_pages: failed to mmap (%s), "
+				"offset = %llu, len = %zu\n",
+				strerror(errno), offset, mmap_len);
+			return saved_errno;
+		}
+
+		if (msync(addr, mmap_len, MS_INVALIDATE)) {
+			saved_errno = errno;
+			log_err("discard_pages: msync failed to free cache "
+				"pages\n");
+
+			if (munmap(addr, mmap_len) < 0)
+				log_err("discard_pages: munmap failed (%s)\n",
+					strerror(errno));
+			return saved_errno;
+		}
+
+		if (munmap(addr, mmap_len) < 0) {
+			saved_errno = errno;
+			log_err("discard_pages: munmap failed (%s), "
+				"len = %zu)\n", strerror(errno), mmap_len);
+			return saved_errno;
+		}
+
+		len -= mmap_len;
+		offset += mmap_len;
+	}
+
+	return 0;
+}
+
+int posix_fadvise(int fd, off_t offset, off_t len, int advice)
+{
+	int ret;
+
+	switch(advice) {
+	case POSIX_FADV_NORMAL:
+	case POSIX_FADV_RANDOM:
+	case POSIX_FADV_SEQUENTIAL:
+		ret = 0;
+		break;
+	case POSIX_FADV_DONTNEED:
+		ret = discard_pages(fd, offset, len);
+		break;
+	default:
+		ret = EINVAL;
+	}
+
+	return ret;
+}
--- a/os/mac/posix.c
+++ b/os/mac/posix.c
@@ -1,4 +1,6 @@
 #include <errno.h>
+#include <fcntl.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <string.h>
 #include <sys/mman.h>
@@ -61,15 +63,30 @@ static int discard_pages(int fd, off_t offset, off_t len)
 	return 0;
 }
 
+static inline int set_readhead(int fd, bool enabled) {
+	int ret;
+
+	ret = fcntl(fd, F_RDAHEAD, enabled ? 1 : 0);
+	if (ret == -1) {
+		ret = errno;
+	}
+
+	return ret;
+}
+
 int posix_fadvise(int fd, off_t offset, off_t len, int advice)
 {
 	int ret;
 
 	switch(advice) {
 	case POSIX_FADV_NORMAL:
-	case POSIX_FADV_RANDOM:
-	case POSIX_FADV_SEQUENTIAL:
 		ret = 0;
+		break;
+	case POSIX_FADV_RANDOM:
+		ret = set_readhead(fd, false);
+		break;
+	case POSIX_FADV_SEQUENTIAL:
+		ret = set_readhead(fd, true);
 		break;
 	case POSIX_FADV_DONTNEED:
 		ret = discard_pages(fd, offset, len);
--- a/os/mac/posix.h
+++ b/os/mac/posix.h
@@ -1,0 +1,11 @@
+#ifndef FIO_MAC_POSIX_H
+#define FIO_MAC_POSIX_H
+
+#define POSIX_FADV_NORMAL       (0)
+#define POSIX_FADV_RANDOM       (1)
+#define POSIX_FADV_SEQUENTIAL   (2)
+#define POSIX_FADV_DONTNEED     (4)
+
+extern int posix_fadvise(int fd, off_t offset, off_t len, int advice);
+
+#endif


helpers.c
```diff
@@ -26,9 +29,68 @@ int sync_file_range(int fd, uint64_t offset, uint64_t nbytes,
...
+ */
+
+#define MMAP_CHUNK_SIZE		(1024 * 1024 * 1024)
+
+int discard_pages(int fd, off_t offset, off_t size)
+{
+	caddr_t *addr;
+	uint64_t chunk_size = MMAP_CHUNK_SIZE;
+
+	printf("fsync'ing file, then invalidating UBC.\n");
```
axboe, line 44: Leftover debug statement?

helpers.c
```diff
@@ -26,9 +29,68 @@ int sync_file_range(int fd, uint64_t offset, uint64_t nbytes,
 }
 #endif
 
+#ifdef CONFIG_MAC_FADVISE_DISCARD
+/**
```
axboe, line 33: Just single ```*``` please

helpers.c
```diff
@@ -26,9 +29,68 @@ int sync_file_range(int fd, uint64_t offset, uint64_t nbytes,
...
+
+int discard_pages(int fd, off_t offset, off_t size)
+{
+	caddr_t *addr;
+	uint64_t chunk_size = MMAP_CHUNK_SIZE;
+
+	printf("fsync'ing file, then invalidating UBC.\n");
+	if (fsync(fd) < 0) {
+		fprintf(stderr, "Cannot fsync file\n");
+		exit(1);
```
axboe, line 47: Needs to return an error here, not exit. It should just return errno, but before you overwrite it by calling fprintf.

helpers.c
```diff
@@ -26,9 +29,68 @@ int sync_file_range(int fd, uint64_t offset, uint64_t nbytes,
...
+	 * mmap the file in 1GB chunks and msync(MS_INVALIDATE).
+	 */
+	while (size > 0) {
+		uint64_t mmap_size = MIN(chunk_size, size);
+
+		addr = mmap((caddr_t)0, mmap_size, PROT_NONE, MAP_SHARED, fd, offset);
+		if (addr == MAP_FAILED) {
+			fprintf(stderr, "Failed to mmap (%s), offset = %llu, size = %llu\n",
+				strerror(errno), offset, mmap_size);
+			return -1;
```
axboe, line 60: You should return errno for the failure cases, not -1. For this one and other spots in this function.

helpers.c
```diff
@@ -26,9 +29,68 @@ int sync_file_range(int fd, uint64_t offset, uint64_t nbytes,
...
+
+		size -= mmap_size;
+		offset += mmap_size;
+	}
+
+	return 0;
+}
+
+int posix_fadvise(int fd, off_t offset, off_t len, int advice)
+{
```
axboe, line 82: I'd probably write this function as:

```
if (advice == POSIX_FADV_DONTNEED)
        return discard_pages(fd, offset, len);
return 0;
```

and avoid needing 'ret' at all.

axboe: Well and the return 0 should be return EINVAL or something like that, should not be pretending success for cases it's not handling.

Developer-Ecosystem-Engineering:
Thank you for the fast review @axboe, will integrate the requested changes!

helpers.c
```diff
@@ -26,9 +29,68 @@ int sync_file_range(int fd, uint64_t offset, uint64_t nbytes,
...
+	int ret = 0;
+
+	if (advice == POSIX_FADV_DONTNEED)
+		ret = discard_pages(fd, offset, len);
+	return ret;
+}
```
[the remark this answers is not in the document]
Developer-Ecosystem-Engineering: Would the project prefer we move this into an os-mac.h?

configure
```diff
@@ -2997,7 +2997,9 @@ fi
 print_config "timerfd_create" "$timerfd_create"
 
 #############################################################################
-
+if test "$targetos" = "Darwin"; then
+    output_sym "CONFIG_MAC_FADVISE_DISCARD"
+fi
```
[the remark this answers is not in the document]
Developer-Ecosystem-Engineering: Whatever the project would prefer; let us know!

helpers.c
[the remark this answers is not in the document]
axboe: It's always nice to encapsulate bits like this into system specific headers or a .c file that's only compiled for that platform, rather than "polluting" generic code with it. Also tends to avoid some of the ifdef hassles, which are a maintenance nightmare generally.

configure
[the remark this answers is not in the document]
axboe: Please be consistent across the project - if the existing __APPLE__ works for you, then just use that rather than have two separate ones. Less confusing that way, otherwise others will read this code and think there's a difference between the two.

Developer-Ecosystem-Engineering:
The latest commit should address all the open feedback, thank you @so much and @axboe for the feedback. Let us know if there is anything else necessary or desired! 

- Remove debug print
- Make all comments a single *
- Use /* */ instead of //
- Return instead of exit
- Ensure errno is populated
- Direct return instead of using a placeholder
- Move to __APPLE__
- Migrate code to os/os-mac.h since it is about macOS

axboe:
You should fold it into a single patch and force push it again, it doesn't make a lot of sense to have a broken state in the upstream tree.

Developer-Ecosystem-Engineering:
> You should fold it into a single patch and force push it again, it doesn't make a lot of sense to have a broken state in the upstream tree.

Not a problem, taking a peek at that failure real quick. Once everyone is satisfied, will squash and force push.

os/os-mac.h
```diff
@@ -117,3 +122,60 @@ static inline bool os_cpu_has(cpu_features feature)
 }
 
 #endif
+
+#if defined (__APPLE__)
```
sitsofe, line 126: You won't need this guard here as `os-mac.h` will only be included on macOS (see https://github.com/axboe/fio/blob/58818df7b0948d36ba9743160a1f570700d2347d/os/os.h#L49 ).

os/os-mac.h
```diff
@@ -117,3 +122,60 @@ static inline bool os_cpu_has(cpu_features feature)
...
+#define MMAP_CHUNK_SIZE		(1024 * 1024 * 1024)
+
+static inline int discard_pages(int fd, off_t offset, off_t size)
+{
+	caddr_t *addr;
+	uint64_t chunk_size = MMAP_CHUNK_SIZE;
+
+	if (fsync(fd) < 0) {
+                int __err = errno; 
+		fprintf(stderr, "Cannot fsync file\n");
```
sitsofe, line 140: I think we would be better off using `log_err()`. Doing a quick grep there isn't a clear standard for how log_err's string should start but perhaps  a prefix of "discard_pages: " would make it clear where the error bubbled up from?

sitsofe (commented):
Just out of interest do you see a difference running the following job with and without your change?

```
./fio --filename=fio.tmp --stonewall --thread --size=1G --filename=fio.tmp --bs=4k \
  --name=create --rw=write \
  --name=cached --rw=read --loops=2 --invalidate=0 \
  --name=invalidated --rw=read --loops=2 --invalidate=1
```

helpers.c
```diff
@@ -26,9 +26,19 @@ int sync_file_range(int fd, uint64_t offset, uint64_t nbytes,
 }
 #endif
 
+#if defined (__APPLE__)
```
sitsofe, line 29: You might be able to move this into `os/os-mac.h` by doing `#define CONFIG_POSIX_FADVISE` over there and providing an implementation in that file...

Developer-Ecosystem-Engineering:
Squashed all commits and force pushed as requested. 

> Just out of interest do you see a difference running the following job with and without your change?
> 
> ```
> ./fio --filename=fio.tmp --stonewall --thread --size=1G --filename=fio.tmp --bs=4k \
>   --name=create --rw=write \
>   --name=cached --rw=read --loops=2 --invalidate=0 \
>   --name=invalidated --rw=read --loops=2 --invalidate=1
> ```

@sitsofe we'd expect to see the removal of cached reads on macOS. 

An example data set 
27.2GiB/s-27.2GiB/s (29.2GB/s-29.2GB/s) -- A cached read, this shouldn't occur 
3290MiB/s-3290MiB/s (3450MB/s-3450MB/s), -- An uncached read

os/os-mac.h
```diff
@@ -117,3 +123,58 @@ static inline bool os_cpu_has(cpu_features feature)
 }
 
 #endif
+
+/*
+ * discard_pages should not be run within Rosetta. Native arm64 or x86 only. 
+ */
+
+#define MMAP_CHUNK_SIZE		(1024 * 1024 * 1024)
```
sitsofe, line 131: Why is the chunk only 1GByte? Is it bad to map a larger amount like 4GBytes on a machine that might only have 8GBytes?

os/os-mac.h
```diff
@@ -117,3 +123,58 @@ static inline bool os_cpu_has(cpu_features feature)
...
+
+static inline int discard_pages(int fd, off_t offset, off_t size)
+{
+	caddr_t *addr;
+	uint64_t chunk_size = MMAP_CHUNK_SIZE;
+
+	if (fsync(fd) < 0) {
```
sitsofe, line 138: Why do you `fsync()` here? Is it to speed up the invalidation in the case that there's cached data that need flushing to disk? If you want flushing semantics is this better than using the `MS_SYNC` flag?

I'm even wondering whether we need to do any flushing at all...

os/os-mac.h
```diff
@@ -117,3 +123,58 @@ static inline bool os_cpu_has(cpu_features feature)
...
+	/*
+	 * mmap the file in 1GB chunks and msync(MS_INVALIDATE).
+	 */
+	while (size > 0) {
+		uint64_t mmap_size = MIN(chunk_size, size);
+
+		addr = mmap((caddr_t)0, mmap_size, PROT_NONE, MAP_SHARED, fd, offset);
+		if (addr == MAP_FAILED) {
+                        int __map_errno = errno;
+			fprintf(stderr, "Failed to mmap (%s), offset = %llu, size = %llu\n",
```
sitsofe, line 154: @Developer-Ecosystem-Engineering I've just looked in `printf(3)` and it makes no mention of `errno` so I wonder why you have to save and restore it...

@axboe do we think that it's possible that printing the error message could end up overwriting `errno`?

helpers.c
```diff
@@ -26,9 +26,19 @@ int sync_file_range(int fd, uint64_t offset, uint64_t nbytes,
...
+int posix_fadvise(int fd, off_t offset, off_t len, int advice)
+{
+	if (advice == POSIX_FADV_DONTNEED)
+		return discard_pages(fd, offset, len);
+        errno = EINVAL;
```
sitsofe, line 34: You can't see it in GitHub but your file has trailing whitespace in it (I can see it when I look at `git log -p` or `git diff HEAD^` - can you remove all instances of this?

os/os-mac.h
```diff
@@ -14,8 +14,14 @@
 #include <machine/endian.h>
 #include <libkern/OSByteOrder.h>
 
+#include <sys/stat.h>
```
sitsofe, line 17: Do you need this include?

os/os-mac.h
```diff
@@ -14,8 +14,14 @@
...
+#include <sys/mman.h>
```
sitsofe, line 18: This include doesn't need to be in a section of its own - can you move to be after `sys/disk.h`?

os/os-mac.h
```diff
@@ -117,3 +123,58 @@ static inline bool os_cpu_has(cpu_features feature)
 }
 
 #endif
+
+/*
+ * discard_pages should not be run within Rosetta. Native arm64 or x86 only.
```
sitsofe, line 128: Is there a way of checking if this is happening? What happens if you DO do the `msync(MS_INVALIDATE)` under Rosetta? :-)

os/os-mac.h
```diff
@@ -117,3 +123,58 @@ static inline bool os_cpu_has(cpu_features feature)
...
+
+	if (fsync(fd) < 0) {
+                int __err = errno; 
+		log_err("%s: Cannot fsync file: %s \n", __func__, strerror(errno));
+                errno = __err;
+		return -1;
+	}	    
+    
+	/*
+	 * mmap the file in 1GB chunks and msync(MS_INVALIDATE).
```
sitsofe, line 146: I think you can skip this comment - it literally describing the loop so it adds very little...

os/os-mac.h
```diff
@@ -117,3 +123,58 @@ static inline bool os_cpu_has(cpu_features feature)
...
+		}
+
+		if (msync(addr, mmap_size, MS_INVALIDATE)) {
+                        int __msync_errno = errno;
+			fprintf(stderr, "msync failed to free cache pages.\n");
+                        errno = __msync_errno;
+			return -1;
+		}
+
+		/* Destroy the above mappings used to invalidate cache - cleaning up */
```
sitsofe, line 167: As above - this comment isn't adding much so you can skip it...

configure
```diff
@@ -2997,7 +2997,6 @@ fi
 print_config "timerfd_create" "$timerfd_create"
 
 #############################################################################
-
```
sitsofe, line 3000: I think you missed reverting the removal of this line.

sitsofe (dismissed):
This is moving in the right direction! However I've been running this PR on my Intel Mac and I didn't see the expected results with this command:

```
./fio --stonewall --size=128M --ioengine=posixaio --filename=fio.tmp --iodepth=64 \
      --bs=4k --invalidate=0 --direct=0 \
      --name=create --rw=write \
      --name=cached --rw=randread --loops=2 \
      --name=invalidated --rw=randread --loops=2 --invalidate=1  | grep groupid= -A 1
create: (groupid=0, jobs=1): err= 0: pid=40797: Sat May 17 19:49:50 2025
  write: IOPS=172k, BW=670MiB/s (703MB/s)(128MiB/191msec); 0 zone resets
--
cached: (groupid=1, jobs=1): err= 0: pid=40798: Sat May 17 19:49:50 2025
  read: IOPS=36.9k, BW=144MiB/s (151MB/s)(256MiB/1775msec)
--
invalidated: (groupid=2, jobs=1): err= 0: pid=40799: Sat May 17 19:49:50 2025
  read: IOPS=40.4k, BW=158MiB/s (165MB/s)(256MiB/1623msec)
```

That cached result seemed too low so I played about and I've got a branch that I can share which shuffles things around and hooks up more of fadvise that shows the expected results:

```
create: (groupid=0, jobs=1): err= 0: pid=42100: Sat May 17 19:57:59 2025
  write: IOPS=180k, BW=703MiB/s (737MB/s)(128MiB/182msec); 0 zone resets
--
cached: (groupid=1, jobs=1): err= 0: pid=42101: Sat May 17 19:57:59 2025
  read: IOPS=216k, BW=842MiB/s (883MB/s)(256MiB/304msec)
--
invalidated: (groupid=2, jobs=1): err= 0: pid=42102: Sat May 17 19:57:59 2025
  read: IOPS=63.8k, BW=249MiB/s (261MB/s)(256MiB/1027msec)
```

@axboe how do we feel about committer email addresses that are the obfuscated "made by the GitHub web interface" type? After the squashing it currently says this:
```
Author: Developer-Ecosystem-Engineering <65677710+Developer-Ecosystem-Engineering@users.noreply.github.com>
```

Developer-Ecosystem-Engineering:
Hi @sitsofe, 

Have most of what you've requested ready, but want to give the project full answers to all the open questions, will deliver a new force push squash with the changes + answers for the remaining open questions above. It takes a little bit for us to work through the qualification on the answers, appreciate the patience.

sitsofe:
Hi @Developer-Ecosystem-Engineering:

> Have most of what you've requested ready, but want to give the project full answers to all the open questions, will deliver a new force push squash with the changes + answers for the remaining open questions above. It takes a little bit for us to work through the qualification on the answers, appreciate the patience.

It's just a bit of a shame the turnaround time is so long because I'd love to see your work go in as soon as possible! It's nice to keep the momentum but I appreciate you need responses from others and that depends on their time.

In the mean time I've put up https://github.com/sitsofe/fio/tree/refs/heads/macos_posix_fadvise which is an idea which builds on your work and moves it into an object of its own. I've made the commit message a bit longer to:
-  Show what's been tested and how things have changed
- Provide references to old commits that do something similar on other platforms
  
What do you think of it? If it looks useful please take from it!

Finally, would it be possible for you to use a Real Name in your commit's? Currently we have this:

```
Author:     Developer-Ecosystem-Engineering <65677710+Developer-Ecosystem-Engineering@users.noreply.github.com>
AuthorDate: Wed Mar 12 15:27:29 2025 -0700
Commit:     Developer-Ecosystem-Engineering <65677710+Developer-Ecosystem-Engineering@users.noreply.github.com>
CommitDate: Wed May 14 09:58:58 2025 -0700
[...]
Signed-off-by: [<groupemail>]@apple.com
```

I've talked to @axboe and he said this:
> What I care about is the identify identifies a person, and:
>
> Developer-Ecosystem-Engineering <[65677710+Developer-Ecosystem-Engineering@users.noreply.github.com](mailto:65677710%2BDeveloper-Ecosystem-Engineering@users.noreply.github.com)>
>
> literally means nothing. So yeah, that needs to be a proper [identity] and email.

So could you:
- Make sure your commits have a real email address (as opposed to GitHub obfuscated one)
- Could you use a Real Name with your commit

(see https://stackoverflow.com/questions/3042437/how-can-i-change-the-commit-author-for-a-single-commit for hints on how to fix up existing commits with new authors)

Thanks!

Developer-Ecosystem-Engineering:
Hi @sitsofe, have read through the changes in your [branch](https://github.com/sitsofe/fio/tree/refs/heads/macos_posix_fadvise) and it looks to incorporate all of the feedback & open questions already. You noticed the test failure and the downstream posix_fadvise issue. It also addresses the feedback about the comments and organization. 

It probably makes sense for the project to take your changes since it incorporates all the feedback, we looked at rebasing on top of it with any changes we would have and there was nothing left to add! It also solves the open feedback on sign-off. 

The other open question above was answered and there is one remaining, which I am working on. 

> If we mmap a 1TByte file fully but don't write to will any extra memory be used aside from set up of the TLB?

sitsofe:
@Developer-Ecosystem-Engineering:

Just for the record - I never intended to steal your thunder but thank you for blessing my additions. I want to see your idea and work go into fio in some form because it solves a genuine behaviour issue on macOS.

> It also solves the open feedback on sign-off.

Not quite, if we're going to go with my branch then I would end up squashing everything into one commit and it is only fair to to credit the person that delivered the solution to the [UBC cache invalidation problem I failed to solve several years ago](https://github.com/axboe/fio/issues/48#issuecomment-328012610). Please could I have a name and email address so I can add a `Suggested-by` / `Co-developed-by` line? If you **really** want it to be the ecosystem address I will use that...

Developer-Ecosystem-Engineering:
Hi @sitsofe 

We need no thunder, it's OK =) Just want to help too. We'd prefer that we use the group identity if you must use anything, but it's perfectly fine to not have it, it's your code base! If you don't want the GitHub one as its a bit verbose, our GitHub handle without the dashes is the email address. 

Must we unmask the Dread Pirate Roberts? =)

Developer-Ecosystem-Engineering:
> If we mmap a 1TByte file fully but don't write to will any extra memory be used aside from set up of the TLB?

When you mmap a 1TByte file without actually writing to it, the primary work is in virtual memory setup. The kernel will only modify the process memory map, but no pages table entries are created until writes and accompanying page faults occur. 

These faults would occur during access time. If the information is in the UBC, then the process mapped address range will use the resident pages.

sitsofe:
> Must we unmask the Dread Pirate Roberts? =)

Gotcha ~~Westley~~ @Developer-Ecosystem-Engineering - your mystique shall continue on.

Re mmap: thanks for the update. I'll see to reworking things a bit Real Soon Now™ and I'll add you to the review so you can see it going past.

sitsofe:
I finally found the time to work on this but I have a bunch of things that I want to bring to people's attention. Here’s a summary:
- Is invalidating all of an otherwise partial page OK?
- Should readahead be hinted at by job type on macOS?
- Should the mmap chunk size be something other than 16GiB?
- There seems to be a bug truncating massive files on macOS 15.6 under APFS (@Developer-Ecosystem-Engineering)

It’s possible to ask fio to invalidate the cache for a range that isn’t page aligned (e.g. by doing `./fio --name=n --offset=2k --size=18k` ) but the mmap based shim has to work with a granularity of a page. The [Linux posix_fadvise man page](https://man7.org/linux/man-pages/man2/posix_fadvise.2.html) says the following about POSIX_FADV_DONTNEED:
>  Requests to discard partial pages are ignored.  It is preferable to preserve needed data than discard unneeded data. If the application requires that data be considered for discarding, then offset and size must be page-aligned.

The [Open Group posix_fadvise documentation](https://pubs.opengroup.org/onlinepubs/000095399/functions/posix_fadvise.html) and the [FreeBSD posix_fadvise man page](https://man.freebsd.org/cgi/man.cgi?query=posix_fadvise) are silent about what happens to partial pages. For the fio macOS posix_fadvise shim I’ve taken the approach of invalidating all of a page that would otherwise only be partially covered by aligning the offset down and the len up. I can change this if people would prefer more Linux-like behaviour.

Since a large mmap doesn’t really cost that much more memory I figured I would set the mmap chunk size to be larger so that fewer mmap/msync/munmap calls would be done thus reducing the overhead of setting things up/tearing them down repeatedly. Here are the results of my testing: 

macOS 15.6 Intel (2GHz i5) msync(MS_INVALIDATE)
| mmap chunk size (GiB) | 0.5 | 1 | 8 | 16 | 128 | 1024 |
| --- | --- | --- | --- | --- | --- | --- |
| linear read | **0.3710356** | 0.4872094 | 0.3823824 | 0.3986368 | 0.5639938 | 1.9420478 |
| rand read | **10.265224** | 13.656658 | 11.046206 | 11.286724 | 11.46629 | 11.436732 |
| gappy | **7.854328** | 10.5910972 | 9.5441084 | 9.5344792 | 9.5729838 | 9.5502052 |
| near empty | 0.0213032 | 0.0185342 | 0.0037254 | 0.0018804 | 0.0002896 | **0.000118** |

macOS 15.6 M4 Pro msync(MS_INVALIDATE)
| mmap chunk size (GiB) | 0.5 | 1 | 8 | 16 | 128 | 1024 |
| --- | --- | --- | --- | --- | --- | --- |
| linear read | 0.0514746 | 0.0448044 | **0.0401286** | 0.0419396 | 0.0646114 | 0.2266038 |
| rand read | 1.3273664 | 1.3098192 | 1.2935882 | 1.3060148 | **1.2855568** | 1.3004732 |
| gappy | 1.1701312 | 1.1473312 | 1.1202244 | **1.1186982** | 1.1388276 | 1.1316176 |
| near empty | 0.0251256 | 0.0101194 | 0.0011742 | 0.0009186 | 0.000453 | **0.000403** |

macOS 15.6 M4 Pro Rosetta msync(MS_INVALIDATE)
| mmap chunk size (GiB) | 0.5 | 1 | 8 | 16 | 128 | 1024 |
| --- | --- | --- | --- | --- | --- | --- |
| linear read | 0.0592442 | 0.0535576 | **0.0443478** | 0.0457978 | 0.0666314 | 0.2284822 |
| rand read | 1.3480976 | 1.333327 | 1.3117098 | 1.3103916 | 1.3127036 | **1.2981126** |
| gappy | 1.1752522 | 1.1550404 | 1.1484232 | 1.1360098 | **1.134033** | 1.143597 |
| near empty | 0.028696 | 0.0207544 | 0.0085264 | 0.0080604 | 0.0063026 | **0.006159** |

Linux AMD EPYC 7742 (2.2Ghz) posix_fadvise([…], POSIX_FADV_DONTNEED)
| Fill I/O pattern | Time (s) |
| --- | --- |
| linear read | 1.6783038 | 
| rand read | 4.6355044 | 
| gappy | 4.0741604 | 
| near empty | 0.0000204 | 

The additional patch to do log the time to do invalidation ([0001-DRAFT-add-cache-invalidation-logging.patch](https://github.com/user-attachments/files/22126284/0001-DRAFT-add-cache-invalidation-logging.patch)) and a script to do the benchmarking ([bench.sh](https://github.com/user-attachments/files/22126297/bench.sh)) are attached. Results are the time it took to invalidate the page cache after performing 8 GBytes of 4 KByte sized reads in a variety of patterns/situations against a 1 TByte sparse file (and different mmap chunk sizes on macOS). Where there are multiple results in a row the quickest time is highlighted in bold.

Summary: The M4 Pro Mac results are extremely quick, the Intel Mac machine was in last place but performed best with a small mmap chunk size and using macOS’ Rosetta doesn’t slow things down by that much. To my surprise, it's only when the page cache is empty does a larger chunk size perform the best! On the Apple Silicon Mac using a chunk size of 16 GiB compared to 1 GiB took a tiny bit less time in the near empty page cache case (which I would hope is the most common scenario) and didn't do badly in other cases so I’ve gone with that but I’m open to suggestions.

While doing the aforementioned tuning, I was shocked to find that macOS can take more than a second to invalidate 8 GBytes of cache loaded randomly. I didn’t remember dropping the page cache on Linux being that slow so I tried the same test on that OS and found it took even longer! Unfortunately time spent invalidating the cache is counted as part of an fio job’s runtime - if you have a job whose runtime is 1 second but invalidating the cache takes 2 seconds, then only a single I/O is done and the bandwidth reported is very low because most of the job’s time was spent invalidating the cache. Fixing this is hard because invalidation happens when a file is opened and further, each time a loop is completed the file is re-opened. For now I’ve just added an ETA status for invalidation so it’s a little bit more visible, although the status is only updated 3 seconds after the first job starts so if the initial invalidation takes less time than that the ETA status won’t show it. I can drop the ETA change if people don't feel it's useful.

macOS’s default readahead seems to be able to turn itself off fairly rapidly. I found that only random/non-sequential jobs that trick readahead into getting a few hits before seeking to a new location show better performance with readahead turned off via fadvise_hint=random. Do people still want the readahead turned off by default on macOS random workloads?

There is a 100% reproducible bug on at least macOS 15.6 where if a large sparse file is created (e.g. 1 Petabyte big), 16 random reads are done from it and the file truncated to 1 Terabyte, then truncation will spin a CPU indefinitely. When in this state, the truncation can’t be aborted/killed even during system shut down (when starting back up macOS asks the user to report an issue). Trying the same test on a file on Linux under XFS (because ext4 limits file sizes to 16 Terabytes) does not show the same behaviour - truncation finishes after a few seconds. The following reproduces the problem on both an Intel Mac and an M4 Mac:
```
rm -f /tmp/huge; truncate -s 1p /tmp/huge
./fio --rw=randread --number_ios=16 --filename /tmp/huge --norandommap --name=n
truncate -s 1T /tmp/huge
```

@Developer-Ecosystem-Engineering can you reproduce the problem too?

axboe:
Apart from that one patch adding the invalidating state, looks fine to me. Maybe we just defer that decision to later and you just drop it from now? Or if you think it really needs to be there, let's here the reasoning :-)

Developer-Ecosystem-Engineering:
> There is a 100% reproducible bug on at least macOS 15.6 where if a large sparse file is created (e.g. 1 Petabyte big), 16 random reads are done from it and the file truncated to 1 Terabyte, then truncation will spin a CPU indefinitely. When in this state, the truncation can’t be aborted/killed even during system shut down (when starting back up macOS asks the user to report an issue). 

We reproduced it! Tracking with 159795525.

sitsofe:
@axboe:
> Apart from that one patch adding the invalidating state, looks fine to me. Maybe we just defer that decision to later and you just drop it from now? Or if you think it really needs to be there, let's here the reasoning :-)

I've become quite attached to that patch while doing this testing these changes :-) However, my sentimentality isn't enough that it has to go in now and given how rarely it shows on the ETA let's drop the patch here (I'll keep it stashed on a branch somewhere).

@Developer-Ecosystem-Engineering:

> We reproduced it! Tracking with 159795525.

Thank you!

sitsofe:
A quick check @Developer-Ecosystem-Engineering, @axboe did you feel this PR needs further changes?

axboe:
Thanks everyone, now merged.

PR #1892, opened 2025-05-07
Add important info about http_host and file formatting to s3 example

avrittrohwer:
Add important info about http_host and file formatting to s3 example

Will save future folks from having to debug the same thing I did.

### DIFF
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -808,8 +808,11 @@ Target file/device
 	For HTTP and S3 access, specify a valid URL path or S3 key, respectively.
 	A filename for path-style S3 includes a bucket name (:file:`/bucket/k/e.y`)
 	while a virtual-hosted-style S3 filename :file:`/k/e.y` does not because 
-	its bucket name is specified in :option:`http_host`.
-
+	its bucket name is specified in :option:`http_host`. In both cases, the
+	filename should begin with a ``/``. The HTTP engine does not automatically
+	add a leading ``/`` when constructing URLs from :option:`http_host` and
+	:option:`filename`.
+ 	
 	The filename "`-`" is a reserved name, meaning *stdin* or *stdout*.  Which
 	of the two depends on the read/write direction set.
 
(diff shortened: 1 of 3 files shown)


axboe:
Commit messages (and signed-off-by) need to go into the commit itself, not be part of the GH PR message.

avrittrohwer:
> Commit messages (and signed-off-by) need to go into the commit itself, not be part of the GH PR message.

Fixed

axboe:
It's not very descriptive. What does "very important" mean?

```
examples/s3: add info about http_host and file formatting

This caused me some headache, let's add some details on how fio expects the
http_host and file to be formatted.

Signed-off-by: Avritt Rohwer <avritt@google.com>
```

I think would be much better.

avrittrohwer:
> It's not very descriptive. What does "very important" mean?
> 
> ```
> examples/s3: add info about http_host and file formatting
> 
> This caused me some headache, let's add some details on how fio expects the
> http_host and file to be formatted.
> 
> Signed-off-by: Avritt Rohwer <avritt@google.com>
> ```
> 
> I think would be much better.

Sure.  Also added some more info on required steps for enabling the http engine while building and added the leading / info to the docs too

axboe:
Thanks! You still missed the ```<>``` around your email, but let's just get it in.

sitsofe:
@avrittrohwer would you be able to update the man page too? See https://github.com/axboe/fio/blob/f4e54d3fa7bf94a17648a4439f06ca7dab4291c0/fio.1#L591-L593 ...

avrittrohwer:
> @avrittrohwer would you be able to update the man page too? See
> 
> https://github.com/axboe/fio/blob/f4e54d3fa7bf94a17648a4439f06ca7dab4291c0/fio.1#L591-L593
> 
> ...

From https://github.com/axboe/fio/blob/f4e54d3fa7bf94a17648a4439f06ca7dab4291c0/README.rst#documentation it seemed like the man page is generated, is that step required manually by editors or taken care of automatically via CI?

Issue #1920, opened 2025-06-13
Add NULL check after calloc() in fileoperations.c:init()


lhywk:
### **Describe**

In the `init()` function of `fileoperations.c`, the return value of `calloc()` is not checked before being dereferenced:

```c
data = calloc(1, sizeof(*data));
data->stat_ddir = DDIR_READ;
```

While this typically works fine, `calloc()` can theoretically return `NULL` in memory-constrained environments, which would lead to a potential NULL pointer dereference.

---

### **Version**

fio-3.40

---

### **Expected behavior**

If `calloc()` fails, the code should gracefully return an error (e.g., `-ENOMEM`) rather than proceeding with a potentially unsafe dereference.

---

### **Actual behavior**

Currently, no NULL check is performed. This could result in undefined behavior (segmentation fault) if memory allocation fails.

---

### **How to reproduce**

This issue is difficult to trigger in typical environments, but can be exposed by running under memory-constrained conditions (e.g., using `ulimit`), or reliably detected by static analysis tools such as Infer.

---

Please let me know if this approach seems reasonable.

I'm happy to revise the patch based on any feedback or suggestions.

axboe:
See https://github.com/axboe/fio/pull/1921

PR #1921, opened 2025-06-13
Fix: add NULL check after calloc() in fileoperations.c init()

lhywk:
Fix: add NULL check after calloc() in fileoperations.c init()

This patch adds a NULL check after `calloc()` in the `init()` function of
fileoperations.c. If memory allocation fails, dereferencing `data` would
result in undefined behavior.
This change ensures a graceful return (-ENOMEM) instead. The fix improves
robustness with no effect on normal execution flow.

Fixes: #1920 

Signed-off-by: Hoyoung Lee <lhywkd22@gmail.com>

### DIFF
--- a/engines/fileoperations.c
+++ b/engines/fileoperations.c
@@ -283,7 +283,9 @@ static int init(struct thread_data *td)
 	struct fc_data *data;
 
 	data = calloc(1, sizeof(*data));
-
+	if (data == NULL)
+		return -ENOMEM;
+	
 	if (td_read(td))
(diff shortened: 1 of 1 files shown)


axboe:
Sorry no, it's very pointless to check a single malloc/calloc invocation and handle the error there, when there are a ton more. Overcommit is a thing, and I've never heard of anyone actually running into a crash due to calloc/malloc failing, regardless of Linux or any other OS.

tldr - not really an issue, and if it was, there'd be a ton more places to fix up.

PR #1939, opened 2025-07-07
engines/http: Add S3 security token support

sfc-gh-rnarubin:
Security tokens are an element of S3 authorization in some environments. This change adds a parameter to allow users to specify a security token, and pass this to S3 requests with the appropriate header.


axboe:
Looks good to me, but please update the ```fio.1``` man page as well.

sfc-gh-rnarubin:
Missed that, thanks. I've pushed an update

Issue #1810, opened 2024-08-26
http ioengine uses deprecated functions


vincentkfu:
The http ioengine no longer builds on newer distributions because it uses deprecated functions. Here is the related `config.log` output:

```
Compiling test case curl-new-ssl
gcc -D_GNU_SOURCE -include config-host.h -Werror -Werror-implicit-function-declaration -o /tmp/fio-conf-14769-21457-3596.exe /tmp/fio-conf-18565-21457-27555.c -lcurl -lssl -lcrypto -lnuma -lz
/tmp/fio-conf-18565-21457-27555.c: In function 'main':
/tmp/fio-conf-18565-21457-27555.c:12:3: error: 'HMAC_CTX_new' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
   12 |   ctx = HMAC_CTX_new();
      |   ^~~
In file included from /tmp/fio-conf-18565-21457-27555.c:2:
/usr/include/openssl/hmac.h:33:33: note: declared here
   33 | OSSL_DEPRECATEDIN_3_0 HMAC_CTX *HMAC_CTX_new(void);
      |                                 ^~~~~~~~~~~~
/tmp/fio-conf-18565-21457-27555.c:13:3: error: 'HMAC_CTX_reset' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
   13 |   HMAC_CTX_reset(ctx);
      |   ^~~~~~~~~~~~~~
/usr/include/openssl/hmac.h:34:27: note: declared here
   34 | OSSL_DEPRECATEDIN_3_0 int HMAC_CTX_reset(HMAC_CTX *ctx);
      |                           ^~~~~~~~~~~~~~
/tmp/fio-conf-18565-21457-27555.c:14:3: error: 'HMAC_CTX_free' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
   14 |   HMAC_CTX_free(ctx);
      |   ^~~~~~~~~~~~~
/usr/include/openssl/hmac.h:35:28: note: declared here
   35 | OSSL_DEPRECATEDIN_3_0 void HMAC_CTX_free(HMAC_CTX *ctx);
      |                            ^~~~~~~~~~~~~
cc1: all warnings being treated as errors
Compiling test case curl-old-ssl
[...]

PR #1944, opened 2025-07-15
engines/http: Add support for range reads

sfc-gh-rnarubin:
The existing HTTP implementation treated blocks as individual objects when reading and writing. This is useful in as far as it enables read and write symmetry -- object stores generally don't allow writes to individual ranges of objects like file systems do with files, so treating blocks as objects is a practical way to replicate block reads and writes.

Reading of object ranges, on the other hand, is widely supported by object stores with the "Range" HTTP header. This change adds a parameter which alters fio's object conventions to issue reads using the block size and offset parameters more like file IO. When enabled, both reads and writes will use the plain filename as the object path to issue IO. Reads will add a "Range: bytes=<start>-<end>" header to the requests, where the start and end positions are determined by the blocksize and offset of the benchmark. Aside from the object path, writes are unchanged for simplicity: the object size is determined by blocksize as before.

### DIFF
--- a/HOWTO.rst
+++ b/HOWTO.rst
@@ -3038,6 +3038,20 @@ with the caveat that when used on the command line, they must come after the
 	turns on verbose logging from libcurl, 2 additionally enables
 	HTTP IO tracing. Default is **0**
 
+.. option:: http_range_header=bool : [http]
+
+	Use :option:`blocksize` for range reads instead of the object size for read
+	I/O.
+
+	By default, the HTTP engine treats :option:`blocksize` as the size of
+	the object to read or write, and appends the block start/end offsets to the
+	:option:`filename` to create the target object path. With this parameter
+	enabled, two changes take place: the object path will instead be the
+        filename directly for both read and write I/O, and blocksize and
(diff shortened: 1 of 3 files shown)


sfc-gh-rnarubin:
I considered making the parameter a "mode" switch instead of a bool, i.e. http_object_mode = block / range / etc. Let me know if that's more clear or preferable and I'll make the change

vincentkfu:
`http_object_mode` would be better if it's possible that there may be more than just `block` and `range` in the future since it allows for easier extensibility while maintaining backward compatibility. I am not familiar with how this ioengine can be used, so will let you be the judge.

By the way, is this an issue for you? https://github.com/axboe/fio/issues/1810

sfc-gh-rnarubin:
I could see an option for multipart upload as well, so perhaps the mode switch will be better. I'll put that together

> By the way, is this an issue for you? https://github.com/axboe/fio/issues/1810

I have not run into this yet, but I can look at updating that area when I have a chance

sfc-gh-rnarubin:
I've updated the parameter to "http_object_mode", which will allow potentially other options in the future

PR #1947, opened 2025-07-20
configure: Only link gnutls with libnfs when on Linux

brad0:
GnuTLS is only used when on Linux.


sitsofe:
Hello @brad0:

Can you give some further justification - clearly you've seen an issue but can you explain it further? Is it so bad to link libnfs against GnuTLS on say FreeBSD or Windows? Would we be better off with another option that says don't link libnfs against GnuTLS and then the user can explicitly set this?

sitsofe:
Closing because we need more information from the reporter to take this further.

@brad0 - please re-open this issue if you're able to supply the requested information. Thank you!

PR #1948, opened 2025-07-21
verify: add --verify_type option for crash-consistent verification

minwooim:
This patch introduces a new --verify_type option to support crash-consistent offline verification scenarios. The primary motivation is to handle cases where write operations are interrupted by device failures or unexpected power loss during the write phase, requiring verification to be limited to data that was properly persisted before the interruption.

Key features:
- New --verify_type=flush option filters verification candidates based on fsync completion timing
- Tracks fsync completion timestamps during write phase
- During verification, excludes writes that completed after the last fsync
- Always includes FUA (Force Unit Access) writes regardless of timing, as they bypass cache and are immediately persistent
- Works with existing verify_state_save mechanism for offline verification

Use case:
This is particularly useful for testing storage durability guarantees in scenarios such as:
1. Simulated power failures during write workloads
2. Device error injection testing

and espeically with such ioengines which directly communicates with the device directly as io_uring_cmd does.

This enhancement enables more realistic testing of data durability in storage systems by ensuring only properly synchronized data is verified after simulated failures.

Examples:

1. write phase

[global]
ioengine=io_uring_cmd
cmd_type=nvme
filename=/dev/ng0n1
rw=write
bs=4k
verify=pattern
verify_pattern=%o

iodepth=32
size=32k
fsync=3

[test]
verify_type=flush
do_verify=0
verify_state_save=1

2. read phase

[global]
ioengine=io_uring_cmd
cmd_type=nvme
filename=/dev/ng0n1
rw=write
bs=4k
verify=pattern
verify_pattern=%o

iodepth=32
size=32k
fsync=3

[test]
verify_type=flush
do_verify=1
verify_only=1
verify_state_load=1


sitsofe (changes requested):
@minwooim:

I've very briefly looked over this and from what you're saying the purpose is to allow the verification of only those writes that happened before an `fsync`? How does this interact with things like trims? Do you really foresee a `verify_type` other than flush?

Some additions that will be needed:
- You've updated the man page but you'll also need to update the HOWO documentation around roughly https://github.com/axboe/fio/blob/master/HOWTO.rst?plain=1#L3984 (yeah I know "why two places"? It's a boring story)
- You'll need to bump FIO_SERVER_VER in server.h (this is mentioned in https://github.com/axboe/fio/blob/master/.github/PULL_REQUEST_TEMPLATE.md )

minwooim:
> I've very briefly looked over this and from what you're saying the purpose is to allow the verification of only those writes that happened before an `fsync`? How does this interact with things like trims? Do you really foresee a `verify_type` other than flush?

Not this time since we already have `trim_verify_zero` option for the TRIM-ed offset ranges.  Rather than excluding the TRIM-ed offsets, we can verify them with zero-data.

> * You've updated the man page but you'll also need to update the HOWO documentation around roughly https://github.com/axboe/fio/blob/master/HOWTO.rst?plain=1#L3984 (yeah I know "why two places"? It's a boring story)
> * You'll need to bump FIO_SERVER_VER in server.h (this is mentioned in https://github.com/axboe/fio/blob/master/.github/PULL_REQUEST_TEMPLATE.md )

Will add these two, Thanks for catching this.

io_u.c
```diff
@@ -2078,12 +2079,37 @@ static void file_log_write_comp(const struct thread_data *td, struct fio_file *f
 	if (!f->last_write_comp)
 		return;
 
+	now_nsec = utime_since_now(&io_u->start_time) * 1000;  /* Convert to nanoseconds */
```
vincentkfu, line 2082: `ntime_since_now` exists

engines/io_uring.c
```diff
@@ -556,6 +556,10 @@ static int fio_ioring_cmd_prep(struct thread_data *td, struct io_u *io_u)
 		io_u_set(td, io_u, IO_U_F_VER_IN_DEV);
 	}
 
+	/* Mark FUA writes for verification state tracking */
+	if (io_u->ddir == DDIR_WRITE && (ld->cdw12_flags[DDIR_WRITE] & (1 << 30)))
```
vincentkfu, line 560: I think this would be clearer if you used `o->writefua` for the condition

vincentkfu: Also, the `sg` ioengine has a writefua flag as well.

io_u.c
```diff
@@ -2078,12 +2079,37 @@ static void file_log_write_comp(const struct thread_data *td, struct fio_file *f
...
 
+static void file_log_flush_comp(struct fio_file *f, struct io_u *io_u)
+{
+	uint64_t completion_time;
+
+	if (!f)
+		return;
+
+	/* Track the last FLUSH completion timestamp */
+	completion_time = utime_since_now(&io_u->start_time) * 1000;
```
vincentkfu, line 2106: `ntime_since_now`
Also, we already calculate a completion time elsewhere for all operations. Is there a way to use that completion time instead of calculating a second one?

vincentkfu:
There is also a `t/verify-state.c` file that will need updating.

verify.h
```diff
@@ -30,6 +30,9 @@ enum {
 	VERIFY_PATTERN,			/* verify specific patterns */
 	VERIFY_PATTERN_NO_HDR,		/* verify specific patterns, no hdr */
 	VERIFY_NULL,			/* pretend to verify */
+	VERIFY_FLUSH,			/* verify with flush timing consideration */
```
vincentkfu, line 33: Why is this change needed?

vincentkfu:
For your commit message there needs to be just one Signed-off-by line.

verify.h
minwooim: Missed removing this one, Thanks for catching this.

sitsofe (changes requested):

minwooim:
Will rebase it!

minwooim:
Will re-create a new PR for this changes.

Issue #1950, opened 2025-07-21
verify state does not properly handle out-of-order write completions


riley-pstg:
**Description of the bug:**
When testing data integrity across power cycles / reboots with `trigger` and `verify_state_load=1`, fio does not properly handle out-of-order completions (e.g. due to tail latency). This can lead to fio attempting to verify a block that was never written.

When saving/loading state files for verification, fio keeps track of only a finite number of recent write completions in `thread_io_list::comps`. The verify code (`verify_state_should_stop()`, specifically) assumes that any write *issued* before the writes recorded in `comps` must have been successfully completed. This is not generally true for iodepth > 1 and if writes are completed out-of-order.

Consider this example: a single write stalls while all other writes complete normally. This condition can persist for an arbitrarily long time, and the stalled write will fall out of the range "covered" by `comps`. Saving state at this point (e.g. via a trigger) and halting the workload (e.g. via power-cycling the machine) will result in that stalled write being verified when the state is loaded despite the fact that the write may have never completed.

It seems like a dynamically-sized data structure is needed to keep track of any completion "gaps" which may be arbitrarily out-of-order. It does not seem generally possible with a fixed-size data structure (but maybe I am missing something).

**Environment**:

```
$ cat /etc/redhat-release 
CentOS Stream release 9
$ uname -a
Linux hyper-ba563a-1 6.11.1-gb9fe45adbd3e #20 SMP Mon Jun  2 21:55:22 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

**fio version**: 3.40

**Reproduction steps**
Note that this was reproduced consistently only by injecting artificial tail latency (1 second every 100 writes) via a proprietary ublk-based software storage driver. It may be possible to construct a similar setup with dm. I was able to reproduce the issue in 174 / 200 tests. It is also important to trim the entire space between each test.

```
$ cat trim.fio 
[global]
iodepth=16
direct=1
group_reporting=1
numjobs=1
ioengine=io_uring
[job0]
rw=trim
filename=/dev/ublkb0
blocksize=1M
size=5%
randseed=43747
new_group=1

$ cat write.fio 
[global]
iodepth=16
direct=1
group_reporting=1
numjobs=1
ioengine=io_uring
rw=randwrite
runtime=240
time_based
verify=md5
[job0]
rw=randwrite
filename=/dev/ublkb0
blocksize=128k
size=5%
randseed=43747
new_group=1

$ cat read.fio
[global]
iodepth=16
direct=1
[...]

PR #1951, opened 2025-07-25
handle out-of-order write completions in verify state (Issue #1950)

noclip-code:
See https://github.com/axboe/fio/issues/1950 for a detailed description of the issue.

Details are in the individual commit messages.

I have seen 0 hits in ~30 runs of the repro setup I have. Without the fix the hit rate was 174 / 200.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -949,6 +949,82 @@ static void handle_thinktime(struct thread_data *td, enum fio_ddir ddir,
 }
 
 /*
+ * Add numberio from io_u to the inflight log.
+ */
+void log_inflight(struct thread_data *td, struct io_u *io_u)
+{
+	int idx, offset;
+
+	if (!td->inflight_numberio || io_u->ddir != DDIR_WRITE) {
+		return;
+	}
+
+	if (io_u->inflight_idx != -1) {
+		log_err("inflight_idx already set: inflight_idx=%d\n",
+			io_u->inflight_idx);
+		abort();
+	}
+
+	if (td->inflight_issued != io_u->numberio) {
+		log_err("inflight_issued does not match: numberio=%"PRIu64", inflight_issued=%"PRIu64"\n",
+			io_u->numberio, td->inflight_issued);
+		abort();
+	}
+
+	/* Walk the inflight list until we find a free slot. */
+	idx = td->next_inflight_numberio_idx;
+	for (offset = 0; offset < td->o.iodepth; offset++) {
+		if (td->inflight_numberio[idx] == INVALID_NUMBERIO) {
+			/*
+			 * The order here is important - we must "protect" this write in the
+			 * inflight list before making it visible in inflight_issued.
+			 */
+			atomic_store_release(&td->inflight_numberio[idx], io_u->numberio);
+			td->next_inflight_numberio_idx = (idx + 1) % td->o.iodepth;
+			io_u->inflight_idx = idx;
+
+			atomic_store_release(&td->inflight_issued, io_u->numberio + 1);
+			dprint(FD_VERIFY, "log_inflight: numberio=%"PRIu64", inflight_idx=%d\n",
+				io_u->numberio, idx);
+			return;
+		}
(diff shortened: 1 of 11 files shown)


verify.h
```diff
@@ -45,8 +45,8 @@ struct verify_header {
 	uint64_t offset;
 	uint32_t time_sec;
 	uint32_t time_nsec;
+	uint64_t numberio;
```
axboe, line 48: This will cause failures if someone attempts to verify an existing verification from a previous version. Which probably is quite fine, but can be unexpected. Should probably just be mentioned in the release notes.

axboe:
I think this looks pretty good. I did notice that you're missing signed-off-by in all your commit messages, though.

noclip-code:
Added "Signed-off-by" to all commits.

verify.h
vincentkfu: This would be a good opportunity to add a version to the verify header with a helpful message displayed if a mismatch is found. Without something like this we'll probably end up having to deal with a lot of users wondering why their previously working verify workloads are failing.

noclip-code: That makes sense. According to git, `verify_header` has not changed for the past 8 years, so there's not much precedent here.

I can address this in a separate PR.

noclip-code: Actually, it looks small enough to simply tack on here. Will push an update.

noclip-code:
Added versioning for `verify_header`. Please take another look. The CI failure looks like a timeout to me - it passed on my fork.

backend.c
```diff
@@ -948,6 +948,82 @@ static void handle_thinktime(struct thread_data *td, enum fio_ddir ddir,
...
 }
 
+/*
+ * Add numberio from io_u to the inflight log.
+ */
+void log_inflight(struct thread_data *td, struct io_u *io_u)
+{
+	int idx, offset;
+
+	if (!td->inflight_numberio || io_u->ddir != DDIR_WRITE) {
```
vincentkfu, line 958: Please drop the curly braces here

backend.c
```diff
@@ -1227,49 +1304,31 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
...
 
-cleanup:
-	free_file_completion_logging(td);
-	log_err("fio: failed to alloc write comp data\n");
-	return 1;
+static void free_inflight_logging(struct thread_data *td)
+{
+	if (td->inflight_numberio) {
+		sfree(td->inflight_numberio);
+	}
```
vincentkfu, line 1350: Drop unnecessary curly braces

verify.c
```diff
@@ -1707,16 +1666,13 @@ struct all_io_list *get_all_io_list(int save_mask, size_t *sz)
 	next = &rep->state[0];
 	for_each_td(td) {
 		struct thread_io_list *s = next;
-		unsigned int comps, index = 0;
 
 		if (save_mask != IO_LIST_ALL && (__td_index + 1) != save_mask)
 			continue;
 
-		comps = fill_file_completions(td, s, &index);
+		// copy inflight info to thread_io_list
```
vincentkfu: Please use C-style comments /* ... */

backend.c
```diff
@@ -1024,6 +1024,25 @@ void invalidate_inflight(struct thread_data *td, struct io_u *io_u)
...
+	int i;
+
+	if (!td->inflight_numberio)
+		return;
+
+	for (i = 0; i < td->o.iodepth; i++) {
+		td->inflight_numberio[i] = INVALID_NUMBERIO;
+	}
+	td->next_inflight_numberio_idx = 0;
+	// Experimental verify can increment io_issues for writes, so catch
```
vincentkfu, line 1041: /* */

backend.c
noclip-code: Thank you for the style pointers. Will fix.

noclip-code:
Gentle ping.

axboe:
Looks good to me!

PR #1953, opened 2025-07-29
fix: prevent null pointer dereference in run_dedupe_threads

lhywk:
### Describe

Fixes a potential null pointer dereference in `run_dedupe_threads()` by adding a null check before calling `fio_sem_remove()`.
### Expected behavior
If `fio_sem_init()` fails (returns `NULL`), the program should skip calling `fio_sem_remove()` to avoid dereferencing a null pointer.
### Actual behavior
When `fio_sem_init()` fails and returns `NULL`, the code still calls `fio_sem_remove(size_lock)`, which dereferences the pointer without validation, leading to undefined behavior or a crash.

This is a minimal and safe fix that prevents a crash without altering the original logic. It improves code robustness by explicitly guarding against a potential null pointer, which aligns with good defensive programming practices.
Thanks for reviewing.

### DIFF
--- a/t/dedupe.c
+++ b/t/dedupe.c
@@ -457,6 +457,8 @@ static int run_dedupe_threads(struct fio_file *f, uint64_t dev_size,
 	total_items = dev_size / blocksize;
 	cur_offset = 0;
 	size_lock = fio_sem_init(FIO_SEM_UNLOCKED);
+	if (size_lock == NULL)
+		return 1;
 
 	threads = malloc(num_threads * sizeof(struct worker_thread));
 	for (i = 0; i < num_threads; i++) {


sitsofe:
Hi @lhywk:
Is this fix in response to a "this happened to me" problem or is this more of a fix to a theoretical issue? The reason I ask is because by my reading if `fio_sem_init()` fails you pretty much shouldn't carry on as you have no semaphore to share between the threads/processes that are spawned after the `fio_sem_init()` line.

lhywk:
Thanks @sitsofe
this issue was identified as a theoretical risk rather than something I encountered in practice.
Given that `fio_sem_init()` failure means no semaphore is available for coordinating the threads, would it make more sense to return early from `run_dedupe_threads()` instead of just guarding the cleanup?
I originally added the null check to prevent a potential crash, but I'm happy to adjust the fix if early exit is the more appropriate behavior here.

sitsofe:
@lhywk:
The bigger question is what is the effort needed (both for you and us) versus reward. The fio project doesn't try and recover from all possible errors and the nature of the project **isn't** a critical internet facing service where crashes on error paths are not allowed. From my perspective if fio is in a situation such as being out of memory, I'm OK with it dereferencing `NULL` and crashing as opposed to making the code more complex so as to do a graceful exit. Further, you are fixing up a test tool (`dedupe.c` is in the `t/` directory) so this fix isn't going to impact the main fio tool most users use (the `dedupe` test tool may not even be packaged up by distro vendors). Given all the above, this isn't what I would have wanted your first contribution to be but still...

Back to your actual question, yes I'd prefer an early return. I did a check and it look like all the errors generated by paths from `fio_sem_init()` already log their errors so I think the following is enough:
- Check if size_lock is NULL
- Return with a non-zero code (but what value to pick?)

As a one off we can accept such a change but if you have time it would provide more value to both you and us for you to tackle one of the triaged/"patches welcome" issues: https://github.com/axboe/fio/issues?q=state%3Aopen%20label%3Atriaged%20OR%20label%3Apatcheswelcome after this :-).

sitsofe:
I would also change the commit title to be:
```
t/dedupe: fix NULL dereference when fio_sem_init() fails
```

lhywk:
@sitsofe Updated the patch to use an early return with `ENOMEM` if `fio_sem_init()` fails.  
Let me know if you’d prefer a different return value.

sitsofe:
@lhywk:
Technically `fio_sem_init()` could fail for reasons other than memory right (although I agree that would be the most likely scenario)?

Please squash these commits together into one commit (I'd recommend using doing an interactive rebase) and update the commit message of the remaining commit so it's just:
```
t/dedupe: fix NULL dereference when fio_sem_init() fails

fio_sem_init() may return NULL if allocation or initialization fails.
Add a null check before calling fio_sem_remove to prevent potential
NULL pointer dereference.

<your signed off by here>
```

sitsofe:
@lhywk additionally you want to bail out as soon as you know that `fio_sem_init()` failed and before you do anything else that might make use of it so I think you need to early return around line 461.

lhywk:
@sitsofe Updated to early return if fio_sem_init() fails, before using size_lock. Thanks for the review.

axboe:
Guys, this makes no sense. First of all, how on earth would fio_sem_init() ever return NULL in there? Answer is it won't. And secondly, if it was an issue, then why not fix at least both spots in that file where fio_sem_init() is called?

This is unnecessary churn.

sitsofe:
@lhywk my apologies - I should have pushed back rather than making work for you but hopefully the practice was still useful.

@axboe the main way I could see this `fio_sem_init()` returning NULL is due to something like `mmap()` failing due to running out of memory. Given the anonymous mmap is tiny I admit this is a very unlikely scenario (and we're in all sort of other trouble). Your point asking why all the call sites weren't fixed is a fair point (I didn't check beyond the PR diff) and I understand your overall reasoning that this is unnecessary. I guess my hope was that this would be a small "gateway" fix to more meaningful changes in the future...

axboe:
In my experience, risk of random breakages for changes like this are way higher than any perceived benefit. And if we'd need to care about malloc/mmap returning NULL, then we'd need a sweeping project to cover everything in fio, not just a single random spot in a tool that's not even part of the fio main executable. Hence I really don't think it's worth wasting time on.

PR #1954, opened 2025-07-30
Sprandom

tomas-winkler-sndk:
Sprandom (SanDisk Pseudo Random) is a method developed to rapidly
precondition large SSDs by leveraging knowledge of physical
over-provisioning (OP). It uses fio to recreate the steady-state OP
distribution of a 100% random write workload by performing
invalidations at region-specific rates during writes.

This approach aims to:

1. Fully populate the L2P table by writing every LBA once.
2. Achieve a steady-state OP distribution across physical media.
3. Prevent physical contiguity in L2P entries, avoiding compressibility.
4. Limit the physical scatter of logically adjacent addresses.

Sprandom accomplishes the first three goals with a single physical
write pass, improving preconditioning efficiency and fidelity.

The patch series suggest a way to integrate sprandom into FIO,
as discussed here: https://github.com/axboe/fio/discussions/1935

Signed-off-by:  Tomas Winkler <tomas.winkle@sandisk.com>


HOWTO.rst
```diff
@@ -1632,6 +1632,24 @@ I/O type
 	space exceeds 2^32 blocks. If it does, then **tausworthe64** is
 	selected automatically.
 
+.. option:: sprandom=bool
+
+        SPRandom is a method designed to rapidly precondition large SSDs for
+        random write workloads. It divides the storage into equally sized regions
+        and calculates the required invalidation rate per region.
+        SPRandom utilizes random writes with **lfrs** distribution.
```
vincentkfu, line 1640: lfsr

sprandom.c
```diff
@@ -0,0 +1,383 @@
...
+{
+	struct buf_output out;
+	int i;
+
+	buf_output_init(&out);
+
+	log_buf(&out, "[");
+	for (i = 0; i < len - 1; i++) {
+		log_buf(&out, "%.2f, ", darray[i]);
+	}
```
vincentkfu, line 40: The coding style we use removes these curly braces when they are not necessary

sprandom.c
```diff
@@ -0,0 +1,383 @@
...
+
+cleanup:
+	free(x_orig);
+	free(y_orig);
+	free(new_x_arr);
+
+	return ret;
+}
+
+/* compute write amplification */
```
vincentkfu, line 262: This comment does not add anything.

sprandom.c
```diff
@@ -0,0 +1,383 @@
...
+{
+	return 0.5 / over_provisioning + 0.7;
+}
+
+static inline double compute_validity(double waf)
+{
+	return 1.0 - (double)1.0 / waf;
+}
+
+static double *compute_validity_dist(unsigned int n_regions, double over_provisioning)
```
vincentkfu, line 273: It seems like this function is carrying out calculations using symbols that are discussed in a paper. However, readers of this code do not have access to the background research from which the code originates. Please write a comment summarizing and justifying the calculations carried out here. For an example of what I am looking for take a look at the long comment in https://github.com/axboe/fio/blob/master/stat.h

filesetup.c
```diff
@@ -1400,6 +1401,16 @@ int setup_files(struct thread_data *td)
 			goto err_out;
 	}
 
+	if (td->o.sprandom) {
+		if (td->o.nr_files != 1) {
+			 log_err("SPrandom supports only one file");
```
vincentkfu, line 1406: `fio: SPrandom supports only one file`

init.c
```diff
@@ -690,6 +690,17 @@ static int fixup_options(struct thread_data *td)
 	if (o->zone_mode == ZONE_MODE_STRIDED && !o->zone_range)
 		o->zone_range = o->zone_size;
 
+	/* SPRandom requires LFSR and disables randommap to enable invalidation rewrites */
+	if (o->sprandom) {
+		if (td_write(td) && td_random(td)) {
+			o->random_generator = FIO_RAND_GEN_LFSR;
```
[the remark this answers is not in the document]
vincentkfu: Should fio abort with an error if the user specifies a 50/50 r/w workload?

HOWTO.rst
tomas-winkler-sndk: Sure, will rephrase.

sprandom.c
tomas-winkler-sndk: it spells the acronym.

sprandom.c
```diff
@@ -0,0 +1,383 @@
...
+	free(y_orig);
+	free(new_x_arr);
+
+	return ret;
+}
+
+/* compute write amplification */
+static inline double compute_waf(double over_provisioning)
+{
+	return 0.5 / over_provisioning + 0.7;
```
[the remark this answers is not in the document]
tomas-winkler-sndk: I guess we will collect more empirical evidence as we go.

pcbuf.h
```diff
@@ -0,0 +1,211 @@
...
+ */
+static inline void pcb_commit(struct pc_buf *cb)
+{
+	cb->commit_head = cb->staging_head;
+}
+
+/**
+ * pcb_pop - Pop a value from the committed buffer.
+ * @cb: pointer to the pc_buf structure.
+ * @out: pinter to the variable to store the popped value.
```
vincentkfu, line 119: pointer

sprandom.c
```diff
@@ -375,9 +381,197 @@ int sprandom_setup(struct sprandom_info *spr_info, uint64_t logical_size, uint64
...
+	dprint(FD_SPRANDOM, "Invalidation[%d] %ld %ld %.04f %d\n",
+		info->current_region,
+		info->region_write_count,
+		info->invalid_count[phase],
+		(double)info->invalid_count[phase] / (double)info->region_write_count,
+		info->invalid_pct[info->current_region]);
+
+}
+
+int sprandom_get_next_offset(struct sprandom_info *info, struct fio_file *f, uint64_t *b)
```
vincentkfu, line 466: I think this function deserves a summary explanation at the top. Something along the lines of:
- This generates offsets to write a region and saves a fraction of the offsets in a ring buffer
- The next region starts by writing all of the saved offsets to achieve the desired fraction of invalid blocks in the previous region. Then the remainder of the current region is filled with new offsets.

vincentkfu:
Also please resolve the build failures identified by our automated tests.

vincentkfu:
I think the approach taken here is still overly influenced by the job file approach described in the OCP presentation which has each region overwrite offsets written in the immediately previous region.

The two-phase circular buffer seems well done but it is a considerable amount of additional code (that will have to be maintained in perpetuity) with only one user. Its use must be very advantageous to justify including it.

I have a simpler alternative that as far as I can tell accomplishes the same result:
- Have each region and its invalidating writes be self contained
- Write each region with offsets from the LFSR to fill (1 - invalidation_fraction) * region_size
- While writing the above, save invalid_fraction of the offsets
- For the remainder of the region, issue writes from the saved list, repeating offsets if necessary

As far as I can tell this produces the same outcome as the two-phase ring buffer approach but is simpler and does not require a special data structure.

steven-sprouse-sndk:
> I think the approach taken here is still overly influenced by the job file approach described in the OCP presentation which has each region overwrite offsets written in the immediately previous region.
> 
> The two-phase circular buffer seems well done but it is a considerable amount of additional code (that will have to be maintained in perpetuity) with only one user. Its use must be very advantageous to justify including it.
> 
> I have a simpler alternative that as far as I can tell accomplishes the same result:
> 
> * Have each region and its invalidating writes be self contained
> * Write each region with offsets from the LFSR to fill (1 - invalidation_fraction) * region_size
> * While writing the above, save invalid_fraction of the offsets
> * For the remainder of the region, issue writes from the saved list, repeating offsets if necessary
> 
> As far as I can tell this produces the same outcome as the two-phase ring buffer approach but is simpler and does not require a special data structure.

We discussed a similar approach, but there were a couple concerns:
1. When writing the "remainder of the region..." we do not want to create a sub-region that has 100% validity. In other words we need to ensure that all writes in a region have the same probability of becoming invalid. 
2. We could create invalidation writes within the 2nd sub-region, but at some point the invalidations of a specific address would come close enough together in time that some drives might be able to do invalidation in their write caches. This might be over estimating how much of an effect this would have but that is one reason we landed on the current approach.

I think if you can think of a way to address (1) and convince ourselves that (2) is not a significant issue, then we can look at this other approach.

vincentkfu:
> > I think the approach taken here is still overly influenced by the job file approach described in the OCP presentation which has each region overwrite offsets written in the immediately previous region.
> > The two-phase circular buffer seems well done but it is a considerable amount of additional code (that will have to be maintained in perpetuity) with only one user. Its use must be very advantageous to justify including it.
> > I have a simpler alternative that as far as I can tell accomplishes the same result:
> > 
> > * Have each region and its invalidating writes be self contained
> > * Write each region with offsets from the LFSR to fill (1 - invalidation_fraction) * region_size
> > * While writing the above, save invalid_fraction of the offsets
> > * For the remainder of the region, issue writes from the saved list, repeating offsets if necessary
> > 
> > As far as I can tell this produces the same outcome as the two-phase ring buffer approach but is simpler and does not require a special data structure.
> 
> We discussed a similar approach, but there were a couple concerns:
> 
> 1. When writing the "remainder of the region..." we do not want to create a sub-region that has 100% validity. In other words we need to ensure that all writes in a region have the same probability of becoming invalid.
> 2. We could create invalidation writes within the 2nd sub-region, but at some point the invalidations of a specific address would come close enough together in time that some drives might be able to do invalidation in their write caches. This might be over estimating how much of an effect this would have but that is one reason we landed on the current approach.
> 
> I think if you can think of a way to address (1) and convince ourselves that (2) is not a significant issue, then we can look at this other approach.

For item 1 it would be straightforward to sprinkle invalidating writes throughout the entire region instead of waiting until near the end. However, this would likely exacerbate item 2 especially for the region with the the highest number of invalid blocks. I cannot think of a way to resolve item 2.

So I think this concern is sufficient to justify including the new ring buffer code. Please include this justification in the comment summarizing this feature.

Finally, please add some unit tests for the two-phase ring buffer. There are examples in https://github.com/axboe/fio/tree/master/unittests

examples/sprandom.fio
```diff
@@ -0,0 +1,32 @@
...
+rw=randwrite
+bs=${FIO_BS}
+blockalign=${FIO_BS}
+direct=1
+norandommap=1
+iodepth=64
+[preconditioning]
+sprandom=1
+spr_op=0.15
+spr_num_regions=100
```
sitsofe, line 32: Could we have a newline at the end of this file?

steven-sprouse-sndk:
@vincentkfu There are still a few commits we'd like to make before this is merged. Can you provide feedback on the fixes that @tomas-winkler-sndk  checked in over the weekend?

examples/sprandom.fio
```diff
@@ -0,0 +1,33 @@
...
+;
+[global]
+ioengine=libaio
+rw=randwrite
+bs=${FIO_BS}
+blockalign=${FIO_BS}
+direct=1
+norandommap=1
+iodepth=64
+[preconditioning]
```
steven-sprouse-sndk, line 29: @tomas-winkler-sndk the HOWTO.rst states that `random_generator` needs to be set to `lfsr`, but `random_generator` is not set in this example.

HOWTO.rst
```diff
@@ -1632,6 +1632,24 @@ I/O type
 	space exceeds 2^32 blocks. If it does, then **tausworthe64** is
 	selected automatically.
 
+.. option:: sprandom=bool
+
+        SPRandom is a method designed to rapidly precondition large SSDs for
+        random write workloads. It divides the storage into equally sized regions
+        and calculates the required invalidation rate per region.
+        SPRandom requires **random_generator** to be set to **lfsr**.
+
```
vincentkfu, line 1641: This should be written to communicate to a user who knows nothing about what this feature does.

How about this?

SPRandom is a method designed to rapidly precondition SSDs for
steady state random write workloads. It divides the device into equally sized regions
and writes the device's entire physical capacity one time selecting offsets so that the regions have a distribution of invalid blocks matching the distribution that occurs at steady state. Default: false.

It doesn't require `random_generator=lfsr` since this will be set for the user if it is unset.

sprandom.c
```diff
@@ -0,0 +1,556 @@
...
+	struct point *new_points_arr = (struct point *)0;
+	unsigned int i;
+	int ret = 0;
+
+	if (points == NULL || resampled_points == NULL)
+		return -EINVAL;
+
+
+	if (num == 0) {
+		log_err("Original points array cannot be empty.\n");
```
vincentkfu, line 318: fio: Original points array...

sprandom.c
```diff
@@ -0,0 +1,556 @@
...
+#include <math.h>
+#include "fio.h"
+#include "file.h"
+#include "sprandom.h"
+
+/*
+ * Model for Estimating Steady-State Data Distribution in SSDs
+ *
+ * This model estimates the distribution of valid data across a flash drive
+ * in a steady state. It is based on the key insight from Desnoyers' research,
```
vincentkfu, line 18: Please add the Desnoyers citation to this comment.

sprandom.c
```diff
@@ -0,0 +1,383 @@
...
+	return ret;
+}
+
+/* compute write amplification */
+static inline double compute_waf(double over_provisioning)
+{
+	return 0.5 / over_provisioning + 0.7;
+}
+
+static inline double compute_validity(double waf)
```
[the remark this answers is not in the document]
vincentkfu: Can you rename this to something like `compute_gc_validity` or `compute_min_validity` so that it does not give the impression that it provides a general validity fraction. It would also be helpful to specify that it returns a proportion.

sprandom.c
```diff
@@ -0,0 +1,556 @@
...
+	}
+
+	acc = 0.0;
+	for (i = 0; i < n_regions; i++) {
+		acc_ratio[i] = acc + blocks_ratio[i];
+		acc = acc_ratio[i];
+	}
+
+	print_d_array("validity_distribution", validity_distribution, n_regions);
+	print_d_array("blocks ration", blocks_ratio, n_regions);
```
vincentkfu, line 476: ratio

sprandom.c
```diff
@@ -0,0 +1,556 @@
...
+ * Return: resampled and reversed validity distribution array or NULL on error.
+ */
+static double *compute_validity_dist(unsigned int n_regions, double over_provisioning)
+{
+	double waf = compute_waf(over_provisioning);
+	double validity = compute_validity(waf);
+	double *validity_distribution = NULL;
+	double *blocks_ratio = NULL;
+	double *acc_ratio = NULL;
+	double acc;
```
vincentkfu, line 440: Can you explain what `blocks_ratio` and `acc_ratio` will be used for? It seems like they are involved in invalidating blocks from previous regions but it is not obvious how they are used.

sprandom.c
```diff
@@ -0,0 +1,556 @@
...
+
+	points = p_alloc(n_regions);
+
+	for (i = 0; i < n_regions; i++) {
+		points[i].x = acc_ratio[i];
+		points[i].y = validity_distribution[i];
+	}
+	print_d_points(points, n_regions);
+
+	ret = sample_curve_equally_on_x(points, n_regions, n_regions,
```
vincentkfu, line 487: Why is resampling needed? Does `linspace()` not produce evenly space values?

sprandom.c
```diff
@@ -0,0 +1,556 @@
...
+	if (new_points_arr == NULL) {
+		log_err("Memory allocation failed for new_points_arr.\n");
+		ret = -ENOMEM;
+		goto cleanup;
+	}
+
+	/* 6. Perform linear interpolation for each new_x to get new_y */
+	for (i = 0; i < num_resampled; i++) {
+		new_points_arr[i].x = new_x_arr[i];
+		new_points_arr[i].y = linear_interp(new_x_arr[i], x_orig, y_orig, num);
```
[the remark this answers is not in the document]
steven-sprouse-sndk: @vincentkfu This approach stems from the way we are interpreting equation that describes the distribution of OP. We construct the curve based on  uniform intervals on the y-axis, but those points are not uniform on the x-axis. The resampling and interpolation is done to find uniform intervals on the x-axis.

malventano:
> * Write each region with offsets from the LFSR to fill (1 - invalidation_fraction) * region_size

> As far as I can tell this produces the same outcome as the two-phase ring buffer approach but is simpler and does not require a special data structure

I believe @vincentkfu was onto something here. Since LFSR is used for the writing of each region, and LFSR is deterministic, there should not be any need to create a memory structure to save off all writes for each region. The previous region's LFSR can simply be played back and the desired percentage of invalidations can be issued in the exact same sequence and with the same addresses as they would have if the memory structure were used.

Issue #1957, opened 2025-08-04
"make" install files to read-only file system on macOS


Kubo326:
**Please acknowledge the following before creating a ticket**

- [ YES ] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
"make" install files to read-only file system.
The issue appears to stem from the use of these paths.
mandir = /usr/share/man
sharedir = /usr/share/fio

**Environment**: macOS Catalina 10.15.7

**fio version**: 3.40

**Reproduction steps**
% ./configure
% make
% sudo make install
install -m 755 -d /usr/local/bin
install fio t/fio-genzipf  t/fio-dedupe t/fio-verify-state ./tools/fio_generate_plots ./tools/plot/fio2gnuplot ./tools/genfio ./tools/fiologparser.py ./tools/hist/fiologparser_hist.py ./tools/hist/fio-histo-log-pctiles.py ./tools/fio_jsonplus_clat2csv /usr/local/bin
install -m 755 -d /usr/share/man/man1
install: chmod 755 /usr/share/man/man1: Read-only file system
install -m 644 ./fio.1 /usr/share/man/man1
install: /usr/share/man/man1/fio.1: Read-only file system
make: *** [install] Error 71

sitsofe:
Hello @Kubo326:
This is not a bug in fio (as fio runs on lots of different platforms) but rather a decision by macOS to mark certain directories off limits.

Generally speaking, if you want a command line program to install itself to a directory other than its hardcoded default (which is your case) it is up to you to tell it so (usually at configure time). Here's a snippet from fio's `./configure --help`:
```
$ ./configure --help
--prefix=               Use this directory as installation prefix
[...]
```

Here's a snippet from from macOS 15.5's `hier` man page:
```
HIER(7)                                               Miscellaneous Information Manual                                              HIER(7)

NAME
     hier – layout of filesystems

DESCRIPTION
     A historical sketch of the filesystem hierarchy.  The modern OS X filesystem is documented in the “File System Programming Guide”
     available on Apple Developer.
[...]
     /usr/         contains the majority of user utilities and applications

[...]
                   local/    executables, libraries, etc. not included by the basic operating system
[...]
```

You can also find other recommendations by doing a [google search for "prefix for macos locally built programs"](https://www.google.com/search?q=prefix+for+macos+locally+built+programs).

It may also be worth your while checking what macOS package managers do with fio:
[...]

PR #1960, opened 2025-08-10
Makefile: fix man and share install paths on macOS

Kubo326:
Fix issue #1957

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -484,14 +484,8 @@ endif
 prefix = $(INSTALL_PREFIX)
 bindir = $(prefix)/bin
 libdir = $(prefix)/lib/fio
-
-ifeq ($(CONFIG_TARGET_OS), Darwin)
-mandir = /usr/share/man
-sharedir = /usr/share/fio
-else
-mandir = $(prefix)/man
+mandir = $(prefix)/share/man
 sharedir = $(prefix)/share/fio
-endif
 
 all: $(PROGS) $(T_TEST_PROGS) $(UT_PROGS) $(SCRIPTS) $(ENGS_OBJS) FORCE
 


sitsofe:
@Kubo326 Looks good! My only recommendation is to reword the commit message:
```
Makefile: fix man and share install paths on macOS

Unlike when bcdf7c5 ("Install man page in /usr/share/man by default on OSX")
was made, recent macOS has /usr/local/share/man in the man path by default.
Further, since macOS 10.11 /usr/share is a read-only directory due to SIP which 
means a default macOS fio install will fail.

Fix the above by changing the macOS install to use $(prefix) for man pages and
shared files the same way other OSes do.

Fixes: https://github.com/axboe/fio/issues/1957

<your signed off by here>
```

That way:
- You make it clear that fix was for macOS
- We teach GitHub to close your issue when your PR is merged

You can fix the message by using
```
git commit --amend
```

and you will be prompted to re-edit it. After you've got it looking the way you want you can replace your existing commit in this pull request by using something like
```
git push origin fix-install-paths --force-with-lease
```

(the exact command you will use may not be exactly that but you get the idea)

sitsofe:
@vincentkfu, @axboe this look good to me.

vincentkfu:
For platforms other than macOS this changes the default `mandir` from `/usr/local/man` to `/usr/local/share/man`. I understand that one of these is typically a symlink to the other but could this change cause problems?

On my Mac `/usr/local/man` does not exist but `/usr/local/share/man` does exist.

axboe:
/usr/local/share/man/ is the right path, so don't think this is much of a problem. The non-share one is more of a relic.

PR #1962, opened 2025-08-12
Add --print_defaults option to display default configuration values

adajani:
This adds a new command line option --print_defaults that outputs all
key FIO default parameter values and exits. This is useful for researchers
who want to understand what defaults FIO assumes when no explicit
configuration is provided.

Signed-off-by: Ahmed Dajani <adajain@iastate.edu>

### DIFF
--- a/init.c
+++ b/init.c
@@ -301,6 +301,11 @@ static struct option l_opts[FIO_NR_OPTIONS] = {
 		.val		= 'A' | FIO_CLIENT_FLAG,
 	},
 	{
+		.name		= (char *) "print_defaults",
+		.has_arg	= no_argument,
+		.val		= 'Z' | FIO_CLIENT_FLAG,
+	},
+	{
 		.name		= NULL,
 	},
 };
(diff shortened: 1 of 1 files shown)


vincentkfu:
The list of options printed out is far from complete.

Is the documentation insufficient?

https://fio.readthedocs.io/en/latest/fio_doc.html#job-description

What about just looking at the value for the `default` field in `struct fio_option`?

https://github.com/axboe/fio/blob/fef641096d925d737d81683626eca52b6e358e52/options.c#L1917

adajani:
The list shows only most used parameters in litritures. The documentation is great and contains everything, but it would be nice to have print default feature at runtime.

axboe:
```fio --cmdhelp=<option>``` already provides this information, and it should be complete as far as the options have a default option.

adajani:
I agree "--cmdhelp" exist, but it didn't show the default ioengine, which motive me to add the print defaults feature.

axboe:
> I agree "--cmdhelp" exist, but it didn't show the default ioengine, which motive me to add the print defaults feature.

It definitely does show that...

axboe:
Closing this one out - ALL options that have a default value, will list it in cmdhelp. This includes ioengine, of course. I don't think there's a need for a separate option for this, it's just unwieldy and doesn't even cover all the options. What may be a primary set of options for you will differ for others.

PR #1972, opened 2025-08-22
fixing missing dependencies

Meiye-lj:
This PR fixes an issue in the Makefile. Specifically, previously, any modifications of files like lib/types.h would not trigger a rebuild of t/verify-state.o. The PR fixes this by including them as additional dependencies..


axboe:
Your commit message is missing:

- An explanation of WHY this change is necessary
- Signed-off-by

Please see other commits in the tree for the expected formatting of commits.

sitsofe (changes requested):

axboe:
Just a heads up, please squash to a single commit before asking for this to be merged again.

axboe:
Oh, and fix the commit message to have some actual why's and reasonings in there, and the signed-off-by as previously mentioned.

PR #1974, opened 2025-08-24
fixing missing dependencies

Meiye-lj:
Fixing missing dependencies

This PR fixes an issue in the Makefile. Specifically, previously, any modifications of files like lib/types.h would not trigger a rebuild of t/verify-state.o. The PR fixes this by including them as additional dependencies. Mainly, T_OBJS and UT_OBJS do not use .d files to record dependencies correctly, unlike OBJS.
The previous PR had to be reopened due to an error during the merge of commits.

Signed-off-by: Jun Lyu lvjun_dnt@outlook.com

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -319,7 +319,7 @@ FIO_OBJS += lex.yy.o y.tab.o
 GFIO_OBJS += lex.yy.o y.tab.o
 endif
 
--include $(OBJS:.o=.d)
+-include $(OBJS:.o=.d) $(T_OBJS:.o=.d) $(UT_OBJS:.o=.d)
 
 T_SMALLOC_OBJS = t/stest.o
 T_SMALLOC_OBJS += gettime.o fio_sem.o pshared.o smalloc.o t/log.o t/debug.o \


sitsofe (changes requested):
@Meiye-lj:

FYI: you can use an interactive git rebase coupled with force pushes to rewrite branches rather than needing to close an old PR and open a new one.

This looks a lot neater. Unfortunately you need to rewrite your commit message because there's currently no Signed-off-by and you also you are missing the section as part of the commit title. I would recommend something like the following (but obviously fill out the template sections):

```
Makefile: fix missing test tool and unit test dependencies

Signed-off-by: Firstname Lastname <user@domain.tld>
```

PS: Your Signed-off-by should include your real name in addition to your email address

Meiye-lj:
> @Meiye-lj:
> 
> FYI: you can use an interactive git rebase coupled with force pushes to rewrite branches rather than needing to close an old PR and open a new one.
> 
> This looks a lot neater. Unfortunately you need to rewrite your commit message because there's currently no Signed-off-by and you also you are missing the section as part of the commit title. I would recommend something like the following (but obviously fill out the template sections):
> 
> ```
> Makefile: fix missing test tool and unit test dependencies
> 
> Signed-off-by: Firstname Lastname <user@domain.tld>
> ```
> 
> PS: Your Signed-off-by should include your real name in addition to your email address

Thanks and updated

PR #1980, opened 2025-09-10
backend: get io_u from requeue if exists

JeongJonghwi:
If the `iodepth=1`, which means only 1 io_u can alive, and also if it is in `td->io_u_requeues`, the `td->io_u_freelist` is empty. With `end_fsync=1` in this, `__get_io_u()` will return `NULL`, because there are no free io_u in `td->io_u_freelist` and `td->runstate` is `FSYNCING`. This leads that the sync operation will be failed with message `end_fsync failed for file <filename>`.

This patch fixes that the io_u's in `td->io_u_requeues` will be returned if `td->runstate == TD_FSYNCING`, and will not be set the `IO_U_F_NO_FILE_PUT` flag, which causes the file's reference count never goes to 0.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -269,7 +269,8 @@ static bool fio_io_sync(struct thread_data *td, struct fio_file *f)
 
 	io_u->ddir = DDIR_SYNC;
 	io_u->file = f;
-	io_u_set(td, io_u, IO_U_F_NO_FILE_PUT);
+	if (td->runstate != TD_FSYNCING)
+		io_u_set(td, io_u, IO_U_F_NO_FILE_PUT);
 
 	if (td_io_prep(td, io_u)) {
 		put_io_u(td, io_u);
--- a/io_u.c
+++ b/io_u.c
@@ -1701,7 +1701,7 @@ struct io_u *__get_io_u(struct thread_data *td)
 		__td_io_u_lock(td);
 
 again:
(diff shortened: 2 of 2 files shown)


vincentkfu:
Please edit the commit message to add the Fixes tag below above your sign-off:

Fixes: 1c83ed15 ("io_u: get io_u from io_u_freelist when TD_FSYNCING")

backend.c
```diff
@@ -269,7 +269,8 @@ static bool fio_io_sync(struct thread_data *td, struct fio_file *f)
 
 	io_u->ddir = DDIR_SYNC;
 	io_u->file = f;
-	io_u_set(td, io_u, IO_U_F_NO_FILE_PUT);
```
vincentkfu, line 272: Will there be a problem here with reference counts if the job is using multiple files and the `io_u` that is in the requeue list refers to a file that is different from the one that is the target of the sync operation?

vincentkfu:
I communicated offline with Jonghwi and suggested cleaning up the file reference in the requeued `io_u` before using it for fsync.

Jonghwi, please rephrase the second paragraph to something like:

```
This patch fixes the issue by obtaining an io_u from the requeue list even when fio is fsyncing.
However, when fio is fysncing, this patch makes sure to decrement the file reference counter
for the file associated with the io_u since it is possible that the fsync will target a different file.
The larger context of this patch is to ensure that file references are correctly counted so that
fio can close all files on exit.
```

PR #1983, opened 2025-09-14
tests: add sprandom test cases

tomas-winkler-sndk:
Add 5 simple sprandom tests. These currently verify that the feature
does not break under normal use and properly handles invalid arguments.


sitsofe:
@tomas-winkler-sndk :
I just happened to do a build on an old Linux install with
```
LDFLAGS="-m32" ./configure --disable-native --extra-cflags="-m32 -march=i386" && make
```
and it generated the following warnings:
```
In file included from fio.h:23:0,
                 from sprandom.c:10:
sprandom.c: In function ‘print_d_array’:
log.h:24:12: warning: the comparison will always evaluate as ‘true’ for the address of ‘out’ will never be NULL [-Waddress]
  if ((buf) != NULL)    \
            ^
sprandom.c:123:2: note: in expansion of macro ‘log_buf’
  log_buf(&out, "[");
  ^~~~~~~
log.h:24:12: warning: the comparison will always evaluate as ‘true’ for the address of ‘out’ will never be NULL [-Waddress]
  if ((buf) != NULL)    \
            ^
sprandom.c:125:3: note: in expansion of macro ‘log_buf’
   log_buf(&out, "%.2f, ", darray[i]);
   ^~~~~~~
log.h:24:12: warning: the comparison will always evaluate as ‘true’ for the address of ‘out’ will never be NULL [-Waddress]
  if ((buf) != NULL)    \
            ^
sprandom.c:127:2: note: in expansion of macro ‘log_buf’
  log_buf(&out, "%.2f]\n", darray[len - 1]);
  ^~~~~~~
sprandom.c: In function ‘print_d_points’:
log.h:24:12: warning: the comparison will always evaluate as ‘true’ for the address of ‘out’ will never be NULL [-Waddress]
  if ((buf) != NULL)    \
            ^
sprandom.c:142:2: note: in expansion of macro ‘log_buf’
  log_buf(&out, "[");
  ^~~~~~~
log.h:24:12: warning: the comparison will always evaluate as ‘true’ for the address of ‘out’ will never be NULL [-Waddress]
  if ((buf) != NULL)    \
            ^
sprandom.c:144:3: note: in expansion of macro ‘log_buf’
   log_buf(&out, "(%.2f %.2f), ", parray[i].x, parray[i].y);
   ^~~~~~~
log.h:24:12: warning: the comparison will always evaluate as ‘true’ for the address of ‘out’ will never be NULL [-Waddress]
  if ((buf) != NULL)    \
            ^
sprandom.c:146:2: note: in expansion of macro ‘log_buf’
  log_buf(&out, "(%.2f %.2f)]\n", parray[len - 1].x, parray[len - 1].y);
  ^~~~~~~
```

Any ideas?

vincentkfu (commented):
Let's just drop this patch? It does not seem related to sprandom. What is liat? There should be an explanation for the changes as well.

tomas-winkler-sndk:
> Let's just drop this patch? It does not seem related to sprandom. What is liat? There should be an explanation for the changes as well.
Dropped, that leftover patch.

PR #1992, opened 2025-10-01
engines/io_uring: validate PI on read

stanleyzhang9:
Currently, the call to validate metadata at the end of fio_ioring_event is unreachable.
Move the call within the success path, and clean up indentation by removing a redundant if statement.


vincentkfu:
Thank you for catching this. `pi_act` is actually checked later on, so it is not necessary to have it here.
We should have some negative error cases in `t/io_uring_pi.py` but I could not come up with a way implement them.

stanleyzhang9:
Removed the unnecessary pi_act check as you pointed out.

PR #1997, opened 2025-10-07
write_hist_log: do not require ZLIB for a CLI client

### LINKED ISSUE #1996: Enabling --write_hist_log=foo fails if ZLIB not enabled
- [ x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
Running a local build of fio (no client/server setup), and attempting to use --write_hist_log=, we get a failure due to this check in init.c (l.1771 at HEAD):

```
#ifndef CONFIG_ZLIB
		if (td->client_type) {
			log_err("fio: --write_hist_log requires zlib in client/server mode\n");
			goto err;
		}
#endif
```

client_type is always non-zero since FIO_CLIENT_TYPE_CLI is 1, _GUI is 2.

**Environment**: Linux version 6.6.64, x86

**fio version**: fio-3.36.0

**Reproduction steps**
```
# cat > ./aio.fio
[global]
direct=1
iodepth=1
thread=1
ioengine=libaio
rw=randread
blocksize=4k
size=1G
numjobs=1
time_based=1
group_reporting=1
ramp_time=2
runtime=2
new_group

[nvme0uq1]
filename=/dev/nvme0n1

# ./fio --write_hist_log=foo --log_hist_msec=100 ./aio.fio
fio: --write_hist_log requires zlib in client/server mode
```

egouriou-google:
write_hist_log: do not require ZLIB for a CLI client

Since `td->client_type` is always non-zero (CLI is 1, GUI is 2), the check `if (td->client_type)` in `init.c` always triggers, causing --write_hist_log to be unusable when ZLIB is not configured.

This patch proposes to replace the check with
```
if (td->client_type != FIO_CLIENT_TYPE_CLI) { ...
```
I am not 100% clear whether this is the right approach to check whether we are NOT in a client-server setup.
Fixes #1996

Signed-off-by: Eric Gouriou <egouriou@google.com>

### DIFF
--- a/init.c
+++ b/init.c
@@ -1769,7 +1769,7 @@ static int add_job(struct thread_data *td, const char *jobname, int job_add_num,
 		const char *suf;
 
 #ifndef CONFIG_ZLIB
-		if (td->client_type) {
+		if (td->client_type != FIO_CLIENT_TYPE_CLI) {
 			log_err("fio: --write_hist_log requires zlib in client/server mode\n");
 			goto err;
 		}


egouriou-google:
I am very fuzzy on fio setups in client-server mode, and how they differ from local-only runs. Naively `FIO_CLIENT_TYPE_CLI` and `FIO_CLIENT_TYPE_GUI` could both refer to client-server mode, just with different types of remote clients. Since `parse_options` appears to set `type` to CLI (https://github.com/axboe/fio/blob/master/init.c#L3176), my best bet is that this denotes a local run. I'd be happy to be corrected.

axboe:
> I am very fuzzy on fio setups in client-server mode, and how they differ from local-only runs. Naively `FIO_CLIENT_TYPE_CLI` and `FIO_CLIENT_TYPE_GUI` could both refer to client-server mode, just with different types of remote clients. Since `parse_options` appears to set `type` to CLI (https://github.com/axboe/fio/blob/master/init.c#L3176), my best bet is that this denotes a local run. I'd be happy to be corrected.

Yeah this looks odd. There are basically two types of CLI clients - standalone run, or a cli client connected to a backend. The latter needs zlib. So I don't think the patch is correct, it should should be type == CLI && remote client. Either that or a new FIO_CLIENT_TYPE_STANDALONE should be added.

egouriou-google:
Closed as resolved by https://github.com/axboe/fio/pull/2033 .

PR #1999, opened 2025-10-08
Remove leftover code for platforms lacking clock_gettime and clockid_t

betonmischer86:
Drop the remaining workaround code for platforms lacking clock_gettime and clockid_t. Produce an error on configure if clock_gettime is unavailable.

Submitted on behalf of @sitsofe.


sitsofe (approved):
@vincentkfu, @axboe this LGTM (and it saves a configure probe).

Issue #1988, opened 2025-09-24
Error compiling with CUDA 13.0


tertumh:
Hello.

### Description of the bug:
I compile FIO with the following parameter

`./configure --enable-cuda --extra-cflags="-I$CUDA_HOME/include"`

Compilation stops at the following errors in the code with CUDA

```c
memory.c: In function ‘alloc_mem_cudamalloc’:
memory.c:253:40: warning: passing argument 2 of ‘cuCtxCreate_v4’ makes pointer from integer without a cast [-Wint-conversion]
  253 |         ret = cuCtxCreate(&td->cu_ctx, CU_CTX_MAP_HOST, td->cu_dev);
      |                                        ^~~~~~~~~~~~~~~
      |                                        |
      |                                        int
In file included from fio.h:69,
                 from memory.c:9:
/usr/local/cuda/include/cuda.h:6275:66: note: expected ‘CUctxCreateParams *’ {aka ‘struct CUctxCreateParams_st *’} but argument is of type ‘int’
 6275 | CUresult CUDAAPI cuCtxCreate(CUcontext *pctx, CUctxCreateParams *ctxCreateParams, unsigned int flags, CUdevice dev);
      |                                               ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
/usr/local/cuda/include/cuda.h:90:45: error: too few arguments to function ‘cuCtxCreate_v4’
   90 | #define cuCtxCreate                         cuCtxCreate_v4
      |                                             ^~~~~~~~~~~~~~
memory.c:253:15: note: in expansion of macro ‘cuCtxCreate’
  253 |         ret = cuCtxCreate(&td->cu_ctx, CU_CTX_MAP_HOST, td->cu_dev);
      |               ^~~~~~~~~~~
/usr/local/cuda/include/cuda.h:90:45: note: declared here
   90 | #define cuCtxCreate                         cuCtxCreate_v4
      |                                             ^~~~~~~~~~~~~~
/usr/local/cuda/include/cuda.h:6275:18: note: in expansion of macro ‘cuCtxCreate’
 6275 | CUresult CUDAAPI cuCtxCreate(CUcontext *pctx, CUctxCreateParams *ctxCreateParams, unsigned int flags, CUdevice dev);
      |                  ^~~~~~~~~~~
make: *** [Makefile:499: memory.o] Error 1
``` 

### Environment
OS : Ubuntu 24.04.3, 
Kernel: 6.8.0-84-generic
[...]

PR #2000, opened 2025-10-09
memory: add cuda 13 support 

maxime-peim:
In the latest cuda driver API, cuCtxCreate takes another parameter CUctxCreateParams.

### DIFF
--- a/configure
+++ b/configure
@@ -180,6 +180,7 @@ libhdfs="no"
 devdax="no"
 pmem="no"
 cuda="no"
+cuda13="no"
 libcufile="no"
 disable_lex=""
 disable_pmem="no"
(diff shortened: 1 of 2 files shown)


vincentkfu:
This will fail if the build environment has a CUDA version before 13.0. We should have a test in the configure script that detects whether this additional parameter is required and uses the new function signature only if it is required.

Here is an example of a similar change: https://github.com/axboe/fio/commit/f95cf49e1d2a8d2babb8aa385fb902d927533644

Also please add to the bottom of the commit message:

Fixes: [1988](https://github.com/axboe/fio/issues/1988)

sitsofe:
@maxime-peim Could you also change the commit's title to something like
```
memory: add cuda 13 support 
```

maxime-peim:
Sure, let me add these

axboe:
That looks better!

PR #2001, opened 2025-10-14
options: add support more POSIX errnos with negative support

minwooim:
Patch series to support masking expected system error with `--ignore_error=`.

### DIFF
--- a/io_u.c
+++ b/io_u.c
@@ -2196,7 +2196,16 @@ static void io_completed(struct thread_data *td, struct io_u **io_u_ptr,
 		}
 	} else if (io_u->error) {
 error:
-		icd->error = io_u->error;
+		if (io_u->flags & IO_U_F_DEVICE_ERROR)
+			icd->error = io_u->error;
+		else {
+			/*
+			 * If @io_u has system error(non-device erorr), it must
+			 * have been negative value.
+			 */
+			icd->error = -io_u->error;
+		}
+
 		io_u_log_error(td, io_u);
 	}
 	if (icd->error) {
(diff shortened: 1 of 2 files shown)


vincentkfu (commented):
This change concerns me because it changes the default behavior for error handling. Most of the time `io_u->error` will be a system error since only `io_uring_cmd` sets `IO_U_F_DEVICE_ERROR`. So this will flip the sign of the error in most cases. This will break jobs that depend on fio's current behavior.

Why can't you just use `--ignore_error=EINVAL` to mask `EINVAL` instead of relying on this patch and using `--ignore_error=-EINVAL`?

minwooim:
The reason I wanted to mask with -EINVAL was exactly what you mentioned: with io_uring_cmd, the NVMe CQ entry’s status code can currently be masked through io_u->error. But you’re right — flipping the error-handling sign for all other cases just to accommodate a single ioengine shouldn’t happen.

> Why can't you just use `--ignore_error=EINVAL` to mask `EINVAL` instead of relying on this patch and using `--ignore_error=-EINVAL`?

The problem is that if IO_U_F_DEVICE_ERROR is set and io_u->error contains the value 22 (0x16), we can’t tell whether that represents EINVAL (22) or an NVMe SGL Offset Invalid error (sct=0x0, sc=0x16).

vincentkfu:
Would you like to split this PR into two pieces? One patch adding more errno values can go in with no problems, but a second one to distinguish between errno values and device errors will require more thought.

minwooim:
> > Why can't you just use `--ignore_error=EINVAL` to mask `EINVAL` instead of relying on this patch and using `--ignore_error=-EINVAL`?
> 
> The problem is that if IO_U_F_DEVICE_ERROR is set and io_u->error contains the value 22 (0x16), we can’t tell whether that represents EINVAL (22) or an NVMe SGL Offset Invalid error (sct=0x0, sc=0x16).

Will do that, @vincentkfu , Thanks for the review.

minwooim:
#2071 is the first PR to go based on the review comment.

PR #2007, opened 2025-10-29
backend.c: fix async ioengine prep/queue direction mismatch in experimental_verify

### LINKED ISSUE #1955: When using libaio/io_uring engine and executing experimental_verify FIO generates writes instead of reads

**Description of the bug:**
When executing verify_only with experimental_verify, FIO generates writes to the device/file.

**Environment**: Linux using libaio engine

**fio version**: 3.33, but it might be valid for master

**Reproduction steps**
Run FIO with verify_only together with experimental_verify. We expect FIO to perform only READ operations for data verification. However, we are observing only WRITE operations. The issue might be related to the experimental_verify procedure: when it changes the I/O direction (from WRITE to READ), it may not trigger the prep io function in the I/O engine

**FIO config**
```
[global] 
ioengine=libaio 
iodepth=32 
direct=1 
readwrite=randwrite 
filename=/dev/vda 
serialize_overlap=1 
experimental_verify=1 
do_verify=1 
verify=md5 
group_reporting 
verify_only 
norandommap

[job] 
```

yehormalikov:
When using --experimental_verify=1 with async ioengines like libaio, there was a critical mismatch between the I/O direction set during prep and the direction used when queuing operations.

The issue occurred because get_io_u() calls td_io_prep() which configures ioengine structures (like iocb) for WRITE operations, but then io_u->ddir is changed from WRITE to READ for verification after prep completes. This left the iocb configured for writes while fio's accounting expected reads, causing incorrect I/O statistics and potential data integrity issues.

Fix by calling td_io_prep() again after changing the data direction in both TRIM and WRITE cases, ensuring async ioengines re-sync their internal structures with the final I/O direction.

Fixes #1955

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -666,12 +666,20 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
 				} else if (io_u->ddir == DDIR_TRIM) {
 					io_u->ddir = DDIR_READ;
 					io_u_set(td, io_u, IO_U_F_TRIMMED);
+					if (td_io_prep(td, io_u)) {
+						put_io_u(td, io_u);
+						continue;
+					}
 					break;
 				} else if (io_u->ddir == DDIR_WRITE) {
 					io_u->ddir = DDIR_READ;
 					io_u->numberio = td->verify_read_issues;
 					td->verify_read_issues++;
 					populate_verify_io_u(td, io_u);
+					if (td_io_prep(td, io_u)) {
+						put_io_u(td, io_u);
+						continue;
+					}
 					break;
 				} else {
 					put_io_u(td, io_u);


vincentkfu:
To the commit message please add

```
Fixes: https://github.com/axboe/fio/issues/1955
```

vincentkfu:
Quickly scanning through the `prep` hooks for all of the built-in ioengines suggests that repeated calls to `prep` do no harm and indeed will resolve the issue in question.

The only problem I found is with the `libhdfs` ioengine which selects file open flags based on data direction. This pattern already may break experimental_verify, so this patch does no harm.

sitsofe:
This looks like a far more elegant solution than the one I was suggesting. The logs will be a little confusing because you'll see multiple preps of the same I/O but it keeps all the extra checks and work out of the main I/O path which is a win. My only note is: @yehormalikov that it looks like your committer address and author address are different - is this intentional?

yehormalikov:
> This looks like a far more elegant solution than the one I was suggesting. The logs will be a little confusing because you'll see multiple preps of the same I/O but it keeps all the extra checks and work out of the main I/O path which is a win. My only note is: @yehormalikov that it looks like your committer address and author address are different - is this intentional?

Thank you for attention. Have already changed committer address

PR #2010, opened 2025-10-31
man: update HOWTO and README links

jwilk:
Fixes: 7c1f6a2f8837 ("docs: rename HOWTO to HOWTO.rst")
Fixes: 79e1e2802e1c ("docs: rename README to README.rst")

### DIFF
--- a/fio.1
+++ b/fio.1
@@ -5294,6 +5294,6 @@ Sample jobfiles are available in the `examples/' directory.
 .br
 These are typically located under `/usr/share/doc/fio'.
 .P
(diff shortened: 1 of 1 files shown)


vincentkfu:
How about `https`?

axboe:
These should point at the git.kernel.org links instead now, and yes use https:// as well.

axboe:
There are a bunch more, I just did all of them.

jwilk:
There's still one `git.kernel.dk` reference left in `engines/ftruncate.c`.

axboe:
Thanks, fixed.

Issue #1618, opened 2023-08-27
ki/Mi/Gi/etc parsed incorrectly when kb_base=1000


kotarou3:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
When (byte) sizes are specified with the ki/Mi/Gi/etc units, they are parsed as 10³B/10⁶B/10⁹B when `kb_base=1000` rather than the expected 2¹⁰B/2²⁰B/2³⁰B.
Notably, if I change the units to kiB/MiB/GiB instead, then they are parsed correctly as 2¹⁰B/2²⁰B/2³⁰B

It seems like ki/Mi/Gi are being parsed as if they didn't have the "i" suffix at all?

**Environment**:
```
$ uname -a
Linux home.kota.moe 6.4.0-2-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.4.4-3 (2023-08-08) x86_64 GNU/Linux
$ lsb_release -d
Description:	Debian GNU/Linux trixie/sid
```

**fio version**: `fio-3.35`

**Reproduction steps**
Blocksize incorrectly parsed as 4000B
```
[global]
kb_base=1000
size=1Mi
blocksize=4ki
rw=randread

[read-4ki-qd1]
iodepth=1
```
`read-4ki-qd1: (g=0): rw=randread, bs=(R) 4000B-4000B, (W) 4000B-4000B, (T) 4000B-4000B, ioengine=psync, iodepth=1`

Blocksize correctly parsed as 4096B:
```
[global]
kb_base=1000
size=1MiB
blocksize=4kiB
rw=randread

[read-4ki-qd1]
iodepth=1
```
`read-4ki-qd1: (g=0): rw=randread, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1`

RobertTheBrucey:
I can see in parse.c:217 that only the cases of (for example k) explicitly accounted for are "kib", "k" and "kb", and it looks like "ki" would be interpreted as "k".
This should be a fairly straightforward fix, and I may try my hand at a pull request for it.

sitsofe:
This is kind of a duplicate of #1396 ...

RobertTheBrucey:
[...]

PR #2011, opened 2025-11-05
Fix incorrect unit prefix examples in manpage

hexiaole1994:
The examples in the manpage for kb_base=1000 and kb_base=1024 had multiple errors in unit prefixes and values:

For kb_base=1000 (IEC/SI standard mode):
- Fix 4 KiB example: use IEC prefix 'ki' instead of 'k' (4k -> 4ki to correctly represent 4096 bytes)
- Fix 1 MiB example: use IEC prefix 'mi' instead of 'm' (1m -> 1mi to correctly represent 1048576 bytes)
- Fix 1 MB example: use SI prefix 'm' instead of 'mi' (1mi -> 1m to correctly represent 1000000 bytes)
- Fix 1 TiB example: use IEC prefix 'ti' instead of 't' (1t -> 1ti to correctly represent 1099511627776 bytes)
- Fix 1 TiB value: 1073741824 -> 1099511627776
- Fix 1 TB example: use SI prefix 't' instead of 'ti' (1ti -> 1t to correctly represent 1000000000000 bytes)
- Fix 1 TB value: 1000000000 -> 1000000000000
- Fix 1 TB prefixes: 1000mi -> 1000g, 1000000ki -> 1000000m

For kb_base=1024 (compatibility mode):
- Fix 1 TiB value: 1073741824 -> 1099511627776
- Fix 1 TiB prefixes: 1024m -> 1024g, 1048576k -> 1048576m (to correctly represent 1099511627776 bytes)
- Fix 1 TB value: 1000000000 -> 1000000000000
- Fix 1 TB prefixes: 1000mi -> 1000gi, 1000000ki -> 1000000mi (to correctly represent 1000000000000 bytes)

All examples now correctly match the actual behavior implemented in parse.c.

Fixes https://github.com/axboe/fio/issues/1618
Related PR https://github.com/axboe/fio/pull/1827


sitsofe:
I myself use `bs=4k` to mean `bs=4096` in fio's default mode as it's convenient even though it could be seen as ambiguous so I'm not sure that should change. Fio also has unexpected behaviour with its units:
```
% rm -f /tmp/fio.tmp; ./fio --minimal --filename /tmp/fio.tmp --kb_base=1024 --bs=4ki --size=4ki --name=go > /dev/null; stat -f'%z' /tmp/fio.tmp
4096
% rm -f /tmp/fio.tmp; ./fio --minimal --filename /tmp/fio.tmp --kb_base=1000 --bs=4ki --size=4ki --name=go > /dev/null; stat -f'%z' /tmp/fio.tmp
4000
```

So care needs to be taken if we are going to update that section. Due to the previous point I don't know if the HOWTO would need updating too.

Issue #2013, opened 2025-11-10
Fio tries to verify whole block device instead of only blocks written and saved in state file


StefVdHaute:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**

In fio 3.40 when executing the following 2 jobs in sequence the second job stops after having verified all the offsets present in the state file.

```ini
[test]
name=writetest
filename=/dev/nvme0n1
rw=randwrite
runtime=20
time_based
verify=crc32c
verify_state_save=1
do_verify=0
```

```ini
[test]
name=writetest
filename=/dev/nvme0n1
rw=randwrite
verify=crc32c
verify_state_load=1
verify_only=1
```

In fio 3.41 the verify job continuous verifying past the written blocks and tries to verify the whole block device.
This of course fails*.

\* This might be unrelated but also seems problematic, writing the whole disk using a similar write-job ensures running the 2 jobs again (with runtime limit) succeeds.

Write whole block device:

```ini
[fill]
name=fill
filename=/dev/nvme0n1
rw=write
verify=crc32c
verify_state_save=1
do_verify=0
```

**Environment**:
- EndeavourOS (arch-based), kernel 6.12.57-1-lts
- Ubuntu 22.04.05 LTS

**fio version**: 3.41

**Reproduction steps**

Execute the following jobs in sequence

```ini
[test]
name=writetest
filename=/dev/nvme0n1
rw=randwrite
runtime=20
time_based
verify=crc32c
verify_state_save=1
do_verify=0
```

```ini
[test]
name=writetest
filename=/dev/nvme0n1
rw=randwrite
verify=crc32c
verify_state_load=1
verify_only=1
```

vincentkfu:
The verify state file feature was overhauled in 3.41 via https://github.com/axboe/fio/pull/1951

@noclip-code would you take a look at this issue?

minwooim:
In offline verification, io_hist_tree is not saved to the state file, so io_u->numberio is always zero in the second job (the verify phase). To address this, do_dry_run() simulates the saved WRITE job and logs io_u entries into io_hist_tree via log_io_piece(). The issue is that io_u->numberio is always zero when passed to log_io_piece().

The following patch works correctly with the two provided job files.

```diff
diff --git a/backend.c b/backend.c
index 684ea0de3bac..e09c210aab57 100644
--- a/backend.c
[...]

PR #2015, opened 2025-11-17
backend: fix zero-numberio in dry-run

minwooim:
In offline verification, io_hist_tree is not saved to the state file, so `io_u->numberio` is always zero in the second job (the verify phase). To address this, `do_dry_run()` simulates the saved WRITE job and logs io_u entries into io_hist_tree via `log_io_piece()`. The issue is that `io_u->numberio` is always zero when passed to log_io_piece().

To reproduce this issue, run the following jobs in turn.

	[test]
	name=writetest
	filename=/dev/nvme0n1
	ioengine=io_uring
	direct=1
	bs=512K
	iodepth=128
	rw=randwrite
	runtime=20
	time_based
	verify=crc32c
	verify_state_save=1
	do_verify=0

	[test]
	name=writetest
	filename=/dev/nvme0n1
	ioengine=io_uring
	direct=1
	bs=512K
	iodepth=128
	rw=randwrite
	verify=crc32c
	verify_state_load=1
	verify_only=1
	verify_dump=1

Fixes: #2013

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -1703,8 +1703,11 @@ static uint64_t do_dry_run(struct thread_data *td)
 		io_u_set(td, io_u, IO_U_F_FLIGHT);
 		io_u->error = 0;
 		io_u->resid = 0;
-		if (ddir_rw(acct_ddir(io_u)))
+		if (ddir_rw(acct_ddir(io_u))) {
+			io_u->numberio = td->io_issues[acct_ddir(io_u)];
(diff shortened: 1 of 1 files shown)


vincentkfu:
Thanks. In fio 3.40 the stopping criterion did not depend on the io_u's `numberio`, but this was changed in 3.41 so that fio now depends on the io_u's `numberio` to know when to stop. So we need to properly set `numberio` when we save metadata in `io_hist_tree`.

PR #2030, opened 2025-12-16
io_uring: don't segfault if pi_chk isn't specified

calebsander:
If `fio` is run with `md_per_io_size` but no `pi_chk` argument, `fio_ioring_io_u_init()` passes `NULL` to `strstr()`, which segfaults. Use the `struct ioring_options` `prchk` flags instead, which have already been parsed from `pi_chk` in `parse_prchk_flags()`.

### DIFF
--- a/engines/io_uring.c
+++ b/engines/io_uring.c
@@ -1593,11 +1593,11 @@ static int fio_ioring_io_u_init(struct thread_data *td, struct io_u *io_u)
 		pi_attr->len = o->md_per_io_size;
 		pi_attr->app_tag = o->apptag;
 		pi_attr->flags = 0;
-		if (strstr(o->pi_chk, "GUARD") != NULL)
+		if (o->prchk & NVME_IO_PRINFO_PRCHK_GUARD)
 			pi_attr->flags |= IO_INTEGRITY_CHK_GUARD;
-		if (strstr(o->pi_chk, "REFTAG") != NULL)
+		if (o->prchk & NVME_IO_PRINFO_PRCHK_REF)
 			pi_attr->flags |= IO_INTEGRITY_CHK_REFTAG;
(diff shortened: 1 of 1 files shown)


vincentkfu:
Can you share the job that triggers this issue?

I think the change is good, actually, but am having trouble understanding how the issue occurs.

`parse_prchk_flags()` relies on `pi_chk` being non-NULL in order to set bits in `prchk`. If checking `pi_chk` fails in `fio_ioring_io_u_init()` how could it have succeeded in `parse_prchk_flags()`?

It might be reasonable also to just change the default for `pi_chk` to an empty string.

calebsander:
The segfault can be reproduced with `--ioengine io_uring` and `--md_per_io_size` but no `--pi_chk`:
```
$ fio --name md --ioengine io_uring --md_per_io_size 1024 --filename /dev/nvme*n*
md: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=io_uring, iodepth=1
fio-3.41
Starting 1 process
fio: pid=138727, got signal=11


Disk stats (read/write):
  ublkb0: ios=0/0, sectors=0/0, merge=0/0, ticks=0/0, in_queue=0, util=0.00%
free(): double free detected in tcache 2
Aborted (core dumped)
```

> parse_prchk_flags() relies on pi_chk being non-NULL in order to set bits in prchk. If checking pi_chk fails in fio_ioring_io_u_init() how could it have succeeded in parse_prchk_flags()?

I'm not sure what you mean. `parse_prchk_flags()` always succeeds, there is no failure case. If `pi_chk` is omitted (`!o->pi_chk`), `o->prchk` will just stay as 0.

vincentkfu:
Ok I see now. Thanks for the fix. I could not trigger the issue with `io_uring_cmd` but it does appear with `io_uring`.

PR #2034, opened 2026-01-10
windows: fix a Y2038 bug by replacing `Int32x32To64` with multiplication

CookiePLMonster:
`Int32x32To64` macro internally truncates the arguments to int32, while `time_t` is 64-bit on most/all modern platforms. Therefore, usage of this macro creates a Year 2038 bug.

I detailed this issue a while ago in a writeup, and spotted the same issue in this repository when updating the list of affected repositories:
<https://cookieplmonster.github.io/2022/02/17/year-2038-problem/>

### DIFF
--- a/os/windows/posix.c
+++ b/os/windows/posix.c
@@ -297,7 +297,7 @@ void Time_tToSystemTime(time_t dosTime, SYSTEMTIME *systemTime)
 	LONGLONG jan1970;
 	SYSTEMTIME tempSystemTime;
 
-	jan1970 = Int32x32To64(dosTime, 10000000) + 116444736000000000;
+	jan1970 = (dosTime * 10000000LL) + 116444736000000000LL;
 	utcFT.dwLowDateTime = (DWORD)jan1970;
 	utcFT.dwHighDateTime = jan1970 >> 32;
 


sitsofe:
@CookiePLMonster Is `time_t` 64 bit on 32 bit Windows Cygwin/MSYS2. I'm assuming it is on 64 bit Windows Cygwin/MSYS2 (but it's worth checking) and there's an argument there's no supported 32 bit Windows after Windows 10 went EOL so perhaps the question is moot... The file in question (`os/windows/posix.c`) contains shims purely for Windows so I'd suggest a commit title like:
```
windows: fix Y2038 bug caused by 32 bit truncation
``` 
because the body of your commit message explains the issue well.

CookiePLMonster:
> Is `time_t` 64 bit on 32 bit Windows Cygwin/MSYS2.

In the majority of cases yes, even on 32-bit Windows it's 64-bit for probably over a decade. But even on platforms where it's 32-bit, this change is OK as the product is 64-bit anyway.



> because the body of your commit message explains the issue well.

Should I reword it or will you do it when merging with a rebase? Both are fine by me.

axboe:
We never rebase, merge or otherwise. So yes would be great if you could just amend your commit and force push it.

CookiePLMonster:
Done.

sitsofe:
(@CookiePLMonster Thank you for reaching out to so many OSS projects with fixes for this issue. And wow: you're also the [PC GTA San Andreas Skimmer fixer](https://cookieplmonster.github.io/2025/04/23/gta-san-andreas-win11-24h2-bug/)!)

Issue #1499, opened 2023-01-29
Reports write I/Os in job stats and logs for -verify_only even though no writes were issued


horshack-dpreview:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
The -verify_only flags allows a verify-only operation even if a write workload is specified. This is expected and described in the manual: "Do not perform specified workload, only verify data still matches previous invocation of this workload." However, after completing a -verify_only run, fio reports that it issued writes for the test, the number of which match the number of writes that would've been issued had fio actually wrote the data (it didn't).

**Environment**:
Ubuntu 22.04.1 LTS

**fio version**: 
3.33-68

**Reproduction steps**
`fio -bs=64K -rw=write -size=10M -filename=/tmp/test.bin -name="test" -verify=xxhash`
then
`fio -bs=64K -rw=write -size=10M -filename=/tmp/test.bin -name="test" -verify=xxhash -verify_only`

Verify-only reports:
`issued rwts: total=160,160,0,0 short=0,0,0,0 dropped=0,0,0,0`

If I enable a log such as bandwidth then it too reports data points implying writes were issued even though they weren't. Here is a sample bandwidth log, run on an actual device and with -size=1GB so that the test runs long enough to yield log samples:
```
610, 1718977, 1, 0, 0
1250, 0, 1, 0, 0
1999, 0, 1, 0, 0
2612, 0, 1, 0, 0
```

vincentkfu:
This seems like mainly a reporting issue. With `--verify_only=1` fio goes through the motions of issuing write operations in order to set everything up for verification.

In backend.c we have:

```
		if (td->o.verify_only && td_write(td))
			verify_bytes = do_dry_run(td);
		else {
			do_io(td, bytes_done);

			if (!ddir_rw_sum(bytes_done)) {
				fio_mark_td_terminate(td);
				verify_bytes = 0;
			} else {
				verify_bytes = bytes_done[DDIR_WRITE] +
						bytes_done[DDIR_TRIM];
			}
		}
```

`do_dry_run()` is:

```
/*
 * Dry run to compute correct state of numberio for verification.
 */
static uint64_t do_dry_run(struct thread_data *td)
{
	td_set_runstate(td, TD_RUNNING);

	while ((td->o.read_iolog_file && !flist_empty(&td->io_log_list)) ||
[...]

PR #2037, opened 2026-01-13
stat: Avoid printing "issued rwts" when "verify_only" is supplied

gautammenghani:
Avoid printing "issues rwts" when user has supplied "verify_only" option, since the IOs are not submitted.

Fixes: #1499


vincentkfu:
What is the harm in having the `issued rwts` line for `verify_only` workloads?

I think if anything we should leave the output alone and just update the `verify_only` documentation to note that any writes reported in the output to be issued are phantom writes.

gautammenghani:
sure, thanks for the feedback. I've updated the man page and the howto doc explaining this

Issue #2043, opened 2026-01-21
SPRandom Behavior Modification To Avoid Cache Invalidation When Cache Size Is Defined


cachyyyk:
I have read the GitHub issues section of [REPORTING-BUGS](https://github.com/axboe/fio/blob/master/REPORTING-BUGS).

In using SPRandom to pre-condition an SSD we noticed that the number of 100% valid blocks was significantly higher than expected when using the examples/sprandom.fio file with spr_num_regions set to 100 or larger.  In looking at the implementation I believe the invalidation writes are sent for a region right after all writes for that region have completed.  If the distance between a region LBA write and the 2nd write to that region LBA to invalidate the previous location is less than the SSD cache size then the invalidation could affect the cache having no affect on the blocks outside the cache.  This can convert the valid percent for a region to a much higher validity creating significant error in the validity distribution for the SSD.

Suggesting the following changes to address this potential issue:
1. Define a new SPRandom variable to allow for an SSD cache size(spr_cs) to be defined.
2. Preserve original SPRandom invalidation behavior if spr_cs is zero(default)
3. If spr_cs is non-zero then region X invalidations are done after region X+1 writes are completed instead of before.  This creates an additional region size of distance between the writes and associated invalidations.
4. Return an error if the defined cache size is greater than the region size and tell the user the maximum number of regions supported with the defined cache size.

In making the above changes the SPRandom expected validity vs the drive validity match better for the higher regions.

Will add associated PR soon to this issue for better discussion.

cachyyyk:
fio version: 3.41
Should be system agnostic but can provide more detail if needed.  Was seen running on Linux Distribution.

cachyyyk:
Adding original implementer for discussion.

@tomas-winkler-sndk @steven-sprouse-sndk

tomas-winkler-sndk:
We're reviewing the proposal in more details, but in general it looks reasonable.

cachyyyk:
> We're reviewing the proposal in more details, but in general it looks reasonable.

@tomas-winkler-sndk Any update on the review of these changes?  Any additional information needed to help with the review?

PR #2044, opened 2026-01-21
SPRandom Cache Size Behavior Implementation

cachyyyk:
SPRandom Cache Size Behavior Implementation

1. Define a new variable spr_cs to set SPRandom Cache Size
2. Preserve original SPRandom invalidation behavior if spr_cs is 
zero(default).
3. If spr_cs is non-zero then region X invalidations are done 
after region X+1 writes are completed instead of before.
This creates an additional region size of distance between
the writes and associated invalidations.
4. Return an error if the defined cache size is greater than 
the region size and tell the user the maximum number of 
regions supported with the defined cache size

Attempting to implement https://github.com/axboe/fio/issues/2043


vincentkfu:
@steven-sprouse-sndk

bbusacker:
@axboe Any chance we can get someone to review this?  The Micron team would like to get it merged if everything looks good.

axboe:
Well, still waiting on @steven-sprouse-sndk to take a look at this. Maybe @tomas-winkler-sndk can help out?

Outside of that, you need to have a single commit in your list of commits, you have for some reason merged master back a few times. So please rebase on master. Secondly, your commit should line break at 72 chars or so, the lines are way too long. Third, and this happens all the time for some reason, people confuse the github PR message for the git commit message. Signed-off-by etc should go in the _git_ commit message, it doesn't mean anything in here.

sprandom.c
```diff
@@ -584,17 +585,21 @@ static int sprandom_setup(struct sprandom_info *spr_info, uint64_t logical_size,
 
 	validity_dist = compute_validity_dist(spr_info->num_regions,
 					      spr_info->over_provisioning);
-	if (!validity_dist)
-		return -ENOMEM;
+	if (!validity_dist) {
```
axboe: Same here

axboe:
Please sort out the review comments I posted. As a bonus, this also makes your main change easier to review. As a submitter, you should always make it as _easy_ as possible for people to review. That increases the likelihood that it does get reviewed, and in turn, applied.

fio.1
```diff
@@ -1469,6 +1469,19 @@ For large devices it is better to use more regions, to increase precision
 and reduce memory allocation. The allocation is proportional to the region size.
 .RE
 .TP
+.B spr_cs=int
+See
+.BR sprandom .
+Specifies the device cache size for SPRandom. Default=0
```
vincentkfu: The documentation should also mention that setting this to a non-zero value defers invalidating writes by one region. It would also make things clearer to explain that the actual cache size is used to ensure that the number of regions was not set too large that it resulted in regions smaller than the cache size.

I would rewrite this along these lines:

- When this is non-zero, delay invalidating writes by one region in order to make sure that the original write is committed to media before its later invalidating write is sent to the device. This deferral prevents the original write and the later invalidating write from being present in the device's cache at the same time (which would allow the device to ignore the original write and prevent sprandom from achieving its target validity fractions).
- The actual cache size is used to ensure that the number of regions is not set so large that the size of a region is smaller than the device cache.

axboe:
Don't have changes, and then a revert of some of those changes. If you want to address the whitespace issues, and that would indeed be great, do that first as a separate commit. Then you have a commit with your changes in it. Either that, or you just ignore the whitespace issues in the current documentation. In which case you'd want to squash those two commits into a single commit. And since the last commit is just fixing issues the main commit should've tackled, you should squash those too.

axboe:
Main commit also still has lines that are way too long, and it's missing a Signed-off-by. The other two do have a Signed-off-by, but no empty line between the commit message and the SOB line. Please use git log or similar in a terminal in the fio git repo to see the expected format.

axboe:
I was referring to the commit message... See previous comments on that. Please squash everything, it's turning into quite a mess at this point.

cachyyyk:
> Please sort out the review comments I posted. As a bonus, this also makes your main change easier to review. As a submitter, you should always make it as _easy_ as possible for people to review. That increases the likelihood that it does get reviewed, and in turn, applied.



> I was referring to the commit message... See previous comments on that. Please squash everything, it's turning into quite a mess at this point.

Thank you for the feedback.  I believe I addressed the review comments.  Sorry about the mess I haven't used git in some of the requested ways before.  So the 3 additional commits should now be squashed into the first, PR is rebased to master, and the only commit here should now have a proper SOB line.

axboe:
Getting closer! But as mentioned a few times, you commit _message_ needs wrapping at 72 chars, or it's unreadable in normal git log. Just go git commit --amend and re-wrap your commit message at 72 chars, then force push that commit. We still need to review the new commit, but at least then the meta data is sane.

sprandom.c
```diff
@@ -606,8 +611,25 @@ static int sprandom_setup(struct sprandom_info *spr_info, uint64_t logical_size,
...
+		log_err("fio: sprandom: spr_cs [%"PRIu64"] must be smaller than"
+			" region_sz [%"PRIu64"] which means [%"PRIu64"] regions"
+			" allowed", spr_info->cache_sz,
+		    region_sz,
+		    (physical_size / spr_info->cache_sz));
+		ret = -EINVAL;
+		goto err;
+	}
+
+	if (spr_info->cache_sz != 0) {
```
axboe: The fio style is generally:

```
if (spr_info->cache_sz) {
[...]
```

axboe:
Just some minor nits, nothing major. I would still like the sprandom folks to sign off on this addition, which is somewhat orthogonal to the actual code in question.

steven-sprouse-sndk:
@cachyyyk I understand the proposal and I think it will do what you want. Just to make sure I understand the implementation. You are not requiring any additional data structures, but just the timing of when the first list is replayed is delayed. So the net effect is there will be more memory in use when this new parameter is specified?

cachyyyk:
> @cachyyyk I understand the proposal and I think it will do what you want. Just to make sure I understand the implementation. You are not requiring any additional data structures, but just the timing of when the first list is replayed is delayed. So the net effect is there will be more memory in use when this new parameter is specified?

Correct.  When spr_cs is non-zero popping from the invalid_buf will not occur until the end of the next region.  So we need roughly 2x of the worst case invalidation size worth of space in this buffer to continue to add to the staging portion of the buffer until the commit portion of the buffer is reduced now at the end of the next region.  For a 32 TB drive, 100 regions, OP of 15% that would change the total allocation from 145.38 MiB to 290.75 MiB.

steven-sprouse-sndk:
@cachyyyk , @vincentkfu  I am good with this feature enhancement.

cachyyyk:
@axboe @steven-sprouse-sndk @vincentkfu Thank you for taking your time to review these changes.

bbusacker:
@axboe @steven-sprouse-sndk @vincentkfu  A shout out from the Micron team.  Great collaboration and engagement!  Thank you so much for your helpfulness as we dotted all the i's and crossed all the t's.  Appreciate you guys!

PR #2045, opened 2026-01-27
Add option to specify job process comm

robertbaldyga:
Some storage systems may exhibit different behaviors depending on the application running on top, identified by its process comm string. It can be useful for fio jobs to present themselves with a specified comm string to trigger those behaviors in a testing setup.


axboe:
You're going to need to make this dependent on a configure detection to not break non-linux.

Issue #2072, opened 2026-03-18
NULL pointer dereference in thread_main() when calling prctl(PR_SET_NAME, o->comm)


Criticayon:
**Please acknowledge the following before creating a ticket**

- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**
Running fio with a malformed or inconsistent job configuration may cause a NULL pointer dereference in thread_main() when calling:
```
prctl(PR_SET_NAME, o->comm);
```
If o->comm is not initialized (NULL), the prctl call eventually triggers a strncpy on a NULL source pointer, resulting in a segmentation fault. This can be reproduced when job initialization partially fails (e.g., invalid blocksize vs dataset range), but threads are still created. The issue is reproducible with AddressSanitizer enabled.

**Environment**: Ubuntu 24.04.3 LTS

**fio version**: fio-3.41

**Reproduction steps**
Create a job file poc.fio with the following content:
```
#
# Test for race-condition DIO-write vs punch_holy�# If race exist dio may rewrite punched block after
# it was allocated to another file, we will catch that, but the
# IO will not really be done and the write latency numbers w chosen randomly
; and writes directed at the wriill only reflect the
# open times.
[global]
create_on_open=1
nrfiles=30
ioengine=dircreate
fallocate=none
filesize=4k
openfiles=1

[tkip error dump
continue_on_error=write
ignore_error=,ENOSPC
error_dump=0
fallocate=none
exitall

# Two threads (dio and punch_hole) operate on single file:'raicer',
# We do not care about data content here
[dio-raicer]
bs=128k 
direct=1
buffered=0 
rw=randwrite
runtime=100
filename=raicer
time_based

[punch_hole-raicer]
bs=4k
rw=randtrim
filename=raicer

# Verifier thread continuously writes to newly allcated blocks
# and verifies written content
[aio-dio-verifier]
create_on_open=1
verify=crc32c-intel
verify_fatal=1
verify_dump=1
verify_backlog=1024
verify_async=4
direct=1
# block size should be equals to fs block size to prevent short writes
bs=4k
rw=randrw
filename=aio-dio-verifier


```
Run:
```
./fio poc.fio
```
Observed result:
<img width="978" height="613" alt="Image" src="https://github.com/user-attachments/assets/26c4cb20-7d27-4c29-af03-3bac2b12e770" />
```
[...]

PR #2073, opened 2026-03-18
backend: guard prctl(PR_SET_NAME) against NULL thread name

Criticayon:
o->comm may be NULL if job initialization fails or the job
structure is only partially initialized before thread creation.

Calling prctl(PR_SET_NAME, NULL) may lead to a NULL pointer
dereference inside strncpy().

Add a NULL check before calling prctl() to prevent the crash.

This issue was originally reported in #2072.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -1823,7 +1823,8 @@ static void *thread_main(void *data)
 	fio_local_clock_init();
 
 #ifdef CONFIG_LINUX
-	prctl(PR_SET_NAME, o->comm);
+	if (o->comm)
+		prctl(PR_SET_NAME, o->comm);
 #endif
 
 	dprint(FD_PROCESS, "jobs pid=%d started\n", (int) td->pid);


vincentkfu:
These are just style issues, but please:

- Use one tab to indent the `if`
- Increase the indent for `prctl` by one tab
- Also, we don't use `//` for comments in this repo. Either remove it or use `/* ... */`
- You could also add a `Fixes: <link to issue>` tag

backend.c
```diff
@@ -1823,6 +1823,7 @@ static void *thread_main(void *data)
 	fio_local_clock_init();
 
 #ifdef CONFIG_LINUX
+if (o->comm)//fix:Add a simple NULL check
```
axboe: What's the point of this comment? Please git rid of it, it reads like there's a fix missing.

Criticayon:
Thanks for the review! I've updated the patch to follow the repository style and added the Fixes tag.

axboe:
Please squash it to a single commit.

Criticayon:
Sorry for the delay. I've updated the PR.

PR #2074, opened 2026-03-22
Support write overlap numberio verification when io_size > size

minwooim:
When io_size > size, sequential writes wrap around and revisit every
block, so each offset ends up written more than once.  fio's
verify_write_sequence feature is supposed to detect the latest write
to each block (via io_hist) and confirm it on readback, but several
independent bugs prevented this from working end-to-end.

Patch 1 makes fio_offset_overlap_risk() return true when io_size > size,
activating the rb-tree backend for io_hist so overlapping io_piece
entries are detected and replaced correctly.

Patch 2 reverted back the previous commit since Patch 1 has led verify
table to be the rb-tree rather than the flist so no longer need to check
`verify` option.

Patch 3 removed considering `rw` and replaced it to `write` since `td_write`
is a superset of `td_rw` to support overlap by updating `total_bytes.`

Patch 4 ensures f->last_pos wraps back to 0 in verify_only=1
mode, so the dry-run write pass in the offline verify phase replays
the full io_size worth of writes rather than stopping at size.

Patch 5 resets nr_done_files when io_size > size so that
keep_running() does not exit the outer iteration loop after the
first pass — without this the second (overlapping) pass never runs.

Patch 6 adds a test suite covering online verification across all four
rw modes (write, randwrite, rw, randrw) with both async and sync engines,
plus a filesize < size edge case.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -1105,7 +1105,7 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
 		total_bytes = max(total_bytes, (uint64_t) td->o.io_size);
 
 	/* Don't break too early if io_size > size */
-	if (td_rw(td) && !td_random(td))
+	if (td_write(td) && !td_random(td))
 		total_bytes = max(total_bytes, (uint64_t)td->o.io_size);
 
 	/*
--- a/backend.c
+++ b/backend.c
@@ -1104,11 +1104,8 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
 	if (td_write(td) && td_random(td) && td->o.norandommap)
 		total_bytes = max(total_bytes, (uint64_t) td->o.io_size);
 
-	/*
-	 * Don't break too early if io_size > size. The exception is when
-	 * verify is enabled.
-	 */
-	if (td_rw(td) && !td_random(td) && td->o.verify == VERIFY_NONE)
+	/* Don't break too early if io_size > size */
+	if (td_rw(td) && !td_random(td))
 		total_bytes = max(total_bytes, (uint64_t)td->o.io_size);
 
 	/*
--- a/io_u.c
+++ b/io_u.c
@@ -382,8 +382,11 @@ static int get_next_seq_offset(struct thread_data *td, struct fio_file *f,
 	/*
 	 * If we reach the end for a rw-io-size based run, reset us back to 0
 	 * and invalidate the cache, if we need to.
+	 *
+	 * If it has possibility of overlapped offsets (io_size > size), we
+	 * should get back to 0 offset to verify overlapped offsets.
 	 */
-	if (td_rw(td) && o->io_size > o->size) {
+	if (td_write(td) && o->io_size > o->size) {
 		if (f->last_pos[ddir] >= f->io_size + get_start_offset(td, f)) {
 			f->last_pos[ddir] = f->file_offset;
 			loop_cache_invalidate(td, f);
--- a/libfio.c
+++ b/libfio.c
@@ -106,7 +106,8 @@ static void reset_io_counters(struct thread_data *td, int all)
 	/*
 	 * reset file done count if we are to start over
 	 */
-	if (td->o.time_based || td->loops > 1 || td->o.do_verify)
+	if (td->o.time_based || td->loops > 1 || td->o.do_verify ||
+			(td->o.io_size > td->o.size))
 		td->nr_done_files = 0;
 }
 
--- a/t/numberio_overlap.py
+++ b/t/numberio_overlap.py
@@ -1,0 +1,842 @@
+#!/usr/bin/env python3
+"""
+# numberio_overlap.py
+#
+# Verification tests for sequential and random write overlap with numberio.
+#
+# When io_size > size, writes wrap around and revisit every offset a second
+# time.  fio_offset_overlap_risk() must return true so that io_hist uses an
+# rb-tree and retains only the latest io_piece per offset.
+#
+# Each TC is run twice: once with the async engine and once with the sync
+# engine selected based on the platform (Linux: io_uring/psync,
+# Windows: windowsaio/sync, other: posixaio/psync).
+#
+# rw=write     offline  TC  1: basic 2× overlap
+#              online   TC  2: do_verify=1
+#
+# rw=randwrite offline  TC  3: norandommap
+#              online   TC  4: norandommap
+#
+# rw=rw        offline  TC  5
+#              online   TC  6
+#
+# rw=randrw    offline  TC  7: norandommap
+#              online   TC  8: norandommap
+#
+# filesize<size offline  TC  9: rb-tree triggered before first I/O
+#
+# nrfiles=2    offline  TC 10: write, 2× overlap across 2 files
+#              online   TC 11: write, do_verify=1 across 2 files
+#              offline  TC 12: randwrite, norandommap
+#              offline  TC 13: write, filesize < size/nrfiles
+#
+# USAGE
+# see python3 t/numberio_overlap.py --help
+#
+# EXAMPLES
+# python3 t/numberio_overlap.py --file /tmp/fio-overlap-test.dat
+# python3 t/numberio_overlap.py --file /tmp/fio-overlap-test.dat -f ./fio
+#
+# REQUIREMENTS
+# Python 3.6
+#
+"""
+import copy
+import os
+import platform
+import sys
+import json
+import time
+import locale
+import logging
+import argparse
+import subprocess
+from pathlib import Path
+from fiotestlib import FioJobCmdTest, run_fio_tests
+from fiotestcommon import SUCCESS_DEFAULT, Requirements
+
+
+class OfflineOverlapVerifyTest(FioJobCmdTest):
+    """
+    Two-phase offline overlap verification test.
+
+    Phase 1 — write:
+        rw=write, io_size = N * size  (N >= 2).
+        Every offset is written N times; the N-th pass has the highest numberio.
+        fio_offset_overlap_risk() must return true so that the rb-tree backend
+        is used and only the latest io_piece per offset is kept.
+
+    Phase 2 — verify:
+        verify_only=1, verify_state_load=1.
+        A dry-run write pass detects overlaps ("iolog: overlap") and advances
+        f->last_pos to wrap back to 0.  The subsequent READ phase issues exactly
+        (size / bs) reads — one per block — checking the latest numberio only.
+
+    Key assertions
+        - Write phase exits 0 and produces io_size bytes written.
+        - Verify phase exits 0 and reports 0 verify_errors.
+        - Verify phase reads exactly size bytes — not io_size bytes — confirming
+          that only the latest numberio is verified once per block.
+    """
+
+    def __init__(self, fio_path, success, testnum, artifact_root, fio_opts,
+                 basename=None):
+        super().__init__(fio_path, success, testnum, artifact_root, fio_opts,
+                         basename)
+        stub = os.path.join(self.paths['test_dir'],
+                            f"{self.basename}{self.testnum:03d}")
+        # write-phase output file (verify phase reuses self.filenames['output'])
+        self.filenames['output_write'] = os.path.abspath(f"{stub}.write.output")
+
+    def setup(self, parameters):
+        """Create test directory; args are built dynamically in run()."""
+        if not os.path.exists(self.paths['test_dir']):
+            os.mkdir(self.paths['test_dir'])
+        self.parameters = []    # satisfy FioExeTest.run() if it ever gets called
+
+    def _filename_str(self):
+        """Return the fio --filename value: colon-joined list for nrfiles > 1."""
+        base = self.fio_opts['filename']
+        nrfiles = self.fio_opts.get('nrfiles', 1)
+        if nrfiles > 1:
+            return ':'.join(f"{base}.{i}" for i in range(nrfiles))
+        return base
+
+    def _fio_args(self, phase):
+        """Return fio CLI args for *phase* ('write' or 'verify')."""
+
+        ioengine = self.fio_opts.get('ioengine', 'io_uring')
+
+        fallocate = "--fallocate=truncate"
+
+        common = [
+            "--name=overlap-verify",
+            f"--ioengine={ioengine}",
+            fallocate,
+        ]
+
+        rw = self.fio_opts.get('rw', 'write')
+
+        common += [
+            f"--filename={self._filename_str()}",
+            f"--bs={self.fio_opts['bs']}",
+            f"--size={self.fio_opts['size']}",
+            f"--io_size={self.fio_opts['io_size']}",
+            f"--verify={self.fio_opts['verify']}",
+            "--output-format=json",
+        ]
+
+        if self.fio_opts.get('norandommap'):
+            common.append("--norandommap=1")
+
+        if phase == 'write':
+            return common + [
+                f"--rw={rw}",
+                "--do_verify=0",
+                f"--output={self.filenames['output_write']}",
+            ]
+        else:
+            # Verify phase replays the same rw mode as the write phase so that
+            # do_dry_run produces an identical offset→numberio mapping.
+            # - randwrite/randrw with norandommap=1: no axmap limit so
+            #   do_dry_run runs to io_size.  Both phases use the same RNG
+            #   sequence because randrepeat=1 is fio's default (options.c
+            #   ".def = 1"), giving identical offset→numberio in io_hist.
+            # - rw/randrw (sequential/random mixed): IOs advance position the
+            #   same way in write and verify phases; only writes go to io_hist.
+            verify_io_size = self.fio_opts.get('verify_io_size',
+                                               self.fio_opts['io_size'])
+            verify_args = [
+                "--name=overlap-verify",
+                f"--ioengine={ioengine}",
+                fallocate,
+                f"--filename={self._filename_str()}",
+                f"--rw={rw}",
+                f"--bs={self.fio_opts['bs']}",
+                f"--size={self.fio_opts['size']}",
+                f"--io_size={verify_io_size}",
+                f"--verify={self.fio_opts['verify']}",
+                "--output-format=json",
+                "--do_verify=1",
+                "--verify_only=1",
+                "--verify_write_sequence=1",
+                f"--output={self.filenames['output']}",
+            ]
+            if self.fio_opts.get('norandommap'):
+                verify_args.append("--norandommap=1")
+            return verify_args
+
+    def _run_one(self, args, stdout_f, stderr_f, ec_f):
+        """Spawn fio with *args* in test_dir; return the Popen object."""
+        command = [self.paths['exe']] + args
+        with open(self.filenames['cmd'], "a",
+                  encoding=locale.getpreferredencoding()) as cmd_f:
+            cmd_f.write(" \\\n ".join(command) + "\n\n")
+        proc = subprocess.Popen(command,
+                                stdout=stdout_f,
+                                stderr=stderr_f,
+                                cwd=self.paths['test_dir'],
+                                universal_newlines=True)
+        proc.communicate(timeout=self.success['timeout'])
+        ec_f.write(f"{proc.returncode}\n")
+        logging.debug("Test %d phase %s: exit %d",
+                      self.testnum, args[-1], proc.returncode)
+        return proc
+
+    def run(self):
+        """Run write phase then verify phase sequentially."""
+
+        # If filesize is specified, pre-create each file at that size so that
+        # real_file_size < size/nrfiles, which is the scenario under test.
+        if 'filesize' in self.fio_opts:
+            base = self.fio_opts['filename']
+            nrfiles = self.fio_opts.get('nrfiles', 1)
+            fnames = [f"{base}.{i}" for i in range(nrfiles)] if nrfiles > 1 \
+                else [base]
+            for fname in fnames:
+                with open(fname, 'wb'):
+                    pass
+                os.truncate(fname, self.fio_opts['filesize'])
+
+        try:
+            with open(self.filenames['stdout'], "w",
+                      encoding=locale.getpreferredencoding()) as stdout_f, \
+                 open(self.filenames['stderr'], "w",
+                      encoding=locale.getpreferredencoding()) as stderr_f, \
+                 open(self.filenames['exitcode'], "w",
+                      encoding=locale.getpreferredencoding()) as ec_f:
+
+                proc_w = self._run_one(self._fio_args('write'),
+                                       stdout_f, stderr_f, ec_f)
+                if proc_w.returncode != 0:
+                    self.output['proc'] = proc_w
+                    self.output['failure'] = \
+                        f"write phase exited with {proc_w.returncode}"
+                    return
+
+                proc_v = self._run_one(self._fio_args('verify'),
+                                       stdout_f, stderr_f, ec_f)
+                self.output['proc'] = proc_v
+
+        except subprocess.TimeoutExpired:
+            self.output['failure'] = 'timeout'
+        except Exception:
+            self.output['failure'] = 'exception'
+            self.output['exc_info'] = sys.exc_info()
+
+    @staticmethod
+    def _load_json(path):
+        """Return parsed JSON from a fio output file, or None on error."""
+        try:
+            with open(path, "r", encoding=locale.getpreferredencoding()) as f:
+                raw = f.read()
+            lines = raw.splitlines()
+            last = len(lines) - lines[::-1].index("}")
+            return json.loads("\n".join(lines[lines.index("{"):last]))
+        except Exception as exc:
+            logging.debug("JSON parse error in %s: %s", path, exc)
+            return None
+
+    def check_result(self):
+        """Check both phases produced the expected byte counts."""
+
+        # basic exit-code / stderr checks via FioExeTest
+        if 'proc' not in self.output:
+            self.failure_reason = self.output.get('failure', 'did not run')
+            self.passed = False
+            return
+
+        if 'failure' in self.output:
+            self.failure_reason = self.output['failure']
+            self.passed = False
+            # still fall through to check JSON if proc is set
+
+        if self.output['proc'].returncode != 0:
+            self.failure_reason += (
+                f" verify phase exited with {self.output['proc'].returncode}")
+            self.passed = False
+
+        # --- write phase: must have issued io_size bytes total ---
+        # For write-only modes (write, randwrite): all bytes are writes.
+        # For mixed modes (rw, randrw): writes + reads together equal io_size.
+        jw = self._load_json(self.filenames['output_write'])
+        if not jw:
+            self.failure_reason += " cannot parse write-phase JSON"
+            self.passed = False
+        else:
+            job_w = jw['jobs'][0]
+            written = job_w['write']['io_bytes']
+            rw = self.fio_opts.get('rw', 'write')
+            if rw in ('rw', 'readwrite', 'randrw'):
+                total_issued = written + job_w['read']['io_bytes']
+                if total_issued != self.fio_opts['io_size']:
+                    self.failure_reason += (
+                        f" write phase: issued {total_issued} bytes "
+                        f"(write={written} read={job_w['read']['io_bytes']}), "
+                        f"expected {self.fio_opts['io_size']} total")
+                    self.passed = False
+                    logging.debug("write job: %s",
+                                  json.dumps(job_w, indent=2))
+            else:
+                if written != self.fio_opts['io_size']:
+                    self.failure_reason += (
+                        f" write phase: wrote {written} bytes, "
+                        f"expected {self.fio_opts['io_size']}")
+                    self.passed = False
+                    logging.debug("write job: %s",
+                                  json.dumps(job_w, indent=2))
+
+        # --- verify phase: no errors, reads exactly size bytes ---
+        jv = self._load_json(self.filenames['output'])
+        if not jv:
+            self.failure_reason += " cannot parse verify-phase JSON"
+            self.passed = False
+        else:
+            job = jv['jobs'][0]
+            read_data = job.get('read', {})
+
+            verify_errors = read_data.get('verify_errors', 0)
+            if verify_errors != 0:
+                self.failure_reason += (
+                    f" verify phase: {verify_errors} verify error(s)")
+                self.passed = False
+
+            # For sequential write-only modes (write) every block is written
+            # exactly once per pass, so verify reads exactly size bytes.
+            # Skip the assertion for:
+            #   - mixed modes (rw, randrw): only written blocks are in io_hist
+            #   - norandommap: random sampling may leave some blocks unvisited
+            rw_mode = self.fio_opts.get('rw', 'write')
+            if rw_mode not in ('rw', 'readwrite', 'randrw') \
+                    and not self.fio_opts.get('norandommap'):
+                expected_read = self.fio_opts['size']
+                read_bytes = read_data.get('io_bytes', 0)
+                if read_bytes != expected_read:
+                    self.failure_reason += (
+                        f" verify phase: read {read_bytes} bytes, "
+                        f"expected {expected_read} "
+                        f"(one pass — latest numberio only)")
+                    self.passed = False
+                    logging.debug("verify job: %s", json.dumps(job, indent=2))
+
+
+class OnlineOverlapVerifyTest(FioJobCmdTest):
+    """
+    Single-phase online overlap verification test (--do_verify=1).
+
+    fio writes and verifies in the same job.  With io_size > size the write
+    phase makes multiple passes over the file (or revisits offsets with
+    norandommap); the verify (read) phase runs after all writes, checking
+    only the latest numberio per block.
+
+    This exercises:
+      - fio_offset_overlap_risk() enabling the rb-tree io_hist when
+        io_size > size, for all rw modes (write, randwrite, rw, randrw).
+      - backend.c: total_bytes for sequential write-only jobs is now bounded
+        by io_size (not size), so the second write pass actually happens.
+      - io_u.c: last_pos wrap-around is now triggered by do_verify=1 (not
+        only verify_only=1), so the verify READ phase starts from offset 0.
+
+    Key assertions
+        rw=write / rw=randwrite (TC 9-10, 17-18):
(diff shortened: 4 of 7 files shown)


vincentkfu:
I like very much the test script that you added. I was not able to give this PR the time it needs today and will be on vacation for the next two weeks. But I will revisit this when I return.

t/numberio_overlap.py
```diff
@@ -0,0 +1,881 @@
...
+        ]
+
+        if self.fio_opts.get('norandommap'):
+            common.append("--norandommap=1")
+
+        if phase == 'write':
+            return common + [
+                f"--rw={rw}",
+                "--do_verify=0",
+                "--verify_state_save=1",
```
vincentkfu: Why do you need to save verify state here and load it in the verify job? Verify state saving is only needed for scenarios like power cut testing. It should not be necessary for jobs that run to completion normally.
When I remove the verify_state_{save,load} options all the tests still pass except test 9. I think this indicates that there is a problem with accounting during replay. It is strange, though, that test 10 passes when test 9 fails since the only difference between the two is the ioengine.

t/numberio_overlap.py
```diff
@@ -0,0 +1,881 @@
...
+        test['fio_opts']['filename'] = args.file
+
+    test_env = {
+        'fio_path':      fio_path,
+        'fio_root':      fio_root,
+        'artifact_root': artifact_root,
+        'basename':      'numberio-overlap',
+    }
+
+    _, failed, _ = run_fio_tests(TEST_LIST, test_env, args)
```
vincentkfu: Can you just call run_fio_tests twice (once with ioengine=io_uring and a second time with ioengine=psync?). Actually this should work across platforms. Take a look at some of the other scripts to see how they choose async and sync ioengines depending on the platform.

As far as I can tell you the test script has duplicate sets of tests for io_uring and psync.

vincentkfu:
Can you also add a patch to add the test script to run-fio-tests.py?

vincentkfu:
Can you also add test cases with multiple files?

t/numberio_overlap.py
minwooim: > Why do you need to save verify state here and load it in the verify job? Verify state saving is only needed for scenarios like power cut testing. It should not be necessary for jobs that run to completion normally. 

Normally, yes, offline verification mostly is useful when crash happens, but it's also super useful to test the data verification in case where extra host behaviors can be put in between the WRITE and VERIFY phase rather than starting the VERIFY phase right after the WRITE phase.  Therefore, I would like to have this feature as well with the same priority just like the on-line verification.

> When I remove the verify_state_{save,load} options all the tests still pass except test 9. I think this indicates that there is a problem with accounting during replay. It is strange, though, that test 10 passes when test 9 fails since the only difference between the two is the ioengine.

I assume that you might have not removed `--verify_only=1` from the VERIFY phase from the script.  With the following simple changes, TC #9 it works.

```diff
diff --git a/t/numberio_overlap.py b/t/numberio_overlap.py
index 0fca6404..148ac39b 100644
--- a/t/numberio_overlap.py
+++ b/t/numberio_overlap.py
@@ -118,7 +118,6 @@ class OfflineOverlapVerifyTest(FioJobCmdTest):
             return common + [
                 f"--rw={rw}",
                 "--do_verify=0",
-                "--verify_state_save=1",
                 f"--output={self.filenames['output_write']}",
             ]
         else:
@@ -144,9 +143,6 @@ class OfflineOverlapVerifyTest(FioJobCmdTest):
                 f"--verify={self.fio_opts['verify']}",
                 "--output-format=json",
                 "--do_verify=1",
-                "--verify_state_load=1",
-                "--verify_only=1",
-                "--verify_state_save=0",
                 "--verify_write_sequence=1",
                 f"--output={self.filenames['output']}",
             ]
```

```bash
minwoo@localhost:~/work/fio.git/t (check-numberio-read-only)$ sudo python3 numberio_overlap.py  --file /tmp/yeah
Artifact directory: numberio-overlap-test-20260413-113313
fio path: /home/minwoo/work/fio.git/t/../fio
Test 1 PASSED 1
Test 2 PASSED 2
Test 3 PASSED 3
Test 4 PASSED 4
Test 5 PASSED 5
Test 6 PASSED 6
Test 7 PASSED 7
Test 8 PASSED 8
Test 9 PASSED 9
Test 10 PASSED 10
Test 11 PASSED 11
Test 12 PASSED 12
Test 13 PASSED 13
Test 14 PASSED 14
Test 15 PASSED 15
Test 16 PASSED 16
Test 17 PASSED 17
Test 18 PASSED 18
18 test(s) passed, 0 failed, 0 skipped
```

t/numberio_overlap.py
minwooim: Will do it shortly.

minwooim:
> Can you also add a patch to add the test script to run-fio-tests.py?

Makes sense.  Will do.

minwooim:
@vincentkfu , Thanks for the review.  Updated accordingly.  Please review.

t/numberio_overlap.py
vincentkfu: Yes,  I removed only the `verify_state_{load,save}` options and left `verify_only`. Why would you remove `verify_only`? It seems to me that doing so defeats the purpose of offline verification.
Why does removing `verify_state_{load,save}` but retaining`verify_only` work for psync but fail for io_uring?

This is what I tried:
```
root@localhost:~/fio-dev/2074/fio# python3 t/numberio_overlap.py
Artifact directory: numberio-overlap-test-20260413-144534
fio path: /root/fio-dev/2074/fio/t/../fio

Running with ioengine=io_uring
Test 1 PASSED 1
Test 2 PASSED 2
Test 3 PASSED 3
Test 4 PASSED 4
Test 5 FAILED:  verify phase exited with 1 5
Test 6 PASSED 6
Test 7 PASSED 7
Test 8 PASSED 8
Test 9 PASSED 9
Test 10 PASSED 10
Test 11 PASSED 11
Test 12 PASSED 12
Test 13 PASSED 13
12 test(s) passed, 1 failed, 0 skipped

Running with ioengine=psync
Test 1 PASSED 1
Test 2 PASSED 2
Test 3 PASSED 3
Test 4 PASSED 4
Test 5 PASSED 5
Test 6 PASSED 6
Test 7 PASSED 7
Test 8 PASSED 8
Test 9 PASSED 9
Test 10 PASSED 10
Test 11 PASSED 11
Test 12 PASSED 12
Test 13 PASSED 13
13 test(s) passed, 0 failed, 0 skipped
root@localhost:~/fio-dev/2074/fio# git diff
diff --git a/t/numberio_overlap.py b/t/numberio_overlap.py
index bdca0c36..9437a20d 100755
--- a/t/numberio_overlap.py
+++ b/t/numberio_overlap.py
@@ -134,7 +134,6 @@ class OfflineOverlapVerifyTest(FioJobCmdTest):
             return common + [
                 f"--rw={rw}",
                 "--do_verify=0",
-                "--verify_state_save=1",
                 f"--output={self.filenames['output_write']}",
             ]
         else:
@@ -160,9 +159,7 @@ class OfflineOverlapVerifyTest(FioJobCmdTest):
                 f"--verify={self.fio_opts['verify']}",
                 "--output-format=json",
                 "--do_verify=1",
-                "--verify_state_load=1",
                 "--verify_only=1",
-                "--verify_state_save=0",
                 "--verify_write_sequence=1",
                 f"--output={self.filenames['output']}",
             ]
```

vincentkfu:
The tests on container platforms are failing likely because io_uring is disabled. On Windows my guess is that the failure is due to an inability to parse the default file `/tmp/fio-numberio-overlap-test.dat`. Perhaps try a file in the current working directory.
I would add an option to the test script to specify the ioengines to use. And then for containers instruct the test script to use psyc and libaio.
https://github.com/axboe/fio/blob/f4e93793cbe3e4703520a9d4e1d6596bb1b43c73/ci/actions-full-test.sh#L29 is an example of how to instruct the runner to pass-through options to a test script.

t/numberio_overlap.py
vincentkfu: My original point is that for these tests you should not need to save/load verify state because this is not a powercut scenario. Why are you saving/loading verify state here?

minwooim: > My original point is that for these tests you should not need to save/load verify state because this is not a powercut scenario. Why are you saving/loading verify state here?

Even if there's no power cutoff situation in this test, gathering the verify @io_u candidates are totally different in between the two verify phases: online and offline.  We should ensure that dry-run to get the verify @io_u instances in the offline verification, we should test offline along with the oneline in the perspective of deciding the verify candidates @io_us.

vincentkfu: I agree that we should test both verify_only={0,1}. The problem is that verify_only=1 fails with ioengine=io_uring when verify_state_{load,save}=0.

minwooim: Yeah, verify_only=1 should fail since there should no saved verify state file.  Let me have a look why `psync` has passed in the workload.

vincentkfu: verify_only=1 does not require loading a verify state file

minwooim: Got your point.  Yes, we can remove the `verify_state_[load|save]` at all in the second verify phase and keep the `--verify_only=1` as it was to verify pre-written data.

Assum with the following options just like TC 5: `--bs=128k --size=1048576 --io_size=4194304`,

The reason why psync has successfully done is that sync ioengine does not keep counting `bytes_issued` in `do_io()` causing all read/write commands (total 32ea) in WRITE phase will be done at once without verify phase started:

```
# fio --name=overlap-verify --ioengine=psync --fallocate=truncate --filename=/dev/nvme0n1 --rw=rw --bs=128k --size=1048576 --io_size=4194304 --verify=crc32c --do_verify=1 --verify_only=0 --verify_write_sequence=1 --debug=io,verify | grep -E 'queue:|loop'

io       906160 queue: io_u 0x577ea29a4ac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x0,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x60000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x80000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x20000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x40000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x60000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xa0000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x80000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xc0000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xa0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xe0000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xc0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xe0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x60000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x0,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x20000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x40000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x60000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x80000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x80000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xa0000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xa0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xc0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xe0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x0,len=0x20000,ddir=1,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xc0000,len=0x20000,ddir=0,file=/dev/nvme0n1
verify   906160 starting loop
io       906160 queue: io_u 0x577ea29a4ac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x60000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0x80000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xa0000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xc0000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       906160 queue: io_u 0x577ea29a4ac0: off=0xe0000,len=0x20000,ddir=0,file=/dev/nvme0n1
```

On the other hands, in an async ioengine, `bytes_isseud` are kept counting so that the following statement breaks the `do_io()` big-loop and everytime 8 * 128k block are issued, out of big-loop and starts verify phase causing corruption the I/O workloads of WRITE phase at all:

Where `do_io()` exits:
```c
    !td->o.read_iolog_file &&
    (!td->o.time_based ||
     (td->o.time_based && td->o.verify != VERIFY_NONE)))
        break;
```

```
# fio --name=overlap-verify --ioengine=io_uring --fallocate=truncate --filename=/dev/nvme0n1 --rw=rw --bs=128k --size=1048576 --io_size=4194304 --verify=crc32c --do_verify=1 --verify_only=0 --verify_write_sequence=1 --debug=io,verify | grep -E 'queue:|loop'

io       907159 queue: io_u 0x5bae29b34ac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x0,len=0x20000,ddir=1,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x60000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x80000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x20000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x40000,len=0x20000,ddir=1,file=/dev/nvme0n1
verify   907159 starting loop
io       907159 queue: io_u 0x5bae29b34ac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
verify   907159 exiting loop
io       907159 queue: io_u 0x5bae29b34ac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x0,len=0x20000,ddir=1,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x60000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x80000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x20000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x40000,len=0x20000,ddir=1,file=/dev/nvme0n1
verify   907159 starting loop
io       907159 queue: io_u 0x5bae29b34ac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
verify   907159 exiting loop
io       907159 queue: io_u 0x5bae29b34ac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x0,len=0x20000,ddir=1,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x60000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x80000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x20000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x40000,len=0x20000,ddir=1,file=/dev/nvme0n1
verify   907159 starting loop
io       907159 queue: io_u 0x5bae29b34ac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       907159 queue: io_u 0x5bae29b34ac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
verify   907159 exiting loop
```

If `do_verify=1` is given to `io_uring`, it will dry-run to gather the candidates and expected numberio 16 from offset 0 (latest), but the WRITE phase actually written to the file and the dry-run is totally different.  The following logs are for the dry-run when `verify_only=1` which is totally different from the WRITE phase above:

```
io       910015 prep: io_u 0x55ea3fffeac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x0,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x60000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x80000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x20000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x40000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x60000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0xa0000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x80000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0xc0000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0xa0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0xe0000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x0,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x20000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0xc0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0xe0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x40000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x60000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x0,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x20000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x40000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x60000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x80000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x80000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0xa0000,len=0x20000,ddir=0,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0xa0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0xc0000,len=0x20000,ddir=1,file=/dev/nvme0n1
verify: bad header numberio 6, wanted 16 at file /dev/nvme0n1 offset 0, length 131072 (requested block: offset=0, length=131072)
io       910015 prep: io_u 0x55ea3fffeac0: off=0xe0000,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0x0,len=0x20000,ddir=1,file=/dev/nvme0n1
io       910015 prep: io_u 0x55ea3fffeac0: off=0xc0000,len=0x20000,ddir=0,file=/dev/nvme0n1
```

I will keep looking into how those two engines are synced up based on `bytes_issued`.

minwooim: If `verify_write_sequence=1` is given, we should assume that all the I/Os in WRITE phase should proceed before VERIFY phase starts.  How about the following changes with this patchset?

Applied this patch to 26425e6c277d6d7c62ed3095e06dc067e049ec60.

```diff
diff --git a/backend.c b/backend.c
index be9061a7cb11..0dbeb5d64a47 100644
--- a/backend.c
+++ b/backend.c
@@ -1119,6 +1119,14 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
        if (!td_rw(td) && td_write(td) && !td_random(td))
                total_bytes = max(total_bytes, (uint64_t)td->o.io_size);

+       /*
+        * verify_write_sequence completes all writes before starting verify.
+        * The full io_size must be written regardless of rw mode, otherwise
+        * async engines may start verifying before all writes are issued.
+        */
+       if (td_write(td) && !td_random(td) && td->o.verify_write_sequence)
+               total_bytes = max(total_bytes, (uint64_t)td->o.io_size);
+
        /*
         * If verify_backlog is enabled, we'll run the verify in this
         * handler as well. For that case, we may need up to twice the

```

vincentkfu: It's backwards to let `verify_write_sequence` affect how much IO is done. The appropriate amount of IO should be done in all cases and then `verify_write_sequence` should determine whether `numberio` is verified or not.

I think the right thing to do is to simply revert 981c372fb797e2edee55309d28b1028fba5e9b68.

```
# git show 981c372fb797e2edee55309d28b1028fba5e9b68
commit 981c372fb797e2edee55309d28b1028fba5e9b68
Author: Ankit Kumar <ankit.kumar@samsung.com>
Date:   Wed Jan 29 01:11:36 2025 +0530

    backend: fix verify issue during readwrite

    In readwrite mode if specified io_size > size, offsets can overlap.
    This will result in verify errors. Add check to handle this case.

    Fixes: d782b76f ("Don break too early in readwrite mode")

    Signed-off-by: Ankit Kumar <ankit.kumar@samsung.com>

diff --git a/backend.c b/backend.c
index f5cfffdb..b75eea80 100644
--- a/backend.c
+++ b/backend.c
@@ -978,8 +978,11 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
        if (td_write(td) && td_random(td) && td->o.norandommap)
                total_bytes = max(total_bytes, (uint64_t) td->o.io_size);

-       /* Don't break too early if io_size > size */
-       if (td_rw(td) && !td_random(td))
+       /*
+        * Don't break too early if io_size > size. The exception is when
+        * verify is enabled.
+        */
+       if (td_rw(td) && !td_random(td) && td->o.verify == VERIFY_NONE)
                total_bytes = max(total_bytes, (uint64_t)td->o.io_size);

        /*
```

With your patch that switches to an RB tree when `io_size > size` the overwrite risk is properly detected and fio will be able to verify the correct sequence number, so the above patch is no longer needed.

You will also be able to simplify the code so that there are no longer separate if statements for sequential read/write and 100% write workloads.

@ankit-sam Do you agree that it's ok to revert this patch now that the overlap risk calculator includes the `io_size > size` condition? Was the only concern addressed by this patch sequence number verification or was there something else?

ankit-sam: Yeah, I think this should be covered by the other fix.

libfio.c
```diff
@@ -106,7 +106,8 @@ static void reset_io_counters(struct thread_data *td, int all)
 	/*
 	 * reset file done count if we are to start over
 	 */
-	if (td->o.time_based || td->loops > 1 || td->o.do_verify)
+	if (td->o.time_based || td->loops > 1 || td->o.do_verify ||
+	    (td->o.io_size && td->o.io_size > td->o.size))
```
ankit-sam: I think it could be simply be written as
```
if (td->o.time_based || td->loops > 1 || td->o.do_verify ||
 (td->o.io_size > td->o.size))
```
But do check I may be wrong.

t/numberio_overlap.py
minwooim: Thanks @ankit-sam , @vincentkfu , you want me to add the revert path in this PR?

ankit-sam: Yes you can revert the commit https://github.com/axboe/fio/commit/981c372fb797e2edee55309d28b1028fba5e9b68

In commit log mention about your commit clearly and that for `io_size > size` we are now using RB tree.

minwooim: Applied, thanks

backend.c
```diff
@@ -1104,6 +1104,14 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
...
 		total_bytes = max(total_bytes, (uint64_t) td->o.io_size);
 
+	/*
+	 * For write-only sequential jobs, allow io_size to control how much to
+	 * write even when verify is enabled.  Unlike rw mode, `bytes_issued`
+	 * only counts writes here, so `io_size` won't be consumed by reads.
+	 */
+	if (!td_rw(td) && td_write(td) && !td_random(td))
+		total_bytes = max(total_bytes, (uint64_t)td->o.io_size);
+
```
vincentkfu: Please merge this `if` with the one below and make it just:

`if (td_write(td) && !td_random(td))`

Also make this a separate commit from the `io_u.c` change.

io_u.c
```diff
@@ -382,8 +382,12 @@ static int get_next_seq_offset(struct thread_data *td, struct fio_file *f,
 	/*
 	 * If we reach the end for a rw-io-size based run, reset us back to 0
 	 * and invalidate the cache, if we need to.
+	 *
+	 * If it has possibility of overlapped offsets (io_size > size), we
+	 * should get back to 0 offset to verify overlapped offsets.
 	 */
-	if (td_rw(td) && o->io_size > o->size) {
+	if ((td_rw(td) || (o->verify != VERIFY_NONE && o->do_verify)) &&
+			o->io_size > o->size) {
```
vincentkfu: Can you replace `(o->verify != VERIFY_NONE && o->do_verify)` with `td_write(td)`? It doesn't make sense for the verify options to affect offset generation.
You would then also be able to drop `td_rw(td)` since `td_write(td)` is a superset of `td_rw(td)`.
Please make this change and the `backend.c` change separate patches. You will need to rewrite the commit message since `verify_only` is no longer relevant here.

vincentkfu:
A note for future reference:

This patch series changes the behavior of `io_size > size` verify write jobs. Previously this situation would have multiple write-verify loops. WIth this patch series, fio completes `io_size` writes and then carries out a single verify phase.

Original behavior: write 2 blocks, verify 2 blocks, write 2 blocks, verify 2 blocks
```
root@localhost:~/fio-dev/fio-canonical# ./fio --name=test --size=8K --io_size=32k --rw=write --verify=crc32c --debug=io | grep queue:
io       151635 queue: io_u 0x55be10614980: off=0x0,len=0x1000,ddir=1,file=test.0.0
io       151635 queue: io_u 0x55be10614980: off=0x1000,len=0x1000,ddir=1,file=test.0.0
io       151635 queue: io_u 0x55be10614980: off=0x0,len=0x1000,ddir=0,file=test.0.0
io       151635 queue: io_u 0x55be10614980: off=0x1000,len=0x1000,ddir=0,file=test.0.0
io       151635 queue: io_u 0x55be10614980: off=0x0,len=0x1000,ddir=1,file=test.0.0
io       151635 queue: io_u 0x55be10614980: off=0x1000,len=0x1000,ddir=1,file=test.0.0
io       151635 queue: io_u 0x55be10614980: off=0x0,len=0x1000,ddir=0,file=test.0.0
io       151635 queue: io_u 0x55be10614980: off=0x1000,len=0x1000,ddir=0,file=test.0.0
```

New behavior: write 8 blocks, verify 2 blocks
```
root@localhost:~/fio-dev/2074/fio# ./fio --name=test --size=8K --io_size=32k --rw=write --verify=crc32c --debug=io | grep queue:
io       151758 queue: io_u 0x55cc1bc9a980: off=0x0,len=0x1000,ddir=1,file=test.0.0
io       151758 queue: io_u 0x55cc1bc9a980: off=0x1000,len=0x1000,ddir=1,file=test.0.0
io       151758 queue: io_u 0x55cc1bc9a980: off=0x0,len=0x1000,ddir=1,file=test.0.0
io       151758 queue: io_u 0x55cc1bc9a980: off=0x1000,len=0x1000,ddir=1,file=test.0.0
io       151758 queue: io_u 0x55cc1bc9a980: off=0x0,len=0x1000,ddir=1,file=test.0.0
io       151758 queue: io_u 0x55cc1bc9a980: off=0x1000,len=0x1000,ddir=1,file=test.0.0
io       151758 queue: io_u 0x55cc1bc9a980: off=0x0,len=0x1000,ddir=1,file=test.0.0
io       151758 queue: io_u 0x55cc1bc9a980: off=0x1000,len=0x1000,ddir=1,file=test.0.0
io       151758 queue: io_u 0x55cc1bc9a980: off=0x0,len=0x1000,ddir=0,file=test.0.0
io       151758 queue: io_u 0x55cc1bc9a980: off=0x1000,len=0x1000,ddir=0,file=test.0.0
```

Fio's documentations says:
```
If writing to a file, fio can verify the file contents after each iteration of the job. 
```

The new behavior seems slightly more consistent with the documentation although the documentation is not clear about how these workloads should behave.

I think it's reasonable to enable this new behavior since the old behavior is available via the `loops` option.

backend.c
minwooim: updated, Thanks

io_u.c
minwooim: Removed the verify condition and updated commit description accordingly.

PR #2082, opened 2026-04-23
sprandom: relax power of 2 block size requirement

vincentkfu:
There is no a priori reason why the SPRandom approach cannot work with block sizes that are not a power of 2. Lift the constraint that allows only power of 2 block sizes when sprandom is enabled.

This allows sprandom to be used when devices are formatted in extended LBA (DIF) mode with 520- or 4160-byte sectors.

@steven-sprouse-sndk
@tomas-winkler-sndk


vincentkfu:
Merged: https://github.com/axboe/fio/commit/a95f84caad2300e73ec9bf7077879f30246f3185

Issue #2084, opened 2026-04-28
cgroup: buffer overflow with long cgroup paths in get_cgroup_root() / write_int_to_file()


linuxholic:
- [x] I have read the GitHub issues section of [REPORTING-BUGS](../blob/master/REPORTING-BUGS).

**Description of the bug:**

fio aborts with a buffer overflow when `cgroup=` is set to a sufficiently long cgroup path.

The issue appears to be caused by fixed-size path buffers in `cgroup.c`:

- `get_cgroup_root()` allocates 64 bytes and then uses `sprintf()`
- `write_int_to_file()` uses a 256-byte stack buffer and then uses `sprintf()`

As a result, long `cgroup=` values can overflow internal buffers and cause `fio` to abort instead of returning a normal error such as `ENOENT` or `ENAMETOOLONG`.

From code inspection, this affects at least:
- long cgroup directory paths
- long cgroup directory + control file path combinations

**Environment**: Linux with cgroup v1 blkio mounted at `/sys/fs/cgroup/blkio`

**fio version**: `fio-3.34`

**Reproduction steps**

Minimal job file:

```ini
[test_throttle]
filename=fio.test
rw=write
bs=4k
ioengine=libaio
iodepth=1
numjobs=1
size=40m
direct=1
cgroup=/agent/staragenxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

**FIx**

I can send a PR with a minimal fix that only replaces the fixed-size path buffers with dynamically sized allocations.

axboe:
Just send a PR, I'm assuming this is still a problem in current -git? Asking as you're testing an ancient version (why?!).

linuxholic:
Yes, this is still reproducible on current -git.

I first hit it on a system that was running fio-3.34, which is why the initial report used that version. I rechecked it locally against current master as well (before the fix), and the same root cause is still there.

I also added a small unit test for the long-path case. Without the fix it aborts with "*** buffer overflow detected ***" (exit code 134), and with the fix the new cgroup tests pass.

I will send a PR shortly.

PR #2085, opened 2026-04-29
cgroup: fix path buffer overflows for long cgroup names

linuxholic:
This fixes the long-path buffer overflows reported in #2084.

## Problem

`cgroup.c` uses fixed-size buffers when constructing:
- the cgroup root path
- the cgroup control file path

With sufficiently long `cgroup=` values, these buffers can overflow and `fio` aborts instead of failing cleanly.

## Fix

- replace the fixed 64-byte cgroup root buffer with dynamically sized allocation
- replace the fixed 256-byte file path buffer with dynamically sized allocation
- preserve existing `cgroup=` behavior and error handling semantics

## Test

Added unit tests for:
- long cgroup root paths
- long cgroup control file paths

Validated in an Ubuntu container:
- without the fix, `cgroup/get-root-long-path` aborts with `*** buffer overflow detected ***`
- with the fix, the new cgroup tests pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -455,6 +455,12 @@ UT_OBJS += unittests/oslib/strlcat.o
 UT_OBJS += unittests/oslib/strndup.o
 UT_OBJS += unittests/oslib/strcasestr.o
 UT_OBJS += unittests/oslib/strsep.o
+ifeq ($(CONFIG_TARGET_OS), Linux)
+UT_OBJS += unittests/cgroup.o
+endif
+ifeq ($(CONFIG_TARGET_OS), Android)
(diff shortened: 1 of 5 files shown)


axboe:
Fix looks good to me, but commit message is missing a ```Signed-off-by``` line at the end.

axboe:
And a some unittest build failures you will want to address.

linuxholic:
I force-pushed an update.

- added the Signed-off-by line
- fixed the unittest build failures caused by discarded const qualifiers in the new cgroup tests

I also revalidated the Linux build and the unit tests in an Ubuntu container.

PR #2086, opened 2026-05-04
verify: skip errored WRITE offset in verify

minwooim:
``get_next_verify()`` has filled up `io_u->rand_seed` in the sequence of
the same of write phase @io_u's.  Let's say that fio faces errored @io_u
with --ignore_error= to suppress the error(expected), the @io_u will be
unlogged in ``io_completed()`` by ``unlog_io_piece()``.  This leads
``get_next_verify()`` skips to advance the random seed for the @io_u
next to the errored @io_u which is excluded.  This caused verification
error.  To fix this issue, @io_u->rand_seed should be filled up in
``do_verify()`` rather than ``get_next_verify()``.

If any WRITE command fails with error (io_u->error != 0), and it's due
to some of expected situation such as resets, those offsets should be
excluded from the verification list since we can't simply guarantee the
written data.

**Online verification**

Online verification can be done by running a single fio session with
`--do_verify=1`.  It can keep the errored numberio information with the
verify list by unlogging the errored io letting `get_next_verify()`
never retrieves the errored @io_u.

**Offline verification**

Offline verification is to keep the verify state as a file and restore
it in the verify phase, and it can be run by:
  - Run two fio sessions (one for write, one for verify) with
    (--verify_state_save=1 and --verify_state_load=1)

After saving a verify state file in WRITE phase, VERIFY phase will read
it and do `do_dry_run()` where the actual skipping the errored @io_u is
done.  Errored @io_u will not be logged into the verify history and will
not be poped by `get_next_verify()` at all in `do_verify()`.

This patch bumped up the verify header version since @inflight[] of the
verify state file has been expanded to keep the errored numberios along
with the existing inflight numberios.  No header attributes changed, but
meaning has been updated.

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -621,6 +621,7 @@ static enum fio_q_status io_u_submit(struct thread_data *td, struct io_u *io_u)
  */
 static void do_verify(struct thread_data *td, uint64_t verify_bytes)
 {
+	uint64_t expected_numberio;
 	struct fio_file *f;
 	struct io_u *io_u;
 	unsigned int i;
@@ -649,6 +650,7 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
 	td_set_runstate(td, TD_VERIFYING);
 
 	io_u = NULL;
+	expected_numberio = UINT64_MAX;
 	while (!td->terminate) {
 		enum fio_ddir ddir;
 		int full;
@@ -675,6 +677,33 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
 			if (get_next_verify(td, io_u)) {
 				put_io_u(td, io_u);
 				break;
+			}
+
+			/*
+			 * Advance verify_state to the seed that was used when
+			 * this block was written.  Any numberio gap between
+			 * @expected_numberio and @io_u->numberio represents
+			 * skipped writes due to errors (e.g., suppressed by
+			 * --ignore_error= in online verification, or skipped
+			 *  offsets from do_dry_run in offline verification).
+			 */
+			if (td->o.verify_header_seed && !td->o.verify_pattern_bytes) {
+				uint64_t seed = 0;
+				uint64_t from;
+
+				if (io_u->numberio < expected_numberio)
+					from = io_u->numberio;
+				else
+					from = expected_numberio;
+
+				for (uint64_t n = from; n <= io_u->numberio; n++) {
+					seed = __rand(&td->verify_state);
+					if (sizeof(int) != sizeof(long *))
+						seed *= __rand(&td->verify_state);
+				}
+
+				io_u->rand_seed = seed;
+				expected_numberio = io_u->numberio + 1;
 			}
 
 			if (td_io_prep(td, io_u)) {
@@ -1207,7 +1236,26 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
 					io_u->rand_seed *= __rand(&td->verify_state);
 			}
 
-			if (verify_state_should_stop(td, td->io_issues[io_u->ddir])) {
+			/*
+			 * Assign numberio for read-only verify workloads. This
+			 * handles two cases: (1) rw=[rand]write with verify_only=1
+			 * and verify_state_load=1, where we replay a prior write
+			 * run; (2) a prior job ran with rw=[rand]write and
+			 * verify_state_save=1, and the current job runs with
+			 * rw=[rand]read and verify_state_load=1 to verify those
+			 * writes.
+			 */
+			if (!td_rw(td) && !(io_u->flags & IO_U_F_VER_LIST))
+				io_u->numberio = td->io_issues[io_u->ddir];
+
+			if (verify_state_should_skip(td, io_u->numberio)) {
+				/* Account for this I/O so we move to the next sequence */
+				td->io_issues[io_u->ddir]++;
+				put_io_u(td, io_u);
+				continue;
+			}
+
+			if (verify_state_should_stop(td, io_u->numberio)) {
 				put_io_u(td, io_u);
 				break;
 			}
@@ -1380,6 +1428,8 @@ static void free_inflight_logging(struct thread_data *td)
 {
 	if (td->inflight_numberio)
 		sfree(td->inflight_numberio);
+	if (td->failed_numberio)
+		free(td->failed_numberio);
 }
 
 static void cleanup_io_u(struct thread_data *td)
@@ -1780,8 +1830,10 @@ static uint64_t do_dry_run(struct thread_data *td)
 		if (td_write(td) && io_u->ddir == DDIR_WRITE &&
 		    td->o.do_verify &&
 		    td->o.verify != VERIFY_NONE &&
-		    !td->o.experimental_verify)
-			log_io_piece(td, io_u);
+		    !td->o.experimental_verify) {
+			if (!verify_state_should_skip(td, io_u->numberio))
+				log_io_piece(td, io_u);
+		}
 
 		ret = io_u_sync_complete(td, io_u);
 		(void) ret;
--- a/io_u.c
+++ b/io_u.c
@@ -2138,6 +2138,29 @@ static void io_completed(struct thread_data *td, struct io_u **io_u_ptr,
 		zbd_recover_write_error(td, io_u);
 
 	/*
+	 * Track failed write I/Os for offline verification (independent of io_u->ipo).
+	 * This works even when do_verify=0 is set.  In case of online
+	 * verification, `unlog_io_piece()` will be called so that the ipo will
+	 * be removed from the `io_hist`.
+	 */
+	if (io_u->error && ddir == DDIR_WRITE && io_u->numberio != INVALID_NUMBERIO) {
+		if (td->failed_numberio_count >= td->failed_numberio_alloc) {
+			unsigned int new_alloc;
+
+			if (td->failed_numberio_alloc)
+				new_alloc = td->failed_numberio_alloc * 2;
+			else
+				new_alloc = 16;
+			td->failed_numberio = realloc(td->failed_numberio,
+						new_alloc * sizeof(uint64_t));
+			td->failed_numberio_alloc = new_alloc;
+		}
+		td->failed_numberio[td->failed_numberio_count++] = io_u->numberio;
+		dprint(FD_VERIFY, "Recorded failed write numberio=%"PRIu64"\n",
+			io_u->numberio);
+	}
+
+	/*
 	 * Mark IO ok to verify
 	 */
 	if (io_u->ipo) {
--- a/verify.c
+++ b/verify.c
@@ -1474,11 +1474,6 @@ int get_next_verify(struct thread_data *td, struct io_u *io_u)
 		free(ipo);
 		dprint(FD_VERIFY, "get_next_verify: ret io_u %p\n", io_u);
 
-		if (!td->o.verify_pattern_bytes) {
-			io_u->rand_seed = __rand(&td->verify_state);
-			if (sizeof(int) != sizeof(long *))
-				io_u->rand_seed *= __rand(&td->verify_state);
-		}
 		return 0;
 	}
 
@@ -1655,7 +1650,8 @@ struct all_io_list *get_all_io_list(int save_mask, size_t *sz)
 			continue;
 		td->stop_io = 1;
 		td->flags |= TD_F_VSTATE_SAVED;
-		depth += (td->o.iodepth * td->o.nr_files);
+		/* Include both inflight and failed I/Os in the state */
+		depth += ((td->o.iodepth + td->failed_numberio_count) * td->o.nr_files);
 		nr++;
 	} end_for_each();
 
@@ -1672,6 +1668,7 @@ struct all_io_list *get_all_io_list(int save_mask, size_t *sz)
 	next = &rep->state[0];
 	for_each_td(td) {
 		struct thread_io_list *s = next;
+		unsigned int total_depth;
 
 		if (save_mask != IO_LIST_ALL && (__td_index + 1) != save_mask)
 			continue;
@@ -1685,7 +1682,15 @@ struct all_io_list *get_all_io_list(int save_mask, size_t *sz)
 		for (int i = 0; td->inflight_numberio && i < td->o.iodepth; i++)
 			s->inflight[i].numberio = cpu_to_le64(atomic_load_acquire(&td->inflight_numberio[i]));
 
-		s->depth = cpu_to_le32((uint32_t) td->o.iodepth);
+		/* Then, append failed I/Os to exclude them from verification */
+		for (unsigned int i = 0; i < td->failed_numberio_count; i++) {
+			s->inflight[td->o.iodepth + i].numberio = cpu_to_le64(td->failed_numberio[i]);
+			dprint(FD_VERIFY, "Added failed numberio=%"PRIu64" to inflight list\n",
+				td->failed_numberio[i]);
+		}
+
+		total_depth = td->o.iodepth + td->failed_numberio_count;
+		s->depth = cpu_to_le32((uint32_t) total_depth);
 		s->numberio = cpu_to_le64((uint64_t) atomic_load_acquire(&td->inflight_issued));
 		s->index = cpu_to_le64((uint64_t) __td_index);
 		if (td->offset_state.use64) {
@@ -1809,6 +1814,18 @@ void verify_free_state(struct thread_data *td)
 		free(td->vstate);
 }
 
+static int verify_u64_cmp(const void *a, const void *b)
+{
+	uint64_t x = *(const uint64_t *) a;
+	uint64_t y = *(const uint64_t *) b;
+
+	if (x < y)
+		return -1;
+	if (x > y)
+		return 1;
+	return 0;
+}
+
 void verify_assign_state(struct thread_data *td, void *p)
 {
 	struct thread_io_list *s = p;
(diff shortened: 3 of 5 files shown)


vincentkfu:
Please fix the CI failure which is just a simple `verify_state_save/load`  job.

backend.c
```diff
@@ -677,6 +678,38 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
...
+				uint64_t seed = 0;
+				/*
+				 * On the first piece (sentinel) start from its own
+				 * numberio so we advance exactly once.  For subsequent
+				 * in-order pieces start from expected_numberio so we
+				 * advance through any gap (failed writes).  For
+				 * out-of-order pieces (rb-tree / overlap-risk) fall
+				 * back to a single advance from the current state.
+				 */
+				uint64_t from = (expected_numberio == UINT64_MAX ||
```
vincentkfu: Please separate the declaration and assignment here and use an if statement for the assignment to improve readability.
The `expected_numberio == UINT64_MAX` condition seems redundant to me. The first time through, the second component of the || is almost guaranteed to evaluate to true. If `io_u->numberio == UINT64_MAX` the first time through the result of the expression will be the same whether `expected_numberio == UINT64_MAX` is present or not.

backend.c
```diff
@@ -677,6 +678,38 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
...
+			 */
+			if (!td->o.verify_pattern_bytes) {
+				uint64_t seed = 0;
+				/*
+				 * On the first piece (sentinel) start from its own
+				 * numberio so we advance exactly once.  For subsequent
+				 * in-order pieces start from expected_numberio so we
+				 * advance through any gap (failed writes).  For
+				 * out-of-order pieces (rb-tree / overlap-risk) fall
+				 * back to a single advance from the current state.
```
vincentkfu: I don't understand how when using the rb-tree the for loop falls back to a single advance. Please explain in more detail.

backend.c
```diff
@@ -677,6 +678,38 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
 				break;
 			}
 
+			/*
+			 * Advance verify_state to the seed that was used when
```
vincentkfu, line 683: Some parts of this are repeated below.
You should also note here that numberio gaps appear for online verification. For offline verification, errored numberios are skipped.

verify.c
```diff
@@ -1672,6 +1668,7 @@ struct all_io_list *get_all_io_list(int save_mask, size_t *sz)
 	next = &rep->state[0];
 	for_each_td(td) {
 		struct thread_io_list *s = next;
+		unsigned int total_depth;
```
vincentkfu, line 1671: Please bump the verify state file version here. The file format is the same but the meaning of depth has changed.

vincentkfu (changes requested):
Please rewrite the commit message for the first patch.

- Move the note about --ignore_error to the first paragraph
- Explain how errored writes are skipped for online verification
- Explain how errored writes are skipped for offline verification
- Add a note that you are changing the verify state file

backend.c
```diff
@@ -1207,7 +1246,17 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
 					io_u->rand_seed *= __rand(&td->verify_state);
 			}
 
-			if (verify_state_should_stop(td, td->io_issues[io_u->ddir])) {
+			if (!td_rw(td) && !(io_u->flags & IO_U_F_VER_LIST))
```
vincentkfu, line 1248: Is this here to handle 100% read workloads with `verify_only=1`

minwooim: Exactly.  It's for read-only verify job where `io_u->numberio` should be set to `td->io_issues[DDIR_READ]`.  If @io_u came from the VER_LIST, `io_u->numberio` has already been set to the WRITE @io_u's numberio same.

To check whether to stop or not here, we should have considered the read-only verify job also which has mis-led with the wrong `numberio`.

verify.c
minwooim: Missed that one.  Will update the header version.

backend.c
minwooim: Unless we put the `io_u->rand_seed` value to `ipo` structure, cases which has overlap-risk has never been verified with `rand_seed` in the verify header which can be tested by `--verify_header_seed=1`.

This comment was all about just keep it go through with no-op for the overlap-risk cases, but it caused confusions at all.  I will revise this code to be run only in case of ` !fio_offset_overlap_risk(td)` with keeping overlap-risk cases as it was.

backend.c
minwooim: In https://github.com/axboe/fio/pull/2086#discussion_r3221009970 review, I will clean up this mess.

backend.c
minwooim: This change is to fix the wrong numberio given to the `verify_state_should_stop()` for the following cases.

**Normal cases (no errors at all)**

If there's no error, online verification without verify state file is okay to go.

```bash
fio --name=yeah --ioengine=io_uring_cmd --cmd_type=nvme --filename=/dev/ng0n1 --rw=write --bs=512 --size=2k --verify=crc32
fio --name=yeah --ioengine=io_uring_cmd --cmd_type=nvme --filename=/dev/ng0n1 --rw=read --bs=512 --size=2k --verify=crc32
```

**Errored cases (error numberio should be stored in verify header)**

If there's errors, verify header is required to retrieve the errroed numberio information with offline verification.

```bash
fio --name=yeah --ioengine=io_uring_cmd --cmd_type=nvme --filename=/dev/ng0n1 --rw=write --bs=512 --size=2k --verify=crc32 --verify_state_save=1
fio --name=yeah --ioengine=io_uring_cmd --cmd_type=nvme --filename=/dev/ng0n1 --rw=read --bs=512 --size=2k --verify=crc32 --verify_state_load=1
```

minwooim:
Updated branch.  Please review.

backend.c
minwooim: Tested with 0037 and this change is required for cases where `loops > 1`

backend.c
```diff
@@ -677,6 +679,40 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
...
+			 * skipped writes due to errors (e.g., suppressed by
+			 * --ignore_error=).
+			 *
+			 * In case where overlap risk exists (rb-tree), @io_u
+			 * is retrieved in an offset order which means we can't
+			 * simply ensure the seed synced here unless we put the
+			 * @rand_seed to the @ipo.
+			 */
+			if (!td->o.verify_pattern_bytes &&
+					!fio_offset_overlap_risk(td)) {
```
vincentkfu: I think it would be better to just check `o->verify_header_seed` instead

backend.c
vincentkfu: Do you agree with my statement above that numberio gaps appear for online verification and that offline verification (`verify_only=1` with `verify_state_load`) will not have numberio gaps because `do_dry_run()` will not reproduce errored offsets.
If so, I think it would be helpful to clarify this in the comment for future people who look at this code.

backend.c
```diff
@@ -677,6 +679,40 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
...
+				uint64_t seed = 0;
+				uint64_t from;
+
+				if (expected_numberio == UINT64_MAX ||
```
vincentkfu: My first review of this patch suggested that checking this sentinel value is not necessary. If you agree, remove this check.

vincentkfu:
I think the first patch does not work for `verify_backlog=X` because numberio is not advanced appropriately when errored writes are skipped.

Edit: `numberio` checking is disabled for `verify_backlog` since we cannot reproduce the sequence of random seeds. So this is a moot point although I think it would be useful to make a note of this somewhere to remind anyone who looks at this code in the future.

backend.c
minwooim: The numberio gap can happen in both situations.  Even in online verification, ipo will be unlogged from the io_hist_tree if error happens in `io_completed()` with the following:

```c
if (io_u->ipo) {                                                    
        /*                                                          
         * Remove errored entry from the verification list          
         */                                                         
        if (io_u->error)                                            
                unlog_io_piece(td, io_u);                           
        else {                                                      
                atomic_store_release(&io_u->ipo->flags,             
                                io_u->ipo->flags & ~IP_F_IN_FLIGHT);
        }                                                           
}
```

In offline verification, as you said, `do_dry_run()` does not log the io_u information to the io_hist_tree so that the numberio will have the gap.

minwooim:
> I think the first patch does not work for `verify_backlog=X` because numberio is not advanced appropriately when errored writes are skipped.
> 
> Edit: `numberio` checking is disabled for `verify_backlog` since we cannot reproduce the sequence of random seeds. So this is a moot point although I think it would be useful to make a note of this somewhere to remind anyone who looks at this code in the future.

Indeed, before this patch, `get_next_verify()` has never been able to check the `rand_seed` since verify_state has already been advanced in WRITE, and the following VERIFY @io_u comes to have advanced verify.  Added some comments on the moved rand_seed code part with the following:

```diff
@@ -677,6 +679,42 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
                                break;
                        }

+                       /*
+                        * Advance verify_state to the seed that was used when
+                        * this block was written.  Any numberio gap between
+                        * @expected_numberio and @io_u->numberio represents
+                        * skipped writes due to errors (e.g., suppressed by
+                        * --ignore_error= in online verification, or skipped
+                        *  offsets from do_dry_run in offline verification).
+                        *
+                        *  Note that it does not apply to --verify_backlog,
+                        *  which puts the verify @io_u among the actual WRITE
+                        *  I/Os by calling `get_next_verify()`.  There is no
+                        *  way to reproduce the PRNG sequence for
+                        *  verify_backlog when write and verifies are
+                        *  interleaved in the same path.  So,
+                        *  --verify_header_seed=1 is not checked for
+                        *  verify_backlog.
+                        */
+                       if (td->o.verify_header_seed) {
+                               uint64_t seed = 0;
+                               uint64_t from;
+
+                               if (io_u->numberio < expected_numberio)
+                                       from = io_u->numberio;
+                               else
+                                       from = expected_numberio;
+
+                               for (uint64_t n = from; n <= io_u->numberio; n++) {
+                                       seed = __rand(&td->verify_state);
+                                       if (sizeof(int) != sizeof(long *))
+                                               seed *= __rand(&td->verify_state);
+                               }
+
+                               io_u->rand_seed = seed;
+                               expected_numberio = io_u->numberio + 1;
+                       }
+
```

verify.c
```diff
@@ -1911,6 +1952,34 @@ int verify_load_state(struct thread_data *td, const char *prefix)
...
+	lo = 0;
+	hi = td->failed_numberio_count;
+	while (lo < hi) {
+		unsigned int mid = lo + (hi - lo) / 2;
+
+		if (td->failed_numberio[mid] == numberio) {
+			dprint(FD_VERIFY, "Skipping failed numberio=%"PRIu64"\n", numberio);
+			return 1;
+		} else if (td->failed_numberio[mid] < numberio)
+			lo = mid + 1;
```
vincentkfu, line 1975: Please add braces above and below this statement.

verify.c
```diff
@@ -1911,6 +1952,34 @@ int verify_load_state(struct thread_data *td, const char *prefix)
...
+		else
+			hi = mid;
```
vincentkfu, line 1977: Please add braces above and below this statement.

io_u.c
```diff
@@ -2137,6 +2137,25 @@ static void io_completed(struct thread_data *td, struct io_u **io_u_ptr,
...
 
+	/*
+	 * Track failed write I/Os for offline verification (independent of io_u->ipo).
+	 * This works even when do_verify=0 is set.  In case of online
+	 * verification, `unlog_io_piece()` will be called so that the ipo will
+	 * be removed from the `io_hist`.
+	 */
+	if (io_u->error && ddir == DDIR_WRITE && io_u->numberio != INVALID_NUMBERIO) {
+		if (td->failed_numberio_count >= td->failed_numberio_alloc) {
+			unsigned int new_alloc = td->failed_numberio_alloc ?
```
vincentkfu: Please separate declaration from assignment and just use an if statement to make the code easier to read.

verify.c
```diff
@@ -1924,10 +1993,12 @@ int verify_state_should_stop(struct thread_data *td, uint64_t numberio)
 		return 0;
 
 	/* If the current seq is lower than the max issued seq, check to make sure
-	 * the write was not inflight.
+	 * the write was not inflight (but exclude failed writes, they should be skipped not stopped).
 	 */
 	if (numberio < s->numberio) {
-		for (i = 0; i < s->depth; i++) {
+		/* Check only the actual inflight array (first td->o.iodepth entries) */
+		int actual_depth = (s->depth > td->o.iodepth) ? td->o.iodepth : s->depth;
```
vincentkfu: Please separate declaration from assignment and use an if statement to make the code easier to read.

backend.c
vincentkfu: Please add a comment here explaining that this is for the case where we have `rw=[rand]write verify_only=1 verify_state_load=1` or we previously ran a job with `rw=[rand]write verify_state_save=1` and are now running it with `rw=[rand]read verify_state_load=1`

vincentkfu:
Ok, I think it would be better the split this PR into one for the first patch and then a second one for the remaining patches.

Please fix up the style issues (adding braces in two places and separating declaration and assignment [also use if instead of ternary expression] in two places).

And then add a comment explaining the code that handles `rw=[rand]read verify=X` use cases.

Finally, I consider using two fio invocations (one to write, one to verify) "offline" verification and using a single fio invocation "online" verification. Please revise the commit message to follow these definitions.

With all this done I think the first patch can go in.

vincentkfu:
This patch triggers failures with `t/verify.py` with `verify=pattern_hdr`.

https://github.com/vincentkfu/fio/actions/runs/26111932917

```
./fio --name=test --filesize=8k --verify=pattern_hdr --rw=write --verify_pattern=123 --debug=io,verify 
fio: set debug option io
fio: set debug option verify
verify   4407  td->trim_verify=0
test: (g=0): rw=write, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
fio-3.42-28-g1d7a
Starting 1 process
io       4409  declare unneeded cache test.0.0: 0/8192
io       4409  fill: io_u 0x5640659c8980: off=0x0,len=0x1000,ddir=1,file=test.0.0
io       4409  prep: io_u 0x5640659c8980: off=0x0,len=0x1000,ddir=1,file=test.0.0
verify   4409  using already filled verify pattern b=1 len=4096
verify   4409  log_inflight: numberio=0, inflight_idx=0
io       4409  queue: io_u 0x5640659c8980: off=0x0,len=0x1000,ddir=1,file=test.0.0
io       4409  complete: io_u 0x5640659c8980: off=0x0,len=0x1000,ddir=1,file=test.0.0
verify   4409  invalidate_inflight: numberio=0, inflight_idx=0
io       4409  fill: io_u 0x5640659c8980: off=0x1000,len=0x1000,ddir=1,file=test.0.0
io       4409  prep: io_u 0x5640659c8980: off=0x1000,len=0x1000,ddir=1,file=test.0.0
verify   4409  using already filled verify pattern b=1 len=4096
verify   4409  log_inflight: numberio=1, inflight_idx=0
io       4409  queue: io_u 0x5640659c8980: off=0x1000,len=0x1000,ddir=1,file=test.0.0
io       4409  complete: io_u 0x5640659c8980: off=0x1000,len=0x1000,ddir=1,file=test.0.0
verify   4409  invalidate_inflight: numberio=1, inflight_idx=0
verify   4409  starting loop
io       4409  declare unneeded cache test.0.0: 0/8192
verify   4409  get_next_verify: ret io_u 0x5640659c8980
io       4409  prep: io_u 0x5640659c8980: off=0x0,len=0x1000,ddir=0,file=test.0.0
verify   4409  verify_state_should_stop numberio=0
io       4409  queue: io_u 0x5640659c8980: off=0x0,len=0x1000,ddir=0,file=test.0.0
io       4409  complete: io_u 0x5640659c8980: off=0x0,len=0x1000,ddir=0,file=test.0.0
verify: bad header rand_seed 0, wanted 46386204153304124 at file test.0.0 offset 0, length 4096 (requested block: offset=0, length=4096)
verify   4409  exiting loop
fio: pid=4409, err=84/file:io_u.c:2316, func=io_u_sync_complete, error=Invalid or incomplete multibyte or wide character
io       4409  close ioengine psync
io       4409  free ioengine psync

test: (groupid=0, jobs=1): err=84 (file:io_u.c:2316, func=io_u_sync_complete, error=Invalid or incomplete multibyte or wide character): pid=4409: Tue May 19 13:17:05 2026
  read: IOPS=1000, BW=4000KiB/s (4096kB/s)(4096B/1msec)
    clat (nsec): min=7193, max=7193, avg=7193.00, stdev= 0.00
     lat (nsec): min=15490, max=15490, avg=15490.00, stdev= 0.00
    clat percentiles (nsec):
     |  1.00th=[ 7200],  5.00th=[ 7200], 10.00th=[ 7200], 20.00th=[ 7200],
     | 30.00th=[ 7200], 40.00th=[ 7200], 50.00th=[ 7200], 60.00th=[ 7200],
     | 70.00th=[ 7200], 80.00th=[ 7200], 90.00th=[ 7200], 95.00th=[ 7200],
     | 99.00th=[ 7200], 99.50th=[ 7200], 99.90th=[ 7200], 99.95th=[ 7200],
     | 99.99th=[ 7200]
  write: IOPS=2000, BW=8000KiB/s (8192kB/s)(8192B/1msec)
    clat (nsec): min=4067, max=25007, avg=14537.00, stdev=14806.82
     lat (nsec): min=16871, max=39304, avg=28087.50, stdev=15862.53
    clat percentiles (nsec):
     |  1.00th=[ 4080],  5.00th=[ 4080], 10.00th=[ 4080], 20.00th=[ 4080],
     | 30.00th=[ 4080], 40.00th=[ 4080], 50.00th=[ 4080], 60.00th=[24960],
     | 70.00th=[24960], 80.00th=[24960], 90.00th=[24960], 95.00th=[24960],
     | 99.00th=[24960], 99.50th=[24960], 99.90th=[24960], 99.95th=[24960],
     | 99.99th=[24960]
  lat (usec)   : 10=66.67%, 50=33.33%
  cpu          : usr=0.00%, sys=0.00%, ctx=0, majf=0, minf=25
     issued rwts: total=1,2,0,0 short=0,0,0,0 dropped=0,0,0,0

   READ: bw=4000KiB/s (4096kB/s), 4000KiB/s-4000KiB/s (4096kB/s-4096kB/s), io=4096B (4096B), run=1-1msec
  WRITE: bw=8000KiB/s (8192kB/s), 8000KiB/s-8000KiB/s (8192kB/s-8192kB/s), io=8192B (8192B), run=1-1msec

Disk stats (read/write):
  vda: ios=0/0, sectors=0/0, merge=0/0, ticks=0/0, in_queue=0, util=0.00%
```

minwooim:
> This patch triggers failures with `t/verify.py` with `verify=pattern_hdr`.
> 
> https://github.com/vincentkfu/fio/actions/runs/26111932917

This issue is due to missing check `!td->o.verify_pattern_bytes` (https://github.com/axboe/fio/pull/2086#discussion_r3250294388).  I've updated the commit with putting this condition to the if.

PR #2088, opened 2026-05-10
t/io_uring: add --hugetlb option for hugetlbfs IO buffers

ming1:
t/io_uring: add --hugetlb option for hugetlbfs IO buffers

Add --hugetlb <path> option to allocate IO buffers from a hugetlbfs file instead of posix_memalign. This allows observing IO performance effects of hugepage-backed buffers (fewer TLB misses, contiguous physical memory).

It is also useful for testing ublk SHMEM_ZC where the client and server share the same physical pages via hugetlbfs.

The hugetlb file is mmap'd with MAP_SHARED | MAP_POPULATE and a bump allocator hands out per-thread buffers. A mutex protects the allocator for concurrent thread initialization.

Signed-off-by: Ming Lei <tom.leiming@gmail.com>

### DIFF
--- a/t/io_uring.c
+++ b/t/io_uring.c
@@ -114,10 +114,18 @@ struct submitter {
 	int per_file_depth;
 	const char *filename;
 
+	void *htlb_buf;		/* pre-assigned hugetlb buffer */
+
 	struct file files[MAX_FDS];
 	unsigned nr_files;
 	unsigned cur_file;
 	struct iovec iovecs[];
+};
+
+struct htlb_alloc {
+	char *path;		/* hugetlbfs file path */
+	void *base;		/* mmap base */
+	size_t size;		/* total mmap size */
 };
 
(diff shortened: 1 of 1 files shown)


axboe:
Looks fine to me, but why not just anon huge pages?

axboe:
Ah for sharing the page(s). Makes sense. Maybe add this to struct submitter and have it be assigned upfront before the thread starts? Then you don't need a mutex for an allocator, you just have the main thread divvy out the memory before the start is started.

ming1:
> Ah for sharing the page(s). Makes sense. Maybe add this to struct submitter and have it be assigned upfront before the thread starts? Then you don't need a mutex for an allocator, you just have the main thread divvy out the memory before the start is started.

Indeed, it is more clever to just add the htlb offset into `struct submitter`,
totally lockless, will do it in V2.

PR #2091, opened 2026-05-19
Introduce crash-aware data verification

minwooim:
Background

When testing NVMe crash consistency via io_uring_cmd, the NVMe
controller bypasses the block layer and handles flush SQEs directly.
Unlike block-layer I/O where a flush drains all preceding writes
before reaching the device, concurrent write SQEs and a flush SQE
submitted to an NVMe controller carry no ordering guarantee. As a
result, after a power loss or intentional reset, only writes covered
by the last flush command are guaranteed to persist — writes submitted
after the most recent flush may or may not survive.

The existing verify flow has no awareness of this crash boundary.
When --verify_state_load is used for offline verification, fio
replays all recorded writes regardless of whether they were fsynced
before the crash. This causes spurious verification failures on
writes that were never guaranteed to persist.

This patch sets fixed the following points:

1. do_dry_run() does not check the termination point, so it logs
io_hist entries beyond what was actually written during the
(possibly interrupted) write phase.
2. When do_dry_run() returns 0 (no verifiable I/Os), the thread is
not marked to terminate, leaving the outer loop spinning.
3. With --verify_only=1 and a loaded state file, keep_running()
can loop indefinitely because the byte limit derived from the
state file may be less than the jobfile limit.

Along with them, this patchset introduced `--verify_policy=` option
with a value named `fsynced` first to limit verification to only the writes
offsets covered by the last fsync.

---

This patchset has tested with the QEMU device with modifications to emit
errors in specific offsets.

Example scenario:
**Write Phase** of `--verify_policy=fsynced`
```
Ws0
Ws1
Ws2
        Wc0
        Wc1
Ws3
        Wc2 (errored)     <---- This will not be verified (skipped)
        Wc3
Fs0
Ws4
Ws5
        Wc4
Ws6
        Wc5
Ws7
        Wc6
        Wc7
        Fc0
Fs1                   <---- Last fsync submission, WRITEs after this submission will not be verified
Ws8
Ws9
        Wc8
        Wc9
        Fc1
<DONE>
```

**Verify Phase**
offsets 0, 1, 3, 4, 5, 6, 7 will be verified.


minwooim:
The first patch of this series is from https://github.com/axboe/fio/pull/2086.

vincentkfu:
If you introduce a new option you need to bump the server version.

HOWTO.rst
```diff
@@ -4156,6 +4156,27 @@ Verification
...
         offset. Default: 0 (disabled).
 
+.. option:: verify_policy=str
+
+        Controls which writes are included during offline verification.  Only
+        takes effect when :option:`verify_state_save` is used to save state
+        during the write phase and a separate job later verifies with
+        :option:`verify_state_load`.
+
+        **none**
```
vincentkfu: Instead of "none" I would call this "completed"

HOWTO.rst
```diff
@@ -4156,6 +4156,27 @@ Verification
...
+                Verify all completed writes (default).
+
+        **fsynced**
+                Only verify writes that were covered by the last fsync.
+                Writes issued after the most recent fsync completion are
+                excluded, because they may not be persistent after a crash or
+                power loss.  fio maintains two in-flight tracking buffers;
```
vincentkfu: This explanation does not seem to match how this feature works.

verify.h
```diff
@@ -32,6 +32,12 @@ enum {
 	VERIFY_NULL,			/* pretend to verify */
 };
 
+/* Values for the verify_policy option (bitmask) */
+enum {
+	VERIFY_POLICY_NONE    = 0,
```
vincentkfu: `VERIFY_POLICY_COMPLETED`

vincentkfu:
Please rewrite the commit message for the first patch according to this format:

- Start with a paragraph explaining what the feature is
- Explain how this patch modifies the verify state feature to accomplish this
- Explain the change made to the io_uring_cmd ioengine to support this feature
- Retain the current final paragraph noting that other ioengines do not require clearing the queue before a flush because the block layer does it

vincentkfu:
Why do you need to save the verify policy to the verify state file? As far as I can tell it is not needed.

During offline verification with a verify state file, jobs stop under two conditions

1. fio tries to queue an inflight write
2. fio tries to queue a write exceeding the max numberio from the original workload

It appears that flush-aware verify state file loading is using condition 2 to signal verify jobs to stop. Please add this explanation to the commit message and in the comment where fio is saving this value.

backend.c
```diff
@@ -1790,6 +1790,14 @@ static bool keep_running(struct thread_data *td)
 		td->o.loops--;
 		return true;
 	}
+
+	/*
+	 * Since `do_dry_run()` might stop before the byte limit which can't be
+	 * terminated by bytes limit.
+	 */
+	if (td->o.verify_only && td->vstate)
```
vincentkfu, line 1800: Will this handle a `verify_state_load` job with `rw=[rand]read`?

The commit message mentions the number of IOs wheereas the comment mentions the number of bytes. Which limit is of concern here or are both limits relevant? Make the comment and the commit message consistent.

vincentkfu:
Why is "backend: check whether to stop in ``do_dry_run()``" necessary if fio is already checking `verify_state_should_stop` in `do_verify`?

backend.c
minwooim: > Will this handle a `verify_state_load` job with `rw=[rand]read`?

Yes, for `rw=[rand]rw` cases with `verify_state_load=1`, it will go through `do_io()` rather than `do_dry_run()`.  In this place, `verify_state_should_stop()` goes early termination when the verify state is exhausted.

> The commit message mentions the number of IOs wheereas the comment mentions the number of bytes. Which limit is of concern here or are both limits relevant? Make the comment and the commit message consistent.

Without this fix, `keep_running()` will return `true` causing another superflous loop since neither the byte limit nor numberios has been reached to the limit.  So, bytes and numberios are both relevant.  Will update the comment correspondingly.

HOWTO.rst
minwooim: Makes sense.  Will update it!

HOWTO.rst
minwooim: Uh, missed that one.  Will update it and review all the other comments also correspondingly.

Thanks for the review.

minwooim:
> Why do you need to save the verify policy to the verify state file? As far as I can tell it is not needed.
> 
> During offline verification with a verify state file, jobs stop under two conditions
> 
> 1. fio tries to queue an inflight write
> 2. fio tries to queue a write exceeding the max numberio from the original workload

You're right, it should have been removed after the whole implementation.  I'll remove the `verify_policy` field from `struct thread_io_list`.

> It appears that flush-aware verify state file loading is using condition 2 to signal verify jobs to stop. Please add this explanation to the commit message and in the comment where fio is saving this value.

Sure, will update.

minwooim:
> Why is "backend: check whether to stop in `do_dry_run()`" necessary if fio is already checking `verify_state_should_stop` in `do_verify`?

`do_dry_run()` runs before `do_verify()` where `io_hist_tree` is built up by replaying the write phase.  Without this, offsets beyond the threshold get logged, but especially with overlap risks, `io_hist_tree` might be corrupted due to beyond offsetse.

engines/io_uring.c
```diff
@@ -1484,6 +1506,14 @@ static int fio_ioring_init(struct thread_data *td)
 		return 1;
 	}
 
+	if (o->writefua && (td->o.verify_policy & VERIFY_POLICY_FSYNCED)) {
+		log_err("fio: writefua=1 is incompatible with verify_policy=fsynced. "
+			"FUA writes are individually committed to storage media "
+			"and do not require fsync coverage tracking. "
+			"Use verify_policy=none(default) instead.\n");
```
vincentkfu: `completed` instead of `none`

t/verify-state.c
```diff
@@ -22,6 +23,7 @@ static void show_s(struct thread_io_list *s, unsigned int no_s)
 	printf("Thread:\t\t%u\n", no_s);
 	printf("Name:\t\t%s\n", s->name);
 	printf("Depth:\t\t%llu\n", (unsigned long long) s->depth);
+	printf("Verify policy:\t%s\n", s->verify_policy & VERIFY_POLICY_FSYNCED ? "fsynced" : "none");
```
vincentkfu: `completed` instead of `none`

HOWTO.rst
vincentkfu: This still needs to be changed

t/verify-state.c
```diff
@@ -22,6 +23,7 @@ static void show_s(struct thread_io_list *s, unsigned int no_s)
 	printf("Thread:\t\t%u\n", no_s);
 	printf("Name:\t\t%s\n", s->name);
 	printf("Depth:\t\t%llu\n", (unsigned long long) s->depth);
+
```
vincentkfu: Please remove this blank line

HOWTO.rst
```diff
@@ -4156,6 +4156,26 @@ Verification
         used to speed up the process of writing each block on a device with its
         offset. Default: 0 (disabled).
 
+.. option:: verify_policy=str
```
vincentkfu, line 4159: Please update fio.1 with these same changes.

verify.c
```diff
@@ -1682,6 +1682,26 @@ struct all_io_list *get_all_io_list(int save_mask, size_t *sz)
...
+		 */
+		if (td->o.verify_policy & VERIFY_POLICY_FSYNCED) {
+			/*
+			 * For `verify_policy=fsynced`, store safe inflight
+			 * threshold numberio to terminate verification due to
+			 * condition 2.  If no fsync has completed yet, store 0
+			 * so that condition 2 stops all verification
+			 * immediately.
+			 */
+			s->numberio = td->safe_inflight_issued ?
```
vincentkfu: Please change this to an `if/else` statement to improve readability.

verify.c
```diff
@@ -1691,7 +1711,7 @@ struct all_io_list *get_all_io_list(int save_mask, size_t *sz)
 
 		total_depth = td->o.iodepth + td->failed_numberio_count;
 		s->depth = cpu_to_le32((uint32_t) total_depth);
-		s->numberio = cpu_to_le64((uint64_t) atomic_load_acquire(&td->inflight_issued));
+
```
vincentkfu: Please remove this blank line

verify.c
```diff
@@ -1990,8 +2010,18 @@ int verify_state_should_stop(struct thread_data *td, uint64_t numberio)
...
 	dprint(FD_VERIFY, "verify_state_should_stop numberio=%"PRIu64"\n", numberio);
-	if (!s)
+	if (!s) {
+		/*
+		 * If verify state is NULL, it means that the current verify
+		 * sessfion is online verification.  We can simply see the cached
+		 * value in @td only in case of --verify_policy=fsynced and there was
+		 * at least one fsync happened.
+		 */
+		if (td->o.verify_policy & VERIFY_POLICY_FSYNCED)
```
vincentkfu: Please add curly braces around multi-line blocks even if they are only a single statement.

backend.c
```diff
@@ -1856,14 +1857,26 @@ static uint64_t do_dry_run(struct thread_data *td)
...
-		io_u->error = 0;
-		io_u->resid = 0;
+		/*
+		 * Check numberio quickly and determinte whether go or no-go
+		 * since write phase might have terminated in the middle of the
+		 * session.
+		 */
 		if (ddir_rw(acct_ddir(io_u))) {
-			io_u->numberio = td->io_issues[acct_ddir(io_u)];
+			numberio = td->io_issues[acct_ddir(io_u)];
```
vincentkfu: What happens if you run a 90/10 read/write workload? You will end up comparing a read IO's numberio against a write IOs numberio and stop prematurely.
Consider moving this to the `if (td_write(td)...)` block below.

minwooim: You're right.  `numberio` should have been given to ``verify_state_should_stop()`` only for `io_issues` of WRITE commands.  I will move it to the below and make `io_u` put if the given `numberio` should be stopped.

I guess this change should have been in commit cd2c17d14432f8608af66a264022648ef028169d (1st) where the actual ``verify_state_should_stop`` has been added.  Will make the change below to the corresponding commit.

```diff
@@ -1895,8 +1883,11 @@ static uint64_t do_dry_run(struct thread_data *td)
                    td->o.do_verify &&
                    td->o.verify != VERIFY_NONE &&
                    !td->o.experimental_verify) {
-                       if (!verify_state_should_skip(td, io_u->numberio) &&
-                           !verify_state_should_stop(td, io_u->numberio))
+                       if (verify_state_should_stop(td, io_u->numberio)) {
+                               put_io_u(td, io_u);
+                               break;
+                       }
+                       if (!verify_state_should_skip(td, io_u->numberio))
                                log_io_piece(td, io_u);

```

t/verify-state.c
minwooim: Done.  Sorry for the churns.

HOWTO.rst
minwooim: Done, will take care of those two as a pair from the next time.

verify.c
minwooim: Done, will keep that.

verify.c
minwooim: Sorry for the churns.  removed!

verify.c
minwooim: Updated

t/verify-state.c
```diff
@@ -12,6 +12,7 @@
 #include "../log.h"
 #include "../os/os.h"
 #include "../verify-state.h"
+#include "../verify.h"
```
vincentkfu: This is not needed

backend.c
```diff
@@ -1212,7 +1253,8 @@ static void do_io(struct thread_data *td, uint64_t *bytes_done)
 				populate_verify_io_u(td, io_u);
 				log_inflight(td, io_u);
 			}
-		}
+		} else if (ddir_sync(io_u->ddir))
+			on_fsync_submitted(td, io_u);
```
vincentkfu, line 1257: Please put {} around `on_fsync_submitted()` since this is part of a conditional statement where another branch uses {}.

verify.c
```diff
@@ -1990,8 +2011,19 @@ int verify_state_should_stop(struct thread_data *td, uint64_t numberio)
 	int i;
 
 	dprint(FD_VERIFY, "verify_state_should_stop numberio=%"PRIu64"\n", numberio);
-	if (!s)
+	if (!s) {
+		/*
+		 * If verify state is NULL, it means that the current verify
+		 * sessfion is online verification.  We can simply see the cached
```
vincentkfu: session

vincentkfu:
I see an error when running the following:

```
vincent@fio-dev:~/fio-dev/2091/fio-verify-policy$ ./fio --name=test --filename=test --filesize=1M --verify_state_save=1 --verify=crc32c --rw=randwrite --verify_policy=fsynced --fsync=16
test: (g=0): rw=randwrite, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
fio-3.3-3118-ga1ce
Starting 1 process

test: (groupid=0, jobs=1): err= 0: pid=6132: Wed Jun  3 12:13:32 2026
  read: IOPS=18.5k, BW=72.1MiB/s (75.6MB/s)(960KiB/13msec)
    clat (usec): min=33, max=120, avg=50.16, stdev=16.87
     lat (usec): min=33, max=120, avg=50.23, stdev=16.89
    clat percentiles (usec):
     |  1.00th=[   35],  5.00th=[   37], 10.00th=[   38], 20.00th=[   40],
     | 30.00th=[   41], 40.00th=[   42], 50.00th=[   44], 60.00th=[   45],
     | 70.00th=[   49], 80.00th=[   59], 90.00th=[   79], 95.00th=[   88],
     | 99.00th=[  111], 99.50th=[  112], 99.90th=[  121], 99.95th=[  121],
     | 99.99th=[  121]
  write: IOPS=14.2k, BW=55.6MiB/s (58.3MB/s)(1024KiB/18msec)
    clat (nsec): min=1723, max=31730, avg=3356.52, stdev=3479.53
     lat (nsec): min=2665, max=37911, avg=4921.44, stdev=4159.64
    clat percentiles (nsec):
     |  1.00th=[ 1736],  5.00th=[ 1832], 10.00th=[ 1928], 20.00th=[ 2040],
     | 30.00th=[ 2352], 40.00th=[ 2448], 50.00th=[ 2544], 60.00th=[ 2672],
     | 70.00th=[ 2800], 80.00th=[ 2992], 90.00th=[ 3728], 95.00th=[ 8768],
     | 99.00th=[21120], 99.50th=[28544], 99.90th=[31616], 99.95th=[31616],
     | 99.99th=[31616]
  lat (usec)   : 2=8.06%, 4=39.11%, 10=2.42%, 20=1.41%, 50=36.49%
  lat (usec)   : 100=11.69%, 250=0.81%
  fsync/fdatasync/sync_file_range/syncfs:
    sync (usec): min=675, max=2148, avg=1091.84, stdev=444.40
    sync percentiles (usec):
     |  1.00th=[  676],  5.00th=[  676], 10.00th=[  750], 20.00th=[  791],
     | 30.00th=[  807], 40.00th=[  807], 50.00th=[  848], 60.00th=[  848],
     | 70.00th=[ 1467], 80.00th=[ 1532], 90.00th=[ 1598], 95.00th=[ 2147],
     | 99.00th=[ 2147], 99.50th=[ 2147], 99.90th=[ 2147], 99.95th=[ 2147],
     | 99.99th=[ 2147]
  cpu          : usr=0.00%, sys=20.00%, ctx=279, majf=0, minf=23
  IO depths    : 1=103.0%, 2=0.0%, 4=0.0%, 8=0.0%, 16=0.0%, 32=0.0%, >=64=0.0%
     issued rwts: total=240,256,0,15 short=0,0,0,0 dropped=0,0,0,0

   READ: bw=72.1MiB/s (75.6MB/s), 72.1MiB/s-72.1MiB/s (75.6MB/s-75.6MB/s), io=960KiB (983kB), run=13-13msec
  WRITE: bw=55.6MiB/s (58.3MB/s), 55.6MiB/s-55.6MiB/s (58.3MB/s-58.3MB/s), io=1024KiB (1049kB), run=18-18msec

Disk stats (read/write):
  vda: ios=0/0, sectors=0/0, merge=0/0, ticks=0/0, in_queue=0, util=0.00%
vincent@fio-dev:~/fio-dev/2091/fio-verify-policy$ ./fio --name=test --filename=test --filesize=1M --verify_state_load=1 --verify=crc32c --rw=randwrite --verify_policy=fsynced --verify_only=1
test: (g=0): rw=randwrite, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
fio-3.3-3118-ga1ce
Starting 1 process
Stop verify because seq 240 >= 240
fio: ioengines.c:343: td_io_queue: Assertion `(io_u->flags & IO_U_F_FLIGHT) == 0' failed.
fio: pid=6144, got signal=6

test: (groupid=0, jobs=1): err= 0: pid=6144: Wed Jun  3 12:13:51 2026
  lat (msec)   : >=2000=99.59%
     submit    : 0=0.0%, 4=0.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
     complete  : 0=0.0%, 4=0.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
     issued rwts: total=0,241,0,0 short=0,0,0,0 dropped=0,0,0,0

Disk stats (read/write):
  vda: ios=0/0, sectors=0/0, merge=0/0, ticks=0/0, in_queue=0, util=0.00%
fio: file hash not empty on exit
```

This stops because the `IO_U_F_FLIGHT` flag is not cleared when `verify_state_should_stop` returns true in `do_dry_run`.

The previous version of this patch avoided this problem because `verify_state_should_stop` was checked before the flag was set and other accounting was carried out. Consider moving the `verify_state_should_stop` check before the flag is set and accounting is carried out but make sure that this stopping criteria is carried out only when appropriate. This seems easier than unwinding the flag and accounting when `verify_state_should_stop` returns true.

minwooim:
@vincentkfu , I've updated the branch with testing.  Thanks for the review.

PR #2094, opened 2026-05-22
Implement Sequence Distribution feature with Striding

vadlakondaswetha:
This PR introduces a new generic random distribution primitive: sequence.

While fio currently provides  primitives for linear, pseudo-random
(randread), and strided access patterns, it lacks a mechanism to simulate
deterministic, local non-linearity within a repeating stride.

This access pattern is required to test modern workloads like:

**LLM Inference Weights Loading**
Reading .safetensors model files which contain a list of tensor files.
Even though the file is read in sequential mode, within a tensor, data
will be requested in a particular order which is not purely sequential.

**Database Engine Log-Merging / LSM-Trees**

Scenarios where specific block indices (like parity blocks or metadata
headers) must be systematically read out-of-order within every repeating
chunk or block group.

Existing option is to use read_iolog.  Using a generated `read_iolog` for multi-terabyte model benchmarks is
sub-optimal as the trace files become massive, unscalable, and lack
dynamic flexibility across varying block sizes. The `sequence`
distribution resolves this by calculating offsets algorithmically on
the fly.

NEW OPTIONS INTRODUCED
----------------------

- random_distribution=sequence:N1,N2,N3 : Intercepts the random offset generator to utilize the pattern sequence.
- random_sequence_stride: Boolean option to modify `sequence` distribution behavior.
> - When disabled (0, default), the sequence loops over absolute block indices.
> - When enabled (1), it switches to a Strided Block Group pattern, advancing the base block index by the sequence length after each cycle to progress through the file.


sitsofe:
@vadlakondaswetha:
(I'm going purely on the description in your PR) What is this providing over something like:
```
./fio --debug=io --name=go --rw=randread --size=64k --zonemode=strided --ioengine=null --zonesize=16k
```
which performs I/O within the zone at random offsets but restricted to a given zone until it's "full" (at which point it moves on to the next zone)? Is it that the precise order of access within a zone is under directly user specified rather than being sequential or random?

vadlakondaswetha:
> @vadlakondaswetha: (I'm going purely on the description in your PR) What is this providing over something like:
> 
> ```
> ./fio --debug=io --name=go --rw=randread --size=64k --zonemode=strided --ioengine=null --zonesize=16k
> ```
> 
> which performs I/O within the zone at random offsets but restricted to a given zone until it's "full" (at which point it moves on to the next zone)? Is it that the precise order of access within a zone is under directly user specified rather than being sequential or random?

Thats right. This provides reading data in a precise order controlled by the user.

Right now, repeating the same sequence across the entire file is controlled by random_sequence_stride option. We can remove this and achieve the same by the combination of random_distribution and zone option. Its a bit complex to configure. Eg:

`
--random_distribution=sequence:2,0,1 --zonemode=strided --zonesize=12k   --zonerange=12k --zoneskip=0
`

sitsofe:
@vadlakondaswetha:
I see - you did mention just what happens in the commit message and I foolishly missed it so that's on me. I'm not promising to re-review (I'm not saying I don't like the PR just that these days I don't have much time and others tend to a more timely job).

1. For a new feature we're going to need man page (`fio.1`) and HOWTO (`HOWTO.rst`) documentation to be part of the commit
2. An example file with comments would be helpful but possibly point 1. will be enough...
3. I'm a bit unsure of this:
    ```c
	if (td->o.random_sequence) {
		free(td->o.random_sequence);
		td->o.random_sequence = NULL;
	}

	td->o.random_sequence = malloc(td->o.random_sequence_nr * sizeof(unsigned int));
	if (!td->o.random_sequence) {
		free(str);
		return 1;
	}
    ```
    because it feels strange to reallocating option values but I'm not that familiar with fio's option parsing so perhaps this is pre-existing pattern.

io_u.c
```diff
@@ -291,6 +291,28 @@ static int __get_next_rand_offset_zoned(struct thread_data *td,
...
+{
+	uint64_t io_count;
+	unsigned int idx;
+
+	if (!td->o.random_sequence_nr)
+		return 1;
+
+	io_count = td->io_issues[ddir];
+	idx = io_count % td->o.random_sequence_nr;
+
```
ankit-sam: extra tab here

t/sequence.py
```diff
@@ -0,0 +1,140 @@
...
+        lines = self.iops_log_lines.split('\n')
+        io_count = 0
+
+        for line in lines:
+            if len(line) == 0:
+                continue
+
+            tokens = line.split(',')
+            offset = int(tokens[4])
+
```
ankit-sam: extra tabs here

t/sequence.py
```diff
@@ -0,0 +1,140 @@
...
+            tokens = line.split(',')
+            offset = int(tokens[4])
+            
+            idx = io_count % seq_len
+            if stride:
+                group_idx = io_count // seq_len
+                expected_block = group_idx * seq_len + expected_seq[idx]
+            else:
+                expected_block = expected_seq[idx]
+
```
ankit-sam: same extra tab here

options.c
```diff
@@ -1340,6 +1340,54 @@ static int parse_zoned_distribution(struct thread_data *td, const char *input,
...
+		return 1;
+	}
+
+	p = str;
+	while ((n = strsep(&p, ",")) != NULL) {
+		if (*n == '\0') {
+			log_err("fio: empty element in random_sequence\n");
+			goto err;
+		}
+		td->o.random_sequence[i++] = atoi(n);
```
ankit-sam, line 1391: Maybe a check for it to only accept valid values i.e. anything negative or anything outside the window (0 to number of IOs) should fail. As I see  these errors 

```
failed getting offset
setting file failed
```

ankit-sam (commented):
Hi @vadlakondaswetha I added a couple of review comments, please check them.

Apart from what @sitsofe mentioned about the documentation and example file, I have few other concerns
1. The sequence values are not checked, so user can pass duplicate values. This breaks  norandommap=0 behavior.
2. For mix workloads, do we need separate sequence like random_sequence_read and random_sequence_write? As in that case both read and write offsets will be identical everytime.
3. Not urgent, but I think it will be great to support ranges like `0-5, 10-20`. we already have options like `bsrange, plids` etc.

options.c
vadlakondaswetha: Thanks for pointing out. Added required validations.

vadlakondaswetha:
> Hi @vadlakondaswetha I added a couple of review comments, please check them.
> 
> Apart from what @sitsofe mentioned about the documentation and example file, I have few other concerns
> 
> 1. The sequence values are not checked, so user can pass duplicate values. This breaks  norandommap=0 behavior.
> 2. For mix workloads, do we need separate sequence like random_sequence_read and random_sequence_write? As in that case both read and write offsets will be identical everytime.
> 3. Not urgent, but I think it will be great to support ranges like `0-5, 10-20`. we already have options like `bsrange, plids` etc.

Thanks for the review.
1. In fio's core init.c (https://github.com/axboe/fio/blob/fio-3.42/init.c#L1025), any use of a non-uniform random distribution automatically overrides and turns off the random map. norandommap gets disabled for sequence pattern also (FIO_RAND_DIST_SEQUENCE != FIO_RAND_DIST_RANDOM)
2. When running mixed workloads with standard non-uniform distribution, fio applies identical distribution formulas to both reads and writes. Same behavior applies for sequence config also. Incase they need to apply different patterns, its better to use different jobs.
3. Thats a good addition. But I would like it keep it out of this PR.

vadlakondaswetha:
> @vadlakondaswetha: I see - you did mention just what happens in the commit message and I foolishly missed it so that's on me. I'm not promising to re-review (I'm not saying I don't like the PR just that these days I don't have much time and others tend to a more timely job).
> 
> 1. For a new feature we're going to need man page (`fio.1`) and HOWTO (`HOWTO.rst`) documentation to be part of the commit
> 2. An example file with comments would be helpful but possibly point 1. will be enough...
> 3. I'm a bit unsure of this:
>    ```c
>    if (td->o.random_sequence) {
>    	free(td->o.random_sequence);
>    	td->o.random_sequence = NULL;
>    }
>    
>    td->o.random_sequence = malloc(td->o.random_sequence_nr * sizeof(unsigned int));
>    if (!td->o.random_sequence) {
>    	free(str);
>    	return 1;
>    }
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    because it feels strange to reallocating option values but I'm not that familiar with fio's option parsing so perhaps this is pre-existing pattern.

Thanks for the review :)

1 & 2: Added required files
3. this is absolutely necessary because fio's option parsing callbacks (.cb) are evaluated incrementally as options are read. If an option is specified multiple times (for example, overridden in a job section or passed again on the CLI), the callback is entirely responsible for storing and managing the value. Since our value is a dynamically sized array of integers (unsigned int *), we must free() any previous allocation before performing a new malloc() to prevent a memory leak during repeated or overridden option evaluations.

vadlakondaswetha:
@sitsofe  and @ankit-sam  - PTAL, replied to your comments. Thanks.

options.c
ankit-sam: `fio --name=sequence --filename=/dev/nvme0n1 --ioengine=io_uring --size=100K --bs=4K --rw=randread --random_distribution=sequence:2,0,1 --debug=io --random_sequence_stride=1`

In this scenario, after some iteration I see
`fio: sequence access block 26 out of bounds for file /dev/nvme0n1`

I saw the checks, and fio stops if it gets anything beyond the `<offset, offset + size>` window, which will happen for any workload, with sequence not starting from 0.

I can't think of any better way to handle this, maybe skipping that offset and trying out the next one until we hit a valid offset?
And in the above example you will probably have to consider sequences like `2,100,1` where 100 is entirely out of bound (0-24).

vadlakondaswetha:
Hi @vincentkfu  / @axboe  , Gentle ping on this PR review. PTAL and let me know if there are any concerns/questions for adding this feature.

fio.1
```diff
@@ -1422,8 +1426,27 @@ Similarly to how \fBbssplit\fR works for setting ranges and percentages
...
+.P
+For a \fBsequence\fR distribution, fio supports specifying a fixed deterministic
+sequence of block indices to be accessed. For example, to access block indices
+2, 0, and 1 in that exact order, the user would do:
+.RS
+.P
+random_distribution=sequence:2,0,1
+.RE
+.P
+This repeats the block access pattern indefinitely. Up to 64 sequence
```
vincentkfu, line 1438: The number of indices is unbounded, no?

HOWTO.rst
```diff
@@ -1625,6 +1628,24 @@ I/O type
 	is given, it'll apply to all of them. This goes for both **zoned**
 	**zoned_abs** distributions.
 
+	For a **sequence** distribution, fio supports specifying a fixed deterministic
+	sequence of block indices to be accessed. For example, to access block indices
+	2, 0, and 1 in that exact order, the user would do::
+
+		random_distribution=sequence:2,0,1
+
+	This repeats the block access pattern indefinitely. Up to 64 sequence
```
vincentkfu, line 1637: Same comment as for `fio.1`

t/sequence.py
```diff
@@ -0,0 +1,140 @@
...
+    parser.add_argument('-o', '--run-only', nargs='+', type=int,
+                        help='list of test(s) to run, skipping all others')
+    return parser.parse_args()
+
+
+def main():
+    """Run sequence tests."""
+    args = parse_args()
+
+    fio_path = str(Path(args.fio).absolute()) if args.fio else 'fio'
```
vincentkfu: I think it would be better to use the location of this script to infer the location of the desired fio executable when none is specified as is done here:

https://github.com/axboe/fio/commit/d72bcf71534b5ac82359e3b987ff109b0de57af0

fio.1
vadlakondaswetha: Earlier, the number of indices are capped at 64 for client/server mode as the network transmission structure requires a fixed array size for serialization. 

Now i made changes to apply the same limit for local mode also to keep it consistent. Documentation already says 64 as the limit.

HOWTO.rst
vadlakondaswetha: replied here: https://github.com/axboe/fio/pull/2094/changes#r3469958394

options.c
vadlakondaswetha: @ankit-sam  - sorry, I missed this comment as it in resolved state.

I updated the code to handle this. The code will now distinguish between normal end-of-file boundary and invalidate index.

- Invalidate index specified in the sequence - Eg 2, 100, 1-> fio will log a error and abort
- End-of-file boundary: If out of bounds is reached because the progressing stride went past the end of file, fio will now handle it as silent EOF and terminate the job normally without logging errors. I am not taking of route of continuing till we find the next valid offset incases like sequence=2,1,0 bs=1M, filesize = 4MB. It will stop after reading 3MB of data.

t/sequence.py
vadlakondaswetha: done.

fio.1
vincentkfu: Actually client/server mode allows transmission of a variable-sized payload for the `buffer_pattern` and `verify_pattern` options, so it would be possible to allow this to be unbounded in both client/server and local mode. Nonetheless, what you have done is reasonable and consistent with how we are handling `random_distribution=zoned`.

Issue #2093, opened 2026-05-21
Potential Use-After-Free in request_client_etas (client.c) during dprint logging


RigelYoung:
**Description of the bug**

During a code audit using a static analysis tool, I found a Use-After-Free (UAF) / Dangling Pointer issue in client.c.

In the `request_client_etas function`, the `eta` pointer can be freed inside the `while (skipped--)` loop via `fio_client_dec_jobs_eta`, but it is subsequently used in a `dprint` statement outside the loop. While evaluating a freed pointer in `dprint (%p)` does not dereference it, it is strictly Undefined Behavior according to the C standard.

**Environment:**
N/A (Identified via Static Analysis)

**fio version:**
Latest master branch

**Reproduction steps**
This is a static analysis finding. The vulnerable execution path in the source code is as follows:

Allocation: eta is allocated dynamically:
```
eta = calloc(1, sizeof(*eta) + __THREAD_RUNSTR_SZ(REAL_MAX_JOBS));
```

Free: The code enters the `while (skipped--)` loop. If `fio_client_dec_jobs_eta(eta, ops->eta)` is called and `eta->pending` reaches `0`, the eta pointer is freed inside that function, and it returns 0.
```
static int fio_client_dec_jobs_eta(struct client_eta *eta, client_eta_op eta_fn)
{
    if (!--eta->pending) {
        eta_fn(&eta->eta);
        free(eta);      // <--- eta is freed here
        return 0;
    }
    return 1;
}
```

This return `value (0)` causes the while loop in `request_client_etas` to break.

Use-After-Free: Right after breaking out of the loop, the freed eta pointer is passed to dprint:

```
while (skipped--) {
    if (!fio_client_dec_jobs_eta(eta, ops->eta))
        break;
}

dprint(FD_NET, "client: requested eta tag %p\n", eta); // <--- UAF / Dangling pointer usage
Suggested Fix:
Move the dprint statement to before the while loop:

Diff
--- a/client.c
+++ b/client.c
@@ -2054,12 +2054,12 @@ static void request_client_etas(struct client_ops const *ops)
                fio_net_send_simple_cmd(client->fd, FIO_NET_CMD_SEND_ETA,
                                        (uintptr_t) eta, &client->cmd_list);
        }
 
+       dprint(FD_NET, "client: requested eta tag %p\n", eta);
+
        while (skipped--) {
                if (!fio_client_dec_jobs_eta(eta, ops->eta))
                        break;
        }
-
[...]

PR #2095, opened 2026-05-22
Fix Use-After-Free in request_client_etas during dprint logging

RigelYoung:
fix issue #2093


axboe:
See comment in issue, this is nonsense.

PR #2097, opened 2026-05-26
backend: remove linux/prctl.h include to fix musl conflict

### LINKED ISSUE #2096: Duplicate definition of prctl_mm_map error when building with musl
Building fio (release 3.42, but also the latest commit) on Alpine 3.23.4, throws the following error:
```
/usr/include/sys/prctl.h:88:8: error: redefinition of 'struct prctl_mm_map'
     88 | struct prctl_mm_map {
        |        ^~~~~~~~~~~~
  In file included from backend.c:35:
  /usr/include/linux/prctl.h:134:8: note: originally defined here
    134 | struct prctl_mm_map {
        |        ^~~~~~~~~~~~
  #/make: *** [Makefile:506: backend.o] Error 1
```

That is because Alpine uses musl, which completely redefines `linux/prctl.h` in `sys/prctl.h`. 
musl suggests [here](https://wiki.musl-libc.org/faq#Q:-Why-am-I-getting-) not to mix userspace and kernel headers.

My suggestion is simply removing the include of `linux/prctl.h` and only keeping `sys/prctl.h`.
This has also been implemented by other projects ([Perfetto](https://android-review.googlesource.com/c/platform/external/perfetto/+/3252813), [LLVM](https://github.com/llvm/llvm-project/issues/139443), [Klipper](https://klipper.discourse.group/t/musl-consider-removing-linux-prctl-h-from-klippy-chelper-pyhelper-c/24501/4)) to fix the same issue.
On my system this works when building with glibc as well as musl.
Alternatively we could add a guard that checks if glibc is used.

If you are fine with this, I will submit a PR.

lucashecht:
Building with musl fails the error `redefinition of 'struct prctl_mm_map'`. This is because musl completely redefines linux/prctl.h in sys/prctl.h. 

This commit removes the linux/prctl.h include and only keeps sys/prctl.h.

Fixes: #2096

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -32,7 +32,6 @@
 #include <pthread.h>
 
 #ifdef CONFIG_LINUX
-#include <linux/prctl.h>
 #include <sys/prctl.h>
 #endif
 


vincentkfu:
We should probably add an Alpine OS container to our CI.

PR #2089, opened 2026-05-13
Io uring/multiple write modes

minwooim:
Introduce multiple wmode(write_mode) options for DDIR_WRITE workload.

PR #2101, opened 2026-05-30
engines/nvme: fix addr/data_len for no data nvme commands

swarnagh6:
assign 'cmd->data_len' to xfer_buflen and 'cmd->addr' to xfer_buf for data transfer commands - read, write and compare opcode. Previously, write zeroes had 'cmd->data_len' assigned to xfer_buflen while the address was set to NULL causing bad address error. Any unrecognised opcode returns -ENOTSUP. This ensures no data transfer opcode - write zeroes, uncor and verify have addr and data_len set to 0.

Fixes: https://github.com/axboe/fio/pull/2089#pullrequestreview-4368826338


vincentkfu:
Please use a `switch` command for this code. I think `switch` would make the intent of this code even clearer. Just `break` for the no data transfer opcodes and add a comment that the buffer address and length are already zero.

PR #2102, opened 2026-06-03
engines/io_uring: Introduce BSG io uring cmd

ljw8161:
Background
  ----------

  This feature requires a kernel patch to work correctly on the latest Linux
  kernel ([eb3f4b7](https://github.com/torvalds/linux/commit/eb3f4b7)). 
  The io_uring_cmd engine was introduced to issue device-specific passthrough
  commands directly through io_uring, bypassing the block layer. Until now
  it was exclusively coupled to NVMe — the cmd_type option accepted only
  "nvme", and the internal data structures, initialization paths, and error
  reporting all assumed NVMe command semantics throughout.

  SCSI devices exposed via the kernel's BSG (Block SCSI Generic) interface
  can also be driven through io_uring using the bsg_uring_cmd path (upstream
  pretty recent), but fio had no way to exercise that path.

  What This Series Does
  ---------------------

  Patch 1 ("Separate NVMe-specific logic") is a pure preparatory refactor
  with no functional changes. It moves all NVMe-specific code — DSM
  allocation, cdw12 flag setup, write-opcode selection, PI flags parsing,
  and the NVMe-only ZBD callbacks (get_zoned_model, report_zones, reset_wp,
  get_max_open_zones, fetch_ruhs) — behind explicit FIO_URING_CMD_NVME
  guards. The intent is to make the engine's cmd_type dispatch structure
  uniform and correct before new device types are introduced.

  Patch 2 ("Add bsg support for io_uring_cmd engine") builds on top of that
  refactor to add a new cmd_type=bsg option. It introduces:

    - engines/bsg.c / engines/bsg.h: a self-contained BSG helper layer
      with big-endian helpers for CDB and data field manipulation.
    - SCSI command construction for the four fundamental I/O directions:
        read  → READ(10)
        write → WRITE(10)
        trim  → UNMAP
        fsync → SYNCHRONIZE CACHE(10)
    - FUA (Force Unit Access) support for read and write via the CDB FUA bit.
    - Device size detection via a READ CAPACITY(10) SG_IO ioctl issued at
      open/get_file_size time, populating the bsg_data per-file structure.
    - CQE result interpretation: the BSG 32-byte CQE packs SCSI device
      status in bits 0–7 and host status in bits 16–23 of big_cqe[0].
    - Error detail reporting that separates SCSI status from host status.
    - The same IORING_SETUP_SQE128 / IORING_SETUP_CQE32 ring flags already
      used by NVMe, since BSG passthrough also requires the extended SQE/CQE
      layout.

  How It Works
  ------------

  When cmd_type=bsg is selected the engine allocates a per-slot bsg_cmd
  array (indexed by io_u->index) at init time. At prep time,
  fio_bsg_uring_cmd_prep() fills the 128-byte SQE cmd payload (a
  struct bsg_uring_cmd / sg_io_v4) with the appropriate CDB, transfer
  direction pointers, and lengths. At completion time the driver reads the
  packed status word from big_cqe[0] and maps it to io_u->error.

  Usage example:
    fio --filename=/dev/bsg/0\\\:0\\\:0\\\:0 \
        --ioengine=io_uring_cmd --cmd_type=bsg \
        --rw=read --bs=4k --iodepth=32

 Testing
 -------
 The BSG io_uring passthrough path returns -ENOMEM
 during data transfers due to a conflict between commit [b520c4e](https://github.com/torvalds/linux/commit/b520c4e) ("block: split
 bio_alloc_bioset more clearly into a fast and slowpath") and the BSG
 io_uring handler commit [7b6d325](https://github.com/torvalds/linux/commit/7b6d325).

 After [b520c4e](https://github.com/torvalds/linux/commit/b520c4e), bio_alloc_bioset() returns NULL immediately when GFP_NOWAIT
 is used and slab allocation fails, bypassing the mempool entirely.
 scsi_bsg_uring_cmd() uses GFP_NOWAIT when IO_URING_F_NONBLOCK is set,
 causing allocation failures. The NVMe path is unaffected as it hardcodes
 GFP_KERNEL.

 The fix is to hardcode GFP_KERNEL in scsi_bsg_map_user_buffer() in the
 kernel, mirroring the NVMe passthrough path behavior.

 ```patch
 diff --git a/drivers/scsi/scsi_bsg.c b/drivers/scsi/scsi_bsg.c
 index e80dec53174e..3cb1f607815b 100644
 --- a/drivers/scsi/scsi_bsg.c
 +++ b/drivers/scsi/scsi_bsg.c
 @@ -91,10 +91,10 @@ static int scsi_bsg_map_user_buffer(struct request *req,
                                               &iter, ioucmd, issue_flags);
               if (ret < 0)
                       return ret;
 -               ret = blk_rq_map_user_iov(req->q, req, NULL, &iter, gfp_mask);
 +               ret = blk_rq_map_user_iov(req->q, req, NULL, &iter, GFP_KERNEL);
       } else {
               ret = blk_rq_map_user(req->q, req, NULL, uptr64(buf_addr),
 -                                     buf_len, gfp_mask);
 +                                     buf_len, GFP_KERNEL);
       }

       return ret;
 ```

I'm planning to post this issue to the kernel community.


ljw8161:
This CI failure is due to the BSG-related UAPI (see link below) not being updated in the CI system.
https://lore.kernel.org/all/20260317072226.2598233-2-yangxiuwei@kylinos.cn/

vincentkfu:
The convention for fio is to define all of the identifiers needed to build for new IO interfaces. Find an appropriate place in your patch to include the definitions needed for building. We cannot rely on the build environment having all of the needed data structures defined.

engines/io_uring.c
```diff
@@ -631,26 +633,47 @@ static int fio_ioring_cmd_prep(struct thread_data *td, struct io_u *io_u)
...
+			populate_verify_io_u(td, io_u);
+			read_opcode = nvme_cmd_compare;
+			io_u_set(td, io_u, IO_U_F_VER_IN_DEV);
+		}
+
+		return fio_nvme_uring_cmd_prep(cmd, io_u,
+				o->nonvectored ? NULL : &ld->iovecs[io_u->index],
+				dsm, read_opcode, ld->write_opcode,
+				ld->cdw12_flags[io_u->ddir]);
+	} else if (o->cmd_type == FIO_URING_CMD_BSG) {
```
ankit-sam: As we are already returning `-EINVAL` for invalid cmd_type earlier I think instead of
```
} else if (o->cmd_type == FIO_URING_CMD_BSG) {
...
```
we can just do
} else {
...

Also we can remove the last `-EINVAL`

engines/io_uring.c
```diff
@@ -1829,6 +1912,23 @@ static int fio_ioring_cmd_open_file(struct thread_data *td, struct fio_file *f)
...
+		struct bsg_data *data;
+		unsigned int bs = 0;
+		unsigned long long last_lba = 0;
+		int ret;
+
+		data = FILE_ENG_DATA(f);
+		if (data == NULL) {
+			data = calloc(1, sizeof(struct bsg_data));
+			ret = fio_bsg_uring_cmd_read_capacity(td, &bs, &last_lba);
```
ankit-sam, line 2020: I think it makes sense to move the data allocation after this, so that if this check fails, we don't need to allocate and free it.

ankit-sam (commented):
Hi @ljw8161 a couple of minor comments.
And can you update the fio documentation `HOWTO.rst` and `fio.1`
And it would be nice to add an example file. Please check `examples/uring-cmd-*`

engines/io_uring.c
ljw8161: Good catch! I've addressed your comments.

engines/io_uring.c
ljw8161: Thanks for the review! I've made the requested changes.

ljw8161:
Hello, @vincentkfu 
Thanks for the feedback. I've added the bsg_uring_cmd struct definition in bsg.h, guarded by #ifndef CONFIG_BSG_URING_CMD, following the same convention used for NVMe in nvme.h (#ifndef CONFIG_NVME_URING_CMD).

ljw8161:
Hello, @ankit-sam !
Thanks for the suggestions. I've updated HOWTO.rst and fio.1 to document the new bsg option for cmd_type, and added an example job file at examples/uring-cmd-bsg.fio and examples/uring-cmd-bsg.png

k831-kim:
Resolved the conflict and updated.

engines/io_uring.c
```diff
@@ -761,55 +753,75 @@ static int fio_ioring_cmd_prep(struct thread_data *td, struct io_u *io_u)
...
 
-	return fio_nvme_uring_cmd_prep(cmd, io_u,
-			o->nonvectored ? NULL : &ld->iovecs[io_u->index],
-			dsm, read_opcode, ld->write_opcode,
-			ld->cdw12_flags[io_u->ddir]);
+		return fio_nvme_uring_cmd_prep(cmd, io_u,
+				o->nonvectored ? NULL : &ld->iovecs[io_u->index],
+				dsm, read_opcode, ld->write_opcode,
+				ld->cdw12_flags[io_u->ddir]);
+	} else {
```
vincentkfu, line 835: This block doesn't belong in this patch.

engines/io_uring.c
```diff
@@ -1018,11 +1031,18 @@ static inline void fio_ioring_setup_pi(struct thread_data *td,
...
 {
 	struct ioring_data *ld = td->io_ops_data;
+	struct ioring_options *o = td->eo;
+	struct fio_file *f = io_u->file;
 
 	if (io_u->ddir == DDIR_TRIM)
 		return;
 
-	fio_nvme_generate_guard(io_u, &ld->ext_opts);
+	if (o->cmd_type == FIO_URING_CMD_NVME)
```
vincentkfu: Please add curly braces { } around this block even though it is only a single line becase the `else` block is multiple lines.

engines/io_uring.c
```diff
@@ -1618,13 +1640,18 @@ static int fio_ioring_init(struct thread_data *td)
...
 
 	}
-	parse_prchk_flags(o);
-	ext_opts = &ld->ext_opts;
-	if (o->pi_act)
-		ext_opts->io_flags |= NVME_IO_PRINFO_PRACT;
-	ext_opts->io_flags |= o->prchk;
-	ext_opts->apptag = o->apptag;
-	ext_opts->apptag_mask = o->apptag_mask;
+	// Currently only support NVMe device
```
vincentkfu: Use /* */ for comments.
I would remove this change from this patch and add a version of this change to your next patch.

engines/io_uring.c
```diff
@@ -1664,40 +1691,47 @@ static int fio_ioring_io_u_init(struct thread_data *td, struct io_u *io_u)
...
-		pi_attr->app_tag = o->apptag;
-		pi_attr->flags = 0;
-		if (o->prchk & NVME_IO_PRINFO_PRCHK_GUARD)
-			pi_attr->flags |= IO_INTEGRITY_CHK_GUARD;
-		if (o->prchk & NVME_IO_PRINFO_PRCHK_REF)
-			pi_attr->flags |= IO_INTEGRITY_CHK_REFTAG;
-		if (o->prchk & NVME_IO_PRINFO_PRCHK_APP)
-			pi_attr->flags |= IO_INTEGRITY_CHK_APPTAG;
-	}
+	// Currently only support NVMe device
```
vincentkfu: /* */ for comments
The idiom used elsewhere in this file for detecting whether metadata is enabled is:

```
       if (o->md_per_io_size && (!ld->is_uring_cmd_eng ||
           (ld->is_uring_cmd_eng && o->cmd_type == FIO_URING_CMD_NVME))) {
```

engines/io_uring.c
```diff
@@ -1716,9 +1750,7 @@ static int fio_ioring_io_u_init(struct thread_data *td, struct io_u *io_u)
 
 static void fio_ioring_io_u_free(struct thread_data *td, struct io_u *io_u)
 {
-	struct nvme_pi *pi = io_u->engine_data;
-
-	free(pi);
+	free(io_u->engine_data);
```
vincentkfu, line 1808: This is a reasonable cleanup but does not belong here.

engines/bsg.c
```diff
@@ -0,0 +1,195 @@
+#include "bsg.h"
+
+static inline uint32_t bsg_uring_cmd_get_be32(uint8_t *buf)
```
vincentkfu: These inline functions duplicate functions defined at the top of sg.c.

Please create another prep patch that creates a new header file (`sg.h`?) containing these functions and include it here and in `engines/sg.c` to define these inline functions.

engines/bsg.c
```diff
@@ -0,0 +1,195 @@
...
+int fio_bsg_uring_cmd_prep(struct bsg_uring_cmd *cmd, struct io_u *io_u,
+			   struct bsg_cmd *bc, bool fua)
+{
+	struct bsg_data *data = FILE_ENG_DATA(io_u->file);
+	unsigned long long offset, nlb;
+	int data_len;
+
+	if (io_u->xfer_buflen & (data->bs - 1)) {
+		log_err("read/write not sector aligned\n");
+		return EINVAL;
```
vincentkfu: Our convention is to return a negative errno

engines/io_uring.c
```diff
@@ -1979,6 +2045,11 @@ static int fio_ioring_cmd_close_file(struct thread_data *td,
 	if (o->cmd_type == FIO_URING_CMD_NVME) {
```
vincentkfu: How about this?
```
if cmd_type == nvme or bsg {
    free(FILE_ENG_DATA(f))
    FILE_SET_ENG_DATA(f, NULL)
} else {
    error message about unsupported command type 
}
```
It doesn't seem necessary to have two separate blocks for nvme and bsg since they do the same thing.

engines/io_uring.c
k831-kim: Yes, I will move to the next commit.

engines/io_uring.c
k831-kim: Ah, I see. I will enclose it with {}.

engines/io_uring.c
k831-kim: The content deals with NVMe‑specific PI processing, such as “NVME_IO_PRINFO_PRCHK_GUARD”. As you mentioned, handling it that way would cause block devices with metadata enabled to be used as well, so it seems preferable to set it to process only NVMe.

engines/io_uring.c
k831-kim: The existing code was NVMe‑specific, and since this was a commit that separated it, I placed it under that commit
should it be split into a separate commit?

engines/io_uring.c
k831-kim: The direction you mentioned seems good! I will change it.

engines/bsg.c
k831-kim: Okay I will change it

engines/bsg.c
k831-kim: Yes, I will add it based on your suggestion.

engines/io_uring.c
vincentkfu: In general a patch should try to accomplish only a single goal. The goal of this patch is to separate out NVMe-specific code. It is not to clean up the code.
That said, this is a very small change and it seems a bit silly to have an entire patch to just this change, so I think ultimately it is ok to keep it in this patch.

k831-kim: Okay, thanks for your opinion

engines/bsg.c
```diff
@@ -0,0 +1,169 @@
...
+		sgio_set_be32((uint32_t) nlb, &bc->unmap_param[16]);
+		cmd->dout_xferp = (uint64_t)(uintptr_t) bc->unmap_param;
+		cmd->dout_xfer_len = data_len;
+		return 0;
+	case DDIR_SYNC:
+		fio_bsg_uring_cmd_init(cmd, bc, io_u, SG_DXFER_NONE);
+		bc->cdb[0] = bsg_cmd_sync_cache_10;
+		break;
+	default:
+		return ENOTSUP;
```
vincentkfu: return a negative errno

engines/bsg.c
```diff
@@ -0,0 +1,169 @@
...
+		sgio_set_be16((uint16_t) data_len - 8, &bc->unmap_param[2]);
+		sgio_set_be64(offset, &bc->unmap_param[8]);
+		sgio_set_be32((uint32_t) nlb, &bc->unmap_param[16]);
+		cmd->dout_xferp = (uint64_t)(uintptr_t) bc->unmap_param;
+		cmd->dout_xfer_len = data_len;
+		return 0;
+	case DDIR_SYNC:
+		fio_bsg_uring_cmd_init(cmd, bc, io_u, SG_DXFER_NONE);
+		bc->cdb[0] = bsg_cmd_sync_cache_10;
+		break;
```
vincentkfu: Can you just return here since there is no need to set the offset and nlb for sync?

engines/io_uring.c
```diff
@@ -761,55 +753,68 @@ static int fio_ioring_cmd_prep(struct thread_data *td, struct io_u *io_u)
...
+			read_opcode = nvme_cmd_compare;
+			io_u_set(td, io_u, IO_U_F_VER_IN_DEV);
+		}
 
-		for (i = 0; i < (int)o->wmode_split_nr; i++) {
-			perc += o->wmode_split[i].perc;
-			if (rand < perc) {
-				uint8_t op = o->wmode_split[i].opcode;
+		if (o->wmode_split_nr > 1 && io_u->ddir == DDIR_WRITE) {
+			unsigned int rand = rand_between(&ld->wmode_state, 0, 99);
```
vincentkfu: Please refactor this block into a separate function. Having so many levels of indentation hurts readability and leads to excessively long lines.

engines/io_uring.c
```diff
@@ -1618,13 +1634,18 @@ static int fio_ioring_init(struct thread_data *td)
...
-	parse_prchk_flags(o);
-	ext_opts = &ld->ext_opts;
-	if (o->pi_act)
-		ext_opts->io_flags |= NVME_IO_PRINFO_PRACT;
-	ext_opts->io_flags |= o->prchk;
-	ext_opts->apptag = o->apptag;
-	ext_opts->apptag_mask = o->apptag_mask;
+	/* Currently only support NVMe device */
+	if (td->o.filename != NULL &&
+	    (strstr(td->o.filename, "nvme") != NULL ||
```
vincentkfu: Instead of inspecting the filename, please use this:
```
       if (o->md_per_io_size && (!ld->is_uring_cmd_eng ||
           (ld->is_uring_cmd_eng && o->cmd_type == FIO_URING_CMD_NVME))) {
```

engines/bsg.c
vincentkfu: Also please go through the rest of the patch to confirm this is fixed

engines/io_uring.c
k831-kim: This content is related to NVMe‑specific PI handling such as “NVME_IO_PRINFO_PRCHK_GUARD”. As you mentioned, if processed that way, block devices with metadata enabled could also be used, which might become a problem. What do you think?

vincentkfu: Yes, io_uring supports NVMe and SCSI devices with metadata. I do not understand your concern.

engines/bsg.c
k831-kim: Changes and verification have been completed.

engines/bsg.c
k831-kim: Thank you for the feedback, I have changed it.

engines/io_uring.c
k831-kim: I made commit about separate function

engines/io_uring.c
k831-kim: Yes, I understand. I have completed the changes according to the feedback.

engines/nvme.c
```diff
@@ -415,6 +415,46 @@ int fio_nvme_uring_cmd_prep(struct nvme_uring_cmd *cmd, struct io_u *io_u,
 	return 0;
 }
 
+int fio_nvme_uring_cmd_prep_split(struct thread_data *td,
```
vincentkfu: Apologies for not being more specific. Please create the helper inside `engines/io_uring.c` and reduce the number of parameters passed to this function. There is no need to pass `read_opcode` and `write_opcode` since this is a write operation and this function will select the opcode.

Inside this helper just define:

```
	struct ioring_data *ld = td->io_ops_data;
	struct ioring_options *o = td->eo;
```

Then there will be no need to pass as parameters

```
o->wmode_split
o->wmode_split_nr
o->deac
ld->wmode_state
```

engines/io_uring.c
```diff
@@ -1618,13 +1634,16 @@ static int fio_ioring_init(struct thread_data *td)
...
 
 	}
-	parse_prchk_flags(o);
-	ext_opts = &ld->ext_opts;
-	if (o->pi_act)
-		ext_opts->io_flags |= NVME_IO_PRINFO_PRACT;
-	ext_opts->io_flags |= o->prchk;
-	ext_opts->apptag = o->apptag;
-	ext_opts->apptag_mask = o->apptag_mask;
+	if (o->md_per_io_size && (!ld->is_uring_cmd_eng ||
```
vincentkfu: This initialization actually still needs to happen when `md_per_io_size == 0` and the device is in extended LBA mode (e,g, 520-byte or 4160-byte sectors). So the correct condition here is:

```
        if ((ld->is_uring_cmd_eng && o->cmd_type == FIO_URING_CMD_NVME) ||
                !ld->is_uring_cmd_eng) {
```

If we are using the regular io_uring ioengine or io_uring_cmd with NVMe then we need to do the PI initialization.

The same change needs to be made in `fio_ioring_io_u_init`.

engines/io_uring.c
```diff
@@ -1664,40 +1683,45 @@ static int fio_ioring_io_u_init(struct thread_data *td, struct io_u *io_u)
...
-		pi_attr->app_tag = o->apptag;
-		pi_attr->flags = 0;
-		if (o->prchk & NVME_IO_PRINFO_PRCHK_GUARD)
-			pi_attr->flags |= IO_INTEGRITY_CHK_GUARD;
-		if (o->prchk & NVME_IO_PRINFO_PRCHK_REF)
-			pi_attr->flags |= IO_INTEGRITY_CHK_REFTAG;
-		if (o->prchk & NVME_IO_PRINFO_PRCHK_APP)
-			pi_attr->flags |= IO_INTEGRITY_CHK_APPTAG;
-	}
+	if (o->md_per_io_size && (!ld->is_uring_cmd_eng ||
```
vincentkfu: My feedback yesterday was wrong. Please use the `if` condition mentioned above that allows PI initialization in extended LBA mode when `md_per_io_size == 0`.

All of the extended LBA PI tests fail with this current condition: https://github.com/vincentkfu/fio/actions/runs/28043964145

k831-kim: I have changed the part you mentioned.

engines/io_uring.c
k831-kim: I have changed the part you mentioned.

engines/nvme.c
k831-kim: As you mentioned, I have organized it in io_uring.c.
However, since io_uring.c is a general I/O path, I don't think NVMe‑specific variables should be passed as parameters, so I redeclared the NVMe‑related variables inside the function.
That said, this could result in duplicate work during the hot‑path preparation stage; I would appreciate any feedback.

engines/io_uring.c
```diff
@@ -713,17 +714,60 @@ static int fio_ioring_prep(struct thread_data *td, struct io_u *io_u)
...
+			}
+
+			dprint(FD_IO, "op selected %u\n", op);
+			return fio_nvme_uring_cmd_prep(cmd, io_u, iov, dsm,
+						       nvme_cmd_read, op,
+						       o->wmode_split[i].cdw12_flag);
+		}
+	}
+
+	return fio_nvme_uring_cmd_prep(cmd, io_u, iov, dsm,
```
vincentkfu: It should be impossible to reach this line of code. So just `return -EINVAL` here.

engines/io_uring.c
```diff
@@ -713,17 +714,60 @@ static int fio_ioring_prep(struct thread_data *td, struct io_u *io_u)
...
 	void *ptr = ld->dsm;
 	unsigned int dsm_size;
-	uint8_t read_opcode = nvme_cmd_read;
+	unsigned int rand = rand_between(&ld->wmode_state, 0, 99);
+	unsigned int perc = 0;
+	int i;
+
+	sqe = &ld->sqes[(io_u->index) << 1];
+	cmd = (struct nvme_uring_cmd *)sqe->cmd;
+	dsm_size = sizeof(*ld->dsm) + td->o.num_range * sizeof(struct nvme_dsm_range);
```
vincentkfu: The `dsm`-related lines are not necessary since they are only used when sending multiple trim ranges. So these calculations can be removed and just pass `NULL` for `dsm` in `fio_nvme_uring_cmd_prep`.

engines/io_uring.c
```diff
@@ -713,17 +714,60 @@ static int fio_ioring_prep(struct thread_data *td, struct io_u *io_u)
...
+	struct nvme_uring_cmd *cmd;
 	struct nvme_dsm *dsm;
 	void *ptr = ld->dsm;
 	unsigned int dsm_size;
-	uint8_t read_opcode = nvme_cmd_read;
+	unsigned int rand = rand_between(&ld->wmode_state, 0, 99);
+	unsigned int perc = 0;
+	int i;
+
+	sqe = &ld->sqes[(io_u->index) << 1];
```
vincentkfu: Consider passing `sqe` and `cmd` as parameters to this function in order to simplify things a bit.

engines/nvme.c
vincentkfu: I think it's ok to pass NVMe-specific parameters. If any other `cmd_type`s gain support for multiple write modes we can change things.

engines/bsg.c
```diff
@@ -0,0 +1,169 @@
...
+
+	blksz = sgio_get_be32(&buf[4]);
+	hlba = sgio_get_be32(buf);
+
+	if (blksz) {
+		*bs = blksz;
+		*max_lba = hlba;
+		ret = 0;
+	} else {
+		ret = EIO;
```
vincentkfu: `-EIO`

vincentkfu:
Also please rebase on the current master branch. This will automatically run additional tests involving the io_uring ioengine.

engines/nvme.c
k831-kim: It seems there will be no problem if we proceed in the direction you mentioned.
I have changed code according to your suggestions.

engines/io_uring.c
k831-kim: I understand it, I have changed return valude

engines/io_uring.c
k831-kim: Oh... Thanks for your comment. The dsm argument that was being passed in the function has been removed, and the value passed to `fio_nvme_uring_cmd_prep` has been changed to NULL.

engines/io_uring.c
k831-kim: Yes, I have made the changes according to your suggestion and deleted the line.

engines/bsg.c
k831-kim: Sorry, I have changed it

PR #1026, opened 2020-06-25
v2 PR : Zone Append command support for NVMe Zoned Namespaces (ZNS)

krish-reddy:
This is a v2 Pull Request for the Zone Append command support for NVMe Zoned Namespaces
The v1 Pull Request is https://github.com/axboe/fio/pull/1021

Review comments for the v1 Pull Request have been taken care of in the v2 Pull Request.
New Opcodes added for libaio and io_uring are removed as per the review comments for the Linux kernel patches in LKML.
Linux Kernel v1 patchset for the Zone Append command support :
	https://lkml.org/lkml/2020/6/17/801
	https://lkml.org/lkml/2020/6/17/802
	https://lkml.org/lkml/2020/6/17/803
	https://lkml.org/lkml/2020/6/17/804

RWF_ZONE_APPEND flag is passed in iocb->aio_rw_flags for libaio IO Engine and sqe->rw_flags for io_uring IO Engine.

Add Zone Append command support for NVMe Zoned Namespaces (ZNS) defined in NVM Express TP4053.

1. Added a new FIO option zone_append.
	When zone_append option is enabled, the existing write path will send Zone Append command with LBA offset as start of the Zone.
2. libaio: support for Zone Append command in libaio IO engine
3. iouring: support for Zone Append command in io_uring IO engine
4. t/zbd: Add support to verify Zone Append (ZNS)) command with libaio, io_uring IO engine tests

This is a RFC and the Linux Kernel interface is under submission and review.
Feedback / Comments will help in improving this further.

PR #2103, opened 2026-06-04
io_uring: Add support for 'zone_append' when using uring_cmd and zonemode ZBD

Krien:
This pull request adds support for zone appends for NVMe ZNS devices. Zone appends enable an iodepth larger than 1 for writes within zones. It only supports uring_cmd as appends can not yet be used in the standard write path (see https://github.com/axboe/fio/pull/1026) and appends can only be enabled when the zonemode is explicitly set to ZBD.

The request introduces the following changes:
1. Add fio_ioring_cmd_finish_zone as a new zone finish command for uring_cmd.
2. Set the finish command in uring_cmd to fio_ioring_cmd_finish_zone instead of the default finish through blkzone. This is needed because appends at iodepth>1 can lead to finish commands, which will fail on the character device.
3. Add a new write_mode for zone_append. The offset needs to be set to the zone start when appends are used.


vincentkfu:
@damien-lemoal @kawasaki Do you have any feedback on this?

engines/nvme.c
```diff
@@ -375,7 +378,14 @@ int fio_nvme_uring_cmd_prep(struct nvme_uring_cmd *cmd, struct io_u *io_u,
 		return -ENOTSUP;
 	}
 
-	slba = get_slba(data, io_u->offset);
+	/* Zone appends must be issued to the start of the target zone */
+	if (cmd->opcode == nvme_zns_cmd_append && io_u->file->zbd_info) {
+		zone_size = io_u->file->zbd_info->zone_size;
+		offset = (io_u->offset / zone_size) * zone_size;
```
damien-lemoal: offset = io_u->offset & (~(zone_size - 1));

Because the zone size is always a power of 2. So let's avoid the expensive divisions.

damien-lemoal (commented):
Looks OK to me, except for one nit.

engines/nvme.c
```diff
@@ -852,6 +852,47 @@ int fio_nvme_reset_wp(struct thread_data *td, struct fio_file *f,
...
+		       uint64_t offset, uint64_t length)
+{
+	struct nvme_data *data = FILE_ENG_DATA(f);
+	unsigned int nr_zones;
+	unsigned long long zslba;
+	int i, fd, ret = 0;
+
+	/* If the file is not yet opened, open it for this function. */
+	fd = f->fd;
+	if (fd < 0) {
```
damien-lemoal, line 876: How can that happen ? The file should already be open since IOs are on-going. No ?

damien-lemoal:
Note: patches on the mailing list would be *a lot* easier to review/comment. The github GUI is really horrendous for reviews :(

engines/io_uring.c
```diff
@@ -102,6 +102,7 @@ enum uring_cmd_write_mode {
 	FIO_URING_CMD_WMODE_UNCOR,
 	FIO_URING_CMD_WMODE_ZEROES,
 	FIO_URING_CMD_WMODE_VERIFY,
+	FIO_URING_CMD_WMODE_ZONE_APPEND,
 };
 
 #define WMODE_SPLIT_MAX	4
```
vincentkfu: This should be increased by one.
Better yet, set `FIO_URING_CMD_WMODE_WRITE = 0`, add `FIO_URING_CMD_WMODE_LAST` at the end of the enum, and `#define WMODE_SPLIT_MAX FIO_URING_CMD_WMODE_LAST`  so that this value is automatically updated.
Fio does something similar with `enum fio_ddir`.

engines/nvme.c
Krien: I agree that it should not happen. I followed the same if statement used in `blkzoned_finish_zone` to ensure consistent behavior. If this is not desirable, I will remove the if.

engines/nvme.c
Krien: Nice!

PR #2106, opened 2026-06-09
server: fix out-of-bounds write zero-terminating received payloads

aizu-m:
out-of-bounds write, 1 byte, server side (peer-controlled TEXT, pdu_len 32):

    fio_net_recv_cmd: buf[buf_len] writes payload+16408, allocation is 33 bytes

The receive path reserves one byte for the terminator (cmd_size = sizeof(cmd) + pdu_len + 1) and then writes the nul at buf[buf_len]. buf_len comes straight off the wire and is never bounded against pdu_len, so a peer can point the write well past the payload and land a single nul anywhere on the heap.

Found reading the framing on the receive side. handle_trigger_cmd() repeats the same mistake with the 16-bit pdu->len from a VTRIGGER. The terminator belongs at the byte the allocation already set aside, so I write it there and drop the wire length entirely.

JOB and TEXT both go through fio_net_recv_cmd(), so --server and --client are both affected; VTRIGGER is server-side. Checked with a guard-page repro: the old buf[buf_len] faults, payload[pdu_len] runs clean, and a normal --client/--server job is unchanged.

### DIFF
--- a/server.c
+++ b/server.c
@@ -443,20 +443,16 @@ struct fio_net_cmd *fio_net_recv_cmd(int sk, bool wait)
 		free(cmdret);
 		cmdret = NULL;
 	} else if (cmdret) {
-		/* zero-terminate text input */
+		/*
+		 * zero-terminate text input at the trailing byte reserved by
+		 * the allocation above, not at the length supplied by the peer
+		 */
 		if (cmdret->pdu_len) {
-			if (cmdret->opcode == FIO_NET_CMD_TEXT) {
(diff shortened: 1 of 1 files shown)


axboe:
You put a full explanation in the PR, and yet the commit message itself is entirely devoid of anything useful. See any other PR or commit in the tree for why this is totally wrong and the opposite of what it should look like. Including things like signed-off-by etc.

Sorry if I'm being brief, but I've had to mention this a thousand times at least, and yet people seemingly blindly submit PRs that violate it. Closing this one, reopen when it conforms to expectations in terms of commit message requirements and formatting.

PR #2112, opened 2026-06-24
parse: ignore leading minus sign when splitting options

### LINKED ISSUE #1657: Bogus warning: nice: max value out of range: 20 (19 max)

**Description of the bug:**

Using `--nice=-20` causes this bogus warning:

``` 
nice: max value out of range: 20 (19 max)
```

Running under strace shows that the value -20 is actually accepted despite the warning:

```
[pid 40039] setpriority(PRIO_PROCESS, 0, -20) = 0
```

**Environment**: Fedora 38

**fio version**: fio-3.36-6-gc11e2 (compiled from git right now, but same behavoir with Fedora and Debian fio packages)

**Reproduction steps**

```
$ sudo ./fio --name=foo --filename=/tmp/fio.tmp --size=1M --nice=-20
nice: max value out of range: 20 (19 max)
foo: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
fio-3.36-6-gc11e2
Starting 1 process
[...]
```

```
$ sudo ./fio --debug=parse --name=foo --filename=/tmp/fio.tmp --size=1M --nice=-20
fio: set debug option parse
parse    40053 dup options
parse    40053 handle_option=name, ptr=foo
parse    40053 __handle_option=name, type=OPT_STR_STORE, ptr=foo
parse    40053 handle_option=filename, ptr=/tmp/fio.tmp
parse    40053 __handle_option=filename, type=OPT_STR_STORE, ptr=/tmp/fio.tmp
parse    40053 handle_option=size, ptr=1M
parse    40053 __handle_option=size, type=OPT_STR_VAL_ZONE, ptr=1M
parse    40053   ret=0, out=1048576
parse    40053 handle_option=nice, ptr=-20
parse    40053 __handle_option=nice, type=OPT_INT, ptr=-20
parse    40053   ret=0, out=18446744073709551596
parse    40053 __handle_option=nice, type=OPT_INT, ptr=20
parse    40053   ret=0, out=20
nice: max value out of range: 20 (19 max)
foo: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=psync, iodepth=1
parse    40053 free options
fio-3.36-6-gc11e2
Starting 1 process
[...]
```

cenhuil:
When parsing an option like --nice=-20, the hyphen is incorrectly identified as a delimiter, splitting the value into two parts: an empty string and '20'. This results in a bogus out-of-range warning: 'nice: max value out of range: 20 (19 max)'.

This commit fixes the issue by ignoring the hyphen if it appears as the very first character, correctly treating it as a negative sign rather than an option delimiter.

Fixes #1657

### DIFF
--- a/parse.c
+++ b/parse.c
@@ -1041,8 +1041,15 @@ static int handle_option(const struct fio_option *o, const char *__ptr,
 			if (o->type != FIO_OPT_STR_MULTI && o->type != FIO_OPT_RANGE) {
 				if (!ptr2)
 					ptr2 = strchr(ptr, ':');
-				if (!ptr2)
+				if (!ptr2) {
 					ptr2 = strchr(ptr, '-');
+					/*
+					 * If the first character is a minus sign, it's
+					 * likely a negative number, not a delimiter.
+					 */
+					if (ptr2 == ptr)
+						ptr2 = strchr(ptr + 1, '-');
+				}
 			}
 		} else if (ptr && o->type == FIO_OPT_FLOAT_LIST) {
 			ptr2 = strchr(ptr, ':');


parse.c
```diff
@@ -1041,8 +1041,15 @@ static int handle_option(const struct fio_option *o, const char *__ptr,
 			if (o->type != FIO_OPT_STR_MULTI && o->type != FIO_OPT_RANGE) {
 				if (!ptr2)
 					ptr2 = strchr(ptr, ':');
-				if (!ptr2)
+				if (!ptr2) {
 					ptr2 = strchr(ptr, '-');
+					/*
+					 * If the first character is a minus sign, it's
+					 * likely a negative number, not a delimiter.
+					 */
```
sitsofe, line 1049: Will `-` ever be a delimiter when we aren't parsing a `FIO_OPT_RANGE`? I think we can only be parsing {`FIO_OPT_STR_VAL`, `FIO_OPT_STR_VAL_TIME, FIO_OPT_INT, FIO_OPT_ULL, FIO_OPT_BOOL, FIO_OPT_STR_SET, FIO_OPT_STR_VAL_ZONE`} at this point...

sitsofe:
@cenhuil  One more request - please could we have a Real Name in front of your email address for both the commit author and Signed-off-by?

parse.c
cenhuil: Yes, - is indeed used as a delimiter for those other types as well.

For example, options like bs, ba, or rate (which are evaluated as FIO_OPT_ULL or FIO_OPT_INT) allow users to specify different values for read, write, and trim. According to the fio documentation and the current parsing logic, these multiple values can be delimited by commas, colons, or hyphens.

Therefore, a command like fio --name=test --bs=4k-8k is perfectly valid and successfully assigns 4k to Read and 8k to Write by using the hyphen as a delimiter.

If we completely remove ptr2 = strchr(ptr, '-'); for types other than FIO_OPT_RANGE, it would break multi-value assignments like bs=4k-8k. That is why we still need to support - as a delimiter here, but merely skip it if it's the very first character (acting as a negative sign).

cenhuil:
> @cenhuil One more request - please could we have a Real Name in front of your email address for both the commit author and Signed-off-by?

I have updated the Author and Signed-off-by fields to use my real name. Thanks!

parse.c
sitsofe: Well this has turned out to be trickier than I first thought! From the documentation for "int":

> If the option accepts an upper and lower range, use a colon ‘:’ or minus ‘-’ to separate such values. See [irange](https://fio.readthedocs.io/en/latest/fio_doc.html#irange).

If someone does `bs=4k-8k` I think they are confused because it looks like a range (so they should have used [bsrange](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-blocksize_range)) but as you stated it will cause reads to be 4k and writes to be 8k. The docs for [blocksize](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-blocksize) define it as taking a comma separated values and not a range. It's similar for [blockalign](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-blockalign) and [rate](https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-rate).

@vincentkfu I'm torn because the docs do describe this strange behaviour ("minus as a separator/delimiter") and taking it out will break backwards compatibility if someone has chosen to use it that way but isn't keeping it confusing? If we do say we keep this behaviour I see this PR as the only way to support negative numbers on things that aren't strings or floats.

vincentkfu: I don't see in the documentation where `-` should be used as a delimiter outside of places where a range is expected. I think treating `-` as a delimiter is just a quirk of the parsing logic. Ideally Fio would print an error message when it encounters a `-` when a comma is expected.
I'm inclined to allow the current undefined (as far as I could find) and confusing behavior to remain and think will take a closer look at the fix as time allows.

sitsofe: > I don't see in the documentation where - should be used as a delimiter outside of places where a range is expected.

Technically the docs [describe this bizarre behaviour for things that take `int` parameter type](https://fio.readthedocs.io/en/latest/fio_doc.html#parameter-types:~:text=use%20a%20colon%20%E2%80%98%3A%E2%80%99-,or%20minus,-%E2%80%98%2D%E2%80%99%20to%20separate%20such). Perhaps we should change that text to leave the range stuff entirely to the `irange` description...

> I'm inclined to allow the current undefined (as far as I could find) and confusing behavior to remain and think will take a closer look at the fix as time allows.

That seems fair.

sitsofe:
@vincentkfu this looks good to me.

PR #2117, opened 2026-07-09
test: Skip libaio tests on Linux if libaio support is not built

bvanassche:
Several tests in the test suite fail on Linux systems where fio is compiled without libaio support (e.g. when libaio-dev is missing). Fix these failures by dynamically skipping tests that use the libaio engine when the compiled fio binary does not support it.

This patch includes the following changes:
- Add a helper `libaio_if_linux` to verify.py, verify_state_save.py, and run-fio-tests.py to check for libaio support on Linux.
- Add the libaio_if_linux requirement to verify.py, verify_state_save.py, and run-fio-tests.py for jsonplus2csv_test.py.
- Handle libaio availability checking in latency_percentiles.py's internal test runner loop.
- Add --skip-req option to verify_state_save.py to ensure it is aligned with other test scripts.
- Harden get_json() in fiotestlib.py to return False instead of raising ValueError when parsing malformed output.

### DIFF
--- a/file.h
+++ b/file.h
@@ -212,6 +212,7 @@ struct fio_mount {
  */
 struct thread_data;
 extern void close_files(struct thread_data *);
+extern bool fio_close_idle_file(struct thread_data *);
 extern void close_and_free_files(struct thread_data *);
 extern uint64_t get_start_offset(struct thread_data *, struct fio_file *);
 extern int __must_check setup_files(struct thread_data *);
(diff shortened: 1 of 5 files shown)


vincentkfu:
Bart, your changes to the test scripts have disappeared but I think this would be cleaner with a `libaio_if_linux` attribute in the `Requirements` class. If this were defined and set in the `Requirements` class there would be no need for the redundant code across the different scripts.

bvanassche:
Thanks Vincent for the feedback. Since the scope of this pull request has changed I will close this pull request and I will open a new pull request when the changes are ready.

PR #2118, opened 2026-07-14
stop experimental verify based on bytes issued, not completed

malikoyv:
In experimental_verify mode, do_verify() decided whether to stop
issuing verify reads by comparing td->bytes_verified against
verify_bytes. bytes_verified is only updated on completion, so with a
queue depth > 1 it lags the issue side. As a result fio kept issuing
read/verify commands to fill the queue even after the last block that
was actually written had already been submitted.

Those extra reads target offsets that were never written (especially
with norandommap, where the write phase does not cover every block),
which the verify phase then reads back as zeroes, producing spurious
"bad magic header 0" verify failures.

Track a new counter, td->verify_bytes_issued, which is incremented at
issue time for both the write->read and trim->read verify conversions,
and use it for the stop condition. Reset it in reset_io_counters()
alongside bytes_verified.

### DIFF
--- a/fio.h
+++ b/fio.h
@@ -378,6 +378,7 @@ struct thread_data {
 	uint64_t io_issues[DDIR_RWDIR_CNT];
 	uint64_t verify_read_issues;
 	uint64_t io_issue_bytes[DDIR_RWDIR_CNT];
+	uint64_t verify_bytes_prepped;
 	uint64_t loops;
 
 	/*
--- a/backend.c
+++ b/backend.c
@@ -710,7 +710,7 @@ static void do_verify(struct thread_data *td, uint64_t verify_bytes)
 				break;
 			}
 		} else {
-			if (td->bytes_verified + td->o.rw_min_bs > verify_bytes)
+			if (td->verify_bytes_prepped + td->o.rw_min_bs > verify_bytes)
 				break;
 
 			while ((io_u = get_io_u(td)) != NULL) {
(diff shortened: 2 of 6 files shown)


fio.h
```diff
@@ -378,6 +378,7 @@ struct thread_data {
 	uint64_t io_issues[DDIR_RWDIR_CNT];
 	uint64_t verify_read_issues;
 	uint64_t io_issue_bytes[DDIR_RWDIR_CNT];
+	uint64_t verify_bytes_issued;
```
vincentkfu: I did use the term "bytes issued" in my comment on the earlier PR, but looking more closely, this is more accurately called "bytes_prepped". Could you rename this to `verify_bytes_prepped`?

vincentkfu:
The last patch of your earlier PR enabled test coverage for `experimental_verify` and also disabled sequence number checks when `norandommap` was enabled for `experimental_verify`. Can you include that patch here as well? Please separate the two components so that we end up with one patch for test coverage and a different patch for the sequence number check.

sitsofe:
Is the sequence verification "only" a problem when you run a random workload that doesn't try to write to different blocks or a workload that might wraparound (e.g. `time_based`, `loops=2` etc.)?

vincentkfu:
Sequence number verification is also problematic when the workload wraps around when `io_size` > `size`, but that is one of the conditions in `fio_offset_overlap_risk`, so I think we are good there.
With `time_based` workloads the verify phase never starts and we have a warning message for that.
`loops=2` seems to work fine in some simple tests I did.

PR #2123, opened 2026-07-31
engines/io_uring_cmd: extend bsg cmd_type with cdb_len, verify, and prefetch

ljw8161:
## Summary

  This series extends the bsg cmd_type of the io_uring_cmd engine, which
  until now hardcoded READ(10)/WRITE(10) and supported only plain
  read/write as the data-direction operations. The changes let users
  select the SCSI CDB length and issue VERIFY and PRE-FETCH commands,
  which is useful for benchmarking devices whose capacity or per-I/O
  transfer size exceeds the 10-byte CDB limits, and for measuring the
  cost of medium verification and read-cache warming.

  The series also consolidates the bsg completion decoding into a single
  place and drops a dead assignment on the bsg submission path. All new
  options default to the current behavior, so existing bsg jobs are
  unaffected.

  ## Changes

  1. **Drop unused `write_opcode` assignment on the bsg path** — cleanup.
     `ld->write_opcode` is only consumed on the nvme path; the bsg path
     fills the CDB directly from `io_u->ddir`, so the assignment never
     had any effect. Simplifies the surrounding write-mode check.

  2. **Consolidate bsg status decoding into `errdetails`** — cleanup.
  The bsg completion was decoded in two places: `fio_ioring_cmd_event()`
  split the packed SCSI status from `cqe->big_cqe[0]` into
  `io_u->error`, and `fio_ioring_cmd_errdetails()` re-derived the SCSI
  and host status from it. The nvme path only does this breakdown once,
  in `errdetails`. Move the bsg path to the same model: `event` now
  stores the raw `cqe->big_cqe[0]` in `io_u->error`, leaving
  `errdetails` as the single place that splits the SCSI status
  (bits 0-7) and host status (bits 16-23).

  3. **Add `cdb_len` option** — selects the CDB length for READ/WRITE
  (and SYNCHRONIZE CACHE). Supported values are 10, 16, and 32; the
  32-byte CDB uses the variable-length CDB (opcode 0x7F). The
  default is `0` (auto): the smallest CDB whose LBA and
  transfer-length fields can hold the request is chosen, escalating
  from 10 to 16 as the LBA or transfer length grows. An explicit
  `cdb_len` rejects requests whose LBA or transfer length would
  overflow the selected CDB's fields rather than silently
  promoting them, so a chosen CDB length is preserved for
  benchmarking.

  4. **Support `write_mode=verify`** — extends the existing (nvme-only)
     option to bsg, issuing SCSI VERIFY as the write-side operation. A
     new `verify_bytchk` option maps to the CDB BYTCHK field:
     `0` = medium verification only (no host transfer), `1` = full
     byte-by-byte comparison, `3` = single-block comparison against the
     range. `writefua` is rejected in this mode since FUA is not defined
     for VERIFY.

  5. **Add `read_mode=prefetch`** — a new option symmetric to
  `write_mode`, issuing SCSI PRE-FETCH for DDIR_READ to warm the
  device read cache without transferring data to the host. Unlike
  SYNCHRONIZE CACHE (exposed via DDIR_SYNC), PRE-FETCH moves data
  toward the cache and belongs on the read side. `readfua` and
  `cdb_len=32` are rejected in this mode: the FUA bit is not defined
  for PRE-FETCH, and SBC defines no 32-byte PRE-FETCH service action.

  Non-default `read_mode` with `cmd_type=nvme`, and `verify_bytchk`
  without `write_mode=verify` (bsg), are rejected at engine init so
  misconfigurations are surfaced instead of silently ignored.

  Documentation (HOWTO.rst, fio.1) is updated for all new options.

### DIFF
--- a/engines/bsg.c
+++ b/engines/bsg.c
@@ -104,7 +104,7 @@ void fio_bsg_uring_cmd_init(struct bsg_uring_cmd *cmd, struct bsg_cmd *bc,
 	if (dxfer_dir == SG_DXFER_TO_DEV) {
 		cmd->dout_xferp = (uint64_t)(uintptr_t) io_u->xfer_buf;
 		cmd->dout_xfer_len = io_u->xfer_buflen;
-	} else {
+	} else if (dxfer_dir == SG_DXFER_FROM_DEV) {
 		cmd->din_xferp = (uint64_t)(uintptr_t) io_u->xfer_buf;
 		cmd->din_xfer_len = io_u->xfer_buflen;
 	}
@@ -155,8 +155,24 @@ static void fio_bsg_varlen_cdb_header(struct bsg_cmd *bc, uint16_t sa)
 	bc->cdb[7] = BSG_VARLEN_CDB_ADDITIONAL_LEN;
 }
 
+static void fio_bsg_set_verify_bytchk(struct bsg_cmd *bc, unsigned int cdb_len,
+				      unsigned int verify_bytchk)
+{
+	/*
+	 * BYTCHK occupies bits 2-1 of byte 1 for VERIFY(10)/(16) and byte 10
+	 * for the 32-byte variable-length VERIFY. BYTCHK=0 leaves the field
+	 * cleared (medium verification only).
+	 */
+	if (cdb_len == 32)
+		bc->cdb[10] |= verify_bytchk << 1;
+	else
+		bc->cdb[1] |= verify_bytchk << 1;
+}
+
 int fio_bsg_uring_cmd_prep(struct bsg_uring_cmd *cmd, struct io_u *io_u,
-			   struct bsg_cmd *bc, bool fua, unsigned int cdb_len)
+			   struct bsg_cmd *bc, bool fua, unsigned int cdb_len,
+			   enum bsg_write_mode wmode,
+			   unsigned int verify_bytchk)
 {
 	struct bsg_data *data = FILE_ENG_DATA(io_u->file);
 	unsigned long long offset, nlb;
@@ -207,6 +223,39 @@ int fio_bsg_uring_cmd_prep(struct bsg_uring_cmd *cmd, struct io_u *io_u,
 		}
 		break;
 	case DDIR_WRITE:
+		if (wmode == BSG_WRITE_MODE_VERIFY) {
+			/*
+			 * VERIFY data direction depends on BYTCHK:
+			 *   0  medium verification only, no host data transfer
+			 *   1  compare the whole transfer against the medium
+			 *   3  compare a single block against the whole range
+			 * BYTCHK 1 and 3 send data to the device.
+			 */
+			int dxfer = verify_bytchk ? SG_DXFER_TO_DEV :
+						    SG_DXFER_NONE;
+
+			fio_bsg_uring_cmd_init(cmd, bc, io_u, dxfer, cdb_len);
+			switch (cdb_len) {
+			case 10:
+				bc->cdb[0] = bsg_cmd_verify_10;
+				break;
+			case 16:
+				bc->cdb[0] = bsg_cmd_verify_16;
+				break;
+			case 32:
+				fio_bsg_varlen_cdb_header(bc, BSG_SA_VERIFY_32);
+				break;
+			}
+			fio_bsg_set_verify_bytchk(bc, cdb_len, verify_bytchk);
+			/*
+			 * BYTCHK=3 compares one block against the entire
+			 * range, so only a single block is transferred while
+			 * the CDB still carries the full block count.
+			 */
+			if (verify_bytchk == 3)
+				cmd->dout_xfer_len = data->bs;
+			break;
+		}
 		fio_bsg_uring_cmd_init(cmd, bc, io_u, SG_DXFER_TO_DEV, cdb_len);
 		switch (cdb_len) {
 		case 10:
--- a/engines/bsg.c
+++ b/engines/bsg.c
@@ -90,13 +90,14 @@ int fio_bsg_uring_cmd_get_file_size(struct thread_data *td, struct fio_file *f)
 }
 
 void fio_bsg_uring_cmd_init(struct bsg_uring_cmd *cmd, struct bsg_cmd *bc,
-			    struct io_u *io_u, int dxfer_dir)
+			    struct io_u *io_u, int dxfer_dir,
+			    unsigned int cdb_len)
 {
 	memset(cmd, 0, sizeof(*cmd));
 	memset(bc->cdb, 0, sizeof(bc->cdb));
 
 	cmd->request = (uint64_t)(uintptr_t) bc->cdb;
-	cmd->request_len = sizeof(bc->cdb);
+	cmd->request_len = cdb_len;
 	cmd->response = (uint64_t)(uintptr_t) bc->sb;
 	cmd->max_response_len = sizeof(bc->sb);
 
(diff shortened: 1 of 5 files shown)


engines/io_uring.c
```diff
@@ -984,16 +984,8 @@ static char *fio_ioring_cmd_errdetails(struct thread_data *td,
...
-		unsigned int host_status = (io_u->error >> 16) & 0xff;
-		if (status) {
-			snprintf(msgchunk, MAXMSGCHUNK, "BSG SCSI Status: 0x%02x; ", status);
-			strlcat(msg, msgchunk, MAXERRDETAIL);
-		}
-		if (host_status) {
-			snprintf(msgchunk, MAXMSGCHUNK, "BSG Host Status: 0x%02x; ", host_status);
-			strlcat(msg, msgchunk, MAXERRDETAIL);
-		}
+		snprintf(msgchunk, MAXMSGCHUNK, "BSG SCSI Status: 0x%02x; ", io_u->error);
```
vincentkfu: It would be better if decoding happened in the same place for the two command types. For nvme, decoding happens here whereas for bsg decoding occurs in `fio_ioring_cmd_event`.

engines/io_uring.c
```diff
@@ -528,6 +529,18 @@ static struct fio_option options[] = {
...
+		.lname	= "SCSI CDB length",
+		.type	= FIO_OPT_INT,
+		.off1	= offsetof(struct ioring_options, cdb_len),
+		.help	= "SCSI CDB length for bsg cmd_type (10, 16, or 32). "
+			  "0 (default) auto-escalates to the smallest CDB that "
+			  "fits the request's LBA and transfer length.",
+		.def	= "0",
+		.category = FIO_OPT_C_ENGINE,
+		.group	= FIO_OPT_G_IOURING,
+	},
```
vincentkfu, line 614: I think it would be slightly better to encode the possible values for `cdb_len` here so that illegal values could be detected when the command line is parsed. I would just do the same thing here that you did for `verify_bytechk`.

engines/io_uring.c
```diff
@@ -1613,6 +1627,14 @@ static int fio_ioring_cmd_init(struct thread_data *td, struct ioring_data *ld)
 	} else if (o->cmd_type == FIO_URING_CMD_BSG) {
 		ld->bc = calloc(td->o.iodepth, sizeof(struct bsg_cmd));
 
+		if (o->cdb_len != 0 && o->cdb_len != 10 &&
```
vincentkfu: This code is no longer necessary if illegal values for `cbd_len` are detected when the option is parsed.

engines/io_uring.c
```diff
@@ -854,12 +879,19 @@ static int fio_ioring_cmd_prep(struct thread_data *td, struct io_u *io_u)
 				ld->cdw12_flags[io_u->ddir]);
 	} else {
 		struct bsg_uring_cmd *cmd;
+		enum bsg_write_variant wvariant;
```
vincentkfu: For NVMe we have been using the word `mode` to refer to different write commands. This patch introduces the word `variant` for this purpose. Is there a compelling reason to deviate from what we have done for NVMe? If not, I think it would be less confusing to use the same nomenclature. Using both `mode` and `variant` might give readers the impression that there is an important distinction between the two. If it's not important, it's better to use the same term.

engines/bsg.c
```diff
@@ -166,7 +166,8 @@ static void fio_bsg_set_verify_bytchk(struct bsg_cmd *bc, unsigned int cdb_len,
 int fio_bsg_uring_cmd_prep(struct bsg_uring_cmd *cmd, struct io_u *io_u,
 			   struct bsg_cmd *bc, bool fua, unsigned int cdb_len,
 			   enum bsg_write_variant wvariant,
-			   unsigned int verify_bytchk)
+			   unsigned int verify_bytchk,
+			   enum bsg_read_variant rvariant)
```
vincentkfu: Since the option name is `read_mode` it would be clearer to use the word mode in the name instead of variant.

engines/io_uring.c
ljw8161: I added the posval entries so invalid values are rejected at parse time, and dropped the value-range check in fio_ioring_cmd_init() as you suggested.

engines/io_uring.c
ljw8161: This code is removed. Thanks

engines/io_uring.c
ljw8161: Thanks for the suggestion.

Patch 2 ("engines/io_uring: drop bsg status parsing from event") moves the bsg path to match nvme:
fio_ioring_cmd_event() now stores the raw cqe->big_cqe[0] in io_u->error, and fio_ioring_cmd_errdetails() is the
single place that splits the SCSI status (bits 0-7) and host status (bits 16-23). Both command types now do their
decoding in errdetails().

engines/io_uring.c
ljw8161: Agreed, there's no meaningful distinction intended between the two terms, so keeping a separate word just risks confusing readers for no reason. In the new version I've dropped "variant" and use "mode" throughout.

engines/bsg.c
ljw8161: Agreed, I renamed the field from rvariant to rmode.

vincentkfu:
It seems that `read_mode=prefetch` returns CONDITION_MET SCSI status 0x04 which means success but Fio treats this as an error return.

```
root@localhost:~/fio-dev/fio-bsg/fio# ./fio --name=test --ioengine=io_uring_cmd --rw=read --number_ios=1 "--filename=/dev/bsg/6\:0\:0\:0" --cmd_type=bsg --read_mode=prefetch
test: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=io_uring_cmd, iodepth=1
fio-3.42-124-g5e9e-dirty
Starting 1 process
fio: io_u error on file /dev/bsg/6:0:0:0: Device-specific error: read offset=0, buflen=4096
fio: io_uring_cmd: /dev/bsg/6:0:0:0: BSG SCSI Status: 0x04;
fio: pid=29286, err=4/file:io_u.c:2057, func=io_u error, error=Interrupted system call

test: (groupid=0, jobs=1): err= 4 (file:io_u.c:2057, func=io_u error, error=Interrupted system call): pid=29286: Mon Aug 10 16:32:02 2026
  cpu          : usr=2.08%, sys=0.00%, ctx=1, majf=0, minf=14
     complete  : 0=50.0%, 4=50.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
     issued rwts: total=1,0,0,0 short=0,0,0,0 dropped=0,0,0,0

```

```
root@localhost:~# sg_seek -p -l 0 -c 1 /dev/sdb
sg_seek: Condition met
```

Please intercept this return value (perhaps in `fio_ioring_cmd_event`) to avoid treating this as an error condition.

ljw8161:
> It seems that `read_mode=prefetch` returns CONDITION_MET SCSI status 0x04 which means success but Fio treats this as an error return.
> 
> ```
> root@localhost:~/fio-dev/fio-bsg/fio# ./fio --name=test --ioengine=io_uring_cmd --rw=read --number_ios=1 "--filename=/dev/bsg/6\:0\:0\:0" --cmd_type=bsg --read_mode=prefetch
> test: (g=0): rw=read, bs=(R) 4096B-4096B, (W) 4096B-4096B, (T) 4096B-4096B, ioengine=io_uring_cmd, iodepth=1
> fio-3.42-124-g5e9e-dirty
> Starting 1 process
> fio: io_u error on file /dev/bsg/6:0:0:0: Device-specific error: read offset=0, buflen=4096
> fio: io_uring_cmd: /dev/bsg/6:0:0:0: BSG SCSI Status: 0x04;
> fio: pid=29286, err=4/file:io_u.c:2057, func=io_u error, error=Interrupted system call
> 
> test: (groupid=0, jobs=1): err= 4 (file:io_u.c:2057, func=io_u error, error=Interrupted system call): pid=29286: Mon Aug 10 16:32:02 2026
>   cpu          : usr=2.08%, sys=0.00%, ctx=1, majf=0, minf=14
>   IO depths    : 1=100.0%, 2=0.0%, 4=0.0%, 8=0.0%, 16=0.0%, 32=0.0%, >=64=0.0%
>      submit    : 0=0.0%, 4=100.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
>      complete  : 0=50.0%, 4=50.0%, 8=0.0%, 16=0.0%, 32=0.0%, 64=0.0%, >=64=0.0%
>      issued rwts: total=1,0,0,0 short=0,0,0,0 dropped=0,0,0,0
>      latency   : target=0.00ns, window=0.00ns, percentile=100.00%, depth=1
> 
> Run status group 0 (all jobs):
> ```
> 
> ```
> root@localhost:~# sg_seek -p -l 0 -c 1 /dev/sdb
> sg_seek: Condition met
> ```
> 
> Please intercept this return value (perhaps in `fio_ioring_cmd_event`) to avoid treating this as an error condition.

Thanks for reporting this.

I've updated fio_ioring_cmd_event() to intercept this: when read_mode=prefetch is in effect and the completion is for a DDIR_READ, CONDITION MET in the device-status byte is now treated as success rather than an error. Other status bits are left untouched, so real failures on the same completion are still surfaced correctly.

PR #2125, opened 2026-08-04
engines/io_uring: add copyright header to bsg files

ljw8161:
Add the copyright header to bsg.c and bsg.h, which provide BSG (Block SCSI Generic) support for the io_uring_cmd engine. The files were previously committed without a license header.

### DIFF
--- a/engines/bsg.c
+++ b/engines/bsg.c
@@ -1,3 +1,9 @@
+/*
+ * Copyright 2026 Samsung Electronics Co., Ltd All Rights Reserved
+ *
+ * For conditions of distribution and use, see the accompanying COPYING file.
+ *
+ */
 #include "bsg.h"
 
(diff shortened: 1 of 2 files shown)


sitsofe:
It is a shame to litter files with copyright headers (when the code wasn't taken from another project and copied verbatim) and I think these days projects like the Linux kernel they actively try to avoid doing this. I understand companies want to protect their work but as fio wasn't started/isn't run entirely by Samsung perhaps some exception can be made?

@vincentkfu , @axboe what are your thoughts?

vincentkfu:
Our lawyers have requested that we include these copyright headers. I view these headers as an acknowledgement of the support that these enterprises provide to maintain and improve Fio, so I am inclined to allow them.

axboe:
FWIW, I don't think there's anything wrong with adding copyright headers.

PR #2134, opened 2026-08-14
Enhance verify_zero with error reporting

sarthak-k:
If verify TRIM fails on verify zero, then dump the buffer of received data if user requested it.


sarthak-k:
@axboe Can you approve the workflows?

axboe:
Usual comments on commit message being bad / wrong. Please fix it up, plenty of example to look at in the git tree. Also style issues, again see any other code.

If people don't even do basic checking of expected git formats or the coding style, then you can't really ask others to spend time on it. Get the basics right first.

PR #2140, opened 2026-08-26
verify: dump the received buffer when zero verify fails

sarthak-k:
verify_zero() logs the file, offset and length of a mismatch, but unlike the header and pattern verify paths it never writes out the data it read. That makes trim_verify_zero and write zeroes verify failures harder to analyze than a regular verify failure, where verify_dump=1 leaves the offending block on disk for inspection.

Dump the received block from verify_zero() when verify_dump is set, so that both callers benefit. There is no point in also dumping an expected buffer, as the expected contents are all zeroes.

### DIFF
--- a/verify.c
+++ b/verify.c
@@ -891,7 +891,7 @@ static int mem_is_zero_slow(const void *data, size_t length, size_t *offset)
 	return !length;
 }
 
-static int verify_zero(struct io_u *io_u)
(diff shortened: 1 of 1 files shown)


sarthak-k:
@axboe Created new one. I ensured that I've followed the guidelines as provided. Can you please review?

PR #2141, opened 2026-09-02
init, backend: keep error output out of the --output file

### LINKED ISSUE #731: JSON output includes error messages
When executing (with fio version 3.12):

```
fio --output=file.json --output-format=json config.fio
```

Where **config.fio** contains the following:

```
[global]
create_on_open=1
kb_base=1000

[10MiB_x_1200_8T]
filesize=10MiB
nrfiles=150
numjobs=8
bs=512KiB
rw=write
openfiles=1
group_reporting=1
file_service_type=sequential
thread=4
directory=fio
```

Sometimes **file.json** contains something as the following:

```
fio: pid=4896, err=5/file:ioengines.c:351, func=td_io_queue, error=Input/output error
fio: pid=4128, err=5/file:ioengines.c:351, func=td_io_queue, error=Input/output error
fio: pid=4420, err=5/file:ioengines.c:351, func=td_io_queue, error=Input/output error
fio: pid=4008, err=5/file:ioengines.c:351, func=td_io_queue, error=Input/output error
fio: pid=4588, err=5/file:ioengines.c:351, func=td_io_queue, error=Input/output error
fio: pid=5192, err=5/file:ioengines.c:351, func=td_io_queue, error=Input/output error
fio: pid=4864, err=5/file:ioengines.c:351, func=td_io_queue, error=Input/output error
{
  "fio version" : "fio-3.12",
  "timestamp" : 1547028960,
  "timestamp_ms" : 1547028960664,
  "time" : "Wed Jan 9 10:16:00 2019",
  "global options" : {
    "kb_base" : "1000",
    "create_on_open" : "1"
  },
  "jobs" : [
    {
```

Those error messages are not JSON so parsing the file causes errors.

prabhakarpujeri:
Fixes #731.

With `--output-format=json`, informational and error messages were written to the same stream as the json document (the `--output` file, or stdout when no file was given), making the result unparseable:

```
fio: io_u error on file /dev/full: ...
fio: pid=..., err=28/..., error=No space left on device
{
  "fio version" : ...
```

An earlier revision of this PR only moved the per-job failure notice to log_err(), which as vincentkfu pointed out is just a bandage: any other log_info() call could still break the json.  This revision does the separation at the point where output is written, so it cannot leak:

- json payload always goes to the primary output stream (f_out) through a dedicated `log_json_buf()`, so the document is the only thing ever written there
- diagnostics (`log_info()`, `log_err()`, debug prints) go to stderr when json output is active, keeping the legacy stderr+file behaviour when it is not
- payloads of other requested formats (`normal`, `terse`, the job banners and progress messages) go through `log_human_buf()`/`log_human()` to the primary stream normally, and to a secondary stream when json owns the primary one: stdout when an `--output` file was given, stderr otherwise
- the eta display moves off stdout while stdout carries json
- client/server mode: the client renders the json document itself, so its stream separation is the same; server text messages land on the client's stderr

Multiple output formats are handled explicitly (requirement from review): `--output-format=json,normal --output=res.json` gives a parseable json file and the usual human text on stdout; `--output-format=json,normal` to stdout gives clean json on stdout with the text on stderr; `terse,json+` puts the json in the file and the terse record on stdout.  `--status-interval` json stays "collated sets" as documented, but every stanza is parseable and nothing else lands on that stream.  Documentation updated (HOWTO, fio.1) to state the contract.

With no json format selected nothing changes, except that the per-job `fio: pid=..., err=...` failure notice is now a real log_err().

The t/ suites that asserted on the old interleave were adapted: new SUCCESS_IGNORE_STDERR (SUCCESS_DEFAULT without the stderr-empty assertion) for the json-mode entries in run-fio-tests.py, client_server.py and verify_state_save.py, and latency_percentiles.py now looks for the terse record on both the .out file and the stdout capture.

Validation against unpatched master (A/B):
- json/json+/json,normal/json,terse × file/stdout × io-error runs: both error lines on stderr, json always parses, `jobs[0].error` still carries the failure, exit codes unchanged
- normal/terse runs without json: output byte-identical modulo timestamps, except the pid= line move above
- client/server runs: json clean on the client, server text on stderr
- full t/run-fio-tests.py run: 52 passed / 11 environment-failed, identical set on master and on this branch


vincentkfu:
This is a small bandage on a much larger issue. Instead of changes that address only a small piece of the problem we need to have a larger change that also addresses the other `log_info()` messages that might make Fio's JSON output illegal.
These are the requirements as I currently see them:
- Fio's JSON output is legal
- Figure out what to do when multiple output formats are requested

vincentkfu:
In the meantime post-processing Fio's output to ignore everything until the first `{` may help work around this issue.

prabhakarpujeri:
Fair point — replaced the band-aid with the actual separation.  With json requested, the document is now the only thing ever written to the stream carrying it (the `--output` file, or stdout): all informational and error output goes to stderr in that case, whatever fires during the run (io_u errors, verify notices, engine failures, debug prints).  That covers the other `log_info()` messengers without playing whack-a-mole with call sites.

For multiple output formats the rule is now: json always owns the output stream; other requested formats go to stdout when a file was given, or stderr when stdout itself carries json.  So `--output-format=json,normal --output=r.json` gives a parseable json file plus the usual human text on stdout, and `json,normal` to stdout gives clean json on stdout with the text on stderr.  `--status-interval` json stays "collated sets" as documented, but each stanza is parseable and nothing else lands on that stream.

t/ suites that asserted on the old interleave were updated (new SUCCESS_IGNORE_STDERR; terse,json+ now reads the terse record from stdout), and the full run-fio-tests profile is unchanged versus master.  Non-json paths are otherwise byte-identical, except the per-job `fio: pid=..., err=...` line which is now a real log_err.

PR #2144, opened 2026-09-04
backend: warn before write to block devices with recognizable content

WengRan:
backend: warn before write to block devices with recognizable content

fio currently protects against writing to mounted devices, but a job
file pointing at the wrong raw block device silently destroys whatever
lives there: partition tables, unmounted file systems, LVM physical
volumes, LUKS headers, swap areas and RAID members are all overwritten
without warning. A single filename typo can be catastrophic and there
is no way to get the data back.
This PR adds a write safety check for block devices, built on libblkid:
- Before any file is opened, every block device targeted by a write
  job is probed (read-only, separate fd) 
- If recognizable content is found, fio reports what the device
  contains in human-readable form, e.g.
  `fio: /dev/sda contains a GPT partition table, writing may damage it`.
- Interactive runs prompt `continue anyway? (y/N)` and abort before
  touching any file unless the user confirms.
- Non-interactive runs (stdin is not a tty, e.g. scripts, CI) only
  warn and continue, so existing automation is unaffected.

Signed-off-by: wengjianing <1528193783@qq.com>

### DIFF
--- a/backend.c
+++ b/backend.c
@@ -2625,6 +2625,149 @@ mounted:
 	return true;
 }
 
+/*
+ * Devices the user declined to overwrite at an earlier job's prompt,
+ * so later jobs and numjobs clones writing them are skipped without
+ * asking again. Only devices that actually triggered a prompt are
+ * recorded; clean devices of a declined job are not blacklisted.
+ */
+#define MAX_DECLINED_DEVS 256
+static const char *declined_devs[MAX_DECLINED_DEVS];
+static unsigned int nr_declined_devs;
+static bool non_interactive_warned;
+
+static bool dev_was_declined(const char *dev)
+{
+	unsigned int i;
+
+	for (i = 0; i < nr_declined_devs; i++)
+		if (!strcmp(declined_devs[i], dev))
+			return true;
+
+	return false;
+}
+
+static void dev_decline(const char *dev)
+{
+	if (dev_was_declined(dev))
+		return;
+
+	if (nr_declined_devs < MAX_DECLINED_DEVS)
+		declined_devs[nr_declined_devs++] = dev;
+}
+
+/*
+ * Warn before writing to block devices that hold data fio is about to
+ * endanger: a partition table, an LVM physical volume, a file system,
+ * or other recognizable content. Each job is prompted separately when
+ * running interactively; a non-interactive run only warns and continues.
+ * Returns true if this job must be skipped: either the user declined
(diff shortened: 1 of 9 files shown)


vincentkfu:
I haven't made up my mind about whether this feature is desirable or not. Users already have to be running as root or use `sudo` to read from or write to a raw block device.
Moreover, the way this is currently implemented is not acceptable because the prompt only appears after Fio has setup the files for each job. If an invocation is running multiple jobs with a stonewall between them, there may be multiple prompts appearing at different points of the invocation which is a poor user experience.
If you want some sort of protection like this, I might consider something like --do-not-write-recognizable-block-device which would be an opt-in check that will cause Fio to abort with an error if a job writes to a recognizable block device, but I would expect that this option would be mostly unused.

WengRan:
> I haven't made up my mind about whether this feature is desirable or not. Users already have to be running as root or use `sudo` to read from or write to a raw block device. Moreover, the way this is currently implemented is not acceptable because the prompt only appears after Fio has setup the files for each job. If an invocation is running multiple jobs with a stonewall between them, there may be multiple prompts appearing at different points of the invocation which is a poor user experience. If you want some sort of protection like this, I might consider something like --do-not-write-recognizable-block-device which would be an opt-in check that will cause Fio to abort with an error if a job writes to a recognizable block device, but I would expect that this option would be mostly unused.

Thanks for the feedback. I'd like to rework this as an opt-in option, disabled by default. I also understand your concern about repeatedly interrupting users in multi-job runs. To avoid that, I can make the check happen before any actual I/O begins: for each job that would write to a recognizable block device, prompt the user once. Since all prompts happen before any job starts its I/O, there would be no interruption in the middle of a run, even with multiple jobs and stonewall. If the user chooses not to continue for a given job, that job is skipped. 
Does that design sound acceptable?

WengRan:
I've just pushed an updated commit (9dbfc8b) to address the earlier feedback.

Summary of changes:
- Reworked the feature as an opt-in option, disabled by default.
- The check now runs before any I/O starts on write jobs, so if an invocation is running multiple jobs with a stonewall between them, warnings and confirmations will only be shown uniformly per-job at the start of invocation, rather than at different points during the invocation.      
       
The workflows are awaiting approval; could a maintainer approve them so CI can run? Thanks!     
                         
8a6f60b68bbf1fa6646c5ebab22d584c

3b7a813a2f9059c7b1d1558071b04505

